PR #219, opened 2022-11-01
chore (refs T22614): move core plugins & Vue (init & config) to clien…

ahmad-demos:
https://yaits.demos-deutschland.de/T22614

- move core plugins & Vue (init & config) to client/js.


salisdemos:
> I thought there should be a folder "tests/frontend" where all the jest-tests live. There the path for configVueLocal then has to be adjusted in all the tests. -> check https://demosdeutschland.slack.com/archives/C0471JB7H16/p1667399839383519 to see where to find them

Since the reintroducing of the tests is pending, we may merge this one here first then...

PR #235, opened 2022-11-04
Re-add frontend tests

spiess-demos:
The frontend tests got lost in the repo split, somehow. This PR adds all files
as [they exist in the archived repo](https://github.com/demos-europe/demosplan/tree/main/tests/frontend/unit). 

As many frontend components have been moved to `client/js` in the meantime, paths have changed. Those are already fixed here (hopefully all). 

~~One error persists, which i appreciate help with: the infamous `SyntaxError: Cannot use import statement outside a module`. There is lots of stackoverflow input on how the `transform` and `transformIgnorePatterns` options in the Jest config can solve this problem, however i couldn't manage to fix it.~~

### Main Changes
where made in `client/js/VueConfigLocal.js` and `client/setup/jest/setup.js` 
- setup.js needed some mocking from externals since there where some resolving issues
- DpVueCore had to be moved (and simplified) to the VueConfigLocal to avoid an Issue with the async/await from hasOwnProps 

The tests themeselfes where only moved and some deleted wich where outdated.

### How to review/test
Run `yarn unit:js` inside the container.

### PR Checklist

- [X] Tests updated/created


salisdemos:
I suggest to close this one, move the Tests for demosplan-ui there, and re-open another PR with only the tests belonging here.

spiess-demos:
As i already fixed a lot of stuff here, i guess it would be better to just move the demosplan-ui stuff out of here.

client/js/VueConfigLocal.js
```diff
@@ -18,12 +18,28 @@ import { createLocalVue, shallowMount } from '@vue/test-utils'
...
 import lscache from 'lscache'
 import PortalVue from 'portal-vue'
 import { VTooltip } from 'v-tooltip'
 import Vuex from 'vuex'
 
+/*
+ * This is copied from DpVueCore.js
+ * Since there is an issue with the hasOwnProp and it has to be loaded via
+ * async/await with makes prblems in the jest context, this looked like a way to
+ * make the tests run without refactor everything
```
hwiem, line 31:
```suggestion
 * This is copied from DpVueCore.js
 * Since there is an issue with the hasOwnProp and it has to be loaded via
 * async/await which causes problems in the jest context, this looked like a way to
 * make the tests run without refactoring everything
```

tests/frontend/unit/DpAutofillSubmitterData.json
```diff
@@ -0,0 +1,496 @@
+[
```
hwiem, line 1: Should we move this file to a `fixtures` subdirectory?

tests/frontend/unit/DpAutofillSubmitterData.spec.js
```diff
@@ -0,0 +1,93 @@
...
+describe('Submitter', () => {
+  it('should be an object', () => {
+    expect(typeof SubmitterComponent).toBe('object')
+  })
+
+  it('should be named DpAutofillSubmitterData', () => {
+    expect(SubmitterComponent.name).toBe('DpAutofillSubmitterData')
+  })
+
+  //  This represents permissions for Bob-HH + Bob-SH
```
hwiem, line 27: Should we remove this comment?

tests/frontend/unit/DpBulkEdit.spec.js
```diff
@@ -0,0 +1,82 @@
...
+
+  let store
+  let actions
+  let getters
+
+  const stubs = {
+    DpTiptap: '<h3>DpLinkModal</h3>'
+  }
+
+  /*
```
hwiem, line 34: Does it make sense to keep this test file in the current state? We should at least remove the commented out code, and the remaining tests are not very useful, so maybe we should just remove the file for now?

tests/frontend/unit/DpClaim.spec.js
```diff
@@ -0,0 +1,74 @@
...
+import shallowMountWithGlobalMocks from '@DpJs/VueConfigLocal'
+
+describe('DpClaim', () => {
+  it('should be an object', () => {
+    expect(typeof DpClaim).toBe('object')
+  })
+
+  it('should be named DpClaim', () => {
+    expect(DpClaim.name).toBe('DpClaim')
+  })
```
hwiem, line 21: Just a side note for the future: I'm not really sure how useful these two tests are, but if we want to keep them, we should move them to a separate file and import them in the test suites.

tests/frontend/unit/DpClaim.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        currentUserId: '1',
+        currentUserName: 'Aktueller Benutzer',
+        lastClaimedUserId: '',
+        entityType: 'fragment'
+      },
+      localVue
+    })
+
+    // Should show open lock if assignedId is not set - lastClaimed doesn't matter.
+    expect(wrapper.vm.status).toEqual({ icon: 'fa-unlock', text: 'statement.fragment.assignment.unassigned' })
```
hwiem, line 40: I think each `expect` statement should be moved to a separate `it` block, but I'm not sure it has to happen in this PR, since we just want to restore the tests asap.

tests/frontend/unit/DpEditFieldMultiSelect.spec.js
```diff
@@ -0,0 +1,93 @@
...
+describe('DpEditFieldMultiSelect', () => {
+  it('is an object', () => {
+    expect(typeof DpEditFieldMultiSelect).toBe('object')
+  })
+
+  it('is named dp-edit-field-multi-select', () => {
+    expect(DpEditFieldMultiSelect.name).toBe('DpEditFieldMultiSelect')
+  })
+
+  /*
```
hwiem, line 29: We should remove the commented out code in this file

tests/frontend/unit/DpFilterModal.spec.js
```diff
@@ -0,0 +1,77 @@
...
+          },
+          getters: getters,
+          actions: actions
+        }
+      }
+    })
+  })
+
+  /*
+   * It('should get data from store', () => {
```
hwiem, line 60: We should remove this commented out code

tests/frontend/unit/Filter.mock.js
```diff
@@ -0,0 +1,111 @@
+/**
+ * (c) 2010-present DEMOS E-Partizipation GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+export const filterList = [
```
hwiem, line 10: I think this file should go into the `__mocks__` directory

tests/frontend/unit/Layers.spec.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * (c) 2010-present DEMOS E-Partizipation GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
```
hwiem, line 9: Shouldn't this file be moved to the `store` subdirectory?

tests/frontend/unit/Layers.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    StubStore.commit('layers/updateApiData', apiData)
+    StubStore.commit('layers/saveOriginalState', JSON.parse(JSON.stringify(apiData)))
+    expect(StubStore.state.layers.apiData).toEqual(StubStore.state.layers.originalApiData)
+
+    StubStore.state.layers.apiData.data.id = 'xxx-xxx-xxx'
+    StubStore.state.layers.apiData.included.splice(0, 1)
+    expect(StubStore.state.layers.apiData).not.toEqual(StubStore.state.layers.originalApiData)
+  })
+
+  /*
```
hwiem, line 64: We should remove this commented out code

tests/frontend/unit/Notify.spec.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * (c) 2010-present DEMOS E-Partizipation GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+import { createLocalVue } from '@vue/test-utils'
```
hwiem, line 10: This should also be moved to the `store` subdirectory

tests/frontend/unit/__mocks__/dplan_api_map_options_admin.json
```diff
@@ -0,0 +1,46 @@
...
+    35000,
+    15000,
+    5000,
+    4000,
+    2000,
+    1000,
+    500,
+    250
+  ],
+  "baselayer": "https://wms.onmaps.de/?key=8ae2a661a3374681f5b409787b17f34c",
```
hwiem, line 44: Should we use basemap here?

tests/frontend/unit/__mocks__/layer_json.mock.js
```diff
@@ -0,0 +1,621 @@
...
+      layers: 'stadtplan',
+      name: 'Stadtkarte ORKa',
+      mapOrder: 3,
+      opacity: 100,
+      procedureId: '45752f51-f68a-11e5-b083-005056ae0004',
+      serviceType: 'wms',
+      type: 'base',
+      isBaseLayer: true,
+      tileMatrixSet: null,
+      url: 'http://geo.sv.rostock.de/geodienste/stadtplan/wms?service=WMS&version=1.1.1&request=GetMap',
```
hwiem, line 442: Is it ok to use these layers?

tests/frontend/unit/drawStyle.spec.js
```diff
@@ -0,0 +1,35 @@
...
+ *    * https://www.google.com/search?rlz=1C1GCEU_deDE821DE822&ei=VeU1XOfDG4SxgweEwL_QBA&q=jest+SyntaxError%3A+Unexpected+token+export+openlayers
+ *    * https://jestjs.io/docs/en/configuration.html#transformignorepatterns-array-string
+ *    * https://medium.com/@compatt84/how-to-test-open-layers-react-components-with-mocha-part-i-9a2ca0458ba1
+ *    * https://medium.com/@compatt84/how-to-test-open-layers-react-components-with-mocha-part-ii-d91d65145bce
+ */
+
+describe('DpOlMap/drawStyle', () => {
+  it('`drawStyle` should return an openLayers Style instance', () => {
+    /*
+     * So far not working :-(
```
hwiem, line 30: We should remove the commented out code (here and above)

tests/frontend/unit/Filter.spec.js
```diff
@@ -0,0 +1,54 @@
+/**
+ * (c) 2010-present DEMOS E-Partizipation GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
```
hwiem, line 9: Shouldn't this file be moved to the `store` subdirectory?

hwiem (changes requested):
I'm a little bit torn between wanting to restore the tests asap and not being happy with the current state of many tests. But I guess restoring the tests is the first step in fixing the state..

hwiem:
...but most importantly, thanks a lot for restoring the tests! :slightly_smiling_face:

tests/frontend/unit/DpAutofillSubmitterData.json
salisdemos: moved in ebea42bf8028deab62c33b12c886a13acfdec18f

tests/frontend/unit/DpAutofillSubmitterData.spec.js
salisdemos: yes. absolutely ebea42bf8028deab62c33b12c886a13acfdec18f

tests/frontend/unit/DpBulkEdit.spec.js
salisdemos: removed in 3fa65a02922197bd820d723212eec72f047df901

tests/frontend/unit/DpClaim.spec.js
salisdemos: I can't tell neither if they are really useful. I have seen them always as a starting point since it is/wasn't always easy to load a component wich get tested here. But why do you want to move them in a separate file?

tests/frontend/unit/Filter.spec.js
salisdemos: right. Moved this and some others 
d05e308de6f395621bc62016363a60924f3e099c

tests/frontend/unit/drawStyle.spec.js
salisdemos: I deleted the whole test-file since after deleting the commented out code no test cases where left 
Florian Salis test: Remove dead code 28 03 2023, 09:44

tests/frontend/unit/__mocks__/layer_json.mock.js
salisdemos: Thanks for pointing that out. I replaces all of them with the basemap
5813dad658afc1f94a46a8c1dfbc3470d7ace597

tests/frontend/unit/DpClaim.spec.js
hwiem: My idea was to make these two tests more generic, so instead of writing them for each component test suite, we could just import them.

hwiem (commented):
One last question

tests/frontend/unit/DpFilterModal.spec.js
```diff
@@ -0,0 +1,58 @@
...
+            filterList: [],
+            procedureId: ''
+          },
+          getters: getters,
+          actions: actions
+        }
+      }
+    })
+  })
+})
```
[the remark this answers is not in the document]
salisdemos: removed in 86f84d835ba147d116b6d1351ded630150045f02

hwiem (approved):
Let's get the tests back into `main` :tada:

PR #436, opened 2022-12-13
F t30460 autoremove caseworkername

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T30460

There are two different ways to determine "authorized" users of a procedure.
One way seems to be to simply ask for user is in the Procedure::getAuthorizedUsers(),
 the other way is to using the ProcedureService::getAuthorizedUsers() (which also uses rocedure::getAuthorizedUsers() in some case).
Obviously this leads to different results of "authorized users".

With this PR I used ProcedureService::getAuthorizedUsers(), to determine if a user is authorized user or not, because
this method is also used by the already existing logic of "unclaiming" users in case of removing them from the authorized users list: https://yaits.demos-deutschland.de/T28838

Therefore all planner of the organisation which is owning the procedure, will be authorized for the procedure, regardless of being listed in "authorized users" in the procedure admin area.

I'm still not sure about that. So let me know if you have any Idea how to solve this.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid:
I'm closing this in favor of a new attempt to get this into the code: https://github.com/demos-europe/demosplan-core/pull/599

PR #449, opened 2022-12-16
feat T29403 refactor keycloak authenticator

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T29403

Description: Move the logic to create or update a user from keycloak data to a new class OzgKeycloakUserLogin.php. Excluded the orga name as an identifier and ensure to use only the orga gwid to identify an orga. Throw an execption if the name "Privatperson" will be used to create a new orga.

### Linked PRs (optional)

https://github.com/demos-europe/demosplan-core/pull/410

### PR Checklist


mussbach (approved):
lgtm, code review only. It is a bit hard to review as you did not only move code but also perdormed some slight changes, but what I yound looked legit

PR #502, opened 2023-01-02
feat: respect the server settings, silence tests

mussbach:
Error reporting for tests should not be hardcoded but respect the server settings

### How to review/test
run tests. There should be less Deprecation warnings

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


graupnerdemos (approved):
While this has merge errors now, my prior concern is not of a high enough priority to block this.

PR #529, opened 2023-01-04
make addons activateable

AlexMenschoid:
This PR is fixing all the little bugs in Core to enable activating addons. It only means, that the project *can* load with an activated addon, not that it will necessarily work.

The addon will not yet do anything as there is more to be done with the FE integration, but it should be fully present after these changes.

### How to review/test
On this branch, it should be possible to install an addon **and** activate it while the project is still working fine. 

### Linked PRs (optional)
- Can only be merged after [this addon PR](https://github.com/demos-europe/demosplan-addon/pull/19)


demosplan/DemosPlanCoreBundle/Logic/FilterUiDataProvider.php
```diff
@@ -25,7 +25,7 @@ class FilterUiDataProvider
 
     public function __construct()
     {
-        $this->relativeFilterNamesPath = 'demosplan/addons/workflow/SegmentsManager/Resources/config/segmentsFilterNames.yaml';
+        $this->relativeFilterNamesPath = 'demosplan/DemosPlanCoreBundle/Resources/config/segmentsFilterNames.yaml';
```
graupnerdemos, line 28: This looks like something that in time wants to be moved to a config loader / compiler pass.

graupnerdemos (approved):
code review only. changes make sense. the comment is a non-blocking observation for the future.

demosplan/DemosPlanCoreBundle/Logic/FilterUiDataProvider.php
AlexMenschoid: true, this should be some part of a configuration. I'll try to keep that in mind as well for future refactorings.

PR #530, opened 2023-01-04
bug(refs T30796): fix extra email addresses disapearing when not persisting the form

muellerdemos:
https://yaits.demos-deutschland.de/T30796

- When not persisting the form but adding extra email addresses: these were removed from the form if in another DpSelect of that form was used to set a new value there.
- Somehow the watcher caused this problem. This led to a confusing data architecture because the initEmails prop and the emails data prop were updated at the same time when in fact initEmails should only contain an array of already persisted emails. This was fixed by removing the watcher, set the default emails data prop to an empty array and fill the data prop with the read-only initEmails prop on mounting the component.

However, it remains unclear why the initEmails were changed when setting a new value in a unrelated DpSelect. This PR fixes the Problem for now but the underlying problem could not be fully determined. This is why I suggest to refactor the DpBasicSettings.vue and administration_edit.html.twig files. From my perspective too much functionality has been wired here together which makes it a bit difficult to understand what is actually going on.

### How to review/test

- Login as FPA user
- navigate to basic settings
- add extra email addresses
- change values of any Select in other accordions (e.g.  Verfahrensschritt Institutionen )
-> the added email addresses should still be there and when persisting the form the added addresses should be displayed

### PR Checklist

Screenshot 2023-01-04 at 16-35-10 Grundeinstellungen New Procedure All Planning Agency Admin Tests BOB-SH Bauleitplanung

### DIFF
--- a/client/js/components/procedure/basicSettings/DpEmailList.vue
+++ b/client/js/components/procedure/basicSettings/DpEmailList.vue
@@ -84,7 +84,7 @@ export default {
 
   watch: {
     initEmails (newVal) {
-      this.newEmails = newVal
+      this.emails = newVal
     }
   },
 
--- a/client/js/components/procedure/basicSettings/DpEmailList.vue
+++ b/client/js/components/procedure/basicSettings/DpEmailList.vue
@@ -70,7 +70,7 @@ export default {
         mail: ''
       },
       itemIndex: null,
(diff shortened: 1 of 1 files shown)


muellerdemos:
https://github.com/demos-europe/demosplan-core/pull/543 solves this so this one can be closed

PR #531, opened 2023-01-04
add portal-vue dependency to dependabot.yml

muellerdemos:
The last version which is supported for vue2 is portal-vue^2.1.7 (which is the one we are using) hence this dependency should be added to dependabot.yml unless we migrate to vue3 in the near future.


graupnerdemos (approved):
Good catch, thx.

PR #534, opened 2023-01-04
feat: ensure that project rollup migrations are performed

mussbach:
Project migrations need to be performed in two parts. The project migrations up to Version 20220914133419 contains all migrations that needs to be executed before the core migration rollup is performed.

### How to review/test
you might perform `bin/<project> dplan:migrate` on an old database that has outstanding migrations before core migration rollup

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Command/Db/MigrateCommand.php
+++ b/demosplan/DemosPlanCoreBundle/Command/Db/MigrateCommand.php
@@ -49,17 +49,32 @@ class MigrateCommand extends CoreCommand
             $db = '--conn='.$db;
         }
 
+
+        $commands = [
+            "dplan:migrations:cache --env={$env}",
+            "doctrine:migrations:sync-metadata-storage {$db} --env={$env}",
+        ];
+
+        $migrationsConfigurationPath = DemosPlanPath::getProjectPath('app/config/project_migrations.yml');
+        $migrationsSyncCommand = 'doctrine:migrations:sync-metadata-storage --configuration ';
+        $migrationsCommand = 'doctrine:migrations:migrate --configuration ';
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Command/Db/MigrateCommand.php
```diff
@@ -49,17 +49,32 @@ public function execute(InputInterface $input, OutputInterface $output): int
...
+
+        $commands = [
+            "dplan:migrations:cache --env={$env}",
+            "doctrine:migrations:sync-metadata-storage {$db} --env={$env}",
+        ];
+
+        $migrationsConfigurationPath = DemosPlanPath::getProjectPath('app/config/project_migrations.yml');
+        $migrationsSyncCommand = 'doctrine:migrations:sync-metadata-storage --configuration ';
+        $migrationsCommand = 'doctrine:migrations:migrate --configuration ';
+        $lastRollupMigration = 'app/Resources/DemosPlanCoreBundle/DoctrineMigrations/2022/09/Version20220914133419.php';
```
graupnerdemos, line 61: this would make more sense as a constant for me.

graupnerdemos (approved):
one minor style suggestion, otherwise looks good.

demosplan/DemosPlanCoreBundle/Command/Db/MigrateCommand.php
mussbach: Even if this is only used in this single context? Hm, I would change it, I do not really care, but feels a bit too much to me

graupnerdemos: It's just a matter of code style. I don't really care either way.

PR #544, opened 2023-01-05
ci: Disallow updates to Symfony 6

graupnerdemos:
Unfortunately, we're not quite ready yet to make the jump to Symfony 6. To keep Dependabot for PHP usable, we should remove these from it's tracking list in order to not fill up our limit of open PRs.

### How to review/test

Check the added ignore rule against the documentation: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -52,6 +52,8 @@ updates:
         versions:
           - ">= 9.a"
           - "< 10"
+      - dependency-name: symfony/*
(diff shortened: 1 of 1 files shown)


mussbach (approved):
lgtm, thx

PR #547, opened 2023-01-05
feat: update test configuration to Elasticsearch 7 syntax

mussbach:
Our configuration to disable elasticsearch in tests needs to be updated to be compatible to ES7

Elasticsearch 7 does not have any types any more so these ones need to be skipped in the configuration file

### How to review/test
reviewdog action should not throw an error regarding elasticsearch configuration /update/ Reviewdog won't run as no php files are touched /update/


graupnerdemos (approved):
lgtm. good catch!

PR #556, opened 2023-01-06
fix(refs T30824): catches LockedByAssignmentException

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T30824

As it was previously uncaught which lead to no readable message being added to the message bag.

This is based on an incident, but I decided to fix it on `main` as that project is going to be updated with `main` very soon anyaway.

### How to review/test
I think a code review should suffice here.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Segment/DraftsInfoController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Segment/DraftsInfoController.php
@@ -13,6 +13,7 @@ namespace demosplan\DemosPlanCoreBundle\Controller\Segment;
 use demosplan\DemosPlanCoreBundle\Annotation\DplanPermissions;
 use demosplan\DemosPlanCoreBundle\Controller\Base\BaseController;
 use demosplan\DemosPlanCoreBundle\Entity\Statement\Statement;
+use demosplan\DemosPlanCoreBundle\Exception\LockedByAssignmentException;
 use demosplan\DemosPlanCoreBundle\Exception\StatementAlreadySegmentedException;
 use demosplan\DemosPlanCoreBundle\Validator\SegmentableStatementValidator;
 use demosplan\DemosPlanStatementBundle\Exception\StatementNotFoundException;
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Controller/Segment/DraftsInfoController.php
```diff
@@ -106,8 +108,8 @@ public function editAction(
         } catch (StatementNotFoundException $e) {
             $this->getMessageBag()->add('error', 'error.statement.not.found');
             throw $e;
-        } catch (StatementAlreadySegmentedException $e) {
-            $this->getMessageBag()->add('error', 'error.statement.already.segmented');
+        } catch (StatementAlreadySegmentedException|LockedByAssignmentException $e) {
+            $this->messageBag->add('error', $e->getMessage());
```
mussbach, line 112: Is it safe to directly output the Exception message here?

AlexMenschoid: I would say so, yes. They both can only be thrown in line 93 `$segmentableStatementValidator->validate($statementId);` and the message generated there is the old explicit one for the `StatementAlreadySegmentedException` and a user friendly one for `LockedByAssignmentException` stating that a user can't edit a statement that is not assigned to them. I think that convey relevant information without any technical details.

mussbach: OK for now, but I feel very uncomfortable about passing Exception messages directly to user. What if a new implementation does not know that it is passed? Perhaps we should introduce something to improve DX that makes it clear that this data is passed to users directly

mussbach (approved):
lgtm

demosplan/DemosPlanCoreBundle/Controller/Segment/DraftsInfoController.php
mussbach: Great, thx, :heart:

mussbach (approved):
lgtm, code review only

PR #557, opened 2023-01-06
Temporarily use GitHub version of demosplan-ui v0.0.6

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T30430

the v0.0.6 release of demosplan-ui did not contain the actual code of that release, as we confused the "prepublish" with the appropriate "prepare" npm lifecycle hook. Until the next release fixes that, this PR requires an interim branch in which demosplan-ui moves webpack, webpack-cli and style-dictionary within its dependencies (leading to a temporarily bloated "yarn.lock" but things will settle as the 0.0.7 npm version of demosplan-ui comes along). 

See https://docs.npmjs.com/cli/v9/using-npm/scripts

also, in a8080d6c27f the two @demos-europe modules were mistakenly designated to be included with babel-loader. This makes no sense as they are already bundled as umd.

### How to review/test
Remove node_nodules, clear yarn cache, install, and build. Open any view ("new statement" eg) with a DpEditor instance. It should be rendered properly.

https://github.com/demos-europe/demosplan-core/pull/478 may be merged after this one here.

### PR Checklist

### DIFF
--- a/client/setup/webpack/moduleRules.js
+++ b/client/setup/webpack/moduleRules.js
@@ -27,8 +27,6 @@ transpiledModules = transpiledModules.concat(
   [
     '@efrane/vuex-json-api',
     '@mapbox', // Ol sub-dependency
-    '@demos-europe/demosplan-ui',
-    '@demos-europe/demosplan-utils',
     'vue-resize'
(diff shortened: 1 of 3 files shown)


salisdemos (approved):
should do

PR #573, opened 2023-01-09
build: Disable virus scan in dev environment

graupnerdemos:
Since the clam setup in the current dev environment is rather brittle disabling the virus scanning on uploads, while technically a security risk, makes it easier to debug and use the remainder of the file upload processes. This should only be a temporary fix and developers should be aware that they can cause harm on their systems if they use testfiles for uploading which contain malware of any kind.

### How to review/test

Uploads on the local dev environment work as expected.

### PR Checklist


mussbach (approved):
lgtm, thx

PR #576, opened 2023-01-09
Use demosplan-ui 0.0.7

spiess-demos:
Now that we have it on npm, we can use it! This might also fix some "strange" behavior in dev environments as demosplan-ui is not build when installing anymore.

### How to review/test
`yarn install` and stuff should work as expected.


hwiem (approved):
lgtm

PR #577, opened 2023-01-09
bug(refs T27124): Change height of menu

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T27124

Both menu and items' container was set to max-height 500, but the menu also has some padding. This has to be added to the max height.

### How to review/test
Go to Master-Töb-Liste and hover over "Spalten ausblenden". You should see all menu items.


salisdemos (approved):
looks good

PR #578, opened 2023-01-09
feat (refs T29639): replace webatlas by basemap

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T29639

Description: This PR includes a core migration to replace the webatlas with basemap.

### How to review/test
- execute migration
- check if in all procedure templates webatlas is replaced by basemap
(Verwalten -> Blaupausen -> select one -> Planungsdokumente und Planzeichnung -> Kartenebenen definieren)
- to check if the master template is also changed, create a new procedure template
- to check if the copyright is changed (Verwalten -> Blaupausen -> select one -> Planungsdokumente und Planzeichnung -> Startkartenausschnitt festlegen)
- check the database (tables: _gis, _procedure_settings and customer)

### PR Checklist


demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/01/Version20230104090531.php
```diff
@@ -0,0 +1,129 @@
...
+        // Replace all Web-Atlas Entries by basemap (only for procedure templates).
+        $this->addSql(
+        'UPDATE _gis AS g
+            INNER JOIN _procedure AS p ON g._p_id = p._p_id
+            SET g._g_name           = :gname,
+                g._g_layers         = :glayer,
+                g._g_url            = :newgurl
+            WHERE g._g_url = :oldgurl AND p._p_master = 1 AND p._p_deleted = 0',
+            [
+                'gname'     => self::NEW_G_NAME,
```
mussbach, line 43: In general it would be better to use variable names that are easily comprehensible

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/01/Version20230104090531.php
```diff
@@ -0,0 +1,129 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+class Version20230104090531 extends AbstractMigration
+{
+    private const NEW_G_NAME = 'basemap';
+    private const NEW_G_LAYER = 'de_basemapde_web_raster_farbe';
+    private const NEW_G_URL = 'https://sgx.geodatenzentrum.de/wms_basemapde';
+    //private const G_PROJECTION_LABEL = 'EPSG:3857';
+    //private const G_PROJECTION_VALUE = '+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs';
+    private const NEW_COPYRIGHT_TEXT = '© basemap.de BKG (www.basemap.de) / LVermGeo SH (www.LVermGeoSH.schleswig-holstein.de)';
```
mussbach, line 16: According to https://sgx.geodatenzentrum.de/web_public/gdz/lizenz/deu/basemapde_web_dienste_lizenz.pdf this needs to be `© basemap.de / BKG 2023`

mussbach (changes requested):
why is this a draft? What is missing?

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/01/Version20230104090531.php
martendemos: Was answered in Ticket: 
`© basemap.de BKG (www.basemap.de) / LVermGeo SH (www.LVermGeoSH.schleswig-holstein.de)`
is exactly what is requested.

mussbach (approved):

mussbach (approved):
thank you

martendemos:
I assume the related US and therefore this PR. I can't find any reason why this is still a draft. Probably this just waited for my assistance.

breierDemos:
> why is this a draft? What is missing?

I was still waiting for answers from the PO's regarding the new name and the new copyright notice.
Furthermore, I was still unsure about the down migration. What if there are already changed maps. These would be changed with a down migration to Web-Atlas.

 Oh and I still found the old copyright notice in some places in the code. Adjusted in [a2790de](https://github.com/demos-europe/demosplan-core/pull/578/commits/a2790de16a4f091fb12c6849736c3d27e3c4f64f)
 I found two json files with old data in them. Do they also have to be adjusted or are they somehow magically generated and change by themselves? One is in [a2790de](https://github.com/demos-europe/demosplan-core/pull/578/commits/a2790de16a4f091fb12c6849736c3d27e3c4f64f) the other is called geojson1.json.

martendemos:
> Furthermore, I was still unsure about the down migration. What if there are already changed maps. These would be changed with a down migration to Web-Atlas.

Good point.
I think we can do delete the down-Migrations, because they are very rarely used at all and therefore haven't a high value.

mussbach:
> Down-Migrations are are very rarely used at all and therefore haven't a great value.

I would strongly disagree. I use them from time to time and they can prevent severe problems with invalid database data and structure

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/01/Version20230104090531.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * @throws Exception
+     */
+    public function down(Schema $schema): void
+    {
+        $this->abortIfNotMysql();
+
+        $this->throwIrreversibleMigrationException('This migration is not revertable because we can\'t
```
mussbach, line 92: Please do not throw an exception, just silently do nothing

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/01/Version20230104090531.php
breierDemos: adjusted in [8c4f137](https://github.com/demos-europe/demosplan-core/pull/578/commits/8c4f1371225fbcc1a3f78c282b31755b79778af8)

mussbach: The comment as comment would be helpful :-)

breierDemos: of course, you are right :)
done in [952fba1](https://github.com/demos-europe/demosplan-core/pull/578/commits/952fba1e2ec34e5ab178385afab8f2a33eff6c55)

breierDemos: Why shouldn't an exception be thrown?

mussbach: There are cases when we need to down migrate several migrations. Consider we are at migration E and want to go back to B for example because an installation failed and you need to fix something in the infrastructure before you could retry the installation again. Of course you could use the backup, but there are cases, where this is complicated or strictly not necessary, as you can quite easily and rapidly fix the issue.
When in this case migration D would throw an exception then the whole process is stopped and there is no way around it.
Down migrations are seldomly used, but when you need them you are really happy if you have them

PR #579, opened 2023-01-09
Do not purge demosplan-utils generated css selectors

spiess-demos:
Some CSS classes required for DpStickyElement were purged because the lib that sets them is located inside demosplan-utils. Until we move the latter into demosplan-ui we have to tell purgeCss to also look for CSS classes in demosplan-utils.

**How to test it**
After building, when scrolling down the assessment table, the side menu should stay fixed to the top of the page (instead of having a random offset top).

### DIFF
--- a/client/setup/config.js
+++ b/client/setup/config.js
@@ -56,7 +56,8 @@ class Config {
         'demosplan/**/*.js.twig',
         'client/**/*.js',
         'client/**/*.vue',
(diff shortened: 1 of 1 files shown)


elmasdemos (approved):
lgtm, code review

gruenbergerdemos (approved):
works and lgtm :+1:

PR #586, opened 2023-01-10
bug(refs T27268): Add hint for no search result

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T27268

There was no hint for no search results when searching for organisations. 

### How to review/test
Login as support and go to Organisationen and search for something with no result. There should be a hint including the searchterm.


elmasdemos (approved):
lgtm, code review

PR #587, opened 2023-01-10
feat: halts jsrouting on 2.8.x

AlexMenschoid:
The update to 3.x requires us to use Symfony 6, which we will not in the next couple of weeks. So we're deactivating the update for the moment.

### Linked PRs (optional)
- Dependabot already alerted us to the update #{511}


graupnerdemos (approved):
Thanks!

PR #592, opened 2023-01-10
fix(refs T30808): Check Permissions for Branding propperly

salisdemos:
One big issue was, that the Branding got requested only beccause
 the permission "field_sign_language_overview_video_edit" was enabled

during investigationg this. I refactored the code a little, to hopefully make it better readable and to not send empty fields

**Ticket:** https://yaits.demos-deutschland.de/T30808

### How to review/test

in bob-sh as support, try to open the "plattformeinstellungen" 
You may doublecheck if the page still works in other projects.

While this fixes, that the page can be visited again, it also reveals a bunch of new errors thrown from demosplan-ui. They have to be fixed in a separate PR.

### PR Checklist


elmasdemos (approved):
thank you for the fix! lgtm, code review

PR #598, opened 2023-01-11
fix broken ES configuration, caused wrong format, caused by merge

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T30885

Wrong format caused by "magic"merge


AlexMenschoid (approved):
yeah, something went wrong there.

lgtm, code review only

PR #599, opened 2023-01-11
F t30460 automatically remove unauthorized case worker

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T30460

This is the second attempt to get this fixed. The first attempt was never merged and had some weird edits going on, so @martendemos and me decided to just redo it and maybe improve a bit on the original design (original #436).

It is possible that a user was assigned as caseworker for a number of statements. Later that user gets removed from the list of authorized users for the procedure. At this point, the user should also no longer show up as caseworker on any statements, so we have to find those statements and remove the caseworker there. 

That's what this does.

### How to review/test
1. Add a user to a procedure
2. Assign that user as caseworker to a statement
3. Remove the user from the list of authorized user for that procedure (in procedure settings)
4. The user should also be removed as the caseworker on the statement

### Linked PRs (optional)
- First Draft #436

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


martendemos:
Unittests are missing, but does not seems to be part of your DOD at the moment.

martendemos (dismissed):
Code review and manual test

demosplan/DemosPlanCoreBundle/Logic/Statement/NonAuthorizedCaseworkerRemover.php
```diff
@@ -0,0 +1,156 @@
...
+            ->unique()
+            ->all();
+    }
+
+    /**
+     * Removes caseworkers from the array of statements.
+     *
+     * @param array<int, Statement> $statementsWithSetCaseWorker
+     */
+    private function unsetCaseWorkers(array $statementsWithSetCaseWorker): array
```
dresslerdemos, line 140: Return type-hint missing.

demosplan/DemosPlanCoreBundle/Logic/Statement/NonAuthorizedCaseworkerRemover.php
```diff
@@ -0,0 +1,156 @@
...
+     */
+    private function getAuthorizedUserNames(Procedure $procedure): array
+    {
+        $ownsProcedureCondition = $this->procedureAccessEvaluator->getOwnsProcedureCondition($procedure);
+        $authorizedUsers = $this->procedureService->getAuthorizedUsers($procedure->getId());
+        $owningUsers = $this->entityFetcher->listEntitiesUnrestricted(User::class, [$ownsProcedureCondition]);
+
+        return $authorizedUsers
+            ->merge($owningUsers)
+            ->map(static fn (User $user): string => $user->getName())
```
dresslerdemos, line 130: I can guess why you need to use the name, but this will result in problems. Having two case workers with the same name is sadly likely enough that it will happen at some point.

dresslerdemos (changes requested):

demosplan/DemosPlanCoreBundle/Logic/Statement/NonAuthorizedCaseworkerRemover.php
AlexMenschoid: Sadly, the caseworker is literally just a string on the statementMeta. So we can only go by the name. The alternative would be to remove caseworkers in the current implementation and re-implement it as a relationship to user. But I don't even know if there isn't a use case where names of persons get added as caseworkers who aren't even users on the plattform.

dresslerdemos: As long as the case worker feature is needed the best way for the user would be to keep the string field but to also allow the usage of a relationship _instead_ (e.g. via dropdown). When a user is removed from the procedure we can automatically remove the case worker relationships.

For the string case worker field I don't think we have another choice but to simply warn the planner on procedure settings edit that there are `n` case workers that he may want to remove manually.

dresslerdemos (changes requested):
https://github.com/demos-europe/demosplan-core/pull/599#discussion_r1071032516

demosplan/DemosPlanCoreBundle/Logic/Statement/NonAuthorizedCaseworkerRemover.php
martendemos: In the process of creating the first attempt to fix this (#436), I asked the PO via the related Ticket, how it should be.
I interpreted the answer like: The name should be removed in case of the user with this name will be removed from the list of authorized users.
I've take this as implicate "ok" to the resulting string comparison and therefore for the possible false positives.

But in order to take over responsibility for the code I wrote, I think you are right.
This kind of logic is to fragile. The wohle "caseworker"-thing should be questioned.
Also I doubt that the automatically removing the name of a caseworker once was a working feature of the system, which means it is a feature request, but not a bugfix.

AlexMenschoid:
After discussing it with PO, we came to the conclusion that this is (most likely) a feature that should be rather removed than anything else. Even if it's not, then doing this automatically is unreliable (as @dresslerdemos pointed out) and also rather unnecessary with regards to the feature itself. Therefore, I'm closing this PR and the corresponding ticket will be rejected.

PR #607, opened 2023-01-12
feature (refs T19171): also update procedureSettings::EmailTitle when updating the procedureName

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T19171

When the procedure name gets updated - the procedure settings are not
updated despite its property emailTitle containing the procedure name as well.
Changed so the procedure-settings-emailTitle gets updated as well when the procedure name
gets updated.
The ProcedureSettingsEmailTitle is used as the subject for emails described in the ticket.

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
```diff
@@ -1345,6 +1345,10 @@ public function administrationEditAction(
                     $inData['r_name'] = $inData['r_externalName'];
                 }
 
+                // also change the procedureSettings EmailTitle as it contains the ProcedureName
+                // used for email subjects when inviting registered institutions
+                $inData['r_emailTitle'] = 'Einladung zur Beteiligung: '.$inData['r_name'];
```
mussbach, line 1350: This should be a transkey

mussbach (changes requested):

dresslerdemos (changes requested):
Maybe I'm missing something, but I think we should not store redundant data in the database, to avoid problems like this one. If the user can't override the `emailTitle` property manually with something custom, then it should be something like `emailTitlePrefix`.

If he can override it, then this solution does not work, as it deletes his custom value when the procedure title is changed.

MoritzMandler:
> If he can override it, then this solution does not work, as it deletes his custom value when the procedure title is changed.

I just found that place:
ProcedureService::addProcedureEntity() line 1063

Working on a solution...

dresslerdemos (changes requested):
I'm concerned about the behavioral complexity, as this is just another hidden feature that is hard to know when developing, testing and using the application.

My suggestion is to allow the usage of an actual `procedureName` placeholder when editing the email title field. The user can then deliberately decide if the procedure name should automatically update in the email title field.

Placeholders (especially their UI) were already implemented in the application at some point.

We can decide if we want to replace the procedure titles initially with the placeholders or if we leave it as is and let the users decide if they want to use them.

MoritzMandler:
03a283058f65724f16bb769d85af530b23a015b9 
the check weather or not the new procedure-name was already set within the emailTitle  prior a procedure-name update is necessary to prevent potential bugs like for example:
Editing the Procedure-Settings EmailTitle first:
```Einladung zur Beteiligung: Procedure I```
gets changed to:
```Einladung zur Beteiligung: Procedure IA*```
then secondly the procedure-name gets updated to ```Procedure IA*```
and replaces every occurrence of ```Procedure I``` within the emailTitle with ```Procedure IA*```  leading to:
```Einladung zur Beteiligung: Procedure IA*A*```

File: demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php [hunk 2]
```diff
@@ -1345,6 +1345,20 @@ public function administrationEditAction(
...
 
+                // also change the procedureSettings EmailTitle as it may contain the old ProcedureName
+                // used for email subjects when inviting registered institutions
+                $emailTitle = $procedureObject->getSettings()->getEmailTitle();
+                if ($procedureObject->getName() !== $inData['r_name']
+                    // the user may have updated the emailTitle already and in this case do not update
+                    && !$this->eMailTitleContainsNewProcedureName($emailTitle, $inData['r_name'])
+                    && $this->eMailTitleContainsOldProcedureName($emailTitle, $procedureObject->getName())
+                ) {
+                    $inData['r_emailTitle'] = str_replace(
```
dresslerdemos, line 1356:
```suggestion
                    $inData['r_emailTitle'] = str_ireplace(
```

demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
```diff
@@ -2872,4 +2886,14 @@ private function validateAdministrationEditInput(array $input): void
             throw new InvalidArgumentException();
         }
     }
+
+    private function eMailTitleContainsNewProcedureName(string $eMailTitle, string $newProcedureName): bool
+    {
+        return str_contains($eMailTitle, $newProcedureName);
```
dresslerdemos, line 2892:
```suggestion
        return str_contains(strtolower($eMailTitle), strtolower($newProcedureName));
```

dresslerdemos (changes requested):

demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
MoritzMandler: We dont need this trans-key at all anymore.

PR #626, opened 2023-01-16
Fix: setVisible is undefined error

elmasdemos:
Description: 
- if the baseLayer in enabled, that means not published, in public detail the console shows an error that setVisible has typeError undefined
- to not be confused, now there is a check in `Map.vue`, which shows a console.warn, that the baseLayer is not published yet, if firstTogglableLayer, which contains the attribute `setVisible`, is undefined

### How to review/test

- Different cases:
- 1.) Go to `Planungsdokumente und Planzeichnung`, click to `Kartenebenen definieren`, scroll to the baseLayers configuration and click `Bearbeiten`, activate the checkbox for `Karteneben veröffentlichen` and save this configuration. Go to the public detail and you will see the baseLayer and no console.warn and no typeError.
- 2.) Go to `Planungsdokumente und Planzeichnung`, click to `Kartenebenen definieren`, scroll to the baseLayers configuration and click `Bearbeiten`, deactivate the checkbox for `Karteneben veröffentlichen` and save this configuration. Go to the public detail and you will not see the baseLayer and a console.warn that the baseLayer is enabled and no typeError.

- Important: There is no request from the customer to this cases. So this is just a fix for the error in the console.


client/js/components/map/publicdetail/Map.vue
```diff
@@ -420,7 +420,12 @@ export default {
         const firstTogglableLayer = this.baseLayers.find(layer => {
           return layer.getProperties().isEnabled === true && layer.getProperties().title !== 'customBaselayerDanmark'
         })
-        firstTogglableLayer.setVisible(true)
+
+        if(typeof firstTogglableLayer !== 'undefined') {
+          firstTogglableLayer.setVisible(true)
+        } else {
+          console.warn('There is no published baseLayer')
```
salisdemos, line 427:
```suggestion
          console.warn('There is no published baseLayer, please go to the admin area and set a baselayer as "published"')
```

salisdemos (changes requested):

client/js/components/map/publicdetail/Map.vue
```diff
@@ -420,7 +420,12 @@ export default {
         const firstTogglableLayer = this.baseLayers.find(layer => {
           return layer.getProperties().isEnabled === true && layer.getProperties().title !== 'customBaselayerDanmark'
         })
-        firstTogglableLayer.setVisible(true)
+
+        if(typeof firstTogglableLayer !== 'undefined') {
```
spiess-demos, line 424: This will be executed if there are baseLayers defined, but none of them is published. Just out of curiosity: did you test how the system behaves if no baseLayer is defined at all?

Apart from that, a minor nitpicky suggestion:

```suggestion
        if (typeof firstTogglableLayer !== 'undefined') {
```

client/js/components/map/publicdetail/Map.vue
elmasdemos: Thank you for the advice. done https://github.com/demos-europe/demosplan-core/pull/626/commits/4c4664def66686d2a8e8745ed430a4ecf1554fc5

salisdemos (dismissed):
Thanks for the improvement

client/js/components/map/publicdetail/Map.vue
elmasdemos: done https://github.com/demos-europe/demosplan-core/pull/626/commits/f105c540478e6563fec85e81b788b9d3382c1754

PR #629, opened 2023-01-16
fix (refs T30865): Count normal statements of procedure

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T30865
Count normal statements of procedure to allow showing a count of normal statements in procedure-list, additionally to count of original-statements.

### How to review/test
In the list of the procedures, should be two counts of statements now, including a tooltip, which discripe

### DIFF
--- a/demosplan/DemosPlanStatementBundle/Repository/StatementRepository.php
+++ b/demosplan/DemosPlanStatementBundle/Repository/StatementRepository.php
@@ -1438,7 +1438,7 @@ class StatementRepository extends FluentRepository implements ArrayInterface, Ob
             ->leftJoin('statement.meta', 'meta')
             ->andWhere('statement.original IS NULL')
             ->andWhere($queryBuilder->expr()->orX(
-            // submitted statements
+                // submitted statements
                 $queryBuilder->expr()->eq('meta.submitUId', ':userId'),
                 // authored statements
                 $queryBuilder->expr()->eq('statement.user', ':userId')
--- a/demosplan/DemosPlanStatementBundle/Repository/StatementRepository.php
+++ b/demosplan/DemosPlanStatementBundle/Repository/StatementRepository.php
@@ -11,6 +11,8 @@
 namespace demosplan\DemosPlanStatementBundle\Repository;
 
 use Carbon\Carbon;
+use DateInterval;
+use DateTime;
 use demosplan\DemosPlanCoreBundle\Entity\Document\Elements;
 use demosplan\DemosPlanCoreBundle\Entity\Document\ParagraphVersion;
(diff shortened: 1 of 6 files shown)


gruenbergerdemos (approved):
works and FE lgtm. 
Just a minor suggestion: If there is enough space, I would place the tooltip next to the heading instead of beneath it.

demosplan/DemosPlanStatementBundle/Repository/StatementRepository.php
```diff
@@ -1446,36 +1450,67 @@ public function getSubmittedOrAuthoredStatements(string $userId): array
...
+     *                            original statements which are no placeholders and not deleted as values
+     */
+    public function getOriginalStatementsCounts(array $procedureIds): array
+    {
+        if ([] === $procedureIds) {
+            return [];
+        }
+
+        $queryBuilder = $this->createStatementsCountsQueryBuilder($procedureIds);
+        $queryBuilder->andWhere($queryBuilder->expr()->isNull('statement.original'));
```
breierDemos, line 1466: getOriginalStatementsCounts() and getStatementsCounts() seem to differ only in this one line, I think this should be handled differently to avoid code duplication.

breierDemos (changes requested):

demosplan/DemosPlanStatementBundle/Repository/StatementRepository.php
martendemos: getOriginalStatementsCounts() was already existing.
I added getStatementsCounts(). To avoid code duplication I extracted the shared logic into createStatementsCountsQueryBuilder().

What do you mean with "handled differently"?
More reducing of the remaining code duplication?

dresslerdemos (approved):
Paired review + diff review. There are some things I'm not sure anymore regarding our best practices (e.g. we could use `non-empty-string` instead of `string`, or we could drop the redirect-only `ProcedureService::getStatementsCounts` method). But such topics need to be discussed in general and do not block this PR.

demosplan/DemosPlanStatementBundle/Repository/StatementRepository.php
breierDemos: Yes I meant reducing the remaining code duplication.

breierDemos: Pair reviewed - not a blocker for now. Could be refactored in a seperate PR.

breierDemos (approved):
Good for now. Maybe the remaining code duplication in getOriginalStatementsCounts() and getStatementsCounts() could refactored in a seperate PR.

PR #636, opened 2023-01-17
bug(refs: T27370): fix createddate and submitteddate in dp public statement list

muellerdemos:
https://yaits.demos-deutschland.de/T27370

The fix of this issue contains 2 approaches:

- The statement object which will be used to transform the data does not contain a `submittedDate` key but a submit key containing the unix timestamp
- The format of the unix timestamp of `statement.submit` and `statement.createdDate` are not consitent. The `createdDate` contains a 13 digit unix timestamp (miliseconds) and the `submit` only a 10 digit timestamp (seconds) hence a calculation is needed for dayjs to display the correct date (reference: https://day.js.org/docs/en/parse/unix-timestamp) by simply dividing the unix timestamp by 1000
- The values passed to dayjs were undefined and there must be some kind of callback in such cases in dayjs to return the current date. Which I assume was the underlying problem for shared statements containing always the current date in the header field.  

### How to review/test

Navigate to the public detail page of a procedure and have a look at submitted and created statements (e.g. draft statements and shared statement tabs and verify the dates in the header fields of `DpPublicStatement.vue `are correct - see also screenshots)

image

image

### PR Checklist

### DIFF
--- a/client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
+++ b/client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
@@ -261,8 +261,8 @@ export default {
       const statementParagraph = (paragraph && paragraph.title) || Translator.trans('notspecified')
       const text = statement.text
 
-      const transformedSubmitDate = submitted === false ? {} : { submittedDate: dayjs(submittedDate).format('DD.MM.YYYY HH:mm') }
-      const transformedCreatedDate = dayjs(createdDate).format('DD.MM.YYYY HH:mm')
+      const transformedSubmitDate = submitted === false ? {} : { submittedDate: dayjs.unix(statement.submit).format('DD.MM.YYYY HH:mm') }
+      const transformedCreatedDate = dayjs.unix(statement.createdDate / 1000).format('DD.MM.YYYY HH:mm')
 
       const transformedPolygon = polygon === '' ? {} : JSON.parse(polygon)
 
@@ -290,6 +290,7 @@ export default {
     },
 
     transformStatements (statements) {
+      console.log(statements.map(s => this.transformStatement(s)))
       return statements.map(s => this.transformStatement(s))
     }
   }
--- a/client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
+++ b/client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
@@ -290,7 +290,6 @@ export default {
     },
 
     transformStatements (statements) {
-      console.log(statements.map(s => this.transformStatement(s)))
       return statements.map(s => this.transformStatement(s))
     }
   }


client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
```diff
@@ -290,6 +290,7 @@ export default {
     },
 
     transformStatements (statements) {
+      console.log(statements.map(s => this.transformStatement(s)))
```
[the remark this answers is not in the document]
muellerdemos: this was done in [931ab6c](https://github.com/demos-europe/demosplan-core/pull/636/commits/931ab6ce2d322687b7e1fffba0395cfae8d64b25)

client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
```diff
@@ -261,8 +261,8 @@ export default {
       const statementParagraph = (paragraph && paragraph.title) || Translator.trans('notspecified')
       const text = statement.text
 
-      const transformedSubmitDate = submitted === false ? {} : { submittedDate: dayjs(submittedDate).format('DD.MM.YYYY HH:mm') }
-      const transformedCreatedDate = dayjs(createdDate).format('DD.MM.YYYY HH:mm')
+      const transformedSubmitDate = submitted === false ? {} : { submittedDate: dayjs.unix(statement.submit).format('DD.MM.YYYY HH:mm') }
+      const transformedCreatedDate = dayjs.unix(statement.createdDate / 1000).format('DD.MM.YYYY HH:mm')
```
spiess-demos, line 265: What is achieved by this operation?

muellerdemos: Well the timestamp of `createdDate` and `submittedDate` are in different formats (one in miliseconds the other in seconds) so to display the correct date this operation was required, otherwise a statement contained a weird created date in the header)

spiess-demos: I dislike the idea of fixing inconsistent data in the frontend. 

> Well the timestamp of createdDate and submittedDate are in different formats (one in miliseconds the other in seconds) 

Are they persisted like that in the database? I do not think so... where does the bug originate from?

muellerdemos: Yes they are persisted like that in the database and DpPubblicStatementList.vue receives the statement data via the templateVars which will be transformed in the function above. Would your proposal be to tell the backend we are going to need consistent data?

mussbach:
@muellerdemos what should be done with this PR?

muellerdemos:
@mussbach oh, this PR is quite outdated by now. As far as I can reconstruct this, we decided that the fix should be applied in the backend

muellerdemos:
this PR can be closed, the fix would not be applied like done here by FE but in BE

PR #639, opened 2023-01-17
fix(refs T31010): Catch undefined maxExtend

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T31010

As "support" Goto
-> /einstellungen/plattform (Plattform-Einstellungen) 
The Map should load without errors

(there may be other Errors in the Console, but they are not related to this Bug)

There are circumstances, where we don't get a maxExtend. Instead of an empty Array we receive `undefined` which was not expected. 

### PR Checklist


elmasdemos (approved):
thanks for the fix! lgtm, code review

PR #646, opened 2023-01-18
fix (refs T30990): Escape any special chars for output

AlexMenschoid:
Ticket: https://yaits.demos-deutschland.de/T30990

In most cases we store texts with already escaped special chars in the database to avoid any later problems. But in some rare cases and for reasons I don't, some special chars make their way into the database. Without the here added line of code (which is present in most if not all other docx exports) those will break the docx generation as that is based on a xml file.
How to review/test

Exporting all statements as a zip with individual docx files should now also produce working docx files for statements with xml breaking special chars (for example '&') in their texts.


breierDemos (approved):
lgtm

PR #648, opened 2023-01-18
bug(refs T27479): Fix bulk edit for documents

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T27479

The given objects in the change event for node selection have changed (in DpTreeListNode), so they also need to be changed in the methods listening to that event. 
Also some refactoring.

### How to review/test
Go to Verfahren -> Dokumente and select some documents. There should be a bulk edit header appear, where you can delete or publish the documents. 

### Related PRs:
https://github.com/demos-europe/demosplan-ui/pull/73

### DIFF
--- a/client/js/components/document/ElementsAdminList.vue
+++ b/client/js/components/document/ElementsAdminList.vue
@@ -84,11 +84,9 @@
 </template>
 
 <script>
-import { hasAnyPermissions, hasOwnProp } from '@demos-europe/demosplan-utils'
+import { DpBulkEditHeader, DpLoading, DpTreeList } from '@demos-europe/demosplan-ui'
+import { dpRpc, hasAnyPermissions, hasOwnProp } from '@demos-europe/demosplan-utils'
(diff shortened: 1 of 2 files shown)


muellerdemos (changes requested):
When selecting single documents of the list, the control menu does indeed appear now, however there is still an problem with the bulk edit checkbox, which does not trigger the rendering of the control menu.

image

image

gruenbergerdemos:
This is another bug, it has the same cause as here https://yaits.demos-deutschland.de/T27441. Unfortunately I don't know how to fix this, so I didn't approach it here. If you have an idea, happy to hear, otherwise I think we should fix it another time.

spiess-demos (approved):
lgtm! code review only, seen it working at @muellerdemos ' machine.

muellerdemos (approved):
These changes are working and hence this PR can be merged. There is another PR https://github.com/demos-europe/demosplan-ui/pull/73 which fixes the underlying Problem

PR #649, opened 2023-01-18
feature (refs T27190): pdf attached to a paraDoc gets exportet even if no chapters exists for that specific paraDoc.

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T27190

Description:
if there was a planningDocument of category paragraph without any chapters but an attached pdf file - this file was excluded
due to a faulty if condition. The document was only attached if the paradoc was not empty.
When exporting a procedure this change assures that a pdf attached to a paraDoc gets exportet even if no chapters exists for that specific paraDoc.

How to test:
Due to a non functional RabbitMQ connection - this can only be testet on dev.

- [ ] Tests updated/created
- [ ] Update documentation


dresslerdemos (approved):
Basically diff review only. I tried to find the reason for the original implementation in the archive, but there was no real insight.

PR #640, opened 2023-01-17
Refactor addon registration and installation

graupnerdemos:
Installing and loading Addons in demosplan is a tedious multistep process.
Some actions happen on the file system, some in Composer, some in Symfony.
Of the actions in Symfony, some need to occur during the container compile
step, some always and some after.

Previously, all of this was more or less mangled into the AddonRegistry
and the AddonInstallFromZipCommand. This commit is a step towards cleaning
things up and making the process a little more understandable and in
consequence hopefully also more maintainable.

As of now, to install an addon zip bundle (installation via pure
composer package is still not fully implemented), the AddonInstallFromZipCommand
has the logic for

- unarchiving the zip to the addons cache (which, as a reminder, is only
  responsible for storing zipped addons and will not be utilized for
  composer-only addons)
- handling the composer "installation" aka the local package registration
  and dependency resolving of the addon
- leaving the system bootable with prepared autoloading for addons

it then passes on to the Registrator which

- takes care of ensuring the addon is valid (not very comprehensive
  yet but that shall be extended upon in due time)
- loads and stores the addon's manifest into the addons.yaml

next, the AddonBuildFrontendCommand takes over and

- leaves again if there is no frontend configuration necessary, or

- installs the required frontend dependencies with Yarn
- runs the production build for the frontend assets assuming the
  addon was packaged using @demos-europe/demosplan-addon

Furthermore, as previously the autoloading, the bundle loading
was extracted into a standalone system.

The AddonRegistry is now an actually runtime usable registry
of all availble addons and configured during container compilation
to receive the AddonInfo instances which replace the AddonInfoProvider
system.

### How to review/test

Install the demospipes addon in it's current state, it should install including frontend stuffs.

PR #658, opened 2023-01-19
fix: removes tagged_iterator argument for addonInfoProviders

AlexMenschoid:
as this is obsolete. in the future we will simply inject the AddonRegistry which then provides the information for all the addons.

### Linked PRs (optional)

**This can only be merged after:** #640 and #683 

**There is a corresponding PR for ALL projects. Please make sure to check them all as well! I'm not listing them here as this is a public repository now.**

**You can see the linked project PRs below** :)


graupnerdemos (dismissed):
@AlexMenschoid thanks. I approved the whole set of these. However, this one has a conflict that needs resolving for merging. My gut feeling is that we already made the change in the test config elsewhere.

PR #670, opened 2023-01-23
Add missing mutation to original statements view

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T31094

While the mutation is called in the event handler, it was nowhere imported. This breaks the whole thing.

### How to review/test
In the "Originalstellungnahmen" view, find a statement that is long enough to have the "Mehr anzeigen +" button below the statement text. This button should be functional.

### PR Checklist


hwiem (approved):
lgtm

PR #672, opened 2023-01-24
Feat(T30784): add street address to the submitter list

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T30784

**Description:** The submitter list currently lacks the information about street and house number. This PR adds a new column called `street` to the table. The `initialOrganisationHouseNumber` and `initialOrganisationStreet` fields are used for this purpose.

- use `DpColumnSelector` to be able to show only selected columns
- default values are: name, organisationAndDepartment and statement
- column headers are stored in the local storage under the `submitterList` key
- the submitter list table is no longer re-sizable, because the `DpColumnSelector` works incorrectly with the re-sizable table. (done here the same way as in the segments table `SegmentsList.vue`)

**How to review/test**
1. login with AHB-A, AHB-SB, FP-A, FP-SB users
2. select a procedure
3. click on **Einreichende** in the right menu


client/js/components/procedure/DpSubmitterList.vue
```diff
@@ -188,15 +208,24 @@ export default {
...
+      if (initialOrganisationStreet) {
+        fullAddress = initialOrganisationHouseNumber ? initialOrganisationHouseNumber + ' ' + initialOrganisationStreet : initialOrganisationStreet
+      }
       if (initialOrganisationPostalCode) {
-        return initialOrganisationCity ? initialOrganisationPostalCode + ' ' + initialOrganisationCity : initialOrganisationPostalCode
+        fullAddress += initialOrganisationStreet ? ', ' : ''
+        fullAddress += initialOrganisationCity ? initialOrganisationPostalCode + ' ' + initialOrganisationCity : initialOrganisationPostalCode
       }
-      return initialOrganisationCity || '-'
+      return fullAddress || '-'
```
spiess-demos, line 220: Within the function, i suggest to drop the `initialOrganisation` from every variable name because it harms readability (also see above suggestion).

client/js/components/procedure/DpSubmitterList.vue
```diff
@@ -162,7 +180,9 @@ export default {
         isCitizen,
         memo,
         submitName,
-        submitterEmailAddress
+        submitterEmailAddress,
+        initialOrganisationHouseNumber,
+        initialOrganisationStreet
```
spiess-demos, line 185: We could rename the variables while destructuring for better readability like:

```suggestion
        submitterEmailAddress: email,
        initialOrganisationHouseNumber: houseNumber,
        initialOrganisationStreet: street
```

Also, alphabetical sorting.

client/js/components/procedure/DpSubmitterList.vue
```diff
@@ -79,17 +88,18 @@ export default {
...
-        { field: 'email', label: Translator.trans('email'), initialMaxWidth: 200 },
-        { field: 'postalCodeAndCity', label: Translator.trans('postalcode') + ' / ' + Translator.trans('city'), initialMaxWidth: 120 },
-        { field: 'organisationAndDepartment', label: Translator.trans('organisation') + ' / ' + Translator.trans('department'), initialMaxWidth: 200 },
-        { field: 'memo', label: Translator.trans('memo'), initialMinWidth: 100, initialMaxWidth: 200 },
-        { field: 'internId', label: Translator.trans('internId.shortened'), initialWidth: 80 },
-        { field: 'statement', label: Translator.trans('id'), tooltip: Translator.trans('id.statement.long'), initialWidth: 40 }
+      headerFieldsAvailable: [
+        { field: 'name', label: Translator.trans('name') },
+        { field: 'email', label: Translator.trans('email') },
+        { field: 'address', label: Translator.trans('city') },
```
spiess-demos, line 94: If this field contains the address, i'd propose to also use the label `address`.

```suggestion
        { field: 'address', label: Translator.trans('address') },
```

spiess-demos (changes requested):
Some rather nitpicky concerns, but another one may be rather fundamental: have you tested the view with realistic data? Does the layout break if long cities and street names meet long names or such? Is the table scrollable if it exceeds the available width?

client/js/components/procedure/DpSubmitterList.vue
muellerdemos: image

Since the address can contain a postalcode+city or a whole address I suggest to make this also visible in the table header like `Adresse / Ort`. I was expecting a city when seeing  `Ort` in the header field but I would also be expecting more than just a postal code when I see the header is `Adresse`

muellerdemos (changes requested):
Just one little remark (besides the other concerns)

sakutademos:
> Some rather nitpicky concerns, but another one may be rather fundamental: have you tested the view with realistic data? Does the layout break if long cities and street names meet long names or such? Is the table scrollable if it exceeds the available width?

Thx for a review! Well, I have tested it with realistic data (different accounts and different data (string length)). The table is scrollable if it exceeds the available width through the use of `overflow-x-auto` class, so the layout should not be broken..

I’ve divided the address into two columns [here](https://github.com/demos-europe/demosplan-core/pull/672/commits/b71acab69b89cb56e6d1e80a829ff8c5a9a3ef25) (**street** and **Postal Code / City**). The first version was not correct. The street address should stay before the postal code in a separate column.

client/js/components/procedure/DpSubmitterList.vue
sakutademos: thx, done!

client/js/components/procedure/DpSubmitterList.vue
sakutademos: done [here](https://github.com/demos-europe/demosplan-core/pull/672/commits/40d83afd9e1edf4cc43d774edcae6d385aff5134)

client/js/components/procedure/DpSubmitterList.vue
sakutademos: thx! I’ve divided the address into two columns [here](https://github.com/demos-europe/demosplan-core/pull/672/commits/b71acab69b89cb56e6d1e80a829ff8c5a9a3ef25) (street and Postal Code / City). The first version was not correct. The street address should stay before the postal code in a separate column.

client/js/components/procedure/DpSubmitterList.vue
```diff
@@ -22,13 +22,21 @@
...
     </a>
 
+    <div class="u-mb text--right">
+      <dp-column-selector
+        :initial-selection="currentSelection"
+        :selectable-columns="selectableColumns"
+        @selection-changed="setCurrentSelection"
+        use-local-storage
+        local-storage-key="submitterList" />
+    </div>
```
spiess-demos, line 32: What if we put both the "export" link and the solumn selector inside one container? With flex  utils we could do it like so:

```suggestion
    <div class="flex flex-items-center">
      <a
        class="u-mb"
        :href="exportSubmitterList">
        <i
          class="fa fa-download"
          aria-hidden="true" />
        {{ Translator.trans('export') }}
      </a>

      <dp-column-selector
        class="flex-item-end"
        :initial-selection="currentSelection"
        :selectable-columns="selectableColumns"
        @selection-changed="setCurrentSelection"
        use-local-storage
        local-storage-key="submitterList" />
    </div>
```

spiess-demos (changes requested):
one last layout suggestion

client/js/components/procedure/DpSubmitterList.vue
sakutademos: good idea, thx! The two elements are placed in the flex-box now https://github.com/demos-europe/demosplan-core/pull/672/commits/dfe689b4efc47dda998f15d21120a65000d4d24e

spiess-demos (dismissed):
Looks good - thanks for refactoring those long var names 😎

muellerdemos (approved):
thanks for optimizing the header fields, it is not confusing anymore with the separation of `address` and `city` - lgtm

PR #683, opened 2023-01-25
Improve addon contract wiring

graupnerdemos:
@AlexMenschoid you know what to do ❤️ 🙈 🙉 🙊 

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Addon/Registrator.php
```diff
@@ -66,6 +65,8 @@ private function refreshAddonsYaml(): void
             'addons' => $this->addons,
         ];
 
+        // TODO: add a "this is autogenerated" note to the file
```
AlexMenschoid, line 69: I did a quick search, but I have no idea how to include that in the dump. We might have to create the file first with the comment and then append the dump, which seems a bit too much trouble for a comment, I think.

demosplan/DemosPlanCoreBundle/Addon/RpcAddonAssetsLoader.php
```diff
@@ -78,9 +78,10 @@ public function isTransactional(): bool
 
     public function validateRpcRequest(object $rpcRequest): void
     {
-        if (!$this->permissions->hasPermission('area_admin_consultations')) {
-            throw AccessDeniedException::missingPermission('area_admin_consultations');
-        }
+        // TODO: This needs a better validation
+        //if (!$this->permissions->hasPermission('area_admin_consultations')) {
+        //    throw AccessDeniedException::missingPermission('area_admin_consultations');
+        //}
```
AlexMenschoid, line 84: We should fix this before we merge it, I'm afraid

demosplan/DemosPlanCoreBundle/Resources/config/services.yml
```diff
@@ -31,6 +31,193 @@ services:
...
+
+#    DemosEurope\DemosplanAddon\Contracts\Entities\EntityInterface:
+#        class:
+
+    DemosEurope\DemosplanAddon\Logic\Rpc\RpcErrorGeneratorInterface:
+        class: demosplan\DemosPlanCoreBundle\Logic\Rpc\RpcErrorGenerator
+
+    DemosEurope\DemosplanAddon\Permission\Validation\PermissionFilterValidatorInterface:
+        class: demosplan\DemosPlanCoreBundle\Permissions\PermissionResolver
+
```
AlexMenschoid, line 220: Is this really necessary? And if so, shouldn't it be in the services.yml of the addon-repo?

AlexMenschoid (changes requested):
I already added a couple of commits for little changes and a couple of things that were already solved but now broken again. I imagine, something went a bit wrong solving merge conflicts.

demosplan/DemosPlanCoreBundle/Addon/Registrator.php
graupnerdemos: It's not that complicated, I just forgot about it. We just dump into a string and prepend the comment, then write the string to the file.

demosplan/DemosPlanCoreBundle/Addon/RpcAddonAssetsLoader.php
graupnerdemos: That's not my Baustellenflatterband-comment. 😇  (I'll have a look.)

demosplan/DemosPlanCoreBundle/Resources/config/services.yml
graupnerdemos: If you have a brilliant idea on how to bind services in the addon repo which the addon repo doesn't and mustn't know about, I'm all ears.

AlexMenschoid (approved):
looks as good as it can be to me. Let's go

PR #708, opened 2023-01-27
feature (refs T25539): display actual date instead of ### in xlsx export

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T25539

Description:
the date gets displayed
as ### if not enough space is provided
to fit in the dateString.
Despite having set a default width - the date
cell needs to be told again about its Width.
Adjusted its alignment to match the others.

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
This sounds like a weird phpSpreadsheet bug. This seems ok for now, hopefully we can remove it at a later date again after the library fixed it on their end.

PR #717, opened 2023-01-30
fix: Repair addons.yaml format

graupnerdemos:
The auto-generation info was lacking a newline and thus broke the loading as the format of the saved yaml was not parsable anymore.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
makes sense :)

PR #728, opened 2023-01-31
chore: fix return type covariance issues

mussbach:
Phpstan complained about invalid covariant method signatures (https://www.php.net/manual/en/language.oop5.variance.php)

### How to review/test
Code review and use phpstan to validate 

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanStatementBundle/Logic/AssessmentExportOptions.php
```diff
@@ -270,10 +270,7 @@ public function get($section)
         return $this->options[$section];
     }
 
-    /**
-     * @return mixed
-     */
-    public function jsonSerialize()
+    public function jsonSerialize(): Collection
```
AlexMenschoid, line 273: Either I'm missing something or this is very wrong. Why should jsonSerialize return a Collection? Especially when it is used in another method that tries to json encode that collection. I think the naming is throwing me for a loop here but I would prefer the naming to get fixed together with the return type.

mussbach: This is just how the current implementation works. I did not want to change anything (yet)

mussbach: @AlexMenschoid may I merge?

mussbach: ping @AlexMenschoid

PR #736, opened 2023-02-01
feat: addon frontend loading

muellerdemos:
This PR enables the usage of the PDF Importer in Frontend. Basically a loading mechanism was implemented to load addons dynamically after receiving the response with the stringified addon component which will be evaluated and automatically binded to the window object. 

### How to review/test

- checkout branch
- install addons zip file (ask a friendly collegue of backend for it ;))
- run: dp:addon:install addons/demosplan-addon-demospipes.zip
- clear cache
- build ewm
- navigate to the import center -> the pdf importer should mount now

### Related PRs:
https://github.com/demos-europe/demosplan-addon-demospipes/pull/21
https://github.com/demos-europe/demosplan-addon-demospipes/pull/20
https://github.com/demos-europe/demosplan-js-addon/pull/14


client/js/components/addon/AddonWrapper.vue
```diff
@@ -0,0 +1,77 @@
...
+  },
+
+  data () {
+    return {
+      component: ''
+    }
+  },
+
+  methods: {
+    getCurrentProperties() {
```
spiess-demos, line 45: Why is a method needed here? Wouldn't it suffice to pass `addonProps` in the template?

client/js/components/addon/AddonWrapper.vue
```diff
@@ -0,0 +1,77 @@
...
+      return this.addonProps
+    },
+
+    loadComponents(hookName) {
```
spiess-demos, line 49: As `loadComponents` is invoked onMounted, could we also drop the param and access `this.hookName` in line 51?

client/js/components/addon/AddonWrapper.vue
```diff
@@ -0,0 +1,77 @@
...
+        .then(response => checkResponse(response))
+        .then(response => {
+          const result = response[0].result
+
+          for (const key of Object.keys(result)) {
+            const addon = result[key]
+            const contentKey = addon.entry + '.umd.js'
+            const content = addon.content[contentKey]
+
+            eval(content)
```
spiess-demos, line 64: Just to ask this here once: we are 100% sure that `content` can be considered "safe" at this point?

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
```diff
@@ -134,48 +131,37 @@ export default {
...
+
+            /**
+             * The evaluation of the response content automatically binds the vue component
+             * to the window object. This way we can implement it in vue's internals to render
+             * the component.
+             */
+            eval(content)
+            this.$options.components[addon.entry] = window[addon.entry].default
+
+            this.asyncComponents.push({
```
spiess-demos, line 157: Why do we use `asyncComponents` here but not in `AddonWrapper`?

muellerdemos: The display the tabs in the import center we need the `asyncComponents` Object to be able to use the `<component :is="...` architecture of the tabs. The `AddonWrapper` has been crafted to mainly serve the purpose to be able to load addon components when the underlying structure is a twig template. Since the `AddonWrapper` is currently only enabled to load a single addon component we do not need the `asyncComponent` structure here because the addon component which is loaded will be automatically set as the component to be displayed.

client/js/components/addon/AddonWrapper.vue
muellerdemos: You are right, we can simply bind the  `addonProps` directly without needing a seperate function. Done in https://github.com/demos-europe/demosplan-core/pull/736/commits/b549b653dbeae08a97e5e3ebe385b097fe374b12

client/js/components/addon/AddonWrapper.vue
muellerdemos: True, we can reduce some lines of code with your suggestion so this was done here https://github.com/demos-europe/demosplan-core/pull/736/commits/283ec1f2756b5b6d46ab5a9cb4035690467ff4f7

spiess-demos: i actually meant `loadComponents() {` and `dpRpc('addons.assets.load', { hookName: this.hookName })`.

muellerdemos: I see, sorry I got that wrong. Dropped the function param for directly usage here: https://github.com/demos-europe/demosplan-core/pull/736/commits/f1f753663dcc8909bac7069263ea540ac49add5d

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
spiess-demos: thanks for the explanation!

spiess-demos (approved):
fe lgtm (code review only)

AlexMenschoid (approved):
BE lgtm, code review only.

PR #737, opened 2023-02-01
Fix(refs: T31100): vue warn by splitting datensätze

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T31100

**Description:**

- the originally problem at begin were the vue warns described in the ticket that for props `documents` and `paragraph` expected an object and got an array
- first set a safety check in backend that if no documents or paragraph exists,
that there is at least one group for the `elementId` available in Frontend, even if its an empty array, 
which is created in `getElementBlock` function and which is set in relation with `elementId` 
- however there exists another error that the assigne.id is null, so that the tab `Datensätze` didnt show the details
- the problem was that that there is no assignee before switching in frontend to the `Datensätze` tab, because
you are available to claim it after switching to this part and seeing which `Datensätze` are available to claim,
otherwise you are not able to claim, so before claiming there is no assignee and you need to check if there is an assignee, than it should take the id of the assignee only in the case that an assignee exists
- another important thing is the computed `isClaim` -> this needs a check if the permission 'feature_statement_assignment' is activated, a statement or fragment is claimed in the case if an assignee exists and the assignee is the logged user so that a statement or fragment(='Datensatz') is claimed
- if this permission is not activated in other projects, the statement or fragment looks like already is claimed,
so that any user can edit this part without claim because in the some projects there is no opportunity that different user could have claimed this statement or fragment
- also I did some refactorings because of sonarcloud issues and cognitive complexity
- I added a confirm message, if the `Begründung` get updated in `Datensätze`

### How to review/test

- Go to the AT in the tab `Datensätze` for example in Robob
- claim a `Datensatz`, change the `Begründung` or `Textabschnitt der Stellungnahme` 
- in the case that there is no `Datensatz`, click to `Datensätze aufteilen` and save this

### PR Checklist


client/js/components/statement/assessmentTable/Fragment.vue
```diff
@@ -542,14 +538,32 @@ export default {
...
+      /*
+      * Some projects do not have the claim opportunity
+      * if the permission 'feature_statement_assignment' is activated,
+      * a statement or fragment is claimed if an assignee exist
+      * and the assignee should be the logged user so that a statement or fragment(='Datensatz') is claimed
+      * after claiming it the statement or fragment is editable for this user
+      * if this permission is not activated, the statement or fragment looks like already is claimed,
+      * so that any user can edit this part without claim because in the projects without this permission,
+      * there is no other opportunity that different user could have claimed this statement or fragment
+      */
```
hwiem, line 551: I don't find the current comment very easy to understand, maybe we could shorten it a little?
We could also consider renaming `isClaimed` to `isEditable`, since the fragment is not actually claimed if the claim feature doesn't exist

```suggestion
      /*
      * The fragment is only editable if 
      * a) in case the permission for the claim feature is active:
      * - the fragment has an assignee
      * - the assignee is the currently logged-in user
      * b) in case the permission for the claim feature is not active:
      * always
      */
```

client/js/components/statement/assessmentTable/Fragment.vue
```diff
@@ -740,25 +734,71 @@ export default {
...
+          fieldToUpdate = this.$refs.considerationAdvice
+        }
+
+        this.updateTag(data, fieldToUpdate, textToBeAdded, updated)
+        this.addTagsToConsideration(textToBeAdded, fieldToUpdate)
+      }
+    },
+
+    updateTag (data, fieldToUpdate, textToBeAdded, updated) {
+      // Then we fire a request for each tag - we return an array of promises to wait until all requests are finished
```
hwiem, line 777: I'd move this comment to where `updateTag()` is called, because right now it's not clear what the "Then" refers to

client/js/components/statement/fragment/SelectDocument.vue
```diff
@@ -93,79 +93,79 @@ export default {
...
       default: () => ({})
     },
 
-    documents: {
-      type: Object,
+    elements: {
       required: true,
-      default: () => ({})
+      type: Array,
+      default: () => []
```
hwiem, line 105: I know you didn't introduce it, so just as a side note: If a prop is required, a default is not needed (as it will never be used)

client/js/components/statement/fragment/SelectDocument.vue
```diff
@@ -190,6 +190,7 @@ export default {
      * @return {boolean}
      */
     elementsHasFiles () {
+      console.log('documents', this.documents[this.currentElementId])
```
hwiem, line 193: I guess this can be removed

hwiem (changes requested):
In general it looks good, just a few small remarks

dresslerdemos (approved):
Disclaimer: Co-authored by me.

elmasdemos:
closed because reopened in another PR because of issues after refactoring

PR #664, opened 2023-01-20
refactor (refs T30603): integrate ForumBundle into CoreBundle

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T30603

In our effort to integrate all Bundles into CoreBundle this PR handles the ForumBundle

### How to review/test
code review or try to interact as moderator with development forum

### PR Checklist

Delete the checkbox if it doesn't apply/isn't necessary.

- [x] Tests updated/created
- [ ] Update documentation
- [x] Link all relevant tickets
- [ ] Move the tickets on the board accordingly

PR #741, opened 2023-02-02
Fix (ref:T31192) : adjust twigs paths

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T31192

Description: 
Twigs under DemosPlanMiscContentBundle are not availabe anymore while it was deactivated in https://github.com/demos-europe/demosplan-core/pull/664. These twigs are still used in core ( needed for all projects ) that's why they have to be moved under core bundle to be accessible.
Since DemosPlanMiscContentBundle is deactivated and the twigs are already under core bundle, this bundle (DemosPlanMiscContentBundle) is removed too.


MoritzMandler (changes requested):
There are still entries within the config.yml of every project regarding the         ```'DemosPlanMiscContent'```
for example: ```projects/regio/app/config/config.yml```
entry: twig: path: ```'%demosplan.project_dir%/templates/bundles/DemosPlanMiscContentBundle': 'DemosPlanMiscContent'```

bakridemos:
more adjustements are needed here, this PR will be redone

PR #761, opened 2023-02-03
F t30967 WIP Introducing Alice Fixtures and better bidirectional relationship handling

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T30967

I've started to see if we could use Alice fixtures to build better fixtures for testing and server population and everything else. It looks partially very promising, but due to the rather convoluted nature of our current data model, we can't use Alice fixtures without an additional PHP pass to glue entities together after the whole generation process happened. 
That seems to me rather against the idea of easy and clean fixtures all in one place. So we have to think about it some more and maybe start with cleaning up our data model.

This also adds better handling for bidirectional relationships as we sometimes tend to ignore the fact, that it is the responsibility of us developer to keep both sides in sync.

### How to review/test
This is **not** ready for review or test. But I very much encourage anyone to give it a read and maybe post some ideas/thoughts/feelings about this.


AlexMenschoid:
We have decided to rather go with Foundry for better fixtures, so this PR is no longer relevant.

PR #762, opened 2023-02-03
feat(refs T31218): Fix docx encoding

graupnerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T31218

The default output encoding of PhpWord is en_US which leads to issues in post processing as they don't have all the beautiful umlauts.

In the future, we might need to make this dependent on the application or project locale, but for now, setting it to german should be enough.

### How to review/test

Export to a docx, open in M$ Word, be happy about the correct language setting. Or just move on with life.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
sounds reasonable, why not? :)

PR #766, opened 2023-02-06
bug(refs T31159): Add check for empty array

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T31159

Map wasn't loading when the procedure extent was empty, because we didn't check for empty arrays.

## How to review/test

Go to Verfahren -> Grundeinstellungen -> Verortung. You should see the map and be able to use the search.


salisdemos (approved):
lgtm

PR #775, opened 2023-02-07
fix: prevent dependabot PRs to update to vue3

muellerdemos:
- Since we are not ready to migrate to vue3 we must prevent dependabot to create update PRs for this dependecny
- The last version we can use is 2.7.14

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,10 @@ updates:
             - ">= 27"
 
       # Is Vue 3 only
+      -   dependency-name: vue
+          versions:
+            - ">= 2.7.14"
+            -
       - dependency-name: vue-loader
         versions:
             - ">= 16"


graupnerdemos (changes requested):
I think this should be `< 3` since there might still be (security) updates to Vue 2.

.github/dependabot.yml
```diff
@@ -22,6 +22,10 @@ updates:
             - ">= 27"
 
       # Is Vue 3 only
+      -   dependency-name: vue
```
hwiem, line 25: The above comment refers to vue-loader I guess, so it should be kept there. Also, the indentation seems off (?).

muellerdemos: Technically the portal-vue dependency falls under this too but you are right, I have added a comment that we can only enable vue3 dependabot PRs if we are ready to migrate to vue 3

Also the indentation of this file was a little off and I have fixed this too.

Both done here: https://github.com/demos-europe/demosplan-core/pull/775/commits/856d2fb78da5911426c031ccaf2eef82e5b7e294

hwiem (approved):
lgtm, thanks!

graupnerdemos (approved):

PR #786, opened 2023-02-08
fix(refs T31275): Fix-refactor measuretools

salisdemos:
after fixing and refactoring the measuretools we have overseen an undefined variable which was forgotten to move to the instance state. Doing this forced me to move several other variables and methods too while this makes the fix a bit large it moves the code in the right direction

 **Ticket:** https://yaits.demos-deutschland.de/T31275

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/map/publicdetail/Map.vue
+++ b/client/js/components/map/publicdetail/Map.vue
@@ -125,10 +125,13 @@ export default {
 
   data () {
     return {
+      activeclickcontrol: null,
       autocompleteOptions: [],
       bPlan: {},
       hasTerritoryWMS: false,
+      dragZoomAlways: new DragZoom({ condition: () => true }),
       map: null,
+      mapSingleClickListener: null,
       measureTools: [
         {
           button: '#measureLineButton',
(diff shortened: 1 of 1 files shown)


hwiem (commented):
Code lgtm + everything seems to work, but I'm getting this warning in the console:
```
vue.esm.js:5140 [Vue warn]: Failed to resolve async component: ()=>w.e(652).then(w.bind(w,652))
Reason: TypeError: Cannot read properties of undefined (reading 'call')
```

hwiem:
> > Code lgtm + everything seems to work, but I'm getting this warning in the console:
> > ```
> > vue.esm.js:5140 [Vue warn]: Failed to resolve async component: ()=>w.e(652).then(w.bind(w,652))
> > Reason: TypeError: Cannot read properties of undefined (reading 'call')
> > ```
> 
> Strange. Can't reproduce it. Wich Browser did You use?

Chrome. But I can't reproduce it on `main`. :woman_shrugging:

PR #787, opened 2023-02-08
Improve display of statement count

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T31258

Showing both numbers (statement and originalStatement count) by default turned out to be not  understandable. To improve this, original statements are now only shown within a tooltip that is triggered when hovering the numbers. Hiding the original statements by default is the better option as users are mostly interested in this metric.

### How to review/test
Checkout, build the frontend, visit admin area with procedures list.

### PR Checklist


salisdemos (approved):
looks good. Thanks for improving that

PR #788, opened 2023-02-08
Use correct import syntax

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T31169

As we moved the file from demosplan-ui into core again, the way it has to be imported has changed.

### PR Checklist


ahmad-demos (approved):
lgtm, codeReview, thanx!

PR #796, opened 2023-02-09
fix(refs T31291, T31301): Catch undefined Reset measuretools

salisdemos:
after fixing and refactoring the measuretools we have overseen another undefined variable. Now everything should work as expected

**Ticket:** https://yaits.demos-deutschland.de/T31291
https://yaits.demos-deutschland.de/T31301

### PR Checklist


hwiem (approved):
:see_no_evil: one does not simply move some code
lgtm, hope this is the last one

PR #802, opened 2023-02-10
fix(refs T31312): Allow access to hashedQuery via API

dresslerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T31312

The resource type seems to have been non-functional since it was created (during a refactoring). Opening it like it is done in this PR is fine, as it can only be accessed via `UserFilterSetResourceType`, that already contains checks to limit access via permission check, the user that created the filter and the procedure in which the filter was created.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm

PR #803, opened 2023-02-10
fix(refs T29992): disable button if statement is not claimed

hwiem:
**Ticket:** https://yaits.demos-deutschland.de/T29992

- if a user hasn't claimed a statement, they should not be allowed to delete it
- previously, only styling was applied in that case, but deleting the statement was still possible

### How to review/test

Go to the statement list and try to delete a statement that is assigned to you as well as a statement that is not assigned to you. The former should be possible, the latter should not.

### PR Checklist


spiess-demos (approved):
lgtm! (code review only)

PR #805, opened 2023-02-10
fix (refs T31312): use correct access modifier in value object

dresslerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T31308

Properties in classes extending from `ValueObject` must be set to `protected` for the `@method` tag to work.

Also used PHP 8+ syntax for properties to keep the code modern.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm

MoritzMandler (approved):
lgtm

PR #810, opened 2023-02-13
fix (refs T31331): move UI-logic into twig

dresslerdemos:
https://yaits.demos-deutschland.de/T31331

Originally the statement entity was converted into an associative array in the backend and the field
`feedback` overridden with the corresponding
translation value before passing the array into the twig file.

However, `getStatement` did no
longer return the array representation of the statement but the statement object instead, which caused a bug. Returning the object is fine, because
we want to handle objects in the backend and avoid unnecessary conversions.

As it would be very bad practice to write a translation value into the `feedback` property of the (now object) statement (as
it may be persisted by accident after refactorings), the approach was changed and the logic determining flags and translation values was moved into the twig file, as this seems to be a frontend-code responsibility anyway.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/templates/bundles/DemosPlanStatementBundle/DemosPlanAssessment/view_statement.html.twig
+++ b/templates/bundles/DemosPlanStatementBundle/DemosPlanAssessment/view_statement.html.twig
@@ -1,17 +1,10 @@
 {% extends '@DemosPlanCore/DemosPlanCore/base.html.twig' %}
 
-{# if there are statement votes ("Mitzeichner"), then the option "Schlussmitteilung versenden" plus corresponding hint will be shown #}
-{% set templateVars.sendFinalEmail = true %}
-{% set templateVars.finalEmailOnlyToVoters = false %}
+{% set sendFinalEmail = true %}
 {% if templateVars.table.statement.feedback == 'snailmail' %}
-  {% set templateVars.table.statement.feedback = "via.post"|trans %}
   {% if templateVars.table.statement.votes is empty %}
-    {% set templateVars.sendFinalEmail = false %}
-  {% else %}
-    {% set templateVars.finalEmailOnlyToVoters = true %}
+    {% set sendFinalEmail = false %}
   {% endif %}
-{% elseif templateVars.table.statement.feedback == 'email' %}
-    {% set templateVars.table.statement.feedback = "via.email"|trans %}
 {% endif %}
 
 {% set statement = templateVars.table.statement %}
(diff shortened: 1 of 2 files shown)


templates/bundles/DemosPlanStatementBundle/DemosPlanAssessment/view_statement.html.twig
```diff
@@ -1,5 +1,19 @@
...
+{% if templateVars.table.statement.feedback == 'snailmail' %}
+  {% set templateVars.table.statement.feedback = "via.post"|trans %}
+  {% if templateVars.table.statement.votes is empty %}
+    {% set templateVars.sendFinalEmail = false %}
+  {% else %}
+    {% set templateVars.finalEmailOnlyToVoters = true %}
+  {% endif %}
+{% elseif templateVars.table.statement.feedback == 'email' %}
+    {% set templateVars.table.statement.feedback = "via.email"|trans %}
+{% endif %}
```
salisdemos, line 15: Please don't use `templateVars` for the vars not coming from the backend anymore.

salisdemos (changes requested):

templates/bundles/DemosPlanStatementBundle/DemosPlanAssessment/view_statement.html.twig
graupnerdemos, line 15: I don't quite understand why this has to happen in Twig. We want to reduce the amount of Twig templating overall and move towards Vue. Yes this is simple logic, but still, logic in Twig is less readable and less performant than logic in PHP.

dresslerdemos: See commit message.

dresslerdemos: Was necessary to avoid refactoring of this twig file and its `include`s.

salisdemos: there are not so much occurrences. And I don't wan't to reduce it technical dept in the controller but produce it here

dresslerdemos: If it can be simplified I'm totally fine with that. @spiess-demos refactored the file, please review.

dresslerdemos: If it can be simplified I'm totally fine with that. @spiess-demos refactored the file, please review.

spiess-demos:
the template in question isn't, nor has ever been, a representation of best practices (many of its contents have been initially copied from the statement detail page without proper understanding of the reasoning behind things).

That's why i suggest to just give the changes a go.

MoritzMandler (approved):
for me this is fine

PR #824, opened 2023-02-14
feature (refs T30295): allow config file for customer creation

dresslerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T30295

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Command/Data/GenerateCustomerCommand.php
+++ b/demosplan/DemosPlanCoreBundle/Command/Data/GenerateCustomerCommand.php
@@ -13,53 +13,107 @@ declare(strict_types=1);
 namespace demosplan\DemosPlanCoreBundle\Command\Data;
 
 use demosplan\DemosPlanCoreBundle\Command\CoreCommand;
+use demosplan\DemosPlanCoreBundle\DependencyInjection\Configuration\CustomerConfiguration;
 use demosplan\DemosPlanCoreBundle\Entity\User\Customer;
 use demosplan\DemosPlanCoreBundle\Exception\EntryAlreadyExistsException;
+use demosplan\DemosPlanCoreBundle\Exception\ViolationsException;
+use demosplan\DemosPlanCoreBundle\Utilities\DemosPlanPath;
 use demosplan\DemosPlanUserBundle\Repository\CustomerRepository;
 use Exception;
+use InvalidArgumentException;
+use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\QuestionHelper;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\Question;
 use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+use Symfony\Component\Yaml\Yaml;
+use function in_array;
+use function is_string;
 
 class GenerateCustomerCommand extends CoreCommand
 {
+    private const CONFIG_OPTION = 'config';
+
     protected static $defaultName = 'dplan:data:generate-customer';
     protected static $defaultDescription = 'Creates a new customer';
+
+    protected QuestionHelper $helper;
+
     /**
-     * @var QuestionHelper
+     * @var list<string>
      */
-    protected $helper;
+    private array $reservedNames;
+
     /**
-     * @var CustomerRepository
+     * @var list<string>
      */
-    private $customerRepository;
+    private array $reservedSubdomains;
 
     public function __construct(
-        CustomerRepository $customerRepository,
+        private readonly CustomerRepository $customerRepository,
+        private readonly ValidatorInterface $validator,
         ParameterBagInterface $parameterBag,
         string $name = null
     ) {
         parent::__construct($parameterBag, $name);
-        $this->customerRepository = $customerRepository;
         $this->helper = new QuestionHelper();
+
+        $existingCustomers = $customerRepository->findAll();
+
+        $this->reservedNames = array_map(
+            static fn (Customer $customer): string => $customer->getName(),
+            $existingCustomers
+        );
+        $this->reservedSubdomains = array_map(
+            static fn (Customer $customer): string => $customer->getSubdomain(),
+            $existingCustomers
+        );
     }
 
+    protected function configure(): void
+    {
+        parent::configure();
+
+        $this->addUsage('foobar'); // FIXME
+
+        $this->addOption(
+            self::CONFIG_OPTION,
+            'c',
+            InputOption::VALUE_REQUIRED,
+            'Provide a YAML config file containing properties to be set in the generated customer. If a config is given it must contain the required parameters (\'name\' and \'subdomain\'). If no config is given the parameters will be asked interactively.'
+        );
+    }
+
+    /**
+     * @throws Exception
+     */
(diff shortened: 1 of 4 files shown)


graupnerdemos (commented):
I'd have figured we'd just add arguments here and pass the required options in command calls from the related setup script. While this is a nice solution it seems more complex than necessary. I may however be overlooking something.

tests/backend/core/Core/Functional/res/customerConfig1.yaml
```diff
@@ -0,0 +1,2 @@
+name: 'Foobar'
```
mussbach, line 1: shouldn't this be an array of customers?

dresslerdemos: Not that I know of. As far as I understood the action to create a customer is very rare, so I assumed the config file would be created for that circumstance and deleted afterwards.

I can refactor it to accept an array if you want.

mussbach: I see. This is a legit implementation, I just thought that this configuration file keeps all customers. This way it might be a bit more resilient and easier to deploy

dresslerdemos: That seems to deviate from "create a customer via config" to "manage customers via config". E.g. if I would see such config for the first time and be told I can add customers to it and they will automatically be created, I'd assume that I can simply remove an entry in the config to remove a customer from the system.

mussbach: Indeed. At least it should not be allowed to be used any more (aka new subdomains_allowed)

dresslerdemos:
State of this PR:

In theory this PR already contains the logic to solve the story and when in a pinch it can be reviewed. The reason why I don't yet recommend it, is that it contains an old state of https://github.com/demos-europe/demosplan-core/pull/854, which may make the review harder than necessary.

https://github.com/demos-europe/demosplan-core/pull/854 has currently problems in receiving information from SonarCloud and is thus blocked. But even if it was merged it would result in (even more) merge conflicts between `main` and this PR, which may also not be helpful for the review.

dresslerdemos:
Conflicts were solved. A proper code review can be done now. The SonarCloud concern is not serious and can be solved by a kindhearted person during the review or will be solved by me at some later point otherwise.

mussbach (commented):
some smaller first remarks

tests/backend/core/Core/Functional/res/customerConfig1.yaml
dresslerdemos: Considering the remaining time I suggest to stick to the current approach for now. We can still add improvements with followup stories.

demosplan/DemosPlanUserBundle/Logic/CustomerService.php
```diff
@@ -12,27 +12,24 @@
...
-    public function __construct(CustomerRepository $customerRepository, GlobalConfigInterface $globalConfig)
-    {
-        $this->customerRepository = $customerRepository;
-        $this->globalConfig = $globalConfig;
+    public function __construct(
+        private readonly CustomerRepository $customerRepository,
+        private readonly GlobalConfigInterface $globalConfig,
+        private readonly ValidatorInterface $validator,
+        private readonly TransactionService $transactionService,
+        private readonly UserRepository $userRepository
```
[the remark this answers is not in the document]
dresslerdemos: done 739ea36eed8ef2be3d06f4eb15db264a211653c8

demosplan/DemosPlanCoreBundle/Command/Data/GenerateCustomerCommand.php
```diff
@@ -13,53 +13,94 @@
...
-        $customer = new Customer($name, $subdomain);
+        $subdomain = $input->getOption(self::OPTION_SUBDOMAIN);
+        if (null === $subdomain) {
+            $subdomain = $this->askSubdomain($input, $output);
+        }
 
         try {
             // create customer
-            $this->customerRepository->updateObject($customer);
+            $customer = $this->customerService->createCustomer($name, $subdomain);
```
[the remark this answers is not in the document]
dresslerdemos: done 739ea36eed8ef2be3d06f4eb15db264a211653c8

demosplan/DemosPlanCoreBundle/Command/ContainerInitCommand.php
```diff
@@ -50,17 +76,127 @@ public function configure(): void
...
+                    }
+
+                    $user = $this->userRepository->getFirstUserByCaseInsensitiveLogin($userLogin);
+                    if (null === $user) {
+                        $user = $this->userService->createMasterUserForCustomer(
+                            $userLogin,
+                            $customer
+                        );
+                    }
+
```
MoritzMandler, line 149: does this user not need an organisation as well?

dresslerdemos: As far I was told this user is only needed for technical reasons (if at all) and thus should not need an organisation. The plan is to test it without creating the user (this is why the `userLogin` field in the configuration is optional). The  code for the creation of the user is only left in in case it does not work otherwise and can be removed if it is not needed.

demosplan/DemosPlanCoreBundle/Command/Data/GenerateCustomerCommand.php
```diff
@@ -13,53 +13,94 @@
...
+        $this->reservedNames = array_column($reservedCustomers, 0);
+        $this->reservedSubdomains = array_column($reservedCustomers, 1);
+    }
+
+    protected function configure(): void
+    {
+        parent::configure();
+        $this->addOption(
+            self::OPTION_NAME,
+            'i',
```
mussbach, line 70: why `i`?

demosplan/DemosPlanUserBundle/Logic/UserService.php
```diff
@@ -1002,6 +1004,19 @@ public function collectRoleStatistics(array $users): array
...
     }
 
+    public function createMasterUserForCustomer(string $userLogin, Customer $customer): User
+    {
+        return $this->userRepository->add([
+            'firstname'    => '',
+            'lastname'     => '',
+            'email'        => $userLogin,
+            'login'        => $userLogin,
+            'password'     => $this->generateRandomString(128, self::HEX_CHARS),
```
mussbach, line 1014: Why not just `bin2hex(random_bytes(15))`? 128 seems a bit heavy btw.

mussbach (changes requested):
looks pretty good overall

demosplan/DemosPlanCoreBundle/Command/Data/GenerateCustomerCommand.php
dresslerdemos: `n` was taken for some reason, `i` didn't feel worse than any of the other 24 choices.

demosplan/DemosPlanUserBundle/Logic/UserService.php
dresslerdemos: I didn't know about `random_bytes`. The length was chosen to match the current length of the hashed passwords.

dresslerdemos: I didn't know about `random_bytes` and can rework it to use that. I'm not sure about the length though, it was chosen to match the current length of the hashed passwords.

mussbach: I see. I would suggest to simplify it with `random_bytes` to reduce maintenance costs, choosing the length is up to you, it might be even helpful to not use the exact same length to have a visual marker that this is no real password hash, but I will not insist to the length

mussbach (approved):
could be improved later on with easier password mechanism

PR #825, opened 2023-02-14
bug(refs: T29346): fix map modal is not toggling

muellerdemos:
https://yaits.demos-deutschland.de/T29346

### Description
- In AssessmentTableCard the reference to the modal was not correct. To make this work it would be needed to reference the modal way deeper (this.$parent.$parent.$parent.$parent... would be the correct reference). However this seemes overly complicated to follow if we need to change this component anytime in the future.

- Hence an approach by emitting a custom event which in the end toggles the modal was a better solution from my perspective.

- DpMapModal listens for the event toggleModal which will actually toggle the modal and the event gets fired in DpAssessmentTableCard

### How to review/test

- login as fpa
- navigate to a-table
- click the marked button (screenshot)
- map modal should open

### PR Checklist

Screenshot 2023-02-14 at 18-58-47 ⚓ T29346 AT Verortete STN Einzeichnung - Ansehen funktioniert nicht


salisdemos (approved):

salisdemos:
I just wonder if the bug is really old or why it appeared now.

client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
```diff
@@ -1175,7 +1175,7 @@ export default {
     },
 
     toggleMapModal (drawingData) {
-      this.$parent.$refs.mapModal.toggleModal(drawingData)
+      this.$root.$emit('toggleModal', drawingData)
```
spiess-demos, line 1178: I'd suggest to use a lightly more specific event name (which includes the term "drawing" pr "map").

spiess-demos (changes requested):

client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
muellerdemos: Makes sense, was done here [219dd87](https://github.com/demos-europe/demosplan-core/pull/825/commits/219dd876358fed66405aac467719509d21f339fa)

muellerdemos:
@salisdemos I'm pretty sure this bug was just unnoticed for a longer period since the reference could simply not work.

spiess-demos (approved):
lgtm!

PR #827, opened 2023-02-15
fix: fixtures should use proper Dependency Injection instead of container

mussbach:
The better we get in using dependency injection and private services the more important it gets to drop the usage of the container an the dynamic service fetching. 
The Fixtures did not use DI in every cases and thus the tests where broken

### How to review/test
Run any test it should not complain about inlined or removed services

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/DataFixtures/ORM/TestData/LoadUserData.php
+++ b/demosplan/DemosPlanCoreBundle/DataFixtures/ORM/TestData/LoadUserData.php
@@ -21,6 +21,7 @@ use demosplan\DemosPlanCoreBundle\Entity\User\OrgaType;
 use demosplan\DemosPlanCoreBundle\Entity\User\Role;
 use demosplan\DemosPlanCoreBundle\Entity\User\User;
 use demosplan\DemosPlanUserBundle\Logic\OrgaService;
+use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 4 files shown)


demosplan/DemosPlanCoreBundle/DataFixtures/ORM/TestData/LoadUserData.php
```diff
@@ -116,6 +117,14 @@ class LoadUserData extends TestFixture
...
 
     public const DEFAULT_PW_HASH = '2308912f941bd13be84578fa73877453f0e3eef455e6674aee57c78cd354fda94087762e9ea4d5e7a56a330f824edf7d125950ce7e5fda7080b34fe1836a0b4e';
+    private OrgaService $orgaService;
+
+
+    public function __construct(EntityManagerInterface $entityManager, OrgaService $orgaService)
+    {
+        parent::__construct($entityManager);
+        $this->orgaService = $orgaService;
+    }
```
dresslerdemos, line 126:
```suggestion
    public function __construct(EntityManagerInterface $entityManager, private readonly OrgaService $orgaService)
    {
        parent::__construct($entityManager);
    }
```

dresslerdemos (approved):
totally mergeable, just one potential improvement

PR #833, opened 2023-02-16
feat (refs T31375): use env var to connect to rabbitmq

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T31375

Rabbitmq parameters are consolidated into `rabbitmq_dsn` to be able to use a single environment variable `RABBITMQ_URL` to define the rabitmq dsn.

### How to review/test
Define a valid `RABBITMQ_URL` env var in your `.env.local` and e.g. import a docx file

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Resources/config/config_dev_container.yml
+++ b/demosplan/DemosPlanCoreBundle/Resources/config/config_dev_container.yml
@@ -40,7 +40,6 @@ parameters:
         - '*.demos-europe.eu'
         - 'cdnjs.cloudflare.com'
 
-    rabbitmq_host: "mq"
     rabbitmq_routing_disabled: true
 
     gateway_url: '/app_dev.php/dplan/login'
--- a/.env
+++ b/.env
@@ -55,7 +55,7 @@ JWT_PASSPHRASE=0699e1b59a14e6fec876c06cb8279190
 ###< lexik/jwt-authentication-bundle ###
 
 ###> php-amqplib/rabbitmq-bundle ###
-RABBITMQ_URL=amqp://guest:guest@localhost:5672
+# You need to define RABBITMQ_URL in your .env.local
 ###< php-amqplib/rabbitmq-bundle ###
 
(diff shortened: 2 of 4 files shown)


breierDemos (approved):
lgtm

demosplan/DemosPlanCoreBundle/Resources/config/config_dev_container.yml
```diff
@@ -40,7 +40,6 @@ parameters:
         - '*.demos-europe.eu'
         - 'cdnjs.cloudflare.com'
 
-    rabbitmq_host: "mq"
```
graupnerdemos, line 43: Shouldn't we set the `rabbitmq_dsn` parameter for the container setup in this file now?

graupnerdemos (changes requested):
I warmly welcome the overlords of the dsn service configuration future. One question though.

demosplan/DemosPlanCoreBundle/Resources/config/config_dev_container.yml
mussbach: No, we cannot, as this file is checked in into the public git

graupnerdemos: That reasoning doesn't exactly match with database, elasticsearch and mailhog being configured here too. 🤔

mussbach: But they are configured without users an passwords. Rabbitmq dns would contain both

mussbach: the variable `rabbitmq_host` does not exist any more

graupnerdemos: Well. We need another seamless solution soon then I guess.

graupnerdemos (approved):

PR #853, opened 2023-02-20
build: Disallow discovery composer plugin

graupnerdemos:
Since we're not adding dependencies very often
and do our own research on what they need when we
configure them, there is no need for this
plugin to run and check if it can help us.

> Since 1.15.0, this library also provides a composer plugin that automatically installs well-known PSR implementations if composer dependencies require a PSR implementation but do not specify which implementation to instal

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
lgtm, code review only

PR #863, opened 2023-02-21
feat: use relative path in addon to be compatible with different environments (2)

martendemos:
Kind of fix of #862


mussbach (changes requested):
you would not find the addon under this path

PR #873, opened 2023-02-23
fix (refs T31518): Use Event that is triggered on manual login

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T31518

Field `last_login` was not updated any more after upgrading the symfony authentication system

### How to review/test
login via login form and the field should be updated

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm

PR #877, opened 2023-02-23
bug(refs T23632): enable automatic name extraction in bimschg antrag x

muellerdemos:
https://yaits.demos-deutschland.de/T23632

- To enable the communication between the upload component and DpNewProcedure, which inherits the AntragX upload component, a store has been introduced to dynamically determine if the name field is required when creating a new procedure. This field is only required when no AntragX file is being uploaded since the procedure name will be extracted from there. The problem is that the name can only be extracted after submitting the form but to submit the form the name field was required.

- Now the store holds the state weather or not the name field is required. The decision for a store has been made because I simply do not like to $root emit events since this is an easy way to lose the ability to read easily through the code and also the $root pattern is not recommended to use since it can be dangerous to use it due to being prone to unwanted behavior in the application.

- Now the vuex store is the single source of truth weather the name field is required or not.

- The form field is only required when not uploading an AntragX file since the name will be extracted from it after submitting the form and the name input would be overwritten anyways so it was simply not useful to have the requirement to add a name when uploading an AntragX.

### How to review/test

- Create a new procedure in bimschgsh
- Upload a AntragX file
- When uploaded the name field should not be required anymore since the name will be extracted from the uploaded file
- When deleting the uploaded file without submitting, the name field should be set to required automatically

### Linked PRs (optional)
- [This is the addon PR](https://github.com/demos-europe/demosplan-addon-bimschg-antrag/pull/3) 

### DIFF
--- a/client/js/store/procedure/NewProcedure.js
+++ b/client/js/store/procedure/NewProcedure.js
@@ -18,9 +18,9 @@ const NewProcedure = {
 
   mutations: {
     /**
-     *
+     * Sets required attribute of a form field dynamically to true/false if the attribute is bound to requireField.
      * @param state {Object}
-     * @param value {Boolean} Sets required form field dynamically to true/false.
+     * @param value {Boolean}
      */
     setRequiredField (state, value) {
       state.requireField = value
--- a/client/js/store/procedure/NewProcedure.js
+++ b/client/js/store/procedure/NewProcedure.js
@@ -1,0 +1,30 @@
+/**
+ * (c) 2010-present DEMOS E-Partizipation GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+const NewProcedure = {
+  namespaced: true,
+
+  name: 'newProcedure',
+
+  state: {
+    requireField: true
+  },
+
+  mutations: {
+    /**
+     *
+     * @param state {Object}
+     * @param value {Boolean} Sets required form field dynamically to true/false.
+     */
+    setRequiredField (state, value) {
+      state.requireField = value
+    }
+  },
+}
+export default NewProcedure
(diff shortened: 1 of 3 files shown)


elmasdemos (approved):
lgtm, code review

client/js/store/procedure/NewProcedure.js
```diff
@@ -0,0 +1,30 @@
...
+
+  state: {
+    requireField: true
+  },
+
+  mutations: {
+    /**
+     *
+     * @param state {Object}
+     * @param value {Boolean} Sets required form field dynamically to true/false.
```
hwiem, line 23: Nitpick: Maybe we could use this comment as a doc block for the mutation (and remove it from the parameter), adjusting it like so:
"Sets `required` attribute of a form field dynamically to true/false if the attribute is bound to `requireField`"?

client/js/store/procedure/NewProcedure.js
```diff
@@ -0,0 +1,30 @@
...
+     */
+    setRequiredField (state, value) {
```
hwiem, line 25: Where is this used?

client/js/store/procedure/NewProcedure.js
muellerdemos: Thanks, I think this is a good idea, done in [b90261c](https://github.com/demos-europe/demosplan-core/pull/877/commits/b90261cccb01b164946b8d07bd379c2fc1bc42b8)

client/js/store/procedure/NewProcedure.js
muellerdemos: This mutation is being called in the bimschg addon in `BimschgAntrag.vue` which is loaded via the addon wrapper in `DpNewProcedure.vue`

hwiem (approved):
lgtm, code review only

PR #890, opened 2023-02-24
Minor Webpack Refactorings

graupnerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T30822

These are some minor refactorings as part of some efforts to make the frontend stack configuration more readable and increase maintainability. The general idea (in following PRs) is to restructure the config and `fe` helper into more meaningful modules and remove parts we don't (need to) use anymore.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


spiess-demos (commented):
When running `yarn dev:<project>`, i get a bunch of 500s originating from all non-legacy js bundles being included directly from root, as seen here:

image

graupnerdemos:
@spiess-demos That is an interesting fallback behaviour which could be labeled a bug in the backend. However, it is fixed now. I forgot to adjust the lookup path to the new directory structure.

spiess-demos (approved):
Works & looks good to me now.

PR #905, opened 2023-03-01
bug(refs T30632): Create baselayers always with layersource

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T30632

Creating the layersource was removed for hidden layers because of performance, but we also need the layer sources for hidden layers. Otherwise when updating the map fields right after loading the page (eg with drawing in the map) we get an error, because not all print layers have a source.

### How to review/test
Login as FPA and create a layer with print, but otherwise hidden and one that is not hidden (under Verfahren -> Planunterlagen und Planzeichnung -> Kartenebenen definieren).

layers

Then login as Privatperson and go to the public site of the procedure. You should be able to draw something in the map immediately without having to change the layers first.


spiess-demos (approved):
Lgtm (code review).

PR #908, opened 2023-03-01
bug (refs T31604): fix occurrences of "Web-Atlas" in blueprints by replacing them with "basemap"

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T31604

Description: "Web Atlas" was not replaced with "basemap" in all procedure templates. 
 This was due to the fact that different urls are used for "Web Atlas", but only one was checked in https://github.com/demos-europe/demosplan-core/pull/578.
 Anyway, this PR should catch every url with "webatlas" in it.

### How to review/test
First: Find a older procedure template with a "Web-Atlas" as map (hint: not all have the name "Web-Atlas", so check if the url contains "webatlas".)
Then execute the migration.
And finally check the founded procedure template again. You should don't find any occurrences of "webaltas" in other procedure templates.

You could also check the database. In the ticket is a statement to find occurrences.

### PR Checklist

### DIFF
--- a/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/03/Version20230301094921.php
+++ b/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/03/Version20230301094921.php
@@ -38,7 +38,7 @@ class Version20230301094921 extends AbstractMigration
     {
         $this->abortIfNotMysql();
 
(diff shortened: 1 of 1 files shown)


mussbach (approved):
lgtm, thx

PR #909, opened 2023-03-01
Add styles to fix layout jump

spiess-demos:
Instead of adding a margin to the expanded item to show its box shadow (which otherwise is overlapped from following items), positioning it relative solves this problem, too. The negative left margin and reduced right padding ensure the item stays within its vertical axis even with added box-shadow.

### How to review/test
 Login as a support user in projects that use `DpTableCard`, and click on a card. It should stay in place also in expalded state.


muellerdemos:
Clean solution, nice - lgtm

muellerdemos (approved):

PR #923, opened 2023-03-02
chore (refs T31667): change import files from demosplan-utils to demo…

ahmad-demos:
https://yaits.demos-deutschland.de/T31667

- change import files from demosplan-utils to demosplan-ui.
- changes in components of Core.
- also the changes in JS Files, package.json & config.


spiess-demos:
At the moment i get a conole error like 
```
DPVueCore.js:23 Uncaught TypeError: (0 , _demos_europe_demosplan_ui__WEBPACK_IMPORTED_MODULE_1__.hasOwnProp) is not a function
    at Object.install (DPVueCore.js:23:111)
```

ahmad-demos:
> At the moment i get a conole error like
> 
> ```
> DPVueCore.js:23 Uncaught TypeError: (0 , _demos_europe_demosplan_ui__WEBPACK_IMPORTED_MODULE_1__.hasOwnProp) is not a function
>     at Object.install (DPVueCore.js:23:111)
> ```

@spiess-demos The problem only in this file, I tried deleting `hasOwnProp` from this file, and re-run the Project, everything works fine, `hasOwnProp` in another place works also fine, I couldn't find the problem.
But with this solution https://github.com/demos-europe/demosplan-core/pull/923/commits/5325e4636b4fd20568fde5255114b221ae4d2a86 I was able to solve the problem, and I tried log `hasOwnProp` in within async function, then works also.
Maybe @graupnerdemos  has a different opinion! Or the reason for this problem only in this file!

spiess-demos:
@ahmad-demos would you try if [the latest changes to the demosplan-ui build](https://github.com/demos-europe/demosplan-ui/pull/114) already fix this issue?

ahmad-demos:
> @ahmad-demos would you try if [the latest changes to the demosplan-ui build](https://github.com/demos-europe/demosplan-ui/pull/114) already fix this issue?

No, unfortunately not!

PR #928, opened 2023-03-03
refactor (refs T31722): add missing exception causes

dresslerdemos:
This does not fix the referenced bug, but helps to find the cause of similar ones in the future.

**Ticket:** https://yaits.demos-deutschland.de/T31722

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
lgtm, thx a lot

PR #929, opened 2023-03-03
Fix(refs:T31701): platform settings are not loading

elmasdemos:
<!-- **Ticket:** https://yaits.demos-deutschland.de/T31701

Description: in the `getRequestPayload` in the return it should be `include` instead of `includes` otherwise there are no relationships in the payload and we get undefined for the requested attributes

### How to review/test
- login as support and go to the platform settings

### PR Checklist


hwiem (approved):
lgtm, code review only

PR #933, opened 2023-03-06
fix (refs T31730): cc email addresses may be a list of emails

mussbach:
Thus, validation is not possible in this case.

### How to review/test
code review is probably best, as login via OSI with an organisation with an emaillist as `$ccEmail2` would trigger the case

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Entity/User/Orga.php
+++ b/demosplan/DemosPlanCoreBundle/Entity/User/Orga.php
@@ -104,10 +104,10 @@ class Orga extends SluggedEntity
     protected $modifiedDate;
 
     /**
+     * Comma separated list of cc-Email addresses
      * @var string|null
      *
      * @ORM\Column(name="_o_cc_email2", type="string", length=4096, nullable=true)
-     * @Assert\Email(message="email.address.invalid")
      */
     protected $ccEmail2;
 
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Entity/User/Orga.php
```diff
@@ -99,15 +107,17 @@ class Orga extends SluggedEntity
...
     protected $modifiedDate;
 
     /**
+     * Comma separated list of cc-Email addresses.
+     *
      * @var string|null
      *
      * @ORM\Column(name="_o_cc_email2", type="string", length=4096, nullable=true)
-     * @Assert\Email(message="email.address.invalid")
      */
```
martendemos, line 121: Is there another way to validate multiple email-addresses?
E.g. split them an loop for each?
If so, I would appreciate a follow up Ticket to create this kind of validation of multiple email-addresses.

dresslerdemos: I think the best way would be to refactor this into a to-many relationship to the `EmailAddress` entity and validate the items via `All`.

mussbach: https://yaits.demos-deutschland.de/T31792

PR #951, opened 2023-03-07
f implement split statement view with dynamic props

muellerdemos:
https://yaits.demos-deutschland.de/T31759

### Can only be tested after demospipes addon was installed
- Since we need the split statement view for enabling the demospipes addon, this commit adds the addon wrapper to load the split statement view
- Also passing the required props via twig into the addon
- Adds AddonWrapper to entrypoint

### DIFF
--- a/templates/bundles/DemosPlanProcedureBundle/DemosPlanProcedure/administration_split_statement.html.twig
+++ b/templates/bundles/DemosPlanProcedureBundle/DemosPlanProcedure/administration_split_statement.html.twig
@@ -7,6 +7,19 @@
         {% set submitTypeOptions = submitTypeOptions|merge([{ label: label, value: value }]) %}
     {% endfor %}
 
+    {% set addonProps = {
+        'procedureId': procedure,
+        'statementId': statementId,
+        'statementExternId': statementExternId,
+        'procedureId': procedure,
+        'procedureName': procedureName,
+        'title': title
+        }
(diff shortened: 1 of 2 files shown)


templates/bundles/DemosPlanProcedureBundle/DemosPlanProcedure/administration_split_statement.html.twig
```diff
@@ -7,6 +7,19 @@
         {% set submitTypeOptions = submitTypeOptions|merge([{ label: label, value: value }]) %}
     {% endfor %}
 
+    {% set addonProps = {
+        'procedureId': procedure,
+        'statementId': statementId,
+        'statementExternId': statementExternId,
+        'procedureId': procedure,
```
elmasdemos, line 14:
```suggestion
        'procedureId': procedure,
        'statementId': statementId,
        'statementExternId': statementExternId,
```

procedureId is double

muellerdemos: Oh thanks for catching this, removed the duplication here [eeb1237](https://github.com/demos-europe/demosplan-core/pull/951/commits/eeb1237e917186797e6b605faac0596739f7fcc9)

elmasdemos (dismissed):
thanks, lgtm, code review

PR #957, opened 2023-03-08
chore: define symfony packages explicitly

mussbach:
Those packages where automatically updated to symfony 6. To be safe we should use Symfony 5 versions instead

### How to review/test
review compser.json

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


graupnerdemos (approved):
Thanks.

PR #970, opened 2023-03-09
feat: provide default values for database

mussbach:
Provide default values to `parameters_default.yml` to be able to build container.

### How to review/test
Code review only, should not be able to break anything

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


graupnerdemos (commented):
This is just so that we can build a service container without creating a valid parameters.yml, right? I'm a little bit torn on the documentation effect of setting the database password empty here. In other software, I've seen bogus values like `replace-me-with-a-safe-password` there, just to remind devs that empty is not an option. However, I could be okay with leaving it like that, too.

mussbach:
Actually I did not paste any password to not receive a shout of sonarcloud. But I see your point

PR #972, opened 2023-03-09
Bug(refs: T31791): fix loadtime not in form request

muellerdemos:
https://yaits.demos-deutschland.de/T31791

- The loadtime is required for the flood control and expects a unix timestamp.
- Unfortunately the loadtime was not set when submitting the new statement form as a guest user.
- Since the form data is set via the vuex store, a mutation has to be added to the form submit function.
- The form data gets reset after submit so it is fine to simply set the loadtime via a mutation but not resetting it manually again.

### How to review/test

- Checkout branch
- Use the statement modal on public detail as guest user
- Verify the form field `r_loadtime` has been set correctly (is not empty when submitting the form)


gruenbergerdemos (approved):
lgtm!

PR #974, opened 2023-03-09
feat: create OpenAPI Data only on development builds

mussbach:
OpenAPT specification only needs to be created on development builds. This also may help in Problems with production asset dumps

### How to review/test
code review might be best

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


graupnerdemos (approved):
We can do this for now. However, since we're also using the OpenAPI output to generate resource information to-be-used in the FE code, this will have to be adjusted at some point. Since the latter is not in use yet, this is fine.

PR #1018, opened 2023-03-17
fix(refs:T31922): expose permission for frontend

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T31922

When refactoring the view to Vue [sic!] that was obviously overlooked. 

Also, some minor spacing tweaks are added (this should do nothing but add the correct and same space between the save/abort button and the preceeding ui elements, with or without the second notification.

### How to review/test
After logging into ewm as Fachplanung-Admin, when creating a new procedure it should be possible to input a token.

### PR Checklist


gruenbergerdemos (approved):
lgtm!

PR #1048, opened 2023-03-22
Remove leftovers from moving utils to demosplan-ui

spiess-demos:
Some occurrences of "demosplan-utils" were forgotten when moving it to demosplan-ui in #923. 
These are removed or changed here.


gruenbergerdemos (approved):
lgtm!

PR #1049, opened 2023-03-22
bug(refs: T31984): prevent unintentional deletion of editor text

muellerdemos:
[Ticket](https://yaits.demos-deutschland.de/T31984)

This PR prevents the user from unintentionally deleting an unsaved text by synchronizing the local statement in `StatementMeta.vue` (which also emits the local statement when saving only metadata) with the statements from store. The editor automatically updates the state of statements in the store when registering an input. This only occurs when a statement has not been segmented already.

### How to review/test
- navigate to detail view of statement
- select statement without segments
- click edit and claim statement
- edit the text of the editor
- use the upper save button for metadata
- reload page and confirm that the text in the editor was saved too
- check if a statement with segments still works like before 

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
@@ -538,10 +538,27 @@ export default {
       this.setContent({ prop: 'slidebar', val: { isOpen: false, showTab: '', segmentId: '' } })
     },
 
+    /**
+     * This prevents the user from unintentionally deleting an unsaved text by synchronizing the local
+     * statement in StatementMeta.vue (which also emits the local statement when saving only metadata)
+     * with the statements from store. The editor automatically updates the state of statements in the
+     * store when registering an input. This only occurs when a statement has not been segmented already.
+     *
+     * @param {object} statement - The local statement of StatementMeta.vue.
+     */
+    canUpdateStatementAttributeFullText (statement) {
+      if (statement.attributes.fullText !== this.statement.attributes.fullText) {
+        statement.attributes.fullText = this.statement.attributes.fullText
+      }
+    },
+
     saveStatement (statement) {
+      this.$on('save', this.canUpdateStatementAttributeFullText(statement))
+
       // The key isManual is readonly, so we should remove it before saving
       delete statement.attributes.isManual
       this.setStatement({ ...statement, id: statement.id })
+
       this.saveStatementAction(statement.id)
         .then(() => {
           dplan.notify.notify('confirm', Translator.trans('confirm.saved'))
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
@@ -420,20 +420,6 @@ export default {
       'toggleSlidebarContent'
     ]),
 
-    /**
-     * This prevents the user from unintentionally deleting an unsaved text by synchronizing the local
-     * statement in StatementMeta.vue (which also emits the local statement when saving only metadata)
-     * with the statements from store. The editor automatically updates the state of statements in the
-     * store when registering an input. This only occurs when a statement has not been segmented already.
-     *
-     * @param {object} statement - The local statement of StatementMeta.vue.
-     */
-    canUpdateStatementAttributesFullText (statement) {
-      if (statement.attributes.fullText !== this.statement.attributes.fullText) {
-        statement.attributes.fullText = this.statement.attributes.fullText
-      }
-    },
-
     checkStatementClaim () {
       if (this.statementClaimChecked === false) {
         this.statementClaimChecked = true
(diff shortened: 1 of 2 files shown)


client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
```diff
@@ -539,9 +553,11 @@ export default {
     },
 
     saveStatement (statement) {
+      this.$on('save', this.canUpdateStatementAttributesFullText(statement))
```
hwiem, line 556: Why is `this.$on('save' ...)` needed? Isn't `saveStatement` called `@save` anyway?

client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
```diff
@@ -420,6 +420,20 @@ export default {
...
 
+    /**
+     * This prevents the user from unintentionally deleting an unsaved text by synchronizing the local
+     * statement in StatementMeta.vue (which also emits the local statement when saving only metadata)
+     * with the statements from store. The editor automatically updates the state of statements in the
+     * store when registering an input. This only occurs when a statement has not been segmented already.
+     *
+     * @param {object} statement - The local statement of StatementMeta.vue.
+     */
+    canUpdateStatementAttributesFullText (statement) {
```
hwiem, line 431: Why did you choose this name for the method? I'd rather expect it to be something like `synchronizeFulltext`

hwiem (changes requested):
Side note: the commit type for bug fixes is `fix`, not `bug` (otherwise one might think you're introducing a bug ;))

client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
```diff
@@ -580,6 +582,20 @@ export default {
...
+    /**
+     * This prevents the user from unintentionally deleting an unsaved text by synchronizing the local
+     * statement in StatementMeta.vue (which also emits the local statement when saving only metadata)
+     * with the statements from store. The editor automatically updates the state of statements in the
+     * store when registering an input. This only occurs when a statement has not been segmented already.
+     *
+     * @param {object} statement - The local statement of StatementMeta.vue.
+     */
+    synchronizeFullText (statement) {
+      if (statement.attributes.fullText !== this.statement.attributes.fullText) {
```
hwiem, line 594: After looking into it a little deeper: This means that when clicking the meta data save button, the statement text is also saved? What was the reasoning behind choosing this solution?

muellerdemos: Because the view contains both scenarios if a statement has segments and if not (the editor only appearing if the statement has no segments). To handle this, the StatementMeta form and the rest was decoupled. Also there is a check to minimalize the request payload by checking if a value has changed and only data which needs to be updated will be part of the request payload.

The problem now is that the `StatementMeta.vue` determines which data needs to be in the request payload by a `local statement ` data prop which contains only initial data. Like that, the store state of `statements` will never contain the updated text field but will reset unsaved changes due to a store reset action within `saveStatements`.

I did not want to refactor this whole view or break all the logic which I find useful indeed. Hence I tried to solve the main problem of a resetted unsaved text but keep all the rest still functional. This is also a candidate for a bigger refactoring imho.

hwiem: > This is also a candidate for a bigger refactoring imho.

Agreed!

For now, would it maybe make sense to use a `dpconfirm` to let the user decide if they want to save the text, too, or not? So if they cancel, nothing is saved and they can reset the statement text if desired before saving the meta data. I don't know if I'm overcomplicating things, but as a user, I would not expect the text to be saved with the meta data save button, so I think we should somehow let them know.

muellerdemos: This is a great idea, this reduces also the possible confusion a user maybe experience with those multiple save buttons in one view.

muellerdemos: Done here (finally): https://github.com/demos-europe/demosplan-core/pull/1049/commits/0582115898dae3a08cef28dc0e889f1976d4a908

hwiem: Isn't it the statement text instead of the recommendation text? I'd suggest something like "Sie haben Änderungen am Text der Stellungnahme vorgenommen. Wenn Sie fortfahren, werden diese ebenfalls gespeichert."

muellerdemos: True, I confused myself because the browser tab displayed "Erwiderungen verfassen" , [a493c54](https://github.com/demos-europe/demosplan-core/pull/1049/commits/a493c543c39cb64227ec4c24f505c74479846c24)

hwiem (approved):
lgtm, code review only

PR #1061, opened 2023-03-24
F T32013 submitters of the same statement are not saved

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T32013

Description: submitters of the same statement are not saved, the resourcetype had to be adjusted


mussbach (approved):
lgtm, code review only

PR #1067, opened 2023-03-24
fix(T32010): solve issue by creating a new organisation

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T32010

**Description:** This PR handles different return types of `UserHandler::addOrga()`. It returns Orga Entity by success or an array if the MissingMandatoryFields is true.

As well as resolve some other issues:
- add method which changes the case of types in relationships (to prevent some warnings)
`ResourcefulApi.js:60 --> The Resource with type 'customer' is sent in lower camel case. Please send as upper camel case.
`
- add additional check when customer id is an empty string (to prevent issues in VUEX store)

### DIFF
--- a/client/js/components/user/DpCreateItem.vue
+++ b/client/js/components/user/DpCreateItem.vue
@@ -210,6 +210,39 @@ export default {
       createUser: 'create'
     }),
 
+    changeTypeToPascalCase (payload) {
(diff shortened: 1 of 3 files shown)


hwiem (dismissed):
lgtm, code review only

PR #1068, opened 2023-03-24
Fix unclaiming of statement

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T32028

**Fixing the bug**
Inside StatementMeta, the statement obj persists its initial state, which it also bubbles back up onSave. So, when loading the view with unassigned statement, and then assigning it, the local statement in StatementMeta still has the unassigned state which is applied to the outer state then, leading to the buggy behavior.

To work around this, we can either only pass the data into StatementMeta which is actually needed by this component, or we can fix it afterwards (this version) - which is a little ugly but as most data is needed within StatementMeta, it is still the better choice.

**Button cleanup**
The claim component generates a link element. To wrap that within a button is not html5 conform, also the link is not really the correct element here. To still have the text label besides the icon, a new prop is used in DpClaim. Also, some flex classes are used to ensure vertical alignment also for the loading state.

**Remove "optimistic claiming"**
The optimistic behavior (that was implemented before) is not the standard behavior within demosPlan. Also, as we show a loading indicator on the claim button, the rest of the UI should be in sync with that. The alternative would have been to also remove the loading indicator, hence hiding the loading completely. However as we do not use this pattern at all, why should we use it here?

### How to review/test
- Best reviewed on a per-commit-basis. 
- The behavior described in the ticket (does only happen if you load the view in unclaimed state) should not persist.
- Before, the edit state of the StatementMeta box was shown as soon as users clicked the claim button. Now, the edit state should only show up after the request (which saves the claiming) resolves.

### PR Checklist

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
@@ -455,6 +455,7 @@ export default {
         .then(response => { checkResponse(response) })
         .then(() => {
           const dataToUpdate = this.setDataToUpdate(true)
+
           this.setStatement({ ...dataToUpdate, id: this.statement.id, group: null })
           dplan.notify.notify('confirm', Translator.trans('confirm.statement.assignment.assigned'))
         })
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
@@ -587,6 +587,7 @@ export default {
     synchronizeAssignee (statement) {
       const oldAssignee = JSON.stringify(statement.relationships.assignee.data)
       const newAssignee = JSON.stringify(this.statement.relationships.assignee.data)
+
       if (oldAssignee !== newAssignee) {
         statement.relationships.assignee.data = this.statement.relationships.assignee.data
       }
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
@@ -638,6 +638,7 @@ export default {
         .then(response => checkResponse(response))
         .then(() => {
           const dataToUpdate = this.setDataToUpdate()
+
           this.setStatement({ ...dataToUpdate, id: this.statement.id, group: null })
           dplan.notify.notify('confirm', Translator.trans('confirm.statement.assignment.unassigned'))
         })
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
@@ -539,6 +539,7 @@ export default {
     },
 
     saveStatement (statement) {
+      this.synchronizeAssignee(statement)
       this.synchronizeFullText(statement)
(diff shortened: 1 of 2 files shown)


client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
```diff
@@ -625,8 +633,11 @@ export default {
         }
       }
       return dpApi.patch(Routing.generate('api_resource_update', { resourceType: 'Statement', resourceId: this.statement.id }), {}, payload)
-        .then((response) => {
-          checkResponse(response)
+        .then((response) => checkResponse(response))
```
salisdemos, line 636: nitpick
```suggestion
        .then(response => checkResponse(response))
```

client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
```diff
@@ -625,8 +633,11 @@ export default {
...
+        .then(() => {
+          const dataToUpdate = this.setDataToUpdate()
+          this.setStatement({ ...dataToUpdate, id: this.statement.id, group: null })
+          dplan.notify.notify('confirm', Translator.trans('confirm.statement.assignment.unassigned'))
```
salisdemos, line 640: nitpick
```suggestion
          const dataToUpdate = this.setDataToUpdate()

          this.setStatement({ ...dataToUpdate, id: this.statement.id, group: null })
          dplan.notify.notify('confirm', Translator.trans('confirm.statement.assignment.unassigned'))
```

client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
```diff
@@ -582,6 +578,20 @@ export default {
...
+    /**
+     * If `this.statement` has changed its assignee (which does not propagate to the
+     * localStatement in StatementMeta), it must be synced back before applying the
+     * StatementMeta data to `this.statement`.
+     * @param {object} statement - The local statement of StatementMeta.vue.
+     */
+    synchronizeAssignee (statement) {
+      const oldAssignee = JSON.stringify(statement.relationships.assignee.data)
+      const newAssignee = JSON.stringify(this.statement.relationships.assignee.data)
+      if (oldAssignee !== newAssignee) {
```
salisdemos, line 590: nitpick
```suggestion
      const newAssignee = JSON.stringify(this.statement.relationships.assignee.data)

      if (oldAssignee !== newAssignee) {
```

client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
```diff
@@ -461,6 +454,8 @@ export default {
       return dpApi.patch(Routing.generate('api_resource_update', { resourceType: 'Statement', resourceId: this.statement.id }), {}, payload)
         .then(response => { checkResponse(response) })
         .then(() => {
+          const dataToUpdate = this.setDataToUpdate(true)
+          this.setStatement({ ...dataToUpdate, id: this.statement.id, group: null })
```
salisdemos, line 458: nitpick
```suggestion
          const dataToUpdate = this.setDataToUpdate(true)

          this.setStatement({ ...dataToUpdate, id: this.statement.id, group: null })
```

salisdemos (dismissed):
legit just really minor nitpicks

PR #1069, opened 2023-03-24
fix (refs T31926): adjust ai api user migration

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T31926

Adjust the migration that brings the AiApiUser to the database. Generate the uuid which reference the user first, then it can be used for relations.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/03/Version20230324132250.php
+++ b/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/03/Version20230324132250.php
@@ -16,6 +16,7 @@ use demosplan\DemosPlanCoreBundle\Entity\User\AiApiUser;
 use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\Migrations\AbstractMigration;
+use Faker\Provider\Uuid;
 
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/03/Version20230324132250.php
```diff
@@ -53,13 +55,12 @@ public function up(Schema $schema): void
...
               provided_by_identity_provider = 0;',
             [
+                'user_id'  => $aiApiUserId,
                 'login'    => AiApiUser::AI_API_USER_LOGIN,
                 'password' => $this->generateNewRandomPassword(),
                 'flags'    => self::FLAGS,
             ]
         );
-
-        $aiApiUserId = $this->getAiApiUserId();
```
mussbach, line 62: couldn't you delete the method as well?

breierDemos: It is needed for the down migration.

mussbach (approved):
lgtm when it is allowed to set the userId manually

PR #1070, opened 2023-03-24
Hide button row

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T31858

There is no practical way of hiding the buttonRow based on permissions - the appearance of form elements is already controlled by twig project overrides. So, to hide the buttonRow for ewm, it is moved within the inner template. Here, it can then be hidden via block, too.

### How to review/test
Nothing should change except for EWM. There should be no buttons anymore when visiting the "Daten der Orga" view anymore.

### Linked PRs
https://github.com/demos-europe/demosplan-project-ewm/pull/44

### PR Checklist


hwiem (approved):
lgtm

PR #1071, opened 2023-03-24
Show statement as text when not in edit mode

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T28146

There is no reason to show a disabled editor, at all. It is more helpful to just show the text.

The border and padding is applied to add structure to the view. It borrows the style from the StatementMeta box.

### How to review/test
In ewm, go to the statement list and open a statement that is not segmented already. unassign it. The statement text should be rendered in a bordered box, without the editor ui.

### PR Checklist


hwiem (approved):
lgtm, code review only

PR #1081, opened 2023-03-28
feature (refs T31926): use new Response instance to prevent null value

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T31926

Description:
use new Response instance for method processRequestParameters
instead of null

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm

PR #1088, opened 2023-03-29
Remove pattern validation for city

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T31969

This PR removes the pattern validation for all "city" form input fields. It was once introduced in DpAutofillSubmitterData without overthinking ux aspects, and has since spread into other form
fields that ask for a city.

It does not add value for users to restrict the input for the "city" field based on assumptions that can prove wrong anytime. It will do little harm to remove the restriction (which will eventually lead to a typo going unnoticed) but it will cause massive trouble for users failing for a too-restrictive validation.

### How to review/test
It should be possible to input all characters into the "city" field.

### PR Checklist

### DIFF
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -144,8 +144,7 @@
                     name="r_orga_city"
                     :label="{
                       text: Translator.trans('city')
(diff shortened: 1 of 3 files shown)


salisdemos (approved):
Thanks. Haven't seen that until now

PR #1092, opened 2023-03-29
fix: update demosplan-ui dependency

spiess-demos:
This update has to be applied to main again bc it got lost in a backmerge.


mussbach (approved):
lgtm, thx

PR #1102, opened 2023-03-30
T30866 administer imprint and privacy settings as customer master user

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T30866

Instead of the administrations of an organisation, the customer master user should do able to edit the imprint text and data protection text of each planning organisation.
An organisation is a "planning" organisation in case of it has one of three orgaTypes (MUNICIPALITY, PLANNING_AGENCY or HEARING_AUTHORITY_AGENCY) within the current customer.
The customer master user is managing the organisation via a special organisation list view, which has to be enhanced to allow showing the required text fields for each planning organisation.
To do so, the OrgaResourceType was enhanced to provide the necessary data along with the logic to update these organisation to allow edit the imprint and data protection text.

### How to review/test
Login as customer master user and navigate to the list of organisation. On each organisation which is a planning organisation (in the current customer), it should be possible to edit imprint and data protection text.
These texts should be displayed in the public area (no login requiered).

### Tasks
The check if an organisation is a planning organisation should probably be covered by a own ResourceType instead of special handling of existing "normal" OrgaResourceType. This has to be modeled and clarified and should result in another Story later on. 

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/ResourceTypes/OrgaResourceType.php
```diff
@@ -215,6 +220,13 @@ protected function getProperties(): array
             $this->createToManyRelationship($this->departments)->readable(false, static function (Orga $orga): TightencoCollection {
                 return $orga->getDepartments();
             }),
+            $this->createAttribute($this->isPlanningOrganisation)->readable(true,
+                function (Orga $orga): bool {
+                    return $orga->hasType(OrgaType::MUNICIPALITY, $this->globalConfig->getSubdomain())
+                        || $orga->hasType(OrgaType::PLANNING_AGENCY, $this->globalConfig->getSubdomain())
+                        || $orga->hasType(OrgaType::HEARING_AUTHORITY_AGENCY, $this->globalConfig->getSubdomain());
```
AlexMenschoid, line 227: This is fine, I just would like to point out that we have created a currentContextProvider (during the whole addon-work) which also provides the currentCustomer. It might make the code a bit better readable as it is clear that we're talking about the current customers subdomain and not any other subdomain from the global config. Just to give you the option in case you wanna try that. This is not a blocking concern.

AlexMenschoid (changes requested):

demosplan/DemosPlanCoreBundle/Entity/User/Role.php
```diff
@@ -105,6 +106,7 @@ class Role extends CoreEntity implements UuidEntityInterface
 
     /**
      * MasterUser eines Mandanten.
+     * Mandanten-Administration.
```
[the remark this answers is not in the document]
martendemos: In general I would agree, but in this case it is very hard to translate such kind of terminology into English.
I actually don't know how to translate this properly 
Customer-Administration? Which would be differ from already used term "customer master user".
So indeed, my intention was to have the actual translation documented directly at the role, to increase the DX and reduce misleading translations

demosplan/DemosPlanCoreBundle/ResourceTypes/OrgaResourceType.php
martendemos: I will try but make another PR for this, to avoid slow down this one.
https://github.com/demos-europe/demosplan-core/pull/1141

demosplan/DemosPlanCoreBundle/Entity/User/Role.php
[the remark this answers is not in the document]
AlexMenschoid: I see. I'm ok with leaving it here as is to be as precise as possible. But I get the feeling that we might need a kind DEMOS specific dictionary to keep all the domain specific terminology consistent. I'll make a note and bring it up in the appropriate meeting.

AlexMenschoid (approved):
lgtm, code review only

demosplan/DemosPlanCoreBundle/ResourceTypes/OrgaResourceType.php
```diff
@@ -58,6 +58,9 @@
  * @property-read End                              $submissionType
  * @property-read End                              $types @deprecated Use {@link OrgaResourceType::$statusInCustomers} instead
  * @property-read End                              $registrationStatuses @deprecated use {@link OrgaResourceType::$statusInCustomers} instead
+ * @property-read End                              $dataProtection
+ * @property-read End                              $imprint
+ * @property-read End                              $isPlanningOrganisation Indicates an organisation as organisation which can be administrate procedures
```
hwiem, line 63:
```suggestion
 * @property-read End                              $isPlanningOrganisation Indicates an organisation as organisation which can administrate procedures
```

hwiem (commented):
I assume it's fine the settings are also available to support users/in other projects?

FE code lgtm

PR #1109, opened 2023-03-31
Hide visibility options if only one option is allowed

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T32027

If there is only one role group which should be able to view faq articles, it makes no sense to force editors to always choose this option. To improve UX, this one group is activated silently, and users may control visibility by using the "Status" toggle.

The permission `feature_institution_participation` which was used to control if one of the options are shown, is not sufficient, so instead of using permissions here, a new configuration parameter `role_groups_faq_visibility` is established - because, technically what is controlled here is not a permission but project configuration.

<details><summary>Possible roadmap (only read if there is a userstory for that!):</summary>

As this is only a bug fix, this PR is scoped to the behavior described above. However when implementing it, we already discovered the further roadmap for improving the faq visibility.

While the visibility of the faq technically is scoped to role groups, it is persisted with a many-to-many relationship faq< - >role, so, to make a faq entry visible for role group GLAUTH, for every role within this role group an entry within _faq_role is made. While this looks clumsy at first, it is ok in terms of later extension (eg we could one day make faq visible to single roles). So, we may leave the model unchanged, but the mapping logic which saves those relation for each of the needed roles is needlessly implemented in different places: In the frontend `availableGroupOptions` (and in different twig files), and in the backend in the FaqHandler `updateFaqFromProperties`. Instead, we could add RoleResourceType as relationship to FaqResourceType and directly save the needed entries from the frontend. Also instead of hardcoding the options within `availableGroupOptions`, we would then get all available roles, filtered by the roleGroups allowed by the config parameter

Possible Frontend Implementation within DpFaqList:

```js
    roleFilters () {
      const filter = {}
      this.activatedRoleGroups.forEach(roleGroup => {
        filter[roleGroup] = {
          condition: {
            path: 'groupCode',
            value: roleGroup,
            memberOf: 'activatedRoles'
          }
        }
      })
      return filter
    },

    async fetchRoles () {
      const response = await dpApi.get(Routing.generate('api_resource_list', { resourceType: 'Role' }),
        {
          fields: {
            Role: ['name', 'groupCode'].join()
          },
          filter: {
            activatedRoles: {
              group: {
                conjunction: 'OR'
              }
            },
            ...this.roleFilters()
          }
        },
        { serialize: true }
      )
      this.roles = response?.data.data
    },
```

We could then use the role group name as label for each option.

</details>

### How to review/test
The current fix should lead to no option being visible within ewm, so faq visibility is controlled only with the "Status" field.

### PR Checklist

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Faq/FaqController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Faq/FaqController.php
@@ -10,7 +10,6 @@
 
 namespace demosplan\DemosPlanCoreBundle\Controller\Faq;
 
-use demosplan\DemosPlanCoreBundle\Resources\config\GlobalConfig;
 use demosplan\DemosPlanCoreBundle\Annotation\DplanPermissions;
 use demosplan\DemosPlanCoreBundle\Controller\Base\BaseController;
 use demosplan\DemosPlanCoreBundle\Entity\Faq;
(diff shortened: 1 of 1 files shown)


AlexMenschoid (approved):
BE looking good to me. Code review only.
And special thanks for translating old german comments! :)

hwiem (approved):
lgtm + works!

PR #1111, opened 2023-03-31
fix(refs T32058): Make check for maxExtent do what's suppose to do

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T32058

After a refactoring/bugfix some checkes to define if a extent is set failed because JS has a weired way to define "true"
(if You check it like `if([]) { /* this code will be executed */ }`, although `[] == false`).

### PR Checklist

### DIFF
--- a/client/js/components/map/map/DpOlMap.vue
+++ b/client/js/components/map/map/DpOlMap.vue
@@ -212,7 +212,8 @@ export default {
       },
       baselayer: '',
       baselayerLayers: '',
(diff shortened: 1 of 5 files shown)


hwiem (approved):
lgtm
I realized that this bug was already fixed in https://github.com/demos-europe/demosplan-core/pull/766, which makes me wonder if there isn't a way of avoiding this kind of "double-fixing bugs" in the future (or if there are other releases that need this fix)

PR #1112, opened 2023-03-31
F remove top level interface from services yaml

bakridemos:
Description: The TopLevelInterface is removed from demosplan addon


MoritzMandler (approved):
lgtm

PR #1124, opened 2023-04-03
fix (refs T32046): simplify sidemenu item extras

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T32046

We knew something like extras would be needed at some point, so we established it as an option. But at least for now, we can keep it a bit more simple which is exactly what this commit does.

This part of the new sidemenu was never used before.


mussbach (approved):
lgtm, thx. Code review and test

PR #1126, opened 2023-04-03
fix (refs T32124): Always show published documents to owning orga

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T32124

Planners may not be able to view published documents in the case when only specific organisations are allowed to view documents

### How to review/test
You need to have only categories published that are also only allowed to be accessed by specific organisations

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, code review only

PR #1127, opened 2023-04-03
refactor (refs T30254): detect invalid state early

dresslerdemos:
`getFieldMapping` did return a non-usable array format for specific cases. One call was adjusted so that an exception is thrown instead of continuing with the invalid return.

Also move `loadFieldMapping()` call up, so that it is loaded even if an empty string is passed as parameter instead of a class string.

**Ticket:** https://yaits.demos-deutschland.de/T30254

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm

PR #1136, opened 2023-04-04
replace env with parameters

shumyachkin:


mussbach (changes requested):
please use lowersnailcase as for all other params. And define a defaultvalue (empty) in parameters_default.yml

demosplan/DemosPlanCoreBundle/Resources/config/parameters_default.yml
```diff
@@ -325,6 +325,15 @@ parameters:
     saml_sp_x509_string: ''
     saml_registration_url: ''
 
+    # knpu_oauth2_client settings
```
mussbach, line 328: You need to define oauth_keycloak_client_id, oauth_keycloak_client_secret, oauth_keycloak_auth_server_url and oauth_keycloak_realm, not the ones defined here

mussbach (changes requested):

graupnerdemos (approved):
This looks correct. However, I think this should go to `main` now?

mussbach:
No, release is fine as we want to test it with the Keycloak in the other environment

mussbach (approved):
lgtm, thx

PR #1138, opened 2023-04-04
fix request format for extra email addresses

muellerdemos:
- The BE needs the allowedSenderAddresses as array not as a string hence this has been adjusted so.


MoritzMandler (approved):
lgtm, thx

PR #1148, opened 2023-04-05
minor (refs T31740): adds logging

AlexMenschoid:
We sometimes have the situation that a user can't log in to the production system and everything we see in the logs seems to be ok. So I'm adding this with the hopes that seeing the actually used strings as well as the result might help us understand that better.


mussbach (approved):
lgtm, thx

PR #1149, opened 2023-04-05
fix invalid request data for extra email addresses

muellerdemos:
- The procedureId may not be sent within the request body due to invalidity and hence must be removed from there.


ahmad-demos (approved):
lgtm, codeReview!

PR #1150, opened 2023-04-05
Fix(refs T31995): no institution data displayed in form fields

muellerdemos:
https://yaits.demos-deutschland.de/T31995

- This was a race condition like scenario caused by the if-else block of the computed `allowedRolesForOrga`
- To prevent this, DpMultiselect should only be rendered if `organisations[this.currentUserOrga.id]` is existant in any form (with value or empty string but not undefined)

### How to review/test

- Login as Lisa Bauer
- navigate to /users/list
- make sure the fields are filled
- make sure there is no console error about undefined values

### PR Checklist

### DIFF
--- a/client/js/components/user/DpUserFormFields.vue
+++ b/client/js/components/user/DpUserFormFields.vue
@@ -134,6 +134,7 @@
         {{ Translator.trans('role') }}*
       </label>
       <dp-multiselect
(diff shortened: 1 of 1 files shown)


ahmad-demos (approved):
lgtm, codeReview + Test, thanks!

PR #1161, opened 2023-04-06
Empty internId on delete last child statement.

martendemos:
Empty internId of original statement in case of last chlid (normal statement) will be deleted.

**Ticket:** https://yaits.demos-deutschland.de/T22439

To "free" the previous given internId and allow to use them again. Extract deleteStatementObject() into own servcie to avoid making the StatementService even bigger. Add new permission to decide if the internId of the related originalSTN should be nulled. This is also the first step do acutally automatically delete the related original statement in case of deleting the last child statement.

### How to review/test
Delete the last child statement of a original-statement with a set internId.
After successful deletion of the child-statement, the interId of the related originalSTN should be emtpy and therefore should be givable to another STN as well.

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
@@ -1898,7 +1898,7 @@ class DemosPlanStatementController extends BaseController
     }
 
     /**
-     * Löschen eines Statements.
+     * Deleting a draftstatement.
      *
      * @param string $_route
      * @param string $procedure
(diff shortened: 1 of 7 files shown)


breierDemos (changes requested):
I imported new statments as described in the ticket and then deleted them. The internId is correctly set to zero in the database. However, if I reload the statement list, the ones that were just deleted are there again, just without the internId. Is that currently what you want?

martendemos:
The just deleted statement is still there? like it was recreated? 
No, that's not how it supposed to be. :eyes:

martendemos:
> So when I click delete, I get a success message and it disappears from the list. When you reload the list, they reappear. In the debugger it also looked like they were being deleted. So yes maybe they will somehow be recreated again?

This sounds like https://yaits.demos-deutschland.de/T28243 to me.
This Bug is not solved yet.

breierDemos:
> > So when I click delete, I get a success message and it disappears from the list. When you reload the list, they reappear. In the debugger it also looked like they were being deleted. So yes maybe they will somehow be recreated again?
> 
> This sounds like https://yaits.demos-deutschland.de/T28243 to me. This Bug is not solved yet.

Okay, but as far as I've seen in the database, the internId is set to zero, which is the goal of this PR. Your tests passed too.
Then the reappearance of the statements is another bug with a new PR. Or not?
Or should the other bug maybe be fixed first and this one wait :thinking:

martendemos:
Yes, the Bug on deleting statement with segments is another bug an will be handled with the named Ticket.
Therefore this is not a regression and the internId was set to null as expected.

breierDemos (approved):

PR #1165, opened 2023-04-06
chore: Delete VueBus

salisdemos:
Remove / Refactor the last ocurences of the Vue Bus to get rid of that bloddy bastard.... finally

I hope I haven't overseen anything. I tried to test it locally and it looked good to me. But we have to test this really good since it was baked in quite nasty. I am not sure what that could break

related https://github.com/demos-europe/demosplan-project-teilhabe/pull/21


client/js/InitVue.js
```diff
@@ -51,6 +50,9 @@ function initialize (components = {}, storeModules = {}, apiStoreModules = [], p
 
         const mountedEvent = new Event('vue-mounted')
         document.dispatchEvent(mountedEvent)
+        setTimeout(() => {
+          window.mounted = true
+        }, 5)
```
spiess-demos, line 55: Why is this necessary?

client/js/components/procedure/basicSettings/DpProcedureCoordinate.vue
```diff
@@ -244,17 +244,26 @@ export default {
...
-    window.Bus.on('wizard:show toggleAnything:clicked', (sender, data) => {
-      //  Only fire when relevant wizard step / toggleAnything toggleId is transmitted
+    const handleWizardShow = data => {
       if (data === Translator.trans('wizard.topic.location') || data === 'procedureLocation') {
         this.$refs.map.updateMapInstance()
       }
+    }
+    //  Listeners are added because the OpenLayers map needs to be initialized on a visible element
+    document.addEventListener('wizard:show', ({ data }) => {
+      //  Only fire when relevant wizard step / toggleAnything toggleId is transmitted
```
spiess-demos, line 254:
```suggestion
      //  Only fire when relevant wizard step is transmitted
```

spiess-demos (commented):
Apart from some nitpicky suggestions and a general question, the changes look reasonable to me. 
The changes in DpPublicLayerListLayer.vue and MapTools.vue may change the behavior of the public map layer controls, as they emit events that are actually listened too (before, wvents were simply emitted the wrong way). We will have to pay attention primarily to the public layer list when testing.

spiess-demos (approved):
lgtm!

client/js/InitVue.js
hwiem: Wouldn't it be a better solution to set the variable in an event listener, like so:

```
document.addEventListener('vue-mounted', () => {
  window.mounted = true
}
```

PR #1166, opened 2023-04-06
chore: Simplify VueCore

salisdemos:
Remove the unused `currentViewport` setting and simplified the hasOwnProp-check to avoid the async loading


ahmad-demos (approved):
lgtm, codeReview, Thanks!

PR #1171, opened 2023-04-11
feat(refs: T30866): use correct permissions to edit orga imprint and data protection

muellerdemos:
https://yaits.demos-deutschland.de/T32219

The wrong permissions were checked for editing orgas imprint and data protection texts, hence was adjusted with this PR.


gruenbergerdemos (approved):
lgtm, code review only.

PR #1182, opened 2023-04-12
fix: remove DpAllowedSenderEmailList.vue from DpBasicSettings

muellerdemos:
### Description
- This is now an addon component and therefore should be removed from the core also this is throwing an error since the component does not exist anymore in the core

### Related
This was overseen in https://github.com/demos-europe/demosplan-core/pull/1179 and hence the import was removed in this PR from the basic settings


salisdemos (approved):
thanks

PR #1184, opened 2023-04-12
F_T31959 adjust core permissions

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T31959

Description: adjust core permissions
    - remove area_organisations_applications_manage from core. This permission will be enabled for the CUSTOMER_MASTER_USER only in the projects where it is allowed

- [X] Tests updated/created


elmasdemos (approved):
lgtm, code review

PR #1193, opened 2023-04-13
fix (refs T30866): Permissions needs to be available by exposing them.

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T32219
Fix for **Ticket:** https://yaits.demos-deutschland.de/T30866

Needed Permissions are not exposed for using in FE.

Fields should be available an editable for customer master user now.

Story PR: https://github.com/demos-europe/demosplan-core/pull/1102


muellerdemos (approved):
thanks - lgtm and tested in pair

PR #1194, opened 2023-04-13
Fix tooltip positioning

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T29863

This PR corrects a few things in terms of tooltips (apart from the bugfix):

- Instead of specifying the overall tooltip margin via Css, it is now applied in the tooltip directive in demosplan-ui via defaultOffset - this fixes a bug with positioning the arrow when a tooltip opens in a corner of its container
- in the map tools components, containers are set relative to the element itself to not cut off tooltip edges
- actual bugfix: if a map tools tooltip opens inside an element that has same or higher z-index than the tooltip z-index, a new class is applied to the tooltip to raise its z-index. It is not applied by default, to let tooltips triggered by elements that are overlapped by fixed elements not overlap those elements. This would be menaged by the tooltip directive itself, but it turns out that v-tooltip is deprecated so hard it even does not have any sufficient documentation. I also could not find an easy way to monkeypatch it or define any callbacks within the options.
- Within SegmentLocationMap.vue, the title attrs are removed in favor of tooltips to have all tool buttons behave similarly
- In StatementSegment.vue, the container option of all tooltips has been set to this.$el to have the tooltip also appear  in fullscreen mode. the same goes for the popover.

### How to review/test
In the "Erwiderung verfassen" view, tooltips should "be lookin nice".

### Tasks to do, before everything will work

- [x] Merge https://github.com/demos-europe/demosplan-ui/pull/159
- [x] Release a new version of demosplan-ui containing the above changes
- [x] Within this PR here, bump demosplan-ui to that new version

### PR Checklist


client/js/components/map/map/DpOlMapEditFeature.vue
```diff
@@ -56,7 +56,11 @@
       type="button"
       @click="toggle"
       data-cy="editButtonDesc"
-      v-tooltip="Translator.trans('explanation.territory.help.edit',{ editTool: Translator.trans('map.territory.tools.edit') })"
+      v-tooltip="{
+        classes: this.tooltipClass,
+        container: this.$parent.$parent.$parent.$el,
```
salisdemos, line 61: You cant know that, but in preparation of the migration to vue3 we have to find other solutions than `.$el` since with vue3 this doesn't work anymore because then there can be more root elements than one. So the explanation here, and my experience. Using `refs` looks like reasonable alternative
https://stackoverflow.com/questions/61600078/vue-3-composition-api-how-to-get-the-component-element-el-on-which-the-comp
even though its explained different here:
https://vuejs.org/api/component-instance.html#el

salisdemos (dismissed):
To be future ready we have to adjust this a little

client/js/components/map/map/DpOlMapEditFeature.vue
spiess-demos: thanks for the hint. i replaced `this.$el` with `this.$refs.<ref>` where appropriate and was even able to remove some container options entirely in 1961b3f1e64c761d3fcae29240889bed1aea54e1

hwiem (dismissed):
lgtm + fixes the bug

PR #1210, opened 2023-04-17
chore: adjust method to match updated interface

mussbach:
Update method signature to math used Interface `AuthenticatorInterface`

### How to review/test
CodeReview should be probably best

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


graupnerdemos (approved):
thanks

PR #1211, opened 2023-04-17
feat: Clear cache out of band

graupnerdemos:
Clearing the cache in-process vs clearing the cache out-of-process yields slightly different results for the next command. In the latter way, the current application state does not affect the cache state.

Changing this is a bit of a gamble as I have seen varying results but it seems that it makes the cache state for the frontend asset build a little more reliable and thus reduces the friction during addon installation.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
let's try it

PR #1219, opened 2023-04-18
bug(refs T28356): Software components with no url

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T28356

The website for some software components is unknown. There should be no link displayed then.

### How to review/test
Go to Impressum -> Lizenzhinweise -> hier and check if the components with unknown websites have no link.


spiess-demos (approved):
Code review only, however this seems to do the drick! Lgtm.

PR #1226, opened 2023-04-19
Fix: Remove double data property

elmasdemos:
### Description: 

- Sonarcloud shows major issue because of double data propöerty
- just remove it to fix it, because it already exists


ahmad-demos (approved):
lgtm, codeReview, Thanks!

PR #1231, opened 2023-04-20
F_remove unnecessary interface defintion

bakridemos:
Description:  

there are no need to add a defintion to this interface in the services yaml.


MoritzMandler (approved):
lgtm - event is created within demospipes - the core as subscriber does not need this as a service

PR #1243, opened 2023-04-24
Feat(refs:T32012): order v bind vue3 migration

elmasdemos:

### Description: 

- this is for the preparation from vue 2 to vue 3 migration
- here there was a warning, that the v-bind attribute was in a wrong order
- it needs to be in the first place or directly after v-if, v-for, eg.

**this PR will be merged into the integration branch: `f_vue3_compat_mode_setup`**

### Linked PRs (optional)

- https://github.com/demos-europe/demosplan-core/pull/1240

### PR Checklist

### DIFF
--- a/client/js/components/statement/publicStatementLists/DpPublicStatement.vue
+++ b/client/js/components/statement/publicStatementLists/DpPublicStatement.vue
@@ -62,7 +62,7 @@
                 v-else-if="item.type === 'link'"
                 v-bind="item.attrs"
                 class="o-link--default u-valign--middle"
-                :href=renderUrl('item.url')>
+                :href="item.url">
                 {{ item.text }}
               </a>
               <h4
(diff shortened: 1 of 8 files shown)


client/js/components/statement/publicStatementLists/DpPublicStatement.vue
```diff
@@ -421,6 +422,10 @@ export default {
         content += `${ln}<br />`
       })
       return DomPurify.sanitize(content)
+    },
+
+    renderUrl (url) {
+      return sanitizeUrl(url)
     }
```
muellerdemos, line 429: I understand the approach but I think this is not neccessary since we are not dealing with user generated input for our links. However I would say we can still keep it but I wonder what we will do with all the other links that are not sanitized in our applications.

muellerdemos (commented):
Just wondering if we want to sanitize url generally/everywhere.

muellerdemos (approved):
thanks for applying our linter rules here - lgtm

client/js/components/statement/publicStatementLists/DpPublicStatement.vue
elmasdemos: removed it, thanks for your advice https://github.com/demos-europe/demosplan-core/pull/1243/commits/1b2191b0f5106018db9f98b4e274e82f2ab038f8

PR #1255, opened 2023-04-26
fix(refs T32449): remove access to segments for ahb

muellerdemos:
https://yaits.demos-deutschland.de/T32034

Also the fetching of segments data should only be fired, when the current user has the permission for segments. Otherwise the response would throw an console error. And also the loading state would not been set to false which prevented the statement text to be rendered instead. To make this component more resilient a finally block has also been added to the fetch function for segments


salisdemos (approved):
legit

martendemos (approved):
manual tested

salisdemos (approved):

PR #1266, opened 2023-04-27
feat: remove rollup migration preparation code

mussbach:
Before performing the database rollup in core demosplan we needed to make sure that the database is in any project in a consistent state. This is not necessary anymore. In fact, on every migration any project migrations later than the rollup migration was downmigrated and after the core migrations reapplied. This PR sets an end to this, as rollup is already performed in most of the existing projects.

### How to review/test
Perform `dplan:migrate -vvv` and notice that no project migrations are down migrated

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, code review only

PR #1267, opened 2023-04-27
chore: reduce authenticator method visibility

mussbach:
The visibility of `getCredentials` may be tightened, as it is only used within class context

### How to review/test
code review should be best, phpstan can help

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, nice little catch

PR #1291, opened 2023-05-02
feat: allow tex to pdf implementation via external service

mussbach:
This PR adds an alternative way to create pdf exports via an external api service. The existing implementation via rabbitmq is still the default one, it is possible to switch via Interface implementation.

### How to review/test
code review might be best. If you have access to the new service you could use it.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMetaAttachments.vue
graupnerdemos, line 1: Is this meant to be changed in here?

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2023/06/Version20230601095828.php
graupnerdemos, line 1: This also doesn't seem like it belongs here.

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanUserController.php
graupnerdemos, line 1: not related to pdf stuff

demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanAssessmentController.php
graupnerdemos, line 1: this seems to be from another change

graupnerdemos (changes requested):
This PR contains a bunch of changes which are completely unrelated to PDF handling and seem to have barged in via merges. @mussbach can you please check that?

mussbach:
@graupnerdemos thx, fixed

demosplan/DemosPlanCoreBundle/Tools/PdfCreatorRabbitmq.php
```diff
@@ -0,0 +1,120 @@
...
+                'Fehler in ImportConsumer:',
+                [$e]
+            );
+            throw new TimeoutException('Timeout ');
+        } catch (Exception $e) {
+            $this->logger->error(
+                'Could not create PDF ',
+                [$e]
+            );
+            throw new Exception('Could not create PDF ');
```
graupnerdemos, line 109: Just once here: I'd prefer having a PdfCreateException and static constructors for the different cases, but I am unsure whether this is part of our coding guidelines or just something I've seen and done a bunch of times.

graupnerdemos (approved):
One minor remark. However, since that is mostly just moved code from the previous solution I don't deem it blocking for this change.

demosplan/DemosPlanCoreBundle/Tools/PdfCreatorRabbitmq.php
mussbach: I am quite sure that this is currently not part of our coding guidelines as this has never been discussed (at least I am not aware of it and I dislike the static constructor). But having specialized Exceptions is better

PR #1294, opened 2023-05-03
F use event interface instead event path

bakridemos:
Description: use event interface instead event path


MoritzMandler (approved):
lgtm

dresslerdemos (approved):

PR #1295, opened 2023-05-03
feat: use default elasticsearch indexing instead of SearchIndexTask

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T28243

SearchIndexTask was once introduced to improve performance. 
This PR disables the reindexing via SerchIndexTask as we want to test, whether the default
indexing via fos elastica listener works and does not lead to
performance issues. When it does, we can reenable this code

### How to review/test
Delete Statement with Segments. It should work now

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


martendemos:
Deleting these methods is not an option?
Do we want mark these methods as deprecated or something?

mussbach:
Usually I like to delete code immediately. My rationale was that we may need to be very quick in case of an incident when we have to rollback this change. When just commenting out we could easily undo the change. After some battle testing in production we could delete a whole bunch of code

mussbach:
Deprecation hints: 92215e5f

AlexMenschoid (commented):
I'm pretty sure that this PR should point towards `main` branch to be tested during the next sprint. I also would rather we delete the code and really thoroughly test it during the sprint, maybe even in different projects, to be sure it works before deploying it. But saying we leave it in like this suggests to me a lack of confidence which means we need more time with it before deploying it.

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
```diff
@@ -2032,23 +1950,11 @@ public function addVote($statementId, User $user)
...
-            $vote = $this->statementVoteRepository->findOneBy([
+            return $this->statementVoteRepository->findOneBy([
                 'user'      => $user->getId(),
                 'statement' => $statementId,
                 'deleted'   => false,
                 'active'    => true, ]);
-
-            // user already voted this statement?
-            if (null === $vote) {
-                $vote = $this->statementVoteRepository->add($data);
```
dresslerdemos, line 2043: `$this->statementVoteRepository->add($data);` really not needed anymore?

dresslerdemos (changes requested):
I don't understand how updating the elasticsearch index works now, but if it does work, then the removal of the code was done properly except for the few things noted in this review.

Please verify that `isElasticsearchAsyncIndexing`, `isElasticsearchAsyncIndexingLogStatus` and `getElasticsearchAsyncIndexingPoolSize` are removed  from `GlobalConfigInterface`.

`ResourceChange::getEntityIdsToUpdateInIndex` (and related code) can be removed too. I'd suggest to do this in a separate PR, but please create a corresponding ticket.

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
mussbach: Good catch, thx! f9c4a333

mussbach:
Indexing works now via default fos elastica listeners defined by `listener: ~` in fos_elastica.yaml. The methods are removed from `GlobalConfigInterface`

martendemos:
soll auch noch in den release_ewm branch

PR #1303, opened 2023-05-03
chore: trim is not allowed to be called with null

mussbach:
Call to trim with null is deprecated

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm

PR #1304, opened 2023-05-03
fix(refs T32450): make radio buttons reactive when initial data is set

muellerdemos:
https://yaits.demos-deutschland.de/T32450

- We are receiving a wrong key from data
- To make the radio buttons toggle the rendering of the institution form fields it was necessary to check for the key `toeb`
- Data send us the key `toeb` but this is outdated. The new term is `institution` hence the key must be checked and adjusted in case it is `toeb`
- This is needed to make the usually read-only props reactive to determine if the statement belongs to an institution or not

### How to review/test

- finish the import of a pdf via the overview page
- confirm the additional form fields are displayed when selected institution (radio button)

### PR Checklist


salisdemos (approved):
legit for that purpose

PR #1305, opened 2023-05-03
bug(refs T32505): center map

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T32505

Center the map around all drawings and zoom to the combined extent. For that, first we need to get the extent from each layer and then combine them into one extent. After that the map can be fitted to the combined extent.

### How to review/test
Go to Verfahren -> Abschnitte -> Erwiderungen verfassen and click on the "Ortsbezug" button. Make some drawings (line, point, polygon), close it and open it again. The map should be centered around the drawings and zooming in (all drawings should be seen).

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
+++ b/client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
@@ -97,6 +97,7 @@ import { DpButtonRow } from '@demos-europe/demosplan-ui'
 import DpOlMap from '@DpJs/components/map/map/DpOlMap'
 import DpOlMapDrawFeature from '@DpJs/components/map/map/DpOlMapDrawFeature'
 import DpOlMapEditFeature from '@DpJs/components/map/map/DpOlMapEditFeature'
+import { extend } from 'ol/extent'
 
 export default {
   name: 'SegmentLocationMap',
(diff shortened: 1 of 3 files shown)


client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
```diff
@@ -233,6 +237,23 @@ export default {
...
     },
 
+    /*
+     * Center the map around all drawings and zoom to the combined extent.
+     */
+    setCenterAndExtent () {
+      const extentPolygon = this.$refs.drawPolygon.getExtent()
+      const extentPoint = this.$refs.drawPoint.getExtent()
+      const extentLine = this.$refs.drawLine.getExtent()
+
```
salisdemos, line 247: can't `this.featuresObject` be used to accomplish thant? As far as I remember, in that var all elements are already combined.

salisdemos (changes requested):

muellerdemos (commented):
This solution lgtm from a functional point of view. Since I haven`t worked a lot with maps so far I could not say more

client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
gruenbergerdemos: `this.featuresObject` only contains the coordinates, but not the extent.

salisdemos: :facepalm: I love fitting wording

PR #1306, opened 2023-05-03
Fix (refs: T32452) attachments in stn detail view not accessable

muellerdemos:
https://yaits.demos-deutschland.de/T32452

- since the claiming or the editing of an already synchronized stn is not allowed we can disable the claiming completely if this is the case
- if a statement is in synchronized status, the claiming elements in the statement list and 
statement detail views have been disabled/will not be rendered
- `templateVars.isSourceAndCoupledProcedure|default('false')` raises console errors since twig passes `1` or `0` as `true` or `false` values but the prop is defined as `Booleans` hence has been adjusted accordingly to `templateVars.isSourceAndCoupledProcedure == 1 ? 'true' : 'false'.`

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
@@ -56,6 +56,7 @@
         <ul class="float--right u-m-0 space-inline-s flex">
           <li class="display--inline-block">
             <dp-claim
+              v-if="!statement.attributes.synchronized"
               class="o-flyout__trigger u-ph-0_25 line-height--2"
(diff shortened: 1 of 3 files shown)


muellerdemos:
@gruenbergerdemos 
> I think we should also disable editing the STN in the detail view if it is synchronized. At the moment you can edit it (if you had claimed it before the synchronization) and get an error when you try to save it.

Actually there already is an implementation for this which just needed some debugging. Done here [995339e](https://github.com/demos-europe/demosplan-core/pull/1306/commits/995339eb462564c8d9c2a585e742d523d2dda050)

gruenbergerdemos (approved):
lgtm!

PR #1310, opened 2023-05-04
Fix(refs: T32450) make radio buttons reactive without clicking in the form

muellerdemos:
https://yaits.demos-deutschland.de/T32450

- Replace key `toeb` with `institution` but keep all other init values
- The key institution was not set correctly if there are init values. If there are, we need to keep all initial values but `toeb` which must be renamed to `institution`

### How to review/test

- Navigate to convert pdf view (from overview eg)
- Click on institution and make sure it still works

### DIFF
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -541,10 +541,16 @@ export default {
   },
 
   methods: {
+    logValue (val) {
+      console.log(val)
+    },
     setInitialValues () {
       this.values = { ...this.initValues }
       // Set default values to ensure reactivity.
-      if (typeof this.values.submitter === 'undefined' || Object.keys(this.values.submitter).length === 0) {
+      if (typeof this.values.submitter !== 'undefined' && typeof this.values.submitter.institution === 'undefined') {
+        Vue.set(this.values.submitter, 'institution',  this.values.submitter.toeb)
+      }
+      if (typeof this.values.submitter === 'undefined' || Object.keys(this.values.submitter).length === 0 ) {
         Vue.set(this.values, 'submitter', {})
         for (const [key, value] of Object.entries(submitterProperties)) {
           Vue.set(this.values.submitter, key === 'toeb' ? 'institution' : key, value)
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -553,7 +553,7 @@ export default {
       if (typeof this.values.submitter === 'undefined' || Object.keys(this.values.submitter).length === 0 ) {
         Vue.set(this.values, 'submitter', {})
         for (const [key, value] of Object.entries(submitterProperties)) {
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -544,10 +544,16 @@ export default {
     setInitialValues () {
       this.values = { ...this.initValues }
       // Set default values to ensure reactivity.
-      if (typeof this.values.submitter === 'undefined' || Object.keys(this.values.submitter).length === 0) {
+      if (typeof this.values.submitter !== 'undefined' && typeof this.values.submitter.institution === 'undefined') {
+        // Since Data sends us the key toeb instead of institution, we need to transform this for now but keep all init values
+        Vue.set(this.values.submitter, 'institution',  this.values.submitter.toeb)
+        Vue.delete(this.values.submitter, 'toeb')
```
salisdemos, line 550: in preparation for vue3 can we omit the Vue.set/delte?
```suggestion
        this.values.submitter.institution = this.values.submitter.toeb
        delete this.values.submitter.toeb
```

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -544,10 +544,16 @@ export default {
...
+      }
+
+      if (typeof this.values.submitter === 'undefined' || Object.keys(this.values.submitter).length === 0 ) {
         Vue.set(this.values, 'submitter', {})
         for (const [key, value] of Object.entries(submitterProperties)) {
-          Vue.set(this.values.submitter, key === 'toeb' ? 'institution' : key, value)
+          Vue.set(this.values.submitter, key, value)
```
salisdemos, line 556: see bove

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -544,10 +544,16 @@ export default {
     setInitialValues () {
       this.values = { ...this.initValues }
       // Set default values to ensure reactivity.
-      if (typeof this.values.submitter === 'undefined' || Object.keys(this.values.submitter).length === 0) {
+      if (typeof this.values.submitter !== 'undefined' && typeof this.values.submitter.institution === 'undefined') {
```
salisdemos, line 547:
```suggestion
      if (hasOwnProp(this.values), 'submitter') && !hasOwnProp(this.values.submitter, 'institution') {
```
(hasOwnProp has to be importet eventually)

salisdemos (changes requested):

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
muellerdemos: doing it like this will make `values.submitter.institution` non-reactive and the trigger to render the additional form fields will not be fired. However in Vue3 this problem does not exist anymore and we can set the data props as usual to make them reactive. Nevertheless, this syntax is not supported in vue3 but in vue3's compat build. This component is also a candidate for a bigger refactoring which I do not want to start before EWM v3.0 is deployed.

salisdemos (approved):

PR #1317, opened 2023-05-05
F T28477 incorrect display of the title when exporting statements

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T28477

Description: 
In the pdf and the doc export, the procedure name is incorrect displayed. Twigs filters are added to display it correctly in the pdf export. 

For the docx export, the quotes conversion while retrieving procedure name is disabled. The htmlspecialchars method has now ENT_NOQUOTES as flag parameter which prevent to encode quotes

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Twig/Extension/ProcedureExtension.php
+++ b/demosplan/DemosPlanCoreBundle/Twig/Extension/ProcedureExtension.php
@@ -12,6 +12,12 @@ namespace demosplan\DemosPlanCoreBundle\Twig\Extension;
 
 use Carbon\Carbon;
 use DateTime;
+use Exception;
+use Psr\Container\ContainerInterface;
+use Psr\Log\LoggerInterface;
+use RuntimeException;
+use Symfony\Contracts\Translation\TranslatorInterface;
+use Twig\TwigFunction;
 use demosplan\DemosPlanCoreBundle\Entity\Procedure\Procedure;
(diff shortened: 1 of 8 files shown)


mussbach (approved):
lgtm, thx

demosplan/DemosPlanCoreBundle/Twig/Extension/ProcedureExtension.php
```diff
@@ -303,6 +303,8 @@ public function getNameFunction($procedure, $type = 'auto')
         try {
             $procedure = $this->getProcedureObject($procedure);
         } catch (Exception $exception) {
+            $this->logger->error('Could not get procedure object', [$exception]);
```
dresslerdemos, line 306: Thank you for the logging, but where does the logger come from?

dresslerdemos (changes requested):
Not sure about `ENT_NOQUOTES` and one concern.

Otherwise looks good.

Code review only.

demosplan/DemosPlanCoreBundle/Twig/Extension/ProcedureExtension.php
bakridemos: Oh thx my bad, fixed hier https://github.com/demos-europe/demosplan-core/pull/1317/commits/4e372f8364f92a280929e03a41f7726c913cf0cc

dresslerdemos (approved):
Code review only

PR #1318, opened 2023-05-05
feature (refs T32200): avoid null for parameter

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T32200

Description:
avoid null for parameter client in FileService when injecting the fileServiceInterface.

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
It looks really weird and I have no clue why this should be needed, but it should do no harm

PR #1319, opened 2023-05-05
feat: #rector Import namespaces

mussbach:
Import namespaces automatically with rector

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Resources/config/bundles.php
+++ b/demosplan/DemosPlanCoreBundle/Resources/config/bundles.php
@@ -1,42 +1,5 @@
 <?php
 
-use Symfony\Bundle\FrameworkBundle\FrameworkBundle;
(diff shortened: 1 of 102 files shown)


demosplan/DemosPlanCoreBundle/Resources/config/bundles.php
graupnerdemos, line 1: We should exclude that file, otherwise symfony flex will flex on us.

graupnerdemos (changes requested):
Rector did very good. One config change should be done though.

demosplan/DemosPlanCoreBundle/Resources/config/bundles.php
mussbach: Indeed. Will change it in the rector config pr and revert the change here 867a9f60

graupnerdemos (approved):

PR #1321, opened 2023-05-05
fix (refs T32108): correctly calculate status percentages

dresslerdemos:
The implementation to generate `$templateVars['statementStatusData']` was overly complex and bug prone with unnecessary performance impact. The goal was to replace it with an aggregation executed directly in the Elasticsearch. To avoid complexity value objects were introduced to pass the backend data into the frontend. This was done without changing the data structure (which would have allowed for additional complexity reduction).

**Ticket:** https://yaits.demos-deutschland.de/T32108

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "demos-europe/edt-dql": "^0.16.0",
         "demos-europe/edt-extra": "^0.16.0",
         "demos-europe/edt-paths": "^0.16.0",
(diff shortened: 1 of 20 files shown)


dresslerdemos:
Sonarcloud concerns result from unrelated code duplications in the `StatetmentService`. I suggest to ignore.

bakridemos (approved):
lgtm

PR #1327, opened 2023-05-08
Remove specific version for demosplan-ui

gruenbergerdemos:
Specific version is not needed anymore.


salisdemos (approved):
Thanks for spotting that

PR #1329, opened 2023-05-08
chore: fix addon route path as it may be called differently

mussbach:
`pwd` might be different depending on how the code is triggered. This PR makes this a bit more robust

### How to review/test
Code review or try to call a addon route in browser directly after cache clear

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


dresslerdemos (approved):
Code review only.

PR #1335, opened 2023-05-09
F adjust dependency version

bakridemos:
Description: 

- not all adjustements made in demosplan are compatible with this branch that's why the dependency version has to be specified

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "demos-europe/edt-dql": "^0.16.0",
         "demos-europe/edt-extra": "^0.16.0",
         "demos-europe/edt-paths": "^0.16.0",
-        "demos-europe/demosplan-addon": "dev-main",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -34,7 +34,7 @@
         "demos-europe/edt-dql": "^0.16.0",
         "demos-europe/edt-extra": "^0.16.0",
         "demos-europe/edt-paths": "^0.16.0",
-        "demos-europe/demosplan-addon": "dev-main",
+        "demos-europe/demosplan-addon": "dev-main#v0.0.3",
```
mussbach, line 37:
```suggestion
        "demos-europe/demosplan-addon": "0.0.3",
```

mussbach (changes requested):
composer.lock needs to be included in PR

mussbach (approved):
lgtm

PR #1339, opened 2023-05-10
chore: Remove unused components

gruenbergerdemos:
These components were moved demospipes and aren't needed here anymore.


muellerdemos (approved):
thanks for removing the duplicates - lgtm

PR #1342, opened 2023-05-11
bug(refs T32727): Add storage key for pagination

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T32727

Add session storage key for currentPage and items per page for list statements and list segments.

## How to test

Go to statement list or segments list and change page and/or items per page. Go to another page and back to the list. Page and items per page should be the same.


client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -557,11 +575,13 @@ export default {
 
     initPagination (data) {
       const dataPag = data.meta.pagination
+      const currentPage = Number(JSON.parse(window.sessionStorage.getItem([this.storageKeyPagination])).currentPage)
```
spiess-demos, line 578: Why are you using sessionStorage (and not localStorage) here? `sessionStorage` [is unique to one browser tab](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) and is cleared when this tab is closed. I think we might very well use `localStorage` here.

gruenbergerdemos: I'm not sure about the reasons for `sessionStorage`. @muellerdemos Do you know more about that?

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -448,8 +448,9 @@ export default {
           }
         })
     },
-    storageKey () {
-      return `${currentUserId}-${pagination}`
+
+    storageKeyPagination () {
+      return `${this.currentUserId}-paginationStatementList`
```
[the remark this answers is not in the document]
gruenbergerdemos: Thanks, done in [60a89e6](https://github.com/demos-europe/demosplan-core/pull/1342/commits/60a89e6fabe3daad4b1d11db0dbd9b955eb7d43e)

client/js/components/procedure/SegmentsList/SegmentsList.vue
gruenbergerdemos: Done in [9af8277](https://github.com/demos-europe/demosplan-core/pull/1342/commits/9af8277525160ae9e7aee3a71b98cac13b5abe87). 
I also updated the pagination storage implementation (I realized we never used the stored values for the API call).

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -498,9 +502,15 @@ export default {
         })
         .then(data => {
           this.isLoading = false
```
muellerdemos, line 504:
```suggestion
        .finally(data => {
          this.isLoading = false
```

I think the loading state should be set to false in a finally block because if the request somehow fails, the loading state will not be set to false.

muellerdemos (changes requested):
Thanks for the upgrade ;) besides one tiny remark this lgtm

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -500,7 +500,7 @@ export default {
         .catch(() => {
           dplan.notify.notify('error', Translator.trans('error.generic'))
         })
-        .then(data => {
+        .finally(() => {
```
muellerdemos, line 503: sorry this was confusing from my side. I actually meant to add a `.finally()` after the `.then()` and set the `loadingState` to false in `finally`. We need the other implementation of `then()` and its logic and I think we can not simply replace the `then()` with a `finally()` here

muellerdemos (changes requested):

client/js/components/procedure/SegmentsList/SegmentsList.vue
gruenbergerdemos: Ah, I see, I didn't get that. Done in [22216f9](https://github.com/demos-europe/demosplan-core/pull/1342/commits/22216f9cab3f6aae63a6e4c9667c1265c0f68fcf)

muellerdemos (approved):
thanks this lgtm

gruenbergerdemos:
To avoid code duplication, I moved the methods regarding the pagination to a mixin. Furthermore, we need a storage key for each procedure, otherwise we get errors if we want to load the storage values from another procedure.

muellerdemos (approved):
thanks for adding a mixin here - lgtm (code review only)

PR #1351, opened 2023-05-12
bug(refs T32572): fix special characters appearing at the end of stri…

muellerdemos:
https://yaits.demos-deutschland.de/T32572

The value `rtl` of the `dir` attribute was the reason that special characters appear at the end of the string even if the input was special characters first.

Relating to https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir?retiredLocale=de, the auto value should be used for data with an unknown directionality, like data coming from user input, eventually stored in a database. Since this is the case here, the attribute's value has been set accordingly to `auto`.

### PR Checklist

### DIFF
--- a/client/js/components/statement/listStatements/ListStatements.vue
+++ b/client/js/components/statement/listStatements/ListStatements.vue
@@ -135,7 +135,7 @@
             <div
               v-tooltip="internId"
               class="o-hellip--nowrap text--right"
-              dir="rtl"
+              dir="auto"
               v-text="internId" />
           </div>
         </template>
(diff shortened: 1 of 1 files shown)


client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -135,7 +135,7 @@
             <div
               v-tooltip="internId"
               class="o-hellip--nowrap text--right"
-              dir="rtl"
```
salisdemos, line 138: why do we need that at all?
May it even work if we drop the prop?

muellerdemos: Turned out that dropping the attribute produces the same result hence done here https://github.com/demos-europe/demosplan-core/pull/1351/commits/9e0456ab0a8d0ee808b55725c3b5903ac14e79e1

salisdemos (approved):

PR #1352, opened 2023-05-12
F adjust dependency version

bakridemos:
Description:

- not all adjustements made in demosplan are compatible with this branch that's why the dependency version has to be specified


mussbach (approved):
when Version 0.0.3 is correct this looks good to me

PR #1363, opened 2023-05-15
feat (refs T32782): simplify WeakPassword check

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T32782

Check password only when authentication is valid. Otherwise a notification may be shown even on invalid password. The password is now validated by using symfonys UserBadge system.

### How to review/test
Code review or try to login with existing user but invalid, too short password like `wer` 

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


martendemos (dismissed):
code review and manual test

PR #1365, opened 2023-05-16
feat(refs T32601): automatically use dates in pdf import confirmation view

muellerdemos:
https://yaits.demos-deutschland.de/T32601

This PR is part of EWM 3.1

- The simplified new statement form component was updated in an addon and must be synced here with the state.
This is mainly intended to reduce the maintenance efforts because the simplified statement form had to be duplicated in addons
- The `values.submitter.date` prop will be passed from data to FE if existant (correct box recognition setup) hence we can use this value to automatically fill in the form field value when confirming the pdf import

### PR Checklist

### DIFF
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -492,7 +492,7 @@ export default {
       isLoading: false,
       isSaving: false,
       values: {
-        authoredDate: this.initialAuthoredDate,
+        authoredDate: this.values.submitter.date ? this.values.submitter.date : '',
         memo: '',
         submittedDate: '',
         tags: [],
@@ -506,10 +506,6 @@ export default {
     escapedUsedInternIds () {
       const specialCharEscaper = /\[|\\|\^|\$|\.|\||\?|\*|\+|\(|\)|\//g
       return this.usedInternIds.map(id => id.replace(specialCharEscaper, (specialChar) => `\\${specialChar}`))
-    },
-
-    initialAuthoredDate () {
-      return this.values.submitter.date ? this.values.submitter.date : ''
     },
 
     internIdsPattern () {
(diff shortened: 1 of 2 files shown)


client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -546,18 +550,28 @@ export default {
...
       }
     },
 
+    /**
+     * @param { Boolean } val
+     * To ensure the reactivity the state of 'values.submitter.institution'.
+     */
+    setInstitutionValue (val) {
+      this.$nextTick(() => {
+        this.$set(this.values.submitter, 'institution',  val)
```
gruenbergerdemos, line 571:
```suggestion
        this.$set(this.values.submitter, 'institution', val)
```

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -492,7 +492,7 @@ export default {
       isLoading: false,
       isSaving: false,
       values: {
-        authoredDate: '',
+        authoredDate: this.initialAuthoredDate,
```
gruenbergerdemos, line 495: Are you sure this can be used here? EsLint says: "The computed property cannot be used in `data()` because it is before initialization.(vue/no-computed-properties-in-data)"

gruenbergerdemos (changes requested):

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
muellerdemos: done https://github.com/demos-europe/demosplan-core/pull/1365/commits/fc71eb1c2b9e9434d39864247e4a11c9ebac1f40

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
muellerdemos: Ah I see, true we should not be doing it like that. I just wanted to avoid a ternary in the data props and use a computed instead but wasn't aware that my implementation could be prone to errors. Done here https://github.com/demos-europe/demosplan-core/pull/1365/commits/fc71eb1c2b9e9434d39864247e4a11c9ebac1f40

PR #1366, opened 2023-05-16
bug(refs T32785): Add css selector to savelist

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T32785

The components for this css were moved to an addon, but the styles are still in core and need to be loaded. Safelist.deep includes the selector's children.

### How to review/test
Split a statement and see if the styles were loaded.


spiess-demos (approved):
lgtm (code review).

PR #1398, opened 2023-05-24
fix (refs T32893): 

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T32893

Fix versioning  by fix creation of entity content change entry.

The user of the newly used TokenInterface is not an instance of User, therefore the condition always fails. Getting an actual User via the userFromSecurityUserProvider solves this.

### How to review/test
Do a change that is tracked. E.g. bulk edit the "step" of a Segment as described in the related ticket.

### PR Checklist

- [ ] Tests updated/created


breierDemos (approved):
lgtm, tested locally and works
thank you!

PR #1399, opened 2023-05-25
bug(refs T32855): Import Vue

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T32855

We need to import Vue here, because otherwise it is not available in the Addon and we get an error, that this function does not exist.

### How to review/test
Verfahren -> STN aufteilen: Edit a segment and click on "Abbrechen" or try to delete a segment.


hwiem (approved):
lgtm, code review only

PR #1403, opened 2023-05-25
refactor(refs T30969): move DpRegisterFlyout

hwiem:
**Ticket:** https://yaits.demos-deutschland.de/T30969

- DpRegisterFlyout is too specific to be included in demosplan-ui
- it was thus moved to dplan and renamed to RegisterFlyout (since only demosplan-ui components should be prefixed with 'Dp')

### How to review/test

- Open the blp project and visit pages that don't require a login. The register flyout ("Registrieren") should be loaded correctly

### Linked PRs (optional)
PR removing DpRegisterFlyout from demosplan-ui:
- https://github.com/demos-europe/demosplan-ui/pull/255

### PR Checklist


spiess-demos (approved):
lgtm (code review).
thanks for the PR!

PR #1405, opened 2023-05-25
fix (refs T32922): Fix versioning by fix getting user on creation of …

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T32922

Actually some times $this->userFromSecurityUserProvider->fromToken($token); can't find a user.
To cover this, we getting the user the old way, in case it was not found.

### How to review/test
Create a new version by changing working person or step.

### PR Checklist

- [ ] Tests updated/created

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/EntityContentChangeService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/EntityContentChangeService.php
@@ -865,13 +865,19 @@ class EntityContentChangeService extends CoreService
      */
     protected function determineChanger(bool $isReviewer): ?object
     {
+        $user = null;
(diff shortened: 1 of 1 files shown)


AlexMenschoid:
SHouldn't we fix `$this->userFromSecurityUserProvider->fromToken($token); ` here instead of adding a fallback? If the method is unreliable than what's the point? And if the method only works in and for certain cases (and this one is not one of those), then we should replace it altogether.

mussbach:
I absolutely agree with you @AlexMenschoid I fixed this but it was not reviewed https://github.com/demos-europe/demosplan-core/pull/1362

mussbach:
Closed in favor of https://github.com/demos-europe/demosplan-core/pull/1362

PR #1406, opened 2023-05-25
fix (refs T32805): only send email to all users within one customer

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T32805

We need to check that any person is in current customer before sending an email as there are organizations like 'Privatperson' special organization that need to be in multiple customers. But not every user is registered in any of these customers.
Moreover, we should check whether the orga is accepted before we send emails to it.

### How to review/test
Register a new citizen in Customer A and send as platform administrator emails to all users of Customer B. The new citizen should not receive any email

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm

PR #1407, opened 2023-05-25
Use distinct filenames for css entry points

spiess-demos:
Before, the getBundleAndRelatedChunkSplits always returned two entrypoins (css.css and publiccss.css) when being asked for css.css only. that led to those styles being included twice. The reason was that strpos('css.css', 'publiccss.css') returns true... well.

### How to review/test
After running `yarn dev:<project>` and then navigating into the planning area of demosPlan, the styles should not only look good (as they always do), but there should be only one css file included in the html.

### DIFF
--- a/config.webpack.js
+++ b/config.webpack.js
@@ -86,8 +86,8 @@ const bundlesConfig = merge(baseConfig, {
   name: 'main',
   entry: () => {
     return {
-      css: config.stylesEntryPoint,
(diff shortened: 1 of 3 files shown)


muellerdemos:
Thanks this is great. I can confirm this is a working solution !

muellerdemos (approved):

PR #1412, opened 2023-05-26
F T32871 new map layer cannot be created

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T32871

Description: The rpc-map-get-capbilities-schema.json path was wrong und had to be adjusted


dresslerdemos (approved):
Code review only.

PR #1413, opened 2023-05-26
Fix(refs:32940): incorrect view by fullscreen

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T32940

### Description: 

- if you click to fullscreen the chosen scale view shall be shown in fullscreen
- the bug is that it shows always the full map content in fullscreen
- remove the function here, which contains a force of a recalculation of the map viewport, 
- so that the fullscreen view cannot be forced to show another view than the chosen view when fullscreen is not clicked

### PR Checklist

### DIFF
--- a/client/js/components/map/map/DpOlMap.vue
+++ b/client/js/components/map/map/DpOlMap.vue
@@ -390,7 +390,6 @@ export default {
       const extent = view.getProjection().getExtent()
       const center = view.getCenter()
 
-      // Force a recalculation of the map viewport size. This should be called when third-party code changes the size of the map viewport.
       this.map.updateSize()
 
       view.fit(extent, {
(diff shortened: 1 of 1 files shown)


spiess-demos (dismissed):
`this.updateMapInstance()` was added by purpose when implementing DpOlMap for its initial usage within the BasicSettings view ("Ortsbezug"). I do not recall the exact reason i added it there, so it may be fine to remove it here... maybe updating the map instance should be triggered by the outer component if necessary anyways.

So long story short, code lgtm.

client/js/components/map/map/DpOlMap.vue
```diff
@@ -391,6 +390,7 @@ export default {
       const extent = view.getProjection().getExtent()
       const center = view.getCenter()
 
+      // Force a recalculation of the map viewport size. This should be called when third-party code changes the size of the map viewport.
```
muellerdemos, line 393: You can remove this comment because it is kind of redundant. You can see the exact same explanation when hovering over the function in your IDE

muellerdemos (changes requested):
Sorry, just a tiny nitpick. Otherwise this solution fixes the bug and looks fine for me.

client/js/components/map/map/DpOlMap.vue
elmasdemos: done https://github.com/demos-europe/demosplan-core/pull/1413/commits/acb630022b689454783c7873a9a770d435850d3d

PR #1437, opened 2023-05-30
fix (refs T32934): Improve performance on check permissions

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T32934

Checking permissions of addons, creates a huge amount of php requests which leads to very low performance.
This logic can modified to directly request via sql and combine these into a few requests.

### How to review/test
Get to the list of Statements. This should be take less than two seconds.

- [ ] Tests updated/created
- [ ] Update documentation


dresslerdemos (approved):
Code review only.

PR #1440, opened 2023-05-31
bug(refs T32601): Set authored date after setting initial values

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T32601

We need to set the authored date after setting the initial values, because otherwise it is always empty.

### How to review/test
Can't be tested locally.


muellerdemos (approved):
For me this seems to be the working solution. lgtm

PR #1450, opened 2023-06-01
chore: Adjust AdministrationPlaces to work with DpDatatable

salisdemos:
after refactoring the DpDatatable, we had to adjust the usage here and there. This changes had to be done, to work with the new version of DpDataTable

This has to be merged in the same Moment, when https://github.com/demos-europe/demosplan-ui/pull/230 get published.

Related to https://github.com/demos-europe/demosplan-ui/pull/275


sakutademos (commented):
Unfortunately, this PR has some issues, so I created a new one here https://github.com/demos-europe/demosplan-core/pull/1487

PR #1452, opened 2023-06-02
chore: Bump required node engine

salisdemos:
Set required node engine to 16 since this is the lowest Version for some dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,7 +180,7 @@
   },
   "engines": {
     "npm": ">= 6.0.0",
-    "node": ">= 14.15.1"
+    "node": ">= 16.0.0"
   },
   "browserslist": [
     "last 5 versions",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -180,7 +180,7 @@
   },
   "engines": {
     "npm": ">= 6.0.0",
-    "node": ">= 14.15.1"
+    "node": ">= 16.0.0"
```
spiess-demos, line 183: Why not the exact version we currently deploy in the dev container? v16.18.1

salisdemos: My thought was, that if someone else wants to setup the project, It could be less than the 16.18.. in theory. But I will change it. Then we are safe on that side.

c21be77b5f9671e3d8d8bdb3961bb220121e01ff

spiess-demos (approved):
lgtm!

PR #1466, opened 2023-06-02
feat(refs T32104): tweak map attribution placeholder

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T32104

This PR changes the way map attributions are handled. It improves UX by implementing preview of attribution settings and enhances attributions to support automatic displaying the current year.

**Default attribution via translation key:** The translation key "map.attribution.default" may now contain {currentYear} and {linkImprint}. This way, projects that require the link to the imprint can carry on using it this way, while projects that want to display the current year can do this magically. Whenever no other attribution is defined, the maps in demosPlan will now display this attribution.

**Preview of attribution in Customer Settings:** As the other map options (layer, layerUrl) are also reactive, the attribution is also implemented as a live preview here.

**Preview of attribution in Procedure Settings:** Here, the attribution is only updated on save, because the input field is further away from the map, hence no direct correlation must be made.

### How to review/test
As a Plattform Admin User, you may use **{year}** within the attribution field for the public index map. This will be replaced by  the current year in the "classic (sh-style)" public index map.

Planners (or users who can administer procedure templates) may use **{year}** within the attribution field for the public detail map, it will be replaced by the current year in the public detail map.

When no attribution is set via one of those views, the "map.attribution.default" string will be rendered. If that string contains **{currentYear}** or **{linkImprint}**, those will be replaced with the relevant content.

### Linked PRs
- https://github.com/demos-europe/demosplan-project-diplanbau/pull/115

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


client/js/components/map/publicdetail/Map.vue
```diff
@@ -1748,14 +1747,24 @@ export default {
       if (PROJECT && PROJECT !== 'robobsh') {
         controls.push(new MousePosition({ className: this.prefixClass('c-map__mouseposition') }))
       }
+
+      // Add either the attribution from procedureSettings or the attribution defined globally via translation key
+      const currentYear = formatDate(new Date(), 'YYYY')
+      let label = ''
```
salisdemos, line 1753:
```suggestion
      const currentYear = formatDate(new Date(), 'YYYY')
      let label = ''

```

templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_public_participation.html.twig
```diff
@@ -65,7 +65,11 @@
                 attributionControl: false
             });
 
-            L.control.attribution({prefix: '{{ map.mapAttribution|default|trans|wysiwyg }}' || Translator.trans('map.attribution.default', { a: function (msg) { return '<a href="/impressum">' + msg + '</a>' }}) }).addTo(map);
+            L.control.attribution({
+                prefix: '{{ map.mapAttribution|default|replace({'{year}': 'now'|date('Y')})|wysiwyg }}' || Translator.trans('map.attribution.default', {
+                    linkImprint: Routing.generate('DemosPlan_misccontent_static_imprint'),
+                    currentYear: {{ 'now' | date('Y') }}
+                })}).addTo(map);
```
salisdemos, line 72:
```suggestion
                prefix: '{{ map.mapAttribution|default|replace({'{year}': 'now'|date('Y')})|wysiwyg }}' || Translator.trans('map.attribution.default', {
                    linkImprint: Routing.generate('DemosPlan_misccontent_static_imprint'),
                    currentYear: {{ 'now' | date('Y') }}
                })
            }).addTo(map);
```

client/js/components/map/map/DpOlMapLayer.vue
```diff
@@ -56,19 +55,44 @@ export default {
...
+  },
+
   methods: {
     addLayer () {
       if (this.map === null) {
         return
       }
 
-      const source = createSourceTileWMS(this.url, this.layers, this.projection, this.attributions, this.map)
+      const source = this.source = createSourceTileWMS(this.url, this.layers, this.projection, this.defaultAttributions, this.map)
```
salisdemos, line 95: when You add `this.soruce` here, I suggest to drop `source` and pass `this.source` to the method instead

client/js/components/map/publicdetail/Map.vue
```diff
@@ -1748,14 +1747,24 @@ export default {
...
-          tipLabel: this.procedureSettings.copyright
-        }))
+        label = this.procedureSettings.copyright.replace('{year}', currentYear)
+      } else {
+        label = Translator.trans('map.attribution.default', {
+          linkImprint: Routing.generate('DemosPlan_misccontent_static_imprint'),
+          currentYear
+        })
       }
+      controls.push(new Attribution({
```
salisdemos, line 1762:
```suggestion

      controls.push(new Attribution({
```

salisdemos (changes requested):

salisdemos (approved):

PR #1467, opened 2023-06-02
fix: tweak export tab colors

spiess-demos:
the active/hover styles were overridden in diplanbau in such a way that there was no chance spotting the currently active tab. also, pointer.

### How to review/test
in the export modal of the "classic" assessment table, the tabs for docx / pdf should clearly indicate active state.

image


sakutademos (approved):
lgtm!

PR #1468, opened 2023-06-02
fix: also define color for hover state

spiess-demos:
Without this fix, in bobhh the :hover declaration of the base link took precedence, coloring the hovered link blue.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_link.scss
+++ b/demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_link.scss
@@ -52,6 +52,7 @@
         text-decoration: underline;
 
         &:hover {
+            color: $dp-color-white;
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_link.scss
```diff
@@ -52,6 +52,7 @@
         text-decoration: underline;
 
         &:hover {
+            color: $dp-color-white;
             color: $dp-color-alt--light-1;
```
elmasdemos, line 56: maybe you wanted just one of them as hover color?

elmasdemos (changes requested):

salisdemos (approved):

elmasdemos (approved):
thank you, lgtm, code review

PR #1475, opened 2023-06-05
bug(refs T33059): Import vue 

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33059

We need to import Vue in the js files, because otherwise we get an TypeError: Vue.delete is not a function. Unfortunately I don't know the reason for that.

### How to review/test
Go to Verfahren -> Abschnitte and filter some tags. Try to delete the filter and see if it works.

### DIFF
--- a/client/js/store/procedure/SegmentFilter.js
+++ b/client/js/store/procedure/SegmentFilter.js
@@ -7,6 +7,8 @@
  * All rights reserved
  */
 
+import Vue from 'vue'
+
 const SegmentFilterStore = {
   namespaced: true,
   name: 'segmentfilter',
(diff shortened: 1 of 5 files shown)


client/js/store/procedure/SegmentFilter.js
```diff
@@ -7,6 +7,8 @@
  * All rights reserved
  */
 
+import Vue from 'vue'
```
salisdemos, line 10: I thnk this is not the propper way to fix that. Please try `this.$delete` instead

salisdemos (changes requested):

salisdemos (approved):
lets hope that does the trick and it only has to be changed with vue3

PR #1482, opened 2023-06-06
bug(refs T33031): Hide upload files when editable is false

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33031

If the statement is not editable we don't need to show the component for upload, but only the attachments already uploaded.

### How to review/test
Couple a procedure and transfer a statement to VHT. On the detail site, you shouldn't see the upload files function, but only the attachments already uploaded.


hwiem (approved):
lgtm!

PR #1485, opened 2023-06-07
fix(refs T33015): implement date validation check

muellerdemos:
https://yaits.demos-deutschland.de/T33015

Only use the submitter date as authored date if the submitter date is in valid format aka `YYYY-MM-DD` otherwise the default value of `''` will be used as authored date

### PR Checklist


gruenbergerdemos (approved):
lgtm!

PR #1486, opened 2023-06-07
feat(refs T29585): Move recommendation modal to addon and insert hook

gruenbergerdemos:

This PR moves the recommendation modal to its corresponding addon.
This PR also moves following components to demosplan-ui for another addon:

- AddonWrapper
- DpBadge
- DpInsertableRecommendation

The reason is that an addon needs to load an addon, which is technically no problem but the AddonWrapperr must be accessible from a non-core source which is in this case demosplan-ui. Besides this requirement it probably is the correct spot to live for the addon wrapper anyways since this is a very generic component.

It is basically the same with DpBadge and DpInsertableRecommendation. Due to the generic nature of those components I'd want to have them a home in demosplan-ui. Also this way we avoid duplicating components in different repositories, which is quite nasty to handle. 

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-ui/pull/473 (**MUST!!!** be merged first before this and all other linked PRs)
https://github.com/demos-europe/demosplan-addon-demospipes/pull/79
https://github.com/demos-europe/demosplan-addon-segment-oracle/pull/1

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -296,16 +295,17 @@ import {
   VPopover
 } from '@demos-europe/demosplan-ui'
 import { mapActions, mapMutations, mapState } from 'vuex'
+import DpAddonWrapper from '../../../../../addons-bkp/cache/demosplan-addon-demospipes-f_T29585_similar_segments_ui/client/DpAddonWrapper.vue'
```
salisdemos, line 298: this looks wrong.
when we move the addonWrapper to dp-ui, we have to be able to use `import DpAddonWrapper from "@demos-europe/demosplan-ui"` here too

salisdemos:
Beside that I am not sure if demosplan-ui is the correct spot:
Why does

>  "the AddonWrapperr must be accessible from a non-core source"
?

salisdemos:
> This PR also moves following components to demosplan-ui for another addon:
> 
> AddonWrapper
> DpBadge
> DpInsertableRecommendation

I don't think that `DpInsertableRecommendation` should live in denmosplan-ui. Its far to specific.

salisdemos (changes requested):

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: adjusted these dev leftovers here: https://github.com/demos-europe/demosplan-core/pull/1486/commits/d149bc96ccca6d37cd5c0e6ed000903b42e9d0d4

Also re-added DpInsertableRecommendations.vue here: https://github.com/demos-europe/demosplan-core/pull/1486/commits/cf1105528444515355d7e72e56542290400514a2

muellerdemos:
I think addons should be self sufficient. And for this case an addon must load another addon (which is a good thing imho if we want a modular system). Also my experience is that we should avoid copying components into addons which is already the case with DpSimplifiedNewStatementForm which exists in 3 repositories as copied component (and this can be really troublesome).

salisdemos:
> DpSimplifiedNewStatementForm which exists in 3 repositories as copied component 

I fully agree. We may should find a way to centralize it somewhere.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -114,11 +114,27 @@
...
-            <addon-wrapper
-              ref="addonWrapper"
-              :addon-props="addonProps"
-              hook-name="recommendation.modal"
-              :ref-component="refRecModal" />
+            <dp-modal
+              ref="recommendationModal"
+              class="recommendation-modal"
+              content-classes="u-2-of-3">
+              <div class="display--inline-block">
```
spiess-demos, line 121:
```suggestion
              <div class="inline-block">
```

client/js/bundles/document/elementsAdminList.js
```diff
@@ -11,7 +11,7 @@
  * This is the entrypoint for elements_admin_list.html.twig
  */
 
-import AddonWrapper from '@DpJs/components/addon/AddonWrapper'
+import AddonWrapper from '@demos-europe/demosplan-ui'
```
spiess-demos, line 14: Now, AddonWrapper is imported from two places. here, from demosplan-ui, and in administrationDashboard.js from core.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -114,12 +114,27 @@
...
               ref="recommendationModal"
-              :procedure-id="procedureId"
-              :segment-id="segment.id"
-              @insert-recommendation="text => modalProps.appendText(text)" />
+              class="recommendation-modal"
+              content-classes="u-2-of-3">
+              <div class="display--inline-block">
+                <h3 class="u-mb">
+                  {{ Translator.trans('segment.recommendation.insert.similar') }}
+                  <dp-icon icon="ai" />
```
spiess-demos, line 124: well... i can't guess how the UI looks, straight from the code... but the AI icon should be reserved for places that are really ai-powered which is only the case for the oracle.recommendation tab. i suggest we look into that tomorrow.

client/js/components/statement/segments/DpInsertableRecommendation.vue
```diff
@@ -1,62 +1,77 @@
 <license>
-  (c) 2010-present DEMOS plan GmbH.
```
spiess-demos, line 2: this file can be removed from core, if i get this correctly?

spiess-demos (changes requested):

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: this has been adjusted here: https://github.com/demos-europe/demosplan-core/pull/1486/commits/dfd06a2e268a73699655a893b43a847de83188b1

client/js/bundles/document/elementsAdminList.js
muellerdemos: we decided to not put the addonwrapper into demosplan-ui hence the path has been adjusted here: 
https://github.com/demos-europe/demosplan-core/pull/1486/commits/0e5c334bf1045388cdc948a51732f401fe8525d3

client/js/components/statement/segments/DpInsertableRecommendation.vue
muellerdemos: Like the addon wrapper we decided to keep this file in the core and copy it to places where we need them

client/js/components/addon/AddonWrapper.vue
```diff
@@ -23,6 +15,15 @@ export default {
...
 
   props: {
+    /**
+     * The addonProps prop will be bound to the addon component to add props dynamically.
+     */
+    addonProps: {
+      type: Object,
+      required: false,
+      default: () => {}
+    },
```
salisdemos, line 25: even though its not new. How does this works. The core should not know anything about the addon..

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -114,12 +114,38 @@
...
+                    hook-name="tag.recommendation.tab"
+                    :addon-props="addonProps"/>
+                </dp-tab>
+                <dp-tab
+                  :label="Translator.trans('segments.recommendations.oracle')"
+                  id="recContent">
+                  <addon-wrapper
+                    hook-name="oracle.recommendation.tab"
+                    :addon-props="addonProps"/>
+                </dp-tab>
```
salisdemos, line 146: we should need ony one addon-wrapper here.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -364,6 +396,11 @@ export default {
 
   data () {
     return {
+      activeId: 'recTags',
+      addonProps: {
+        segmentId: this.segment.id,
```
salisdemos, line 401: I think this should fetched via store. We should design it like so, that we change the core as less as possible when writing an addon

package.json
```diff
@@ -47,7 +47,7 @@
     "lint:scss": "stylelint {demosplan/DemosPlanCoreBundle/Resources/client/scss,projects}/**/*.scss"
   },
   "dependencies": {
-    "@demos-europe/demosplan-ui": "^0",
+    "@demos-europe/demosplan-ui": "^0.1.12",
```
salisdemos, line 50: I assume this should stay as is

salisdemos (changes requested):

demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_modal.scss
```diff
@@ -84,4 +84,12 @@
...
         height: auto;
     }
+    &.recommendation-modal {
+        li:hover {
+            background-color: $dp-color-background-light;
+            .badge {
+                background-color: $dp-color-main;
+            }
+        }
+    }
```
spiess-demos, line 94: This must be solved in another way. I guess you're styling the recommendation items here? as the markup is located inside the addon, we must not define styles in core. This would be easy to implement once we have tailwind colors... however for now there is no easy way i guess :(

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: The addon wrapper could fetch more than one addon from a single hook but we can not add multiple hooks to a single addon wrapper. Since these addons will be provided by two different addon repositories (and therefore have 2 different hooks) we need the addon wrapper twice

client/js/components/addon/AddonWrapper.vue
muellerdemos: Sometimes addon components need props which and therefore those props must be set dynamically. This way we can pass down required props directly from the wrapper to the corresponding addon component. However what will be done with those props in the addon is something the core does not know before the addon is mounted

package.json
muellerdemos: I changed this back to a non specific version here: https://github.com/demos-europe/demosplan-core/pull/1486/commits/2a273471cb6e6a28185ccf5e72f046d0b3972817

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: But isn't it like we need the id to fetch a single segment from the store? Currently the addon requires the segment.id from its corresponding parent to set the mutation in the store. I'm not quite sure how we can design this otherwise.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: the ai icon has been removed here https://github.com/demos-europe/demosplan-core/pull/1486/commits/564be88fd18b61f2f5f15b60268a73569a079c64 and will be added in the corresponding addon directly

demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_modal.scss
muellerdemos: Hm, what would your proposal be in this case? I can't come up with a better idea than either hardcode these styles in a style block for each addon that has these `li` and `badge` classes or to sopy the needed scss (but we would need all those scss variable too in that case?)

client/js/components/addon/AddonWrapper.vue
salisdemos: but how do we know which props to pass...
I would like to discuss somewhere if we can find another solution wich is more flexible

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
salisdemos: and therefor both addons should mount on the same hook. Thats the idea behind it.
we don't need a hook-name for each repos. We should have several dedicated hooks in core where any addon can register to.

spiess-demos: and the DpTag component would then be the outermost component for each addon?

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -414,6 +450,10 @@ export default {
       return this.segment.relationships.comments?.data?.length || 0
     },
 
+    isAiTabActive () {
```
spiess-demos, line 453: not needed anymore.

client/js/components/statement/segments/DpInsertableRecommendation.vue
```diff
@@ -1,62 +1,77 @@
...
+for more information see the license file.
 
-  All rights reserved
+All rights reserved
 </license>
 
 <template>
-  <li class="flex items-start space-inline-xs u-pr-0_5">
-    <div class="min-width-m">
+  <li class="flex flex-items-start space-inline-xs u-pr-0_5">
```
spiess-demos, line 11:
```suggestion
  <li class="flex items-start space-inline-xs u-pr-0_5">
```

client/js/components/statement/segments/DpInsertableRecommendation.vue
```diff
@@ -1,62 +1,77 @@
...
+      v-if="fromOtherProcedure"
+      class="min-width-m">
       <i
-        v-if="fromOtherProcedure"
         v-tooltip="{
           boundariesElement: body,
-          content: Translator.trans('segment.recommendation.other.procedure') + ': ' + procedureName,
-          classes: 'z-ultimate'
+          content: Translator.trans('segment.recommendation.other.procedure') + ': ' + procedureName ?? '',
+          classes: 'u-z-super'
```
spiess-demos, line 30:
```suggestion
          classes: 'z-ultimate'
```

client/js/components/statement/segments/DpInsertableRecommendation.vue
```diff
@@ -1,62 +1,77 @@
...
+    </div>
+    <div
+      class="flex-grow"
       v-cleanhtml="recommendationText" />
     <div class="flex flex-nowrap space-inline-s">
       <button
         class="btn--blank o-link--default"
         :aria-label="Translator.trans(isExpanded ? 'dropdown.close' : 'dropdown.open')"
-        v-tooltip="{ boundariesElement: body, content: Translator.trans(isExpanded ? 'dropdown.close' : 'dropdown.open'), classes: 'z-ultimate' }"
+        v-tooltip="{ boundariesElement: body, content: Translator.trans(isExpanded ? 'dropdown.close' : 'dropdown.open'), classes: 'u-z-super' }"
```
spiess-demos, line 50:
```suggestion
        v-tooltip="{ boundariesElement: body, content: Translator.trans(isExpanded ? 'dropdown.close' : 'dropdown.open'), classes: 'z-ultimate' }"
```

client/js/components/statement/segments/DpInsertableRecommendation.vue
muellerdemos: adjusted here 
https://github.com/demos-europe/demosplan-core/pull/1486/commits/17efc08386b512ea985610420f48c61e3437585d

client/js/components/statement/segments/DpInsertableRecommendation.vue
muellerdemos: done here https://github.com/demos-europe/demosplan-core/pull/1486/commits/42e402551ecb94e9b7ffb98f4e5726c3d4f0e73b

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: Well I see what was the issue and did a little refactoring here:
https://github.com/demos-europe/demosplan-core/pull/1486/commits/93bc79d6b3b83d1df89db41e1e55a5fe69c96482
https://github.com/demos-europe/demosplan-core/pull/1486/commits/118622924c1cd6775a8000a1e8a5778b914ba333
https://github.com/demos-europe/demosplan-core/pull/1486/commits/467c192120f6c5f6dbaac1508b324cd4aeb31b88
https://github.com/demos-europe/demosplan-core/pull/1486/commits/e84bd07d3be068875f66d2299b28106a869c2134

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: has been removed here https://github.com/demos-europe/demosplan-core/pull/1486/commits/564be88fd18b61f2f5f15b60268a73569a079c64

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -601,6 +642,10 @@ export default {
         })
     },
 
+    setActiveTabId () {
+      this.activeId = this.activeId === 'recContent' ? 'recTags' : 'recContent'
```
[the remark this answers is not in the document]
muellerdemos: done here https://github.com/demos-europe/demosplan-core/pull/1486/commits/6e2efd2cbbb2314afcd24f632765199fc3d29353

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -543,12 +585,26 @@ export default {
         })
     },
 
+    loadAsyncComponents (hookName) {
```
salisdemos, line 588: isn't this a generic method which should live in `libs` next to `loadAddonComponents` ??

and then onMounted we only do something like 
```
  this.asyncComponents = loadAsyncComponents(hookName)
```

salisdemos (changes requested):

client/js/lib/addon/loadAddonComponents.js
```diff
@@ -21,3 +22,16 @@ export default function loadAddonComponents (hookName) {
...
+  for (const addon in addons) {
+    eval(addon)
+    Vue.options.components[addon.entry] = window[addon.entry].default
+
+    return {
+      name: addon.entry,
+      options: addon.options ?? ''
+    }
+  }
+}
```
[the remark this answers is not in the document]
salisdemos: I think otherwise we alway get only one addon. Or not?

client/js/lib/addon/loadAddonComponents.js
```diff
@@ -0,0 +1,34 @@
...
+  const params = {
+    hookName: hookName
+  }
+
+  const addons = []
+
+  return dpRpc('addons.assets.load', params)
+    .then(response => checkResponse(response))
+    .then(response => {
+      const result = response[0].result
```
salisdemos, line 14: can't test it, but shouldn't be `const addons = []` within the `then` ?

And if it doesn't work accessing `Vue` here, we could pass it in the Method. But I think it will work like it is...

salisdemos (changes requested):

client/js/lib/addon/loadAddonComponents.js
muellerdemos: this https://github.com/demos-europe/demosplan-core/pull/1486/commits/34048d60da081d14a985a4bd06fae3a6ab8eb573 is fixing the addon loading mechanism per se. However there is now a new demosplan-ui error...

demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_modal.scss
```diff
@@ -84,4 +84,14 @@
         margin: $inuit-base-spacing-unit 0 0;
         height: auto;
     }
+
+    // can be removed in favor of implementing a contrast prop in DpLabel
```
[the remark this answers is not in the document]
muellerdemos: thanks, has been adjusted here https://github.com/demos-europe/demosplan-core/pull/1486/commits/0332f8806d20133938e029e0ddceb8365d6b552a

client/js/lib/addon/loadAddonComponents.js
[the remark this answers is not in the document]
muellerdemos: this has been fixed here https://github.com/demos-europe/demosplan-core/commit/34048d60da081d14a985a4bd06fae3a6ab8eb573 too

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: for refs we can acces them like that

spiess-demos (changes requested):

client/js/components/statement/segments/DpInsertableRecommendation.vue
```diff
@@ -1,27 +1,47 @@
...
+for more information see the license file.
 
-  All rights reserved
+All rights reserved
 </license>
 
 <template>
-  <li class="flex items-start space-inline-xs u-pr-0_5">
-    <div class="min-width-m">
+  <li class="flex flex-start space-inline-xs u-pr-0_5">
```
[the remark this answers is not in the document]
muellerdemos: done here https://github.com/demos-europe/demosplan-core/pull/1486/commits/879bb1cf4d93b1f0a72ac5a2a4043746196b0891

spiess-demos (dismissed):
lgtm (code review only).

client/js/lib/addon/loadAddonComponents.js
```diff
@@ -0,0 +1,33 @@
+import { checkResponse, dpRpc } from '@demos-europe/demosplan-ui'
+import Vue from 'vue'
+
+export default async function loadAddonComponents (hookName) {
```
salisdemos, line 4: I guess if this is async, we need an await somewhere

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -774,6 +819,12 @@ export default {
           this.selectedAssignee = this.assignableUsers.find(user => user.id === this.segment.relationships.assignee.data.id)
         }
       })
+
+    Promise.resolve(loadAddonComponents('oracle.recommendation.tab'))
```
salisdemos, line 823: if I get it right, we could use an `await` within loadAddonComponents (see comment below) and then we should not need the Preomise.resolve here anymore (which looks a bit strange)

But if the change leads to issues. let me know and we leave it like it is.

salisdemos (changes requested):

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
muellerdemos: Not so sure why but without the Promise.resolve, the asyncComponents the returned array of addons in will be passed to vue as an Promise object. I can at the same time, log the addons object in the `loadAddonComponents.js` file which logs not a Promise but the actual array. If I keep it as it is now, it works. However I think I have implemented all needed changes in the core for this PR besides that.

salisdemos (approved):

PR #1488, opened 2023-06-08
bug(refs T33015): check if authored date is valid when converting pdfs

muellerdemos:
https://yaits.demos-deutschland.de/T33015

- Since I forgot that the datepicker only accepts the format 'DD.MM.YYYY', the implementation of passing the correct value format was missing and added now.


hwiem (approved):
lgtm

PR #1490, opened 2023-06-08
F t29585 adding addon controller paths

AlexMenschoid:

This new (optional) parameter in the addon-manifest allows addons to specify the locations of controllers so that the core can add the routes from annotations to the core router. This opens up the possibility of non-standard namespaces for controllers. Addons can either specifiy one path as a string or multiple paths as an array. If the parameter is not used, it defaults to the Symfony standard of `/src/Controller` which is the behaviour we used before. All addons should still work the same.

### How to review/test
Installing an addon withouth the new `controller_paths` parameter should work as before. But if there is one or more values in that parameter, routes from controllers in these locations should show up in the router. You can easily check after installing and activating the addon via the command `ewm debug:router <ROUTE NAME FROM CONTROLLER ANNOTATION>`.


demosplan/DemosPlanCoreBundle/Addon/AddonRoutingLoader.php
```diff
@@ -56,10 +58,18 @@ protected function configureRoute(Route $route, ReflectionClass $class, Reflecti
      */
     private function addControllers(mixed $addonInfo, RouteCollection $routeCollection): void
     {
-        $controllerDir = $addonInfo->getInstallPath().self::PATH_TO_CONTROLLERS_FROM_ADDONROOT;
-        $controllerPath = AddonPath::getRootPath($controllerDir);
-        if ('' !== $addonInfo->getInstallPath() && is_dir($controllerPath)) {
-            $routeCollection->addCollection($this->load($controllerPath));
+        // Without any installPath, we can't register any routes, so we can skip this step
+        if ('' === $addonInfo->getInstallPath()) {
+            return;
```
graupnerdemos, line 63: Should we log this? Ideally, an exception should be thrown before we even end up here, but if we do, a silent return may not be easy to find if this fails in production.

graupnerdemos (changes requested):

demosplan/DemosPlanCoreBundle/Addon/AddonRoutingLoader.php
```diff
@@ -36,7 +36,9 @@ public function loadRoutes(): RouteCollection
     {
         $routeCollection = new RouteCollection();
         foreach ($this->addonRegistry->getAddonInfos() as $addonInfo) {
-            $this->addControllers($addonInfo, $routeCollection);
+            if ($addonInfo->isEnabled()) {
```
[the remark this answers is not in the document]
AlexMenschoid: For me, it was enough to clear the cache after enabling the addon to force a frontend-build, which we have to do anyway after enabling an addon.

Also, we need to check if an addon is enabled first as we would otherwise add functional routes that are not allowed but available.

demosplan/DemosPlanCoreBundle/Addon/AddonRoutingLoader.php
AlexMenschoid: After looking at this a bit, I realized that `install_path` can never be an empty string as we set it like this:
''install_path' => 'addons/vendor/'.$addonName,'

Now the question becomes, do we want to add a better evaluation of this in this PR or would that be better handled in a follow-up improvement. If you don't mind, I would prefer to put that in a PR of its own as it has not really anything to do with this one.

demosplan/DemosPlanCoreBundle/Addon/AddonRoutingLoader.php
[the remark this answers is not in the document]
MoritzMandler: I tested it myself again - this works fine.

demosplan/DemosPlanCoreBundle/Addon/AddonRoutingLoader.php
mussbach: Id' suggest to delete this if clause because it never does anything it seems to do and do the error handling in another PR as it seems to be non-trivial in this case

demosplan/DemosPlanCoreBundle/Addon/AddonRoutingLoader.php
```diff
@@ -56,10 +58,18 @@ protected function configureRoute(Route $route, ReflectionClass $class, Reflecti
      */
     private function addControllers(mixed $addonInfo, RouteCollection $routeCollection): void
     {
-        $controllerDir = $addonInfo->getInstallPath().self::PATH_TO_CONTROLLERS_FROM_ADDONROOT;
```
mussbach, line 59: You could delete local constant PATH_TO_CONTROLLERS_FROM_ADDONROOT now

demosplan/DemosPlanCoreBundle/Addon/ManifestConfiguration.php
```diff
@@ -27,6 +27,11 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootChildren->scalarNode('description')->defaultValue('')->end();
         $rootChildren->scalarNode('entry')->isRequired()->end();
         $rootChildren->scalarNode('permissionInitializer')->isRequired()->end();
+        $rootChildren->arrayNode('controller_paths')
+            ->defaultValue(['/src/Controller'])
```
mussbach, line 31: Is the initial `/` correct here?

mussbach: I saw it working with an existing addon

demosplan/DemosPlanCoreBundle/Addon/AddonRoutingLoader.php
mussbach: @graupnerdemos @AlexMenschoid could you please reach a decision so this PR could be merged? We need this one for a recent Addon

graupnerdemos: We can add the evaluation at a later point, as it's not too critical at the moment.

PR #1491, opened 2023-06-08
fix (refs T33143): uses parent statement submit date

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T33143

If the entity is a segment, it now uses the submit date from the parent statement and not its own date.


mussbach (dismissed):
lgtm, thx

mussbach:
Please double check the branch. Should this be main (might well be)

AlexMenschoid:
Thanks for looking out, but main is correct. This is only relevant for newer versions of the demospipes-addon which should only become relevant next sprint.

PR #1495, opened 2023-06-09
feat (refs T33162): fix year replacement of copyright

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T33162

Change the placeholfer for the current year from {year} to {currentYear}. It's more meaningful and {currentYear} is already used in the database, which also causes the error that {currentYear} was not replaced because the modified code assumed {year}.

### How to review/test
code review
check the web interface of diplanbau, it should now show 2023 instead of {currentYear} in the copyright of maps.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
be looks good to me

salisdemos (approved):

PR #1504, opened 2023-06-12
fix (refs T33144): Avoid interpreting a 0 as timestamp.

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T33144

Actually authored date can be null/empty.
On trying to get the timestamp of auhtoredDate of a Statement/Segment, 0 will be returned in case of no date was set.
Because of this is actually a valid timestamp, it will be interpreted as '01.01.1970', which is not correct or intended.
To solve this, I simply use another method, which returns me the already to string converted authored date, without this "special" kind of interpretation and returning simple an empty string in case of no date is set.

+ Removing the now obsolete conversion from the export logic.

### How to review/test
Export statements with an empty authored date.

- [ ] Tests updated/created


breierDemos (approved):
lgtm

PR #1505, opened 2023-06-12
feat(refs T33153): Revert custom label for multiselect

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33153

Vue multiselect uses here option and not props.
Also correct wrong color name.

Do not merge before https://github.com/demos-europe/demosplan-ui/pull/290


sakutademos (dismissed):
lgtm!

PR #1506, opened 2023-06-12
fix(T33012): issue with an empty tags dropdown

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33012

**Description:**  This PR fixes the issue with an empty tags dropdown. The options for the DpMultiselect need to be passed correctly, with the data structure: {id: '', name: ''}.

- convert the data before passing it to the multiselect


gruenbergerdemos (commented):
I think this should be fixed with https://github.com/demos-europe/demosplan-ui/pull/290, because the group related props (`group-label` etc) are not passed to vue multiselect and we don't have the options available via the props in the slot.

PR #1507, opened 2023-06-12
F remove fs margin

spiess-demos:
This corrects some margin-top in fullscreen mode, as well as a strangely misplaced icon.

### How to review/test
in ewm, go to the "Erwiderungen verfassen"-view of a statemet  that already has segments.

- the button which opens the edit mode should be not padded strangely. it should look this way:
image

- proceed to fullscreen mode. there should not be any margin-top above the light blue area


sakutademos (approved):
lgtm

hwiem (approved):
lgtm!

PR #1512, opened 2023-06-13
chore(refs T32012): Remove unnecessary Use of vuex

salisdemos:

For some reason we initiallized vuex a couple of times.
In the End we don't have to init it at all because the vuex-json-api-lib does it for us.

In vue3 initializing it more than once throws an error. That didn't happen in vue2


muellerdemos (approved):
looks legit to me

PR #1515, opened 2023-06-13
bug(refs T33031): Hide upload files when editable is false (#1482)

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33031

Cherry pick from release (#1482), because the fix is also wanted in the release now.


hwiem (approved):
lgtm

PR #1518, opened 2023-06-13
fix (refs T33109): Fix adding of similar statement submitters

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T33109

Flush similar statement submitter on create a new statement, to ensure added submitters, will be stored in DB.

### PR Checklist
- [x] Tests updated/created


breierDemos (approved):
lgtm code review only

PR #1531, opened 2023-06-14
chore(refs T19759): Use the new way to init Vue3

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T19759

Use the new way to init Vue3
https://v3-migration.vuejs.org/migration-build.html 
Step 8. Update app entry to use [new global mounting API](https://v3-migration.vuejs.org/breaking-changes/global-api.html#a-new-global-api-createapp).


client/js/InitVue.js
```diff
@@ -79,7 +57,45 @@ function initialize (components = {}, storeModules = {}, apiStoreModules = [], p
...
+    app.component('NotifyContainer', NotifyContainer)
+    app.component('DpAccordion', DpAccordion)
+    app.component('DpFlyout', DpFlyout)
+
+    Object.values(components).forEach(comp => {
+      if (comp) {
+        app.component(comp.name, comp)
+      } else {
+        console.log('at least one is undefined', components)
+      }
```
muellerdemos, line 93:
```suggestion
    Object.keys(components).forEach(comp => {
      if (components[comp]) {
        app.component(components[comp].name, components[comp])
      } else {
        console.log(`${components[comp]} is undefined}`, components)
      }
```

muellerdemos: Just out of curiosity, wouldn't it be beneficial if we could determine the problematic component by adjusting this logic a little bit like suggested?

muellerdemos (changes requested):
Nice, config looks legit to me, however to let the build finish without errors I had to adjust `client/js/components/statement/assessmentTable/DpFilterModal.vue`:58

from: `id="userFilterSets"v-model="selectedUserFilterSet"` to:
`id="userFilterSets"
v-model="selectedUserFilterSet"`

This error is also present in the target branch `f_vue3_compat_mode_setup` hence I suggest to adjust this here.

client/js/components/statement/assessmentTable/DpFilterModal.vue
```diff
@@ -187,7 +188,7 @@
         </div>
 
         <!-- Button row -->
-        <div class="text--right space-inline-s">
+        <div class="text-right space-inline-s">
```
muellerdemos, line 191: i think this should also be
```suggestion
        <div class="text--right...
```
like in :79

muellerdemos (changes requested):

salisdemos:
> Nice, config looks legit to me, however to let the build finish without errors I had to adjust `client/js/components/statement/assessmentTable/DpFilterModal.vue`:58
> 
> from: `id="userFilterSets"v-model="selectedUserFilterSet"` to:
> 
> ```
> id="userFilterSets"
> v-model="selectedUserFilterSet"
> ```
> 
> This error is also present in the target branch `f_vue3_compat_mode_setup` hence I suggest to adjust this here.

done here 27d61e0284fccb8e840d288579f0044595ddd5d8

client/js/components/statement/assessmentTable/DpFilterModal.vue
spiess-demos: It is `text-right`, as the respective utility classes [have been moved to tailwind](https://github.com/demos-europe/demosplan-core/pull/1544).

muellerdemos (approved):
concerns were addressed and this lgtm

PR #1533, opened 2023-06-14
fix: Change check for response status

salisdemos:
For some reasons, the good old wy to check against OK doesn't work on one instance. checking against 200 is better anyway


mussbach (approved):
lgtm, thx. Code Review and I saw it work

PR #1541, opened 2023-06-15
chore: provide loggedin to security user to be able to use app.user.loggedin in twig

mussbach:
Some twig templates use app.user.loggedin to receive the login status of a user. With the refactoring to a distinct SecurityUser this was broken.

### How to review/test
Call any page in dev mode where app.user.loggedin is used in twig like paragraph detail page in public detail

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


graupnerdemos (approved):
Awesome. Thanks.

PR #1544, opened 2023-06-15
Use Tailwind utility classes

spiess-demos:
After enabling Tailwind in demosplan-core, we can drop our handcrafted utility classes. 

As a first step, this PR switches to Tailwind utility classes for all cases where it is possible to do so without further preparation (because the classes are designed the same way and map to static properties).

For utility classes that depend on design token, and for those that need responsive modifiers, we first have to do some configuration in Tailwind, later on.

**How to review**

Possibly this PR is best reviewed on a per-commit basis.

**Changed utilities:**

- caps -> uppercase
- cursor--move -> cursor-move
- cursor--default -> cursor-default
- cursor--grab -> cursor-grab
- cursor--help -> cursor-help
- cursor--pointer -> cursor-pointer
- flex-basis-20 -> basis-1/5
- flex-column -> flex-col
- flex-content-between -> justify-between
- flex-content-end -> justify-end
- flex-content-start -> justify-start
- flex-grow -> grow
- flex-inline -> inline-flex
- flex-item-end -> ml-auto
- flex-items-center -> items-center
- flex-items-end -> items-end
- flex-items-start -> items-start
- flex-shrink-2 -> shrink-2
- fontstyle--normal -> not-italic
- overflow-word-break -> break-words
- position--absolute -> absolute
- position--inherit ->  wech
- position--relative ->  relative
- position--sticky -> sticky
- text-decoration-none -> no-underline
- text-decoration-underline -> underline
- text-decoration-underline--hover -> hover:underline visibility--hidden -> invisible
- u-valign--top -> align-top
- u-valign--bottom
- u-valign--middle
- u-valign--baseline
- u-valign--text-bottom
- u-valign--text-top
- u-valign--sub
- display--none -> hidden
- display--block -> block
- display--inline-block -> inline-block
- display--inline -> inline
- display--flex -> flex
- display--table -> table
- display--block-palm -> removed
- cf -> flow-root
- float--right -> float-right
- float--left -> float-left
- opacity-7 -> opacity-70
- text--[left,center,right,justify] -> text-[left,center,right]


demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_statement.scss
```diff
@@ -170,7 +170,7 @@
             }
         }
 
-        a.display--none {
+        a.hidden {
             display: none;
         }
```
salisdemos, line 175: can't we remove this since it comes as utility from tailwind?

muellerdemos (approved):
thanks for so much work, however these changes look reasonable to me but I am not sure how to review more than 300 files better than this

demosplan/DemosPlanCoreBundle/Resources/client/scss/utility/_float.scss
```diff
@@ -9,20 +9,10 @@
 
 //  Utility classes: float
 
-.float--right {
-    float: right !important;
-}
+// @TODO when refactoring to tailwind, we must first define breakpoints in Tailwind.
 
-.float--right-lap-up {
+.float-right-lap-up {
```
salisdemos, line 14: couldn't find any occurrence. Can You tell me where its used?

salisdemos (changes requested):

demosplan/DemosPlanCoreBundle/Resources/client/scss/utility/_float.scss
spiess-demos: - DpPublicSurvey
- ConsolidateModal
- listentry_header
- list_users_statements

demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_statement.scss
spiess-demos: i'm pretty sure this is some "stupid but important" specificity hack - in the statement modal there are plenty `nojs-display--block` things which mess around with specificity.

However as i could not even track down [the dark origin of this code](https://github.com/demos-europe/demosplan/commit/ac73f863ec33c21253eef46a0b08323a3b4934a7), i removed it! 03a24ce2711f8c5135d6b5c5e325f1f92aaadfcb

salisdemos (dismissed):

PR #1551, opened 2023-06-16
feat: remove redundant components from core

muellerdemos:
Both PDF Import and Email Import components are not needed anymore in the core hence should be removed which is basically what this PR is doing.


salisdemos (approved):
Thanks

PR #1555, opened 2023-06-16
rector: update code to use php 8.1 features

mussbach:
This PR applies rector set `LevelSetList::UP_TO_PHP_81` to the codebase

### How to review/test
run phpstan


mussbach:
I would suggest to ignore the phpstan level 2 issues, as they have nothing to do with the changes of this PR

PR #1562, opened 2023-06-19
fix(refs T33234): Remove incorrect late permission evaluation

graupnerdemos:
**Ticket:** https://yaits.demos-deutschland.de/33234

The addon hook configurations do not yet provide permission information. If we were to evaluate those, we should filter the scripts in the RPC method instead of loading all scripts on the users' machine and evaluating on the client side.

In addition, previously no import tabs would have been loadable from addons at all if the excel import were disabled.

Yes, I know this PR adds a TODO. Please be gentle with me. I created a [matching Ticket](https://yaits.demos-deutschland.de/T33332).

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/procedure/AdministrationImport/AdministrationImport.vue
+++ b/client/js/components/procedure/AdministrationImport/AdministrationImport.vue
@@ -109,7 +109,6 @@ export default {
   computed: {
     availableImportOptions () {
       return [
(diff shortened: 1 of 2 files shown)


salisdemos (approved):
Thanks. And yes, we have to do that.

PR #1565, opened 2023-06-19
fix: Allow complete permission evaluation without valid user

graupnerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33234

We cannot evaluate the permissions via the db without an existing user and need to revert to evaluating directly with edt

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


dresslerdemos (approved):
Code review only.

PR #1566, opened 2023-06-19
refactor: prepare for EDT 0.18 update

dresslerdemos:
**Commits can be reviewed individually if preferred**

Prepares dplan for EDT >0.18.

Previously to fetch entities via resource types (i.e. automatically check for access violations) you would call the `EntityFetcher` with your resource type instance. The `EntityFetcher` would call some methods of the resource type to check for access violations and then generate DQL queries to fetch the actual entities.

Now to fetch entities via resource types you call the resource type directly, which will check for access violations internally and uses the corresponding entity repository to fetch the actual entities.

This approach has multiple advantages:

* it places the code to fetch entities where it belongs: into the corresponding repository. Resource types still focus on access management and leave the fetching details to the repository implementation.
* it is also a preparation to reduce the surface area of resource types, as there will be no need to expose methods to check access violations when they are used inside the resource type only.
* this allows to define a different data source for each resource type implementation. Currently all resource types will use Doctrine and DQL to fetch entities. However, by handling the fetching of entities inside the resource type this can be adjusted where needed, which makes it possible to define a resource type that uses Elasticsearch, Redis or anything else as data source.

Please note that this PR is still a preparation for EDT >0.18 only and the full advantages may not be in effect yet.

### How to review/test

What I did:

* phpstan level 1 runs without concerns locally (`bin/ewm dp:phpstan -l 1`)
* some testing in local EWM (procedure list, statement list + search, segment list + search + filtering)

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation: https://github.com/demos-europe/demosplan-documentation/pull/40


demosplan/DemosPlanCoreBundle/Logic/Faq/FaqService.php
```diff
@@ -20,12 +20,13 @@
 use demosplan\DemosPlanCoreBundle\Entity\User\Role;
 use demosplan\DemosPlanCoreBundle\Entity\User\User;
 use demosplan\DemosPlanCoreBundle\Exception\CustomerNotFoundException;
-use demosplan\DemosPlanCoreBundle\Logic\ApiRequest\EntityFetcher;
```
martendemos, line 23: It seems, the entityFetcher is still needed here?!
`return $this->entityFetcher->listEntitiesUnrestricted($className, $conditions, [$sortMethod]);`
in line 168

martendemos (changes requested):
- One missing call found in FaqService.php
- Documentation was not adjusted.

dresslerdemos:
@martendemos: previously I forgot to create the PR for the documentation branch, here it is: https://github.com/demos-europe/demosplan-documentation/pull/40

PR #1567, opened 2023-06-20
fix remove unused permission

bakridemos:
Description: 
remove the permission 'feature_export_statement_with_datasheet': this permission ist not used and the datasheets related logic is already moved from core, it will be used as an addon in the relevant projects.

- [X] Update documentation


mussbach (approved):
lgtm, thx

PR #1573, opened 2023-06-20
fix remove duplicated method

bakridemos:
Description: 

- move the method 'generateDownloadFilename' to a new service 'NameGenerator.php'
- remove duplicated method

Note: these changes are not made randomly, this method can be used by an addon in the future, otherwise it was duplicated and it will be better to move it to a service


demosplan/DemosPlanCoreBundle/Controller/Forum/DemosPlanReleaseController.php
```diff
@@ -841,7 +845,7 @@ public function exportReleaseAction(Environment $twig, string $releaseId)
         $response->setContent($bom.$response->getContent());
         $filename = 'export_'.$part.'_'.date('Y_m_d_His').'.csv';
         $response->headers->set('Content-Type', 'text/csv');
-        $response->headers->set('Content-Disposition', $this->generateDownloadFilename($filename));
+        $response->headers->set('Content-Disposition', $pdfNameService->generateDownloadFilename($filename));
```
mussbach, line 848: Why did you choose PdfNameService? In this case it is used to generate a name for a csv file which seems weird

mussbach: I don't think so as in this case the service name would be weird when generating Filenames for pdf files. I think it needs to be something neutral

bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/1573/commits/ec3512bd9a27ab7cfbcb347b1555ce7ef448cced

demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
```diff
@@ -1669,8 +1672,12 @@ protected function replacePhaseByPhaseNameForVotedStatementList(array $statement
      *
      * @throws Exception
      */
-    protected function createPdfDraftStatement($draftStatementList, $type, Procedure $procedure, $itemsToExport = null)
-    {
+    protected function createPdfDraftStatement(
+        $draftStatementList, $type,
+        Procedure $procedure,
+        $itemsToExport = null,
+        NameGenerator $nameGenerator
```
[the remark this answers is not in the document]
bakridemos: fixed https://github.com/demos-europe/demosplan-core/pull/1573/commits/150e36db2db282d7a4da63f9a3a47aeadc87ff98

demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
```diff
@@ -1669,8 +1672,12 @@ protected function replacePhaseByPhaseNameForVotedStatementList(array $statement
      *
      * @throws Exception
      */
-    protected function createPdfDraftStatement($draftStatementList, $type, Procedure $procedure, $itemsToExport = null)
-    {
+    protected function createPdfDraftStatement(
```
mussbach, line 1675: You need to modify the method calls as well when you change the signature

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
bakridemos: sorry fixed here https://github.com/demos-europe/demosplan-core/pull/1573/commits/058f6f64501dd13389225679e170f073ed202a6e

mussbach (changes requested):
As this introduces a shiny, tiny and easy testable class you should add a test

bakridemos:
> As this introduces a shiny, tiny and easy testable class you should add a test

done in https://github.com/demos-europe/demosplan-core/pull/1573/commits/05f1648cb481f70abe5d19928a66c415a94042bc

PR #1574, opened 2023-06-21
chore: check phpstan levels 0-2 for PRs

mussbach:
At the moment phpstan is only checked for level 2, failures in level 0 or 1 are not detected, which is counterintuitive

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach:
Regarding to the documentation the rules should be cumulative "Levels are cumulative - for example running level 5 also gives you all the checks from levels 0-4." (https://phpstan.org/user-guide/rule-levels) Then there was some other issue we need to examine

PR #1575, opened 2023-06-21
bug(refs T33391): Fix z-index for tooltip

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33391

- Return empty string instead of null if we don't need any tooltip classes, because otherwise v-tooltip gets internal errors
- Return 1 for z-index if all elements have z-index auto and stop recursion if outer element is reached (html)

### How to review/test
Go to Verfahren -> Planungsdokument und Planzeichnung -> Verfahren verorten. The tooltip for the edit map features (eg Auswählen und bearbeiten) should work (switch between the different tooltips) and there should be no warnings or errors in the console.

### DIFF
--- a/client/js/components/map/map/DpOlMapEditFeature.vue
+++ b/client/js/components/map/map/DpOlMapEditFeature.vue
@@ -210,11 +210,7 @@ export default {
     getZIndex (element) {
       const z = window.getComputedStyle(element).getPropertyValue('z-index')
       if (isNaN(z)) {
-        if (element.nodeName !== 'HTML') {
-          return this.getZIndex(element.parentNode)
-        } else {
-          return 1
-        }
+        return (element.nodeName === 'HTML') ? 1 : this.getZIndex(element.parentNode)
       }
 
       return z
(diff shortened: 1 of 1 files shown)


client/js/components/map/map/DpOlMapEditFeature.vue
```diff
@@ -210,7 +210,11 @@ export default {
     getZIndex (element) {
       const z = window.getComputedStyle(element).getPropertyValue('z-index')
       if (isNaN(z)) {
-        return this.getZIndex(element.parentNode)
+        if (element.nodeName !== 'HTML') {
+          return this.getZIndex(element.parentNode)
+        } else {
+          return 1
+        }
```
spiess-demos, line 217: Certainly not a blocker but i find ternaries to be more readable by now. Also, i'd suggest to move the critical condition at the start of the expression.

```suggestion
        return (element.nodeName === 'HTML') ? 1 : this.getZIndex(element.parentNode)
```

spiess-demos (approved):
thanks for fixing this. one non-blocking suggestion only.

spiess-demos (approved):
lgtm!

PR #1576, opened 2023-06-21
fix remove unused permission

bakridemos:
Description: remove unused permission

- the permission 'feature_create_procedure_from_XBauleitplanungMessage' is not used anywhere, otherweise the Xbauleitplanung logic is removed from core and is used as addon in the relevant projects.


MoritzMandler (approved):
lgtm

PR #1579, opened 2023-06-22
chore: only try to add constraint when it does not exist

mussbach:
When performing the migration on a just created database the index already exists. Obviously there are other cases as well, where the migration fails due to already existing constraint. We need to check the existence before creating it.
This PR alters an existing migration as it already leads to problems on multiple environments.

### How to review/test
You may manually delete the Migration Version20230601095828.php from the migrations database and perform migrations

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm

PR #1580, opened 2023-06-22
chore (refs T33424): avoid unnecessary user set

mussbach:
Ticket: https://yaits.demos-deutschland.de/T33424

Request keys are always set with value null which leads to an user set that is unnecessary. Moreover it lead to weird exceptions on one (and only one) environment. This way it is cleaner anyhow.

### How to review/test
Save procedure phases in procedure detail settings

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm

PR #1581, opened 2023-06-22
fix(T33011): access getCapabilities directly in the FE utilizing externalAPI method

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33337

**Description:** This PR replaces the `dpRpc` method with a new method called `externalApi` for the `getCabalities` request, which makes API calls through the FE:

- request headers to external API should not include any default headers configured for the demosplan API

**! If the map servers are not configured to allow access to everyone, retrieving the `getCapabilities` through the FE will not be possible !**
- a specific error message for users has been added for such cases where indicating potential CORS issue (`No 'Access-Control-Allow-Origin' header is present on the requested resource`) 

### How to review/test
Verfahrensdetailseite > interaktive Karte > eigener Layer
FP Role: Verfahren wählen > Planungsdokumente und Planzeichnung > Kartenebenen definieren > Kartenebenen

### Linked PRs 
- should be merged together with [PR](https://github.com/demos-europe/demosplan-ui/pull/318)


gruenbergerdemos (dismissed):
lgtm and works

client/js/components/map/admin/LayerSettings.vue
```diff
@@ -395,29 +395,30 @@ export default {
       }
 
       const url = this.handleUrlParams(this.url)
+      const hasWMTSType = url.toLowerCase().includes('wmts')
       let parser = null
-      dpRpc('map.get_capabilities', { url })
+      externalApi(url)
```
salisdemos, line 400:
```suggestion

      externalApi(url)
```

client/js/components/map/admin/LayerSettings.vue
```diff
@@ -395,29 +395,30 @@ export default {
...
             this.version = this.currentCapabilities.version
             this.serviceType === 'wmts' ? this.extractDataFromWMTSCapabilities() : this.extractDataFromWMSCapabilities()
           }
         })
         .catch(err => {
           dplan.notify.error(Translator.trans('maplayer.capabilities.fetch.error'))
-          console.log(err)
+          if (err.code === 'ERR_NETWORK') {
+            dplan.notify.error(Translator.trans('maplayer.capabilities.fetch.error.cors.policy'))
+          }
```
salisdemos, line 415:
```suggestion

          if (err.code === 'ERR_NETWORK') {
            dplan.notify.error(Translator.trans('maplayer.capabilities.fetch.error.cors.policy'))
          }

```

translations/messages+intl-icu.de.yml
```diff
@@ -2016,6 +2016,9 @@ maplayer.locked: "Kartenebene gesperrt"
 maplayer.show: "Kartenebene einblenden"
 maplayer.show/hide: "Kartenebenen ein/ausblenden"
 maplayer.capabilities.fetch.error: "Beim Laden der Ressourcen ist ein Fehler aufgetreten. Bitte überprüfen Sie Ihre Eingaben."
+maplayer.capabilities.fetch.error.cors.policy: |
+    Falls das Problem weiterhin auftritt, wenden Sie sich bitte an den Betreiber des Dienstes mit folgendem Hinweis:
+    Es soll sichergestellt werden, dass ein CORS-Header für den Dienst gesetzt ist. Dies wird von der GDI-DE im Kapitel 4.7.1 empfohlen.
```
salisdemos, line 2021:
```suggestion
maplayer.capabilities.fetch.error.cors.policy: |
    Der angegebene Dienst konnte nicht geladen werden. Bitte stellen Sie sicher, dass die URL richtig angegeben wurde.
    
    Falls das Problem weiterhin auftritt, wenden Sie sich bitte an den Betreiber des Dienstes mit folgendem Hinweis:
    Es soll sichergestellt werden, dass ein CORS-Header für den Dienst gesetzt ist. Dies wird von der GDI-DE im Kapitel 4.7.1 empfohlen.
```

salisdemos (changes requested):

translations/messages+intl-icu.de.yml
sakutademos: I thought that it would be more readable to have two messages instead of one. The first message informs that something went wrong, please check the inputs. The second message should instruct to contact the map server if the problem occurs again...

image
image

gruenbergerdemos:
@salisdemos I also replaced the ajax call in the public detail, could you have a look at that, too [4503c4b](https://github.com/demos-europe/demosplan-core/pull/1581/commits/4503c4bec75683a91414d85eff64341540b41bf9)

translations/messages+intl-icu.de.yml
salisdemos: I think we should make one message, but with a blank line.  (you can use `<br />`)

client/js/components/map/publicdetail/Map.vue
```diff
@@ -607,25 +601,18 @@ export default {
...
-        this.deleteCustomAjaxHeaders()
         const layerArray = Array.isArray(layer.attributes.layers) ? layer.attributes.layers : layer.attributes.layers.split(',')
         const url = this.addGetCapabilityParamToUrl(layer.attributes.url)
-        $.ajax({
-          dataType: 'xml',
-          url: url || '',
-          async: false,
-          success: response => {
+        externalApi(url)
+          .then((response) => {
```
salisdemos, line 607:
```suggestion
          .then(response => {
```

Code looks legit, but to be safe: Did You test that?

salisdemos (changes requested):

client/js/components/map/publicdetail/Map.vue
gruenbergerdemos: I did, but only a hacky version with WMS layers instead of WMTS, because I couldn't find a WMTS source, that works. Do you know one?

gruenbergerdemos: Ok, I found one, something does not work right. I'm on it.

salisdemos: Please check first if it worked before. Maybe it doesn't have anything todo with the refactoring. Then we should create a bug Ticket but not try to fix it within this PR

gruenbergerdemos: It worked before, but now it does not. We get a different response, but I don't understand why. The `OperationsMetadata` is missing (I'm not sure if they are imported) and we get an error calling this function 
` options = optionsFromCapabilities(result, {
              layer: layerArray[0] || '',
              matrixSet: layer.attributes.tileMatrixSet
            })`

gruenbergerdemos: This layer works for testing: https://isk.geobasis-bb.de/mapproxy/webatlasde_wmts/service?service=WMTS&request=GetCapabilities&REQUEST=GetCapabilities

translations/messages+intl-icu.de.yml
sakutademos: I used the YAML block scalar syntax with a literal style indicator (` | `). It allows preserving line breaks and indentation. But in this case, it does not work as expected, as well as if I add a `br` tag..

image

It could be achieved with the `DpInlineNotification`, but it may not be displayed as an error message on the right side below

salisdemos: Ok. I see. I was sure that it works within the flash-messages :-/ Then letzt take Your approach with two messages. But then I suggest to make the second one as warning-Level

client/js/components/map/admin/LayerSettings.vue
```diff
@@ -395,29 +395,33 @@ export default {
...
 
       const url = this.handleUrlParams(this.url)
+      const hasWMTSType = url.toLowerCase().includes('wmts')
       let parser = null
-      dpRpc('map.get_capabilities', { url })
+
+      externalApi(url)
         .then(response => {
-          this.serviceType = (response.data['0'].result.type === 'wmts') ? 'wmts' : 'wms'
+          this.serviceType = hasWMTSType ? 'wmts' : 'wms'
```
salisdemos, line 403: why is the "old way" not working anymore?

sakutademos: because the response has changed, we no longer receive the expected type in a response as it was before (through the BE). we receive only an xml document. So, we need to find out alternative methods to determine the type. I think that the URL is a good option since it contains the type information

salisdemos: one should not work when unconcentrated. :facepalm:

salisdemos (approved):

PR #1586, opened 2023-06-22
chore (refs T33426): avoid eternal login loop

mussbach:
https://yaits.demos-deutschland.de/T33426

Any problems during userdata save in welcome page lead to an infinitive redirect loop. We need to be more defensive in this case as this is reported to happen but hard to debug or trigger by purpose.

### How to review/test
Code review might be best as it is hard to trigger this error. Perhaps you may return a non User object during save

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm, code review

PR #1592, opened 2023-06-22
Fix T30158 statements can be accessed without logging

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T30158

Description:  

- use fileProcedureAction instead fileAction to access files.

- add bool parameters 'strictCheck' which define if a procedureId is available and has to be checked if true ( here is fileProcedureAction used ) or not if false ( here is fileAction used ).

- adjust route name in twigs

### DIFF
--- a/client/js/components/map/publicdetail/controls/legendList/DpLayerLegend.vue
+++ b/client/js/components/map/publicdetail/controls/legendList/DpLayerLegend.vue
@@ -29,7 +29,7 @@
           <a
             :class="prefixClass('c-map__group-item display--block')"
             target="_blank"
-            :href="Routing.generate('core_file', { hash: planPdf.hash })"
+            :href="Routing.generate('core_file_procedure', { hash: planPdf.hash })"
             :title="planPdfTitle">
             <i
               :class="prefixClass('fa fa-download')"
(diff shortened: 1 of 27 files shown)


client/js/components/map/publicdetail/controls/legendList/DpLayerLegend.vue
```diff
@@ -29,7 +29,7 @@
           <a
             :class="prefixClass('c-map__group-item display--block')"
             target="_blank"
-            :href="Routing.generate('core_file', { hash: planPdf.hash })"
+            :href="Routing.generate('core_file_procedure', { hash: planPdf.hash })"
```
mussbach, line 32: you need to add procedureId when using this route

mussbach (changes requested):

client/js/components/map/publicdetail/controls/legendList/DpLayerLegend.vue
bakridemos: ja my bad, it's missing in other Vues too, i ll try to fix otherwise i ll search for Frontend help

bakridemos: @mussbach all concerns should be handled and gefixed.

mussbach:
Branch needs to be adjusted. Most probably to release_ewm

bakridemos:
This pr is closed because of wrong base branch, it's replaced with https://github.com/demos-europe/demosplan-core/pull/1641

PR #1594, opened 2023-06-23
chore (refs T33309): add missing Footer hook/selector.

ahmad-demos:
https://yaits.demos-deutschland.de/T33309

- add missing Footer hook/selector.

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/base_footer.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/base_footer.html.twig
@@ -7,6 +7,7 @@
 
             {% if hasPermission('area_main_static_documents') %}
                 <a class="c-footer__link" href="{{ path("DemosPlan_misccontent_static_documents") }}"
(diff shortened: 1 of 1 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/base_footer.html.twig
```diff
@@ -7,33 +7,38 @@
 
             {% if hasPermission('area_main_static_documents') %}
                 <a class="c-footer__link" href="{{ path("DemosPlan_misccontent_static_documents") }}"
+                   data-cy="static_documents"
```
spiess-demos, line 10: Some time ago we discussed naming for those hooks and came up with some kind of namespacing "per component", to make tests more readable. for those footer links, i'd suggest

```suggestion
                   data-cy="footerNav:staticDocuments"
```
or the like. The other changes have to be adapted then, too.

spiess-demos (changes requested):

spiess-demos (approved):
lgtm!

PR #1607, opened 2023-06-26
Fix:(T33457): permanently expanded basic settings

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33457

**Description:** After enabling Tailwind in demosplan-core and refactoring our handcrafted utility classes, the `cf` class was replaced with the `flow-root` class, which does not function correctly in this context. However, it looks like everything is working fine without these two classes. So, the flow-root class was removed.

### How to review:
Verwalten > Grundeinstellungen > Einstellung einklappen


gruenbergerdemos (approved):
lgtm and works

PR #1610, opened 2023-06-27
Use higher specificity for showing elements within wizard

spiess-demos:
As the tailwind display classes are configured to use !important, the jQuery show() method is now overruled by them. To keep the fix local, the show() and hide() functions were substituted with custom functions which apply the !important.

### How to review/test
The wizard (in "Grundeinstellungen" of a procedure) should show up and work "as intended".

### DIFF
--- a/client/js/lib/procedure/DPWizard.js
+++ b/client/js/lib/procedure/DPWizard.js
@@ -84,12 +84,13 @@ export default function DpWizard () {
         $items.find('.o-wizard__content').removeClass('is-active')
 
         $currentItem.find('legend').addClass('is-active-toggle')
-        $currentItem.find('.o-wizard__btn--next').show()
+        this.showElement($currentItem.find('.o-wizard__btn--next'))
+        this.showElement($wizardElements)
         $currentItem
           .addClass('o-wizard--active')
           .find('.o-wizard__content')
           .append(
-            $wizardElements.show().attr('aria-hidden', false)
+            $wizardElements.attr('aria-hidden', false)
           )
           .addClass('is-active')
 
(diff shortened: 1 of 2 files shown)


client/js/lib/procedure/DPWizard.js
```diff
@@ -103,15 +104,30 @@ export default function DpWizard () {
...
+        $wizardElements.attr('aria-hidden', true)
+        this.hideElement($wizardElements)
+        this.hideElement($form.find('.o-wizard__btn--next'))
         return this
       },
 
+      hideElement: function ($elements) {
+        const elements = $elements.toArray()
+        elements.forEach((element) => {
+          element.setAttribute('style', '')
```
salisdemos, line 127: shouldn't here (and in the showElement-method) still be the `aria-hidden` handling?

salisdemos (changes requested):
I want to vueify this. When can we do that? :sob:

spiess-demos:
> I want to vueify this. When can we do that? sob

I want to change the ui from accordions into a fixed toc navigation and sections that scroll into view.

client/js/lib/procedure/DPWizard.js
spiess-demos: switching the `aria-hidden` value is untouched by the changes, it's still happening in line 93 and 118.

salisdemos: ok. I see. I expected it within the method. But I understand now, that You don't want to set it for other elements than $wizardelements.

salisdemos (approved):

PR #1613, opened 2023-06-27
fix(refs T33506): filter out errorous items

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T33506

Due to a refactoring in the backend, the statementStatusData json contains elements without all needed properties. To quickfix the issue, those items are filtered out. However, it seems like this needs further investigation.

### How to review/test
The procedure dashboard should show some charts.

### PR Checklist


muellerdemos (approved):
the hotfix looks fine but I agree we should investigate the origin of this - code review only

PR #1614, opened 2023-06-27
fix(refs T33507): raise specificity

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T33507

The good old toggleAnything conflicts with tailwind specificity here. So, it has to raise specificity. Also, unneeded flow-root classes are removed.

### How to review/test
The "Schlagworte" table should work + look as before.

### PR Checklist


sakutademos (approved):
works, lgtm!

PR #1617, opened 2023-06-28
chore: updates lcobucci dependency

AlexMenschoid:
to avoid composer inconsistencies:
`Problem 1
     - lcobucci/clock is locked to version 3.1.0 and an update of this package was not requested.
     - lcobucci/clock 3.1.0 requires php ~8.2.0 -> your php version (8.1.12) does not satisfy that requirement.
Problem 2
     - lcobucci/clock 3.1.0 requires php ~8.2.0 -> your php version (8.1.12) does not satisfy that requirement.
     - lexik/jwt-authentication-bundle v2.19.0 requires lcobucci/clock ^1.2|^2.0|^3.0 -> satisfiable by lcobucci/clock[3.1.0].
     - lexik/jwt-authentication-bundle is locked to version v2.19.0 and an update of this package was not requested.`

### How to review/test
Without this change there should be an error while trying to run `composer i` on `main`. With it, there should not be.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2214,16 +2214,16 @@
         },
         {
             "name": "doctrine/doctrine-bundle",
-            "version": "2.10.0",
+            "version": "2.10.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/DoctrineBundle.git",
-                "reference": "b2ec6c2668f6dc514e8bf51257d19c7c19398afe"
+                "reference": "f9d59c90b6f525dfc2a2064a695cb56e0ab40311"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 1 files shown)


mussbach (changes requested):
could you please add `lcobucci/clock` as version `^3.0` to composer.json as well? Otherwise this problem will reoccur

AlexMenschoid:
> could you please add `lcobucci/clock` as version `^3.0` to composer.json as well? Otherwise this problem will reoccur

Maybe I'm missing something here, but why? It is not a direct dependency of our software and with this update the doctrine-bundle, for the time being, intentionally set it to the 3.0 release. I don't really see how this would reoccur or why we should add it to our list of dependencies if it isn't.

AlexMenschoid:
I'm merging this as is, because we need it. This has been cleared with @mussbach and changes may be made in further PRs

PR #1621, opened 2023-06-29
Fix(T33521): add notification for invalid service type

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33521

**Description:** This PR adds an additional check to handle projects that do not support WMTS service type. It checks if the entered URL has a WMTS service type and the required permission `feature_map_wmts`. If the check fails, an error message is displayed to inform the user that the WMTS service is not supported.

- clear the dropdowns and `this.currentCapabilities` when a new URL is entered
- add an additional check on the public detail page to prevent creating a new layer if no `currentCapabilities` exist

### How to review/test


salisdemos (approved):
looks legit.

PR #1628, opened 2023-06-30
fix (refs: T33353): adds fix to core

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T33353

This has been already fixed in the new demosplan-ui dependency. But as that is not yet really included in this older branch, we add the fix like this to the core.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


salisdemos (approved):
lgtm. works in current demosplan-ui Version. Should work here as well

PR #1632, opened 2023-06-30
chore: fix platform requirements syntax

mussbach:
We need to set platform requirements to ensure that vendor dependencies remain compatible with php 8.1 even when composer update is performed on a php 8.2 machine. Otherwise problems like #1617 will occur.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, thanks! I didn't think about fixing the platform instead of the specific dependency.

PR #1634, opened 2023-07-03
fix adjust events

bakridemos:
Description:  these events are related to the Datasheets logic ( which is removed from core ), they were added recently and have to be adjusted 

- extend DPlanEvent
- add missing event interface implementation


mussbach (approved):
lgtm, code review only

PR #1637, opened 2023-07-03
feature (refs T33554): fix procedure export

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T33554

Description: ZipExportService should return void, but return int.

### How to review/test
just check it in browser

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm - tested and works fine

mussbach (changes requested):
shouldn't this be based on release?

jalalifardemos:
I don't know, should be Release?

AlexMenschoid:
Something is strange here, there are too many changes in this PR and too many commits that don't sound like they belong here. Could it be that this was branched off `release_diplanbau` for some reason?

jalalifardemos:
Yes, I changed just ZipExportService, I don't know these changes come from where

jalalifardemos:
Close because of wrong base branch.
replace with https://github.com/demos-europe/demosplan-core/pull/1639

PR #1639, opened 2023-07-03
feature (refs T33554): fix procedure export

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T33554

Description: ZipExportService should return void, but return int.

### How to review/test
just check it in browser

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
lgtm, thx

PR #1641, opened 2023-07-04
fix (ref:T30158) Define Statements files accessibility 

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T30158

Description:  

- use fileProcedureAction instead fileAction to access files.

- add bool parameters 'strictCheck' which define if a procedureId is available and has to be checked if true ( here is fileProcedureAction used ) or not if false ( here is fileAction used ).

- adjust route name in twigs and add procedure Id for routing


mussbach (approved):
interesting PR title, but code lgtm

bakridemos:
> interesting PR title, but code lgtm

oh my bad, it's fixed now

PR #1642, opened 2023-07-04
fix(refs T33576): use correct id

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T33576

If users click on the input label, at least the input should be focused - this way they won't experience it as a bug if the cursor turns to a pointer when hovering labels.

### How to review/test
As a planner org, go to "Daten der Organisation". When clicking the Label for the orga slug, the respective input field should be focused.

### PR Checklist


ahmad-demos (approved):
lgtm, codeReview!

PR #1645, opened 2023-07-04
feature (refs T33519): add agency main email address to service output

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T33519 

Description: DemosPlanProcedureContrlooer::getProcedureWithPhaseNames() must return string but null.

### How to review/test
test in browser

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
@@ -2143,7 +2143,7 @@ class DemosPlanProcedureController extends BaseController
 
         $emailTextAdded = '';
         if ($this->permissions->hasPermission('feature_email_invitable_institution_additional_invitation_text')) {
-            $agencyMainEmailAddress = isset($procedureAsArray['agencyMainEmailAddress']) && $procedureAsArray['agencyMainEmailAddress'] !== null ? $procedureAsArray['agencyMainEmailAddress'] : '';
+            $agencyMainEmailAddress = isset($procedureAsArray['agencyMainEmailAddress']) && null !== $procedureAsArray['agencyMainEmailAddress'] ? $procedureAsArray['agencyMainEmailAddress'] : '';
             $emailTextAdded = $this->generateAdditionalInvitationEmailText($publicAffairsAgents, $organization, $agencyMainEmailAddress);
         }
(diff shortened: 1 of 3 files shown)


bakridemos (changes requested):
i don't really think the adjustements in the addPhaseNames method are necessary, it's not the right context and i don't really understand why you need the internalPhases here ?  i think we could fix this in the Controller with a simple Check, if the agencyMainEmailAddress is null or not and try to pass an empty string if its null.

bakridemos (commented):
it's look good to me, but since we did this together i suggest that you ll ask someone else to review this pr.

demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
```diff
@@ -2143,7 +2143,8 @@ public function administrationMemberListAction(
 
         $emailTextAdded = '';
         if ($this->permissions->hasPermission('feature_email_invitable_institution_additional_invitation_text')) {
-            $emailTextAdded = $this->generateAdditionalInvitationEmailText($publicAffairsAgents, $organization, $procedureAsArray['agencyMainEmailAddress']);
+            $agencyMainEmailAddress = isset($procedureAsArray['agencyMainEmailAddress']) && null !== $procedureAsArray['agencyMainEmailAddress'] ? $procedureAsArray['agencyMainEmailAddress'] : '';
+            $emailTextAdded = $this->generateAdditionalInvitationEmailText($publicAffairsAgents, $organization, $agencyMainEmailAddress);
```
mussbach, line 2147:
```suggestion
            $emailTextAdded = $this->generateAdditionalInvitationEmailText($publicAffairsAgents, $organization, $procedureAsArray['agencyMainEmailAddress'] ?? '');
```

mussbach (commented):
This can be simplified

demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
jalalifardemos: done in [cb1a76c](https://github.com/demos-europe/demosplan-core/pull/1645/commits/cb1a76c566ac32de04160e486908541c32a6ae84)

mussbach (approved):
lgtm, code review only

PR #1648, opened 2023-07-04
fix(T33575): issue with the table when a row is expanded

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33575

**Description:** This PR fixes an issue with the table when a row is expanded.

After some refactoring an `!important` declaration was added to the `display:block` property in the `is-active `and `is-visible` classes, which causes an issue when `is-active` is added to the `table-row`, because it has higher priority. 

it was necessary to add also the `!important` declaration to the '.table-row' class to fix the bug.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_toggle.scss
+++ b/demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_toggle.scss
@@ -100,7 +100,7 @@
                 animation-fill-mode: forwards;
 
                 &.table-row {
-                    display: table-row;
+                    display: table-row !important;
                 }
             }
 
@@ -119,7 +119,7 @@
                 transform: translateY(0);
 
                 &.table-row {
-                    display: table-row;
+                    display: table-row !important;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


sakutademos (commented):
@spiess-demos was it necessary to add the `!important` declaration for the `display: block ` property in the `is-active` and `is-visible` classes? here: https://github.com/demos-europe/demosplan-core/commit/791ac9c076ecdcc5184023832fbc2bd7e065a134

spiess-demos:
> @spiess-demos was it necessary to add the `!important` declaration for the `display: block ` property in the `is-active` and `is-visible` classes? here: [791ac9c](https://github.com/demos-europe/demosplan-core/commit/791ac9c076ecdcc5184023832fbc2bd7e065a134)

Sadly yes. Otherwise there were issues with specificity in the "Schlagworte" table.

Luckily both the "Schlagworte" table and the "Institutionen" table are not using the latest table pattern (DpDataTable), so i assume over time we can get rid of this unnecessary specificity battle.

spiess-demos (approved):
lgtm (code review only)

PR #1650, opened 2023-07-04
feat: Adds dplanPermissions attribute

AlexMenschoid:
This PR adds an attribute equivalent of the old dplanPermissions annotation and includes it as the new first choice in the permission listener. If both an attribute and an annotation are found, the attribute takes precedent. But otherwise the old annotation system works just the same as before.

I consider this to be preparation for switching out the old annotation for the new attribute, which can now be either done bit by bit or via a bigger one time refactoring.

### How to review/test
All routes should work just as before. Otherwise a code review would be best. Please let me know if you have suggestion on how to improve this.


demosplan/DemosPlanCoreBundle/EventListener/CheckPermissionListener.php
```diff
@@ -72,21 +73,10 @@ public function onControllerRequest(ControllerEvent $event)
...
-                    E_USER_DEPRECATED
-                );
-
-                return;
-            }
+            $dplanPermissions = $this->getDplanPermissions($reflectionMethod);
 
             // perform initialize with permissions from annotation
-            $this->initializeService->initialize($dplanPermissions->getPermissions());
+            $this->initializeService->initialize($dplanPermissions);
```
breierDemos, line 79: In initialize the function checkPermissions() is called. In checkPermissions() a log message seems not matching too.

breierDemos (changes requested):

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
```diff
@@ -1039,8 +1039,8 @@ public function checkPermissions($permissions): void
             }
         } else {
             // Gib Devs einen Hinweis aus, dass hier die Rechte nachgearbeitet werden müssen
```
breierDemos, line 1041:
```suggestion
            // Give devs a hint that the permissions here need to be reworked
```

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
```diff
@@ -1039,8 +1039,8 @@ public function checkPermissions($permissions): void
...
-            $this->logger->info('Dieser Bereich hat kein explizites Permission angegeben! '
-                        .'Bitte @DplanPermissions mit einem zu prüfenden Recht annotieren.', \debug_backtrace(0, 4));
+            $this->logger->info('Dieser Bereich hat keine explizite Permission angegeben! '
```
breierDemos, line 1042:
```suggestion
            $this->logger->info('This area has no explicit permission specified! '
```

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
```diff
@@ -1039,8 +1039,8 @@ public function checkPermissions($permissions): void
...
+                        .'Bitte per Attribut #[DplanPermissions] oder Annotation @DplanPermissions mit einem zu prüfenden Recht versehen.', \debug_backtrace(0, 4));
```
breierDemos, line 1043:
```suggestion
                        .'Please provide a permission to be checked using the attribute #[DplanPermissions] or annotation @DplanPermissions.', \debug_backtrace(0, 4));
```

breierDemos (changes requested):
Some suggestions for translation from german to english.

breierDemos (approved):
lgtm :)

PR #1651, opened 2023-07-04
fix (refs T33506): Fix statement status data by adding missing necessary 'freq' key.

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T33506
(Related Ticket: https://yaits.demos-deutschland.de/T32108)

Fix statement status data by adding missing necessary 'freq' key.
Also only add only $aggregations['status'] which is useful for the User, defined in form_options.yml

I do not understand, where some key-value-pairs are came from. like '1' : '1' or '0' : '0' for example.
I just filter these, but actually I would prefer to understand where and why they come from. Is there a deeper problem with ES or something?

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
@@ -451,7 +451,6 @@ class DemosPlanProcedureController extends BaseController
                 $statusValue = $aggregationBucket['value'];
                 $statusCount = $aggregationBucket['count'];
                 $statementStatusData[$statusValue]['count'] = $statusCount;
-                $statementStatusData[$statusValue]['freq'] = $priorityInitialValues;
 
                 // add link with filterhash to assessment table
                 if (0 < $statusCount) {
--- a/demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
@@ -445,20 +445,22 @@ class DemosPlanProcedureController extends BaseController
         }
 
         // save status counts
-        $esResultMeta = $statementQueryResult->getFilterSet();
-        $aggregations = $esResultMeta['filters'];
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
```diff
@@ -445,20 +445,22 @@ protected function getStatementStatusData(
...
-                $statementStatusData[$statusValue]['url'] = $this->generateAssessmentTableFilterLinkFromStatus(
-                    $statusValue,
-                    $procedureId,
-                    'statement'
-                );
+            if (array_key_exists($aggregationBucket['value'], $statementStatuses)) {
+                $statusValue = $aggregationBucket['value'];
+                $statusCount = $aggregationBucket['count'];
+                $statementStatusData[$statusValue]['count'] = $statusCount;
+                $statementStatusData[$statusValue]['freq'] = $priorityInitialValues;
```
breierDemos, line 454: Are you sure you need to set the priorityInitialValues here. I think this happens already in line 432. There they are added for each status. And in the if statement they will be added for the same statuses again. When I remove this line, the result seems the same.

breierDemos (changes requested):
I also don't understand where some of the key value pairs come from. I also have a '1' : '1' but no '0' : '0' for example.
I also removed quick fix from https://github.com/demos-europe/demosplan-core/pull/1613/files to test and your fix seems to work. The Quick Fix could therefore be removed again if desired.

martendemos:
> I also don't understand where some of the key value pairs come from. I also have a '1' : '1' but no '0' : '0' for example. I also removed quick fix from https://github.com/demos-europe/demosplan-core/pull/1613/files to test and your fix seems to work. The Quick Fix could therefore be removed again if desired.

Yes it could be, but FE wants it to remain in the code, which is okay for me.

demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
martendemos: You're right. thx.
09788e334349618283248317725d663d569dd288

breierDemos (approved):
lgtm

PR #1652, opened 2023-07-04
F adjust user entity

bakridemos:
Description: 

adjust implemented classes: it is better when the User entity implement one interface (the AddonUserInterface) which extends the other needed interfaces.

other needed adjustements are made in demosplan addon too


AlexMenschoid (dismissed):
lgtm, code review only

PR #1657, opened 2023-07-05
fix(refs T33353): Change way of replacing \n\r with blanks

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33353

For some reason, the way it works atm on main doesn't worked here (#1628). So we had to adjust the fix a bit.

### How to review/test

Paste Text from Word (not office365) into some editor (e.g. "Textbausteine Anlegen"). All blank spaces should be preserved.

As a Dev, this can't be tested, only reviewed since there is no real word on linux.

### PR Checklist

### DIFF
--- a/client/js/components/core/DpEditor/libs/handleWordPaste.js
+++ b/client/js/components/core/DpEditor/libs/handleWordPaste.js
@@ -272,7 +272,7 @@ function buildListAsHtmlString (list) {
  * @return {string}
  */
 function prepareDataBeforeParsingMso (slice) {
-  const x = slice
+  return slice
     // Strip head
     .replace(/<head>(.|\n|\r)*?<\/head>/mi, '')
     // Strip line breaks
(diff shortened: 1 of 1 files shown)


client/js/components/core/DpEditor/libs/handleWordPaste.js
```diff
@@ -273,10 +273,10 @@ function buildListAsHtmlString (list) {
  */
 function prepareDataBeforeParsingMso (slice) {
   return slice
-    // Strip line breaks
-    .replace(/(&nbsp;|\r|\n)/gmi, ' ')
     // Strip head
     .replace(/<head>(.|\n|\r)*?<\/head>/mi, '')
+    // Strip line breaks
+    .replace(/(\r\n)+?/gmi, ' ')
```
AlexMenschoid, line 279: Did you drop the `&nbsp;` intentionally? On first glance, I feel like we should keep that as well, just in case.

salisdemos: no. It just doesn't makes any sense here. If Someone inserted a nonebreaking-space, it shall remain and will be handled by tiptap.

AlexMenschoid (approved):
lgtm, code review

ahmad-demos (approved):
lgtm, codeReview

PR #1658, opened 2023-07-05
bug (refs T33551): Fix the format of emails for released statements.

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T33551

This PR changes the format of released statements.

### How to review/test
code review, please check also if this is the right file to adjust the email and if this is the right way to adjust the format. The format should be the same as in https://yaits.demos-deutschland.de/T33551#789241.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/send_notification_email_for_released_statement.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/send_notification_email_for_released_statement.html.twig
@@ -3,11 +3,8 @@
 {%- set count = templateVars.statements|length -%}
 Guten Tag,
 
-im Verfahren "{{ templateVars.procedure.name|default('') }}"
-{{ 'email.text.notification_for_released_statement.1'|trans({ count: count }) }} freigegeben.
-
-Bitte prüfen Sie die {{ 'email.text.notification_for_released_statement.2'|trans({ count: count }) }}
-und reichen Sie sie bei dem Verfahrensträger ein.
+im Verfahren "{{ templateVars.procedure.name|default('') }}" {{ 'email.text.notification_for_released_statement.1'|trans({ count: count })|replace({ "\n": "" }) }} freigegeben.
+Bitte prüfen Sie die {{ 'email.text.notification_for_released_statement.2'|trans({ count: count })|replace({ "\n": "" }) }} und reichen Sie sie bei dem Verfahrensträger ein.
 
 Freundliche Grüße
 


bakridemos (approved):
lgtm, code review only

templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/send_notification_email_for_released_statement.html.twig
```diff
@@ -3,11 +3,8 @@
 {%- set count = templateVars.statements|length -%}
 Guten Tag,
 
-im Verfahren "{{ templateVars.procedure.name|default('') }}"
-{{ 'email.text.notification_for_released_statement.1'|trans({ count: count }) }} freigegeben.
-
-Bitte prüfen Sie die {{ 'email.text.notification_for_released_statement.2'|trans({ count: count }) }}
-und reichen Sie sie bei dem Verfahrensträger ein.
+im Verfahren "{{ templateVars.procedure.name|default('') }}" {{ 'email.text.notification_for_released_statement.1'|trans({ count: count })|replace({ "\n": "" }) }} freigegeben.
```
salisdemos, line 6:
```suggestion
im Verfahren "{{ templateVars.procedure.name|default }}" {{ 'email.text.notification_for_released_statement.1'|trans({ count: count })|replace({ "\n": "" }) }} freigegeben.
```

salisdemos, line 6: I don't get from where `\n` appears, that it has to be stripped away

salisdemos (changes requested):

mussbach:
@breierDemos what happens with this one

templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/send_notification_email_for_released_statement.html.twig
breierDemos: I think the line breaks come from the `translations.json` somehow? At least the `'\n'` match the line breaks in the `messages+intl-icu.de.yml`. See: `'email.text.notification_for_released_statement.1'`. But I don't know how to pack the contents of `'email.text.notification_for_released_statement.1' `in `the messages+intl-icu.de.yml` into one line? I have seen that other email texts are available completely as translation keys. Should this perhaps be adopted here too? See: `'fragment.notify.email.reviewer.admin.new'`.

salisdemos:
@breierDemos should we leave it like it is, or do You want to move the texxt to the translation file as You suggested?

PR #1661, opened 2023-07-05
Fix(T33134): improve accessibility

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33134

**Description:** This PR adds an aria status role (role=”status”) for the better accessibility, as well as changes the default value of the `this.isVisible` variable to resolve a bug where the role was only visible after changing the tab


spiess-demos (approved):
Lgtm! Code review only, however.

PR #1662, opened 2023-07-05
F remove duplicated current user interface

bakridemos:
Description: 

In order to use the interfaces in demosplan and in the addons (as contracts), we did till now tried to make and use one only interface which normaly located in demosplan addon ( which make it available for the core and the addons ) and in some cases we moved the already existed interfaces from core to demosplan addon.

We do have now two Identical CurrentUserInterfaces, and only one was used. To avoid confusion and for better code quality the duplicated one is removed and only the demosplan addon CurrentUserInterface will be used


config/services.yml
```diff
@@ -45,6 +45,9 @@ services:
     DemosEurope\DemosplanAddon\Contracts\CurrentContextProviderInterface:
         class: demosplan\DemosPlanCoreBundle\Logic\CurrentContextProvider
 
+    DemosEurope\DemosplanAddon\Contracts\CurrentUserInterface:
```
mussbach, line 48: Is this really necessary?

mussbach (changes requested):
adjustment to new contracts version is missing, otherwise lgtm, one question

config/services.yml
bakridemos: > Is this really necessary?

it is jup, otherwise we have the usual error that no DemosEurope\DemosplanAddon\Contracts\CurrentUserIn terface service found and that we have to alias it to the demosplan existing service

mussbach: weird this should be possible to handle globally

bakridemos:
> adjustment to new contracts version is missing, otherwise lgtm, one question

it's not really needed here, the demosplan addon CurrentUserInterface exist already and i did have to do anything there, it's was already adjusted

mussbach (approved):

PR #1663, opened 2023-07-06
Use local pdflatex api & docx import container instead of java service

mussbach:
A step towards smaller containers away from fat containers is to use several containers for distinct services. This PR introduces and activates two new containers for docx import and pdf export.

### How to review/test
This is the last PR of a series of PRs that needs to be merged first. When they are you should restart your containers and docx import and pdf export should just work. You can verify that the new containers are used when you see a request in the symfony toolbar for HTTP Client

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Tools/DocxImporterHttp.php
```diff
@@ -53,6 +53,7 @@ public function importDocx(File $file, string $elementId, string $procedure, str
             ];
         } catch (Exception $e) {
             $this->logger->error('Error while creating docx with http: '.$e->getMessage());
+            $this->logger->error('Response body: '. $response->getContent(false));
```
AlexMenschoid, line 56: It is very possible that `$response` is undefined here, which would trigger an error while trying to log.Maybe wrap it in a check to only log when present.

AlexMenschoid (changes requested):

demosplan/DemosPlanCoreBundle/Tools/DocxImporterHttp.php
mussbach: good catch, thx f25b5bda

AlexMenschoid (dismissed):
lgtm, code review only

PR #1666, opened 2023-07-06
Prevent 2px jump of button

spiess-demos:
This PR fixes the 2px jump of the "Geltungsbereich" and "Planzeichnung" buttons.

1px-jump

.btn assigns `border: 1px solid transparent;`, so if the `border` shorthand specifies the color only, this actually means `border: none $dp-color-highlight initial;`, at least to my understanding.

### How to review/test
Buttons should have consistent sizing when activated and deactivated.


salisdemos (approved):
thanks

PR #1679, opened 2023-07-07
chore (refs T31115): add Translator.trans to dp-select options.

ahmad-demos:
https://yaits.demos-deutschland.de/T31115

- add Translator.trans to dp-select options.
- remove Translator.trans from demosplan-ui


spiess-demos (approved):
lgtm! code review only.

PR #1682, opened 2023-07-07
feature (refs T32782): check for weak passwords only after succesful login

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T32782

Description:
The CheckPassportEvent has many different listeners which do not belong to our Code.
They use priorities from 2000 down to 0. On login failure an Exception is thrown preventing further events being triggered.
 Our own listener needs to be called last - only after successful login - to check for weak passwords - so a negative priority was added.

### How to review/test
try to log in with an existing user and a wrong pw... check the flashbag messages.

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm

PR #1685, opened 2023-07-07
fix(refs T33640): Bump demosplan-ui to 0.1.8-rerelease

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33640
With the release of demosplan-ui@0.1.8 there came changes along which broke the whole application.
To fix that, we realeased a new Version `0.1.8.rerelease` which is technically before 0.1.8. That on the other hand forces us to set a fixed version in the package.json (that 0.1.8 doesn't get fetched)


spiess-demos (approved):
lgtm - code review only, again.

PR #1686, opened 2023-07-07
feat: Remove unused properties in StatementService.php

breierDemos:
I removed two unused properties from the StatementService.

### How to review/test
Check if the properties are still used

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, couldn't find any usage of the two in this class either.

PR #1704, opened 2023-07-12
feat(refs T12070):  New back to top button

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T12070

This replaces the old Back to top Button.

On every Page in the admin-area this button should appear. But only after scrolling 100vh.
The Position "allways" fit. Even after resizing and on small screens and with a large menu and at the bottom of the page.

### How to review/test

Goto a admin-page with enogh content (like a-table with some statements.) Scroll 100vh and see if the btt-Button does what its suppose to do.

### Linked PRs (optional)

:warning: 
Only works after https://github.com/demos-europe/demosplan-ui/pull/364 is Published.
--> https://github.com/demos-europe/demosplan-core/pull/1712

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


client/js/components/button/BackToTopButton.vue
```diff
@@ -0,0 +1,79 @@
...
+  mounted () {
+    this.calculateSizes()
+    this.calculatePosition()
+
+    window.addEventListener('resize', () => {
+      this.calculateSizes()
+      this.calculatePosition()
+    })
+
+    window.addEventListener('scroll', () => {
```
spiess-demos, line 73: Not a blocker: We could try to use the `throttle` utility function to prevent scroll jank, both here and on the resize handler binding. 
See client/js/directives/scroller.js:53

client/js/components/button/BackToTopButton.vue
```diff
@@ -0,0 +1,79 @@
...
+    variant="outline"
+    color="primary"
+    :class="{ 'hide-visually': hide }"
+    hide-text
+    rounded
+    icon="arrow-up"
+    icon-size="large"
+    :style="buttonPosition"
+    @click="scrollTop">
+  </dp-button>
```
spiess-demos, line 14:
```suggestion
    @click="scrollTop" />
```

client/js/components/button/BackToTopButton.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <dp-button
+    :aria-label="Translator.trans('backToTop')"
+    :title="Translator.trans('backToTop')"
+    variant="outline"
+    color="primary"
```
spiess-demos, line 6: Does not have to be set, as color defaults to primary.

client/js/components/button/BackToTopButton.vue
```diff
@@ -0,0 +1,79 @@
...
+    :class="{ 'hide-visually': hide }"
+    hide-text
+    rounded
```
spiess-demos, line 9: Nitpick: alphabetic order

client/js/components/button/BackToTopButton.vue
```diff
@@ -0,0 +1,79 @@
...
+      window.scrollTo(0, 0)
+    }
+  },
+
+  mounted () {
+    this.calculateSizes()
+    this.calculatePosition()
+
+    window.addEventListener('resize', () => {
+      this.calculateSizes()
```
spiess-demos, line 69: `calculateSizes()` should also run if dimensions change in other ways (eg. if accordions are triggered). [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/Resize_Observer_API) is a good way to implement that. Also see scroller implementation: client/js/directives/scroller.js:58

spiess-demos (changes requested):

client/js/components/button/BackToTopButton.vue
salisdemos: changed in 9487307521c058b398e728dc4b72f98556c11350

client/js/components/button/BackToTopButton.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <dp-button
+    :aria-label="Translator.trans('backToTop')"
+    :title="Translator.trans('backToTop')"
```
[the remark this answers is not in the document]
salisdemos: Thanks. 0bfc2123c5b8766178cb2a9300c480d0cd947160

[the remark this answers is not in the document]
salisdemos: and good to know 43506d4921056913e9a2d761d967eb0ac33b1cb8

client/js/components/button/BackToTopButton.vue
salisdemos: 755482495d4967cdd66fd9a3a1f8fd5770c41d62

client/js/components/button/BackToTopButton.vue
salisdemos: 755482495d4967cdd66fd9a3a1f8fd5770c41d62

templates/bundles/DemosPlanCoreBundle/DemosPlanCore/base.html.twig
```diff
@@ -140,19 +140,15 @@
                         {% block component_part %}{% endblock component_part %}
 
                         {% block base_backtotop %}
-                            <p class="layout__item text-right u-mt">
-                                <a href="#" data-scroll-to-top="{# This attribute is used as a hook to append an EventListener in GlobalEventListener.js #}">
```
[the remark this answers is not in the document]
salisdemos: thanks for spotting that 7bc7afa4a456ee4be707ff31b433c3971e8deab6

client/js/components/button/BackToTopButton.vue
salisdemos: like so? 4459b273c89d8522f096db624c2098747e1600a2

client/js/components/button/BackToTopButton.vue
```diff
@@ -0,0 +1,79 @@
...
+      footerHeight: 0,
+      positionFromLeft: 0,
+      scrollPos: 0,
+      windowHeight: 0
+    }
+  },
+
+  computed: {
+    hide () {
+      return this.scrollPos < this.windowHeight
```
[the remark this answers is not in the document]
salisdemos: Now it appears at 70%. For me, now it feels smooth. 4459b273c89d8522f096db624c2098747e1600a2

templates/bundles/DemosPlanCoreBundle/DemosPlanCore/base.html.twig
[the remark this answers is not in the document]
salisdemos: now it should be there 3df22f34162bc01bb8b8a144a61d8bb8a9ffa95f (I guess I took the wrong hash. and  at least My push failed)

client/js/components/button/BackToTopButton.vue
spiess-demos: You are right, there is no observable advantage of the throttling here. It rather introduces a visual bug: when scrolling really fast to the very bottom of the page, the button is mistakenly placed on top of the footer. In summary i think it is better to go without throttling. However we could use the [passive event listener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) flag to [indicate that the handler will not cancel the scroll event](https://developer.chrome.com/blog/passive-event-listeners/).

client/js/components/button/BackToTopButton.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <dp-button
+    :class="{ 'hide-visually': hide }"
+    hide-text
+    icon="arrow-up"
+    icon-size="large"
+    rounded
+    :style="buttonPosition"
+    :text="Translator.trans('back.to.top')"
```
spiess-demos, line 9: Hopefully last thing: the key must be renamed also in the messages.de...yaml.

spiess-demos (changes requested):

spiess-demos (approved):
Lgtm - Thanks!

PR #1712, opened 2023-07-13
chore: Bump demosplan-ui to v0.1.9

salisdemos:
I need this straight away for my user Story, so I can't wait for dependabot.
And we had to change the version resolving in the package.json, since it haad to be fixed before due to a buggy situation


spiess-demos:
As this contains some breaking changes regarding scss paths, i'd be happy to contribute here.

PR #1714, opened 2023-07-13
fix: add missing JSON:API deletion permission check

dresslerdemos:
So far to handle JSON:API deletion requests it was required to check permissions inside the `delete` method manually. This however is error-prone, as the necessity is not explicit enough and was in fact already forgotten in `ProcedureNews-` and `SignLanguageOverviewVideo` resource types.

This commit adds a new method
(`getRequiredDeletionPermissions`) that must be implemented by deletable resource types, making the need to define the needed permissions more explicit.

The new method was used in the `ProcedureNews`- and `SignLanguageOverviewVideo` resource types to check for editing permissions, which should be sufficient for their use-cases.

The usage of the new `getRequiredDeletionPermissions` method was added in a way, that prevents the
`BeforeResourceDeletionEvent` event from being dispatched. Until now the event was dispatched even in case of missing permissions, which may potentially have resulted in bugs.

The adjusted `DeletableDqlResourceTypeInterface` is still in the core, so I assume adjustments of addons are not necessary.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
lgtm, thx, nice and clean. Code review and phpstan check

PR #1723, opened 2023-07-14
Include Tailwind as postCss plugin

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T27593

To be able to make full use of all the cool [Tailwind 3 features](https://tailwindcss.com/blog/tailwindcss-v3), it is necessary to make Tailwind run in the demosplan build, instead of just purging a "complete" Tailwind Css file. There are two main reasons to do so:

1. There is no practical way of having Tailwind render all of its css. The possible combinations of [breakpoints, pseudo-elements, and pseudo-selectors](https://tailwindcss.com/docs/hover-focus-and-other-states) would lead to a filesize that is simply not manageable.
2. Purging a huge pile of Css is way slower than just letting Tailwind do its job on a chosen set of source files.

The Tailwind config requires the demosplan-ui Tailwind config and makes some necessary tweaks. Stylelint is adapted to not error for `@tailwind` expressions. 

:tada: **Improvements**

- We can now use [breakpoints, pseudo-elements, and pseudo-selectors](https://tailwindcss.com/docs/hover-focus-and-other-states) in the Tailwind syntax
- Watcher is sufficient during development (when working on Vue components)
- Faster build time

:warning: **Known limitations:**

- Tailwind is not able to recognize classes within our messages.*.yml files, probably because of the escaped characters.
- Currently there is no Tailwind integration for addons. Tailwind classes used inside addon markup does only have an effect if the same class is also found in markup recognized by Tailwind (demosplan-core + -plugins, project templates, demosplan-ui).

### How to review/test
`yarn install`, `yarn watch:<project>` -> change a core component by adding something like `hover:shadow-lg` to an element, and reload the page. The styles for adding the shadow onHover should be in the compiled css.

### Linked PRs
https://github.com/demos-europe/demosplan-ui/pull/361 paved the way for this PR by configuring Tailwind to use  borderRadius, boxShadow, screens, and zIndex values from our design tokens - ensuring they are on par with the scss variables also rendered from those tokens.

### PR Checklist

- [ ] Update documentation

### DIFF
--- a/stylelint.config.js
+++ b/stylelint.config.js
@@ -25,7 +25,7 @@ module.exports = {
     'scss/at-import-no-partial-leading-underscore': null,
     'scss/at-rule-conditional-no-parentheses': null,
     'scss/at-rule-no-unknown': [
-      true,
+      true, // Allow @tailwind expression while keeping the rest of the rule
       {
         'ignoreAtRules': [
           'extends',
--- a/stylelint.config.js
+++ b/stylelint.config.js
@@ -11,8 +11,7 @@
 module.exports = {
   extends: 'stylelint-config-standard-scss',
   ignoreFiles: [
-    'demosplan/DemosPlanCoreBundle/Resources/client/scss/inuit-5/*.scss',
-    'client/ui/tokens/**/*.scss'
+    'demosplan/DemosPlanCoreBundle/Resources/client/scss/inuit-5/*.scss'
   ],
   rules: {
     'at-rule-empty-line-before': null,
@@ -25,6 +24,15 @@ module.exports = {
     'scss/at-extend-no-missing-placeholder': null, // Fontawesome does not provide placeholders for individual icons
     'scss/at-import-no-partial-leading-underscore': null,
     'scss/at-rule-conditional-no-parentheses': null,
+    'scss/at-rule-no-unknown': [
+      true,
+      {
+        'ignoreAtRules': [
+          'extends',
+          'tailwind'
+        ]
+      }
+    ],
     'scss/comment-no-empty': null,
     'scss/dollar-variable-colon-space-after': null,
     'scss/dollar-variable-empty-line-before': null,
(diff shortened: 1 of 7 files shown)


stylelint.config.js
```diff
@@ -25,6 +24,15 @@ module.exports = {
...
     'scss/at-rule-conditional-no-parentheses': null,
+    'scss/at-rule-no-unknown': [
+      true,
+      {
+        'ignoreAtRules': [
+          'extends',
+          'tailwind'
+        ]
+      }
+    ],
```
salisdemos, line 35: May You add a comment here, why thats neccessary?

tailwind.config.js
```diff
@@ -0,0 +1,19 @@
...
+const config = {
+  ...require('@demos-europe/demosplan-ui/tailwind.config'),
+  content: [
+    './client/js/**/!(generated|legacy|store)/*.{js,vue}',
+    './demosplan/plugins/**/Resources/**/*.twig',
+    './node_modules/@demos-europe/demosplan-ui/dist/*.js',
+    './projects/**/templates/**/*.twig',
+    './templates/bundles/DemosPlanCoreBundle/**/*.twig'
+  ],
+  safelist: [] // Can be removed when demosplan-ui drops the /./ pattern in the Tailwind config
```
salisdemos, line 16: When we now generate the tailwind styles here in the core, do we still need the /./ Pattern indemosplan-ui? Or can it be removed in a next step?

salisdemos (commented):
I think thats absolutely a step in the right direction. Thanks a lot for the effort!

tailwind.config.js
spiess-demos: It can be removed, this is correct.

stylelint.config.js
spiess-demos: sure! cd2ec5c5d019fda155f1f0c368c09404f3983798

graupnerdemos (approved):
Wonderful! ❤️

salisdemos (approved):

PR #1726, opened 2023-07-14
fix: adapt padding to larger select spacing

spiess-demos:
This restores a correct spacing between the select element and the save/abort controls.

image


salisdemos (approved):
thanks

PR #1733, opened 2023-07-17
chore(refs  T32012): Change access to demosplan-ui as interims state

salisdemos:

While demosplan-ui doesn't run in vue3 or at least in compat-mode, we have to access the compoents uncompiled to avoid 
version-missmatch errors.

this is just necessary during developmentphase and has to be switched back after demosplan-ui can be build with vue3


salisdemos:
Not necessary anymore....

PR #1735, opened 2023-07-17
fix(refs T33784): fixes sortMethod handling

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T33784

Something must have gone wrong here recently. There is no sortMethod here to be used and the identifier condition is most definitely not a sortMethod. So I added it to the conditions and left the sortMethod empty.

### How to review/test
Publishing any procedure documents will fail without this, but should work with this change.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


dresslerdemos (approved):
Code review only.

PR #1736, opened 2023-07-17
Fix(T33838): the list style issue 

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33838

**Description:** This PR fixes the list style issue by adding specific styling for `o`l and `ul` tags.

(When the list-style property of the `ol` tag is set to the `initial` value, it displays rounded bullet points. Therefore, was necessary  to set it to the `decimal` value separately)

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_tiptap.scss
+++ b/demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_tiptap.scss
@@ -250,14 +250,16 @@
                 font-weight: bold;
             }
 
+            ul, ol {
+                padding-left: 1rem;
+            }
+
             ul {
                 list-style: disc;
-                padding-left: 1rem;
             }
 
             ol {
                 list-style: decimal;
-                padding-left: 1rem;
             }
 
             li > p,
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_tiptap.scss
```diff
@@ -250,8 +250,13 @@
                 font-weight: bold;
             }
 
-            ul,
+            ul {
+                list-style: disc;
+                padding-left: 1rem;
```
spiess-demos, line 255: Nitpick: back in the days it was a best practice to define a block with properties that are shared by multiple selectors first, followed by the blocks with the styles that are different, like so:

```scss
ul,
ol {
    // Things that ul and ol have in common
}

ul {
  // unique styles
}

ol {
  // unique styles
}
```

This is only to enhance readability, i think that the compiler will produce the same output for both versions.

spiess-demos (approved):
lgtm, however 1 nitpick.

demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_tiptap.scss
hwiem: Since @sakutademos is out of office, I allowed myself to make the adjustment: 55d6767a308acac1dd9239df6e63ce9550f2ac87

spiess-demos (approved):
Thanks! Lgtm!

PR #1740, opened 2023-07-18
fix(refs T33737): use correct user for assignee check

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T33737

There has been a rework of the token storage as well as the used user entity to move our application closer to Symfony standards. In that process, the token storage now returns a securityUser instead of the regular user entity. So we need to get the corresponding user entity via the UserFromSecurityUserProvider to restore the logic.

### How to review/test
Without this fix, a user wont be able to edit a statement even when they are the assignee. With this fix it should work as intended.


mussbach (changes requested):
While this would perfectly work I suggest the solution in #1746

AlexMenschoid:
The implementation in #1746 seems to me the better way to go, so I'm closing this.

PR #1741, opened 2023-07-18
feat: Adjustments to PostProcedureUpdatedEvent

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T32776

This PR adjusts the PostProcedureUpdatedEvent, the event will now store the updated procedure, the not updated procedure and has methods to get and determine the updated values as array of the update process.

### How to review/test
code review

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (changes requested):
I withdraw my approve as you need to update the addon contracts first before you can use new interface

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
```diff
@@ -12,21 +12,66 @@
...
+    protected Procedure $procedureBeforeUpdate;
+    protected Procedure $procedureAfterUpdate;
 
-    public function __construct(Procedure $procedure)
+    public function __construct(Procedure $procedureBeforeUpdate, Procedure $procedureAfterUpdate)
     {
-        $this->procedure = $procedure;
+        $this->procedureBeforeUpdate = $procedureBeforeUpdate;
+        $this->procedureAfterUpdate = $procedureAfterUpdate;
     }
```
dresslerdemos, line 30: Can be simplified via [constructor properties](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion). You can combine this with the usage of [`readonly` properties](https://www.php.net/manual/en/language.oop5.properties.php).

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
```diff
@@ -12,21 +12,66 @@
...
-        return $this->procedure;
+        return $this->procedureBeforeUpdate;
+    }
+
+    public function getProcedureAfterUpdate(): Procedure
+    {
+        return $this->procedureAfterUpdate;
+    }
+
+    public function getModifiedValues(): array
```
dresslerdemos, line 42: Typehint missing.

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
```diff
@@ -12,21 +12,66 @@
...
+    {
+        return $this->determineModifiedValues($this->procedureBeforeUpdate, $this->procedureAfterUpdate);
+    }
+
+    private function determineModifiedValues(EntityInterface $oldEntity, EntityInterface $newEntity): array
```
dresslerdemos, line 47: Typehint missing.

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
```diff
@@ -1123,10 +1123,6 @@ public function updateProcedureObject(Procedure $procedureToUpdate)
             $sourceProcedure->setSettings($sourceProcedureSettings);
 
             $procedure = $this->procedureRepository->updateObject($procedureToUpdate);
```
dresslerdemos, line 1125: I don't understand this approach. `updateObject` simply validates and persists the changes that must already be present in `$procedureToUpdate`. So cloning it before will result in two instances without differences after the update.

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
```diff
@@ -12,21 +12,66 @@
...
+
+            $oldValue = $property->getValue($oldEntity);
+            $newValue = $property->getValue($newEntity);
+
+            if ($oldValue !== $newValue) {
+                if (is_object($oldValue) && is_object($newValue)) {
+                    $modifiedSubValues = $this->determineModifiedValues($oldValue, $newValue);
+                    if ([] !== $modifiedSubValues) {
+                        $modifiedValues[$propertyName] = $modifiedSubValues;
+                    }
```
dresslerdemos, line 65: I'd suggest to somehow keep the information that there are two different instances set in the two entities.

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
dresslerdemos, line 42: Why is this method not used in this diff?

dresslerdemos (changes requested):

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
breierDemos: adjusted in [3412c03](https://github.com/demos-europe/demosplan-core/pull/1741/commits/3412c03af57cc1035ebac82fe51bb21d79ef2e59)

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
breierDemos: added in [2426b23](https://github.com/demos-europe/demosplan-core/pull/1741/commits/2426b238e6a21626369a0c2acd1901a00afa6995)

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
breierDemos: added in [2426b23](https://github.com/demos-europe/demosplan-core/pull/1741/commits/2426b238e6a21626369a0c2acd1901a00afa6995)

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
breierDemos: Currently the method is only called in the context of `ProcedureService::resetMapHint()`. There the `mapHint` is set under the procedure settings. I tested this case locally and debuged into the event, and indeed no changes between the procedures. I replaced the use of the method `ProcedureService::updateProcedureObject()` in [eb48b27](https://github.com/demos-europe/demosplan-core/pull/1741/commits/eb48b2765f099161068d696dc3a6fab0c1e16747) with `ProcedureService->updateProcedure()`. I added a `todo` to remind that the `ProcedureService::updateProcedureObject()` has to be refactored. What do you think.

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
breierDemos: How do you mean? For ProcedureSettings::mapHint the result would be: `['settings' => ['mapHint' => ['old' => 'old text', 'new' => 'new Text']]]` so it exist the old and the new value.

breierDemos:
> I withdraw my approve as you need to update the addon contracts first before you can use new interface

@mussbach sry for wrong commit message, composer.lock updated in [3795616](https://github.com/demos-europe/demosplan-core/pull/1741/commits/379561697e78119c90a301c5bf531201011b4358)

demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
dresslerdemos: I reviewed the method more with general usage and reusability in mind. So for example if someone calls this method with procedures like this:
```
$originalProcedure = $procedure;
$procedureClone = clone $procedure;
$settingsClone = clone $procedure->getSettings();
$procedureClone->setSettings($settingsClone);
```
My understanding is that the method would not detect any differences, even though there are different `ProcedureSettings` instances set inside the procedures. This may cause problems at some point. My suggestion is to at least explain the behavior (i.e. no support for different instances if they are otherwise equal) in the docblock of the public method, but leave it at that.

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
```diff
@@ -1112,21 +1112,15 @@ public function updateProcedure($data)
     public function updateProcedureObject(Procedure $procedureToUpdate)
     {
         try {
+            // todo: refactor this method so that there are differences between sourceProcedure and updatedProcedure
```
dresslerdemos, line 1115: We don't allow new TODOs in the code.

dresslerdemos (changes requested):
I don't understand the intention of the changes that are not related to the event.

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
breierDemos: removed in [32a9944](https://github.com/demos-europe/demosplan-core/pull/1741/commits/32a9944df961a25bce58dce1761fc4e17782d668)

dresslerdemos (approved):
There was no way found to avoid `updateProcedure` and its IDE problems. Thus approved.

PR #1746, opened 2023-07-19
chore refs (T33737): Real user object should be stored in Token

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T33737

The SecurityUser should only be used during login process. Afterward it should be swapped with the real user entity so that it could be used later on via injection of the TokenStorage

This is an alternative approach to #1740 to fix the root problem

### How to review/test
Login should work as well as the problem described in ticket

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
I thought about swapping the user in the token, but wasn't sure if we needed the securityUser still somewhere. I like this better as it is not as local as my change. I'm closing my PR 👍🏼

PR #1748, opened 2023-07-19
Fix(refs:33886): up scroll button doesnt work sometimes if display size changes

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33886

### Description: 

- the up scroll button doesnt work sometimes due to the display size 

### PR Checklist


client/js/components/button/BackToTopButton.vue
```diff
@@ -52,7 +52,7 @@ export default {
       this.scrollPos = document.documentElement.scrollTop
       const fromBottom = (this.contentHeight - 10 - this.footerheight > this.scrollPos) ? 10 : -this.contentHeight + 10 + this.footerheight + this.scrollPos
 
-      this.buttonPosition = `bottom: ${fromBottom}px; left: ${this.positionFromLeft}px; position: fixed`
+      this.buttonPosition = `bottom: ${fromBottom}px; left: ${this.positionFromLeft}px; position: sticky; z-index: 1`
```
salisdemos, line 55: when I test this, the Positioning is a bit off and the button sits on the content and not aligned with the menu anymore. Is it the same when you test it? OR is it myabe some issue with my "on the fly" tweak in the browser?
Screenshot_2023-07-21_12-03-09-back-to-top

salisdemos (changes requested):

client/js/components/button/BackToTopButton.vue
```diff
@@ -52,7 +53,7 @@ export default {
       this.scrollPos = document.documentElement.scrollTop
       const fromBottom = (this.contentHeight - 10 - this.footerheight > this.scrollPos) ? 10 : -this.contentHeight + 10 + this.footerheight + this.scrollPos
 
-      this.buttonPosition = `bottom: ${fromBottom}px; left: ${this.positionFromLeft}px; position: fixed`
+      this.buttonPosition = `bottom: ${fromBottom}px; left: ${this.positionFromLeft -80}px`
```
salisdemos, line 56: I can't test it, but I think 80 is a bit much. And maybe we have to change the `containerElement` too, not only the ``observer. DId You check that?

client/js/components/button/BackToTopButton.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <dp-button
+    class="sticky z-above-zero"
```
salisdemos, line 3: I think I found the reason, why the position is not a good Idea. We should move it to the right again
Screenshot_2023-07-24_09-51-19-grundeinstellungen-btt

salisdemos (changes requested):

salisdemos (approved):

PR #1757, opened 2023-07-20
B t33736 entwurf gespeichern

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T33736

 Description: User Token need to have departmentID. We need department to create new draft statement

### How to review/test
Code review or browser

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid:
Technically, this would have fixed it. For some reason, calling the departments before adding the user to the tokenstorage then made the departments appear if the user was later fetched. Not sure why. But as this is not the solution for the actual problem, we should close this in favor of {1769}

PR #1767, opened 2023-07-24
feature: add missing getProcedure method

jalalifardemos:
Description: getProcedure Method missed in procedureUpdateEvent

### How to review/test
we don't have any access to any page with this Error


mussbach (approved):
lgtm, thx

PR #1770, opened 2023-07-24
fix(refs T33917): fixes the lookup of available projections

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T33917

The lookup if the current projection is actually in the list of available projections happened before the available projections were properly filtered. Now it happens after, which should fix this bug after newly saving the gis layer.
Proper credit: I was just the hands while @salisdemos was directing me.

### How to review/test
With a current database, you can try to reproduce the bug in the linked ticket with/without this fix. Otherwise a code review might be enough.

### DIFF
--- a/client/js/components/map/admin/LayerSettings.vue
+++ b/client/js/components/map/admin/LayerSettings.vue
@@ -275,8 +275,6 @@ export default {
     },
 
     extractDataFromWMSCapabilities () {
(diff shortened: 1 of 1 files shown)


salisdemos (approved):
if legal, I would approve... ;-)

PR #1779, opened 2023-07-25
bug(refs T33913): Remove hint for changing profile data

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33913

The user can't change his profile data, so this text is unnecessary. We need to put the content of the embed in a new file and include this, because blocks inside an embed can't be overridden.

### How to review/test
Login as Privatperson and go to Verfahren -> Stellungnahmen -> Mitzeichnen. There should be no block about changing the user data over the hint box.


hwiem (approved):
lgtm!

PR #1780, opened 2023-07-26
bug(refs T33913): Remove hint for changing user data

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33913

The user can't edit his data, so the hint is unnecessary. We need to move the content of the embed to a new file and include it, because nested blocks inside an embed can't be overridden.


templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/includes/new_public_participation_statement_confirm_content.html.twig
```diff
@@ -0,0 +1,171 @@
...
+
+        <dt class="{{ 'layout__item u-1-of-4 u-mb u-mr'|prefixClass }}">{{ "city.name"|trans }}</dt>
+
+        <dd class="{{ 'layout__item u-7-of-12 u-ml'|prefixClass }}">{{ templateVars.user.address.city|default("----") }}</dd>
+    {% endif  %}
+
+</dl>
+
+<div class="{{ 'layout__item u-2-of-3 u-1-of-1-lap-down u-pl-0-lap-down u-mb-0_5-lap-down'|prefixClass }}">
+    {% block vote_data %}
```
spiess-demos, line 37: I feel that `vote_data` is not the ideal name. How about `user_data_hint`?

spiess-demos (changes requested):
Only one nitpick.

templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/includes/new_public_participation_statement_confirm_content.html.twig
gruenbergerdemos: Good point, I agree. Done in [bcd9cec](https://github.com/demos-europe/demosplan-core/pull/1780/commits/bcd9ceca8a5e2d3d9ce0724afd947f452a0a3f32)

spiess-demos (approved):
lgtm! (code review)

PR #1783, opened 2023-07-27
feat (T32776): correct parameter type for internal event method

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T32776

In the original implementation of `PostProcedureUpdatedEvent::determineModifiedValues()` it was assumed that every possible object value implements the `EventInterface`. It is not so. 
The method PostProcedureUpdatedEvent::determineModifiedValues() is a recursive function. The initialize call handles always the two ProcedureEntities, but inside the function, for each property that is an object, the function is possibly called recursively. And these objects are not always an entity and do not implement the EntityInterface.

### How to review/test
code review

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, code review only

PR #1784, opened 2023-07-28
feature (T33926): change endfor in statement, to understand statement…

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T33926

Description: change endfor in twig, otherwise `vote` is every time `false`

### How to review/test
review or browser

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (changes requested):
I do not think that this is a valid fix. What happens if a statement is voted by multiple persons? It seems to me that with your change the vote button or the message that it was already voted is now shown for every voter

jalalifardemos:
Without this change, `hasVoted` is every time `false`.
`{% if vote.uId == userId %} {% set hasVoted = true %} {% endif %}` check userId before `for` statement

PR #1791, opened 2023-07-31
chore: keycloak group role string needs to be configurable

mussbach:
Keycloak group role string needs to be configurable to be able to grant different permissions to different projects.

### How to review/test
Code review or add a new client to keycloak and try to grant different roles as in other client

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


martendemos (approved):
code review only

PR #1794, opened 2023-08-01
bug(refs T34112): Import missing component

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34112

Import missing component, so it can be loaded in twig. 

### How to review/test
Go to Verfahren -> Dokumente. There should be a box for uploading files.


elmasdemos (approved):
thanks for the fix! lgtm, code review

PR #1799, opened 2023-08-04
fix (refs T34139): allow resource type code reusabiliy

dresslerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34139

Must be merged with https://github.com/demos-europe/demosplan-addon/pull/69

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/ValueObject/APIPagination.php
```diff
@@ -22,7 +24,7 @@
  * @method self   setSortBy(string $sortBy)
  * @method self   setSortDirection(string $sortDirection)
  */
-class APIPagination extends ValueObject
+class APIPagination extends ValueObject implements ApiPaginationInterface
```
mussbach, line 27: The methods defined in the docblock above are now mostly defined in the ApiPaginationInterface and therefore mus be implemented in this class, otherwise the class needs to be declared abstract, which doesn't suit for ValueObjects afaik

AlexMenschoid:
I don't understand the usage of a trait here, tbh. I understand traits as a tool for horizontal inheritance, but as this one is in the core I don't really see it being used by anything else than the DplanResourceType. Am I overlooking something?

Btw: This is not a blocker, I'm just very curious as it looks rather deliberate and I would like to understand the reasoning behind it.

demosplan/DemosPlanCoreBundle/Logic/ApiRequest/ResourceType/DplanResourceTypeTrait.php
```diff
@@ -0,0 +1,197 @@
...
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\Service\Attribute\Required;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+/**
+ * Provides helper class properties and corresponding injection/initialization for {@link DplanResourceType}.
+ *
+ * Do not add any unrelated properties or methods to this trait.
+ */
+trait DplanResourceTypeTrait
```
mussbach, line 40: Not so sure whether such a huge trait is really useful. I think I would rather inline it. Where would you reuse this?

mussbach (changes requested):
thank you for this solution. Some remarks

demosplan/DemosPlanCoreBundle/Logic/ApiRequest/ResourceType/DplanResourceTypeTrait.php
dresslerdemos: The code in the trait is neither reusable as trait nor when inlined. The goal is to keep the `DplanResourceType` more maintainable and also discoverable by someone having to adjust it, than it would be if the code was inlined like it was before.

demosplan/DemosPlanCoreBundle/ValueObject/APIPagination.php
dresslerdemos: Bummer. Done in ba849603faf490e6fa73b6729042ac0146449ad7.

dresslerdemos:
Tested locally in EWM procedure, statement and segment list/filters. Edit works too.

mussbach (approved):
lgtm, thx, pair programming review

PR #1801, opened 2023-08-04
Fix(T34174): issue with lists and their numbering

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34174

**Description:** This PR fixes the issue with lists and their numbering by adding styling for both `ul` and `ol` tags in the AT (Abwägungstabelle). The problem was that the initial value for `ol` did not work as expected, displaying bullets instead of numbers. As a solution, this PR sets the appropriate styling for each tag to resolve the numbering problem.


hwiem (approved):
lgtm

PR #1802, opened 2023-08-04
Fix(refs:34175): docx export is empty

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34175

### Description: 

- the computed  `selectionContainsUnclaimedFragments` cannot be used in every project, only in projects where the permission `area_statements_fragment` is activated
- so the error said for this computed function:  `Reduce of empty array with no initial value`

### PR Checklist

### DIFF
--- a/client/js/components/statement/assessmentTable/DpSelectedItemsStatements.vue
+++ b/client/js/components/statement/assessmentTable/DpSelectedItemsStatements.vue
@@ -173,19 +173,15 @@ export default {
     },
 
     selectionContainsUnclaimedFragments () {
-      if (hasPermission('area_statements_fragment')) {
-        return Object.keys(this.selectedElements)
-          .map(id => this.fragments[id] || [])
-          .map(storeFragment => storeFragment.fragments || [])
-          .reduce((acc, fragments) => {
-            return [...acc, ...fragments]
-          })
-          .filter(fragment => {
-            return (fragment.assignee.id !== this.currentUserId) || (fragment.departmentId && fragment.departmentId !== '')
-          }).length > 0
-      } else {
-        return false
-      }
+      return Object.keys(this.selectedElements)
+        .map(id => this.fragments[id] || [])
+        .map(storeFragment => storeFragment.fragments || [])
+        .reduce((acc, fragments) => {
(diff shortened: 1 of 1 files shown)


client/js/components/statement/assessmentTable/DpSelectedItemsStatements.vue
```diff
@@ -173,15 +173,17 @@ export default {
...
-      return Object.keys(this.selectedElements)
-        .map(id => this.fragments[id] || [])
-        .map(storeFragment => storeFragment.fragments || [])
-        .reduce((acc, fragments) => {
-          return [...acc, ...fragments]
-        })
-        .filter(fragment => {
-          return (fragment.assignee.id !== this.currentUserId) || (fragment.departmentId && fragment.departmentId !== '')
-        }).length > 0
+      if (hasPermission('area_statements_fragment')) {
```
hwiem, line 176: Shouldn't it also work if in line 350 we put the permission check first? If it evaluates to false, this code should not be executed.
Should we keep the permission check here, I'd suggest to return `false` if it fails.

elmasdemos: I agree with you. I recognized it after pushing this solution

elmasdemos: I mean the return false I would like add here but the check I would like keep it here

hwiem: Could you explain why? :)

hwiem: No, I'd change the code in line 350 from

```
if (this.selectionContainsUnclaimedFragments && hasPermission('area_statements_fragment')) {
  this.triggerWarning('warning.edit.selection.fragments.not.claimed')
  isAllowed = false
}
```

to

```
if (hasPermission('area_statements_fragment') && this.selectionContainsUnclaimedFragments) {
  this.triggerWarning('warning.edit.selection.fragments.not.claimed')
  isAllowed = false
}
```

so in case the permission check fails, the computed is not even executed.

PR #1822, opened 2023-08-09
Fix(T34155): console issue: uncaught (in promise) DOMException

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34155

**Description:** This PR fixes the console issue: Error: Uncaught (in promise) DOMException.

The problem occurs on the AT page after clicking the link "Originalstellungnahme" in the flyout menu. This action redirects the user to the "Originalstellungnahme" table, and a hash (with the first statement's id) is added to the URL. The issue appears when the hash is not the last element in the URL. In this case, the `window.location.hash` property takes everything after the '#' symbol, including any characters until the end of the URL.

This becomes problematic when there are additional page parameters, such as r`_limit=25&page=1` appended after the hash. These parameters should not be included in the hash. One approach to resolving this issue is by checking whether the string includes a hash and then removing everything after the '?' separator, which separates the base path from the query parameters.


gruenbergerdemos (approved):
lgtm

PR #1823, opened 2023-08-09
chore (refs T34139): fix connection to data services

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T34139

`AiApiUser` needs to be attached to a database user as it is needed by doctrine actions. Moreover, addon permissions are checked directly via database and therefore a database user needs to exist anyhow

Along the way code duplication in FunctionalUsers is reduced and Addon interfaces used.

### How to review/test
try to import a pdf

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm

PR #1835, opened 2023-08-11
fix(refs T34274): Safely generate imprint route

graupnerdemos:
By using Symfony's router to generate the url to be displayed as final imprint url we can be much more sure that the displayed url will be the one users would be expecting there.

**Ticket:** https://yaits.demos-deutschland.de/T34274

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm

PR #1840, opened 2023-08-14
upgrade demosplan-ui to the latest 0.1.11 version

sakutademos:
Description: To fix this [bug](https://yaits.demos-deutschland.de/T34111) need to upgrade demosplan-ui to the latest version, which is 0.1.11


gruenbergerdemos (approved):
lgtm

PR #1841, opened 2023-08-14
update yarl.lock file with the right version of the demosplan-ui

sakutademos:
**Description:** It should remain ` "@demos-europe/demosplan-ui@^0"` instead of the specific version `"@demos-europe/demosplan-ui@^0.1.11"` in the **yarn.lock** file. It should be same as in the package.json file.


gruenbergerdemos (approved):
Oh, that's right, thanks!

PR #1842, opened 2023-08-14
Fix(T34174): bulleted and numbered lists issue

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34174

**Description:** This PR fixes the issue with lists and their numbering by adding styling for both `ul` and `ol` tags. The problem was that the initial value for `ol` did not work as expected, displaying bullets instead of numbers. As a solution, this PR sets the appropriate styling for each tag to resolve the numbering problem.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_public-statement.scss
+++ b/demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_public-statement.scss
@@ -55,4 +55,12 @@
         font-weight: $bold-font-weight;
         font-family: $bold-font-family;
     }
+
+    ul {
+        list-style: disc;
+    }
+
+    ol {
+        list-style: decimal;
+    }
 }


demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_public-statement.scss
```diff
@@ -55,4 +55,12 @@
...
         font-family: $bold-font-family;
     }
+
+    ul {
+        list-style: disc;
+    }
+
+    ol {
+        list-style: decimal;
+    }
```
salisdemos, line 65: is this an issue only in the public statement lists or is it more general?
If so, shouldn't it be within `__content-item` ?

sakutademos: So you mean these changes have to be applied to the `.c-public-statement__content-item` class? If so, it does not work..

The issue appears in the AT (Abwägungstabelle), TipTap and the public statement component (these problems are recognized). They each have distinct classes.. Because of that I'm applying styling for `ul` and `ol` to these components individually (There are other PRs, this for example: https://github.com/demos-europe/demosplan-core/commit/891c8f5ff2a23f0e227fc1eda04024937fdd1720)

If I implement it more globally, there might be issues in other places where list styles should not be applied... or am I wrong?

salisdemos: sad, that @spiess-demos is not available. So we may fix that, the way You started... I think we should do it the other wayy round, but if I recall it correct, @spiess-demos wanted to implement it the way You fixed it..

salisdemos (approved):

PR #1843, opened 2023-08-14
feature (refs T33365): Set DocX language to german

MoritzMandler:
Ticket:
 https://yaits.demos-deutschland.de/T33365

Description:
DocX exports should use the static method `getPreConfiguredPhpWord()` which sets the language and other things like spelling correctly to german.
If instead a new instance is created via new PhpWord() like prior this PR, these settings need to be made separately.
Changed all places to implement the static preconfig mehtod.

### How to review/test
Try to download DocX files and check their language settings

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, code review only

PR #1852, opened 2023-08-15
Fix:(T34286) incorrect time in the statement versions

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34286

**Description:** This PR fixes the problem with dates in the statement versions. It employs the `createdDate` for the last item within the `draftStatementVersions` array to display the `createdDate` instead of using `versionDate`.

- issue with sorting by `createdDate` has also been resolved (BE)


AlexMenschoid (approved):
BE lgtm, code review only

salisdemos (approved):
As far as I can tell it looks fine.

PR #1854, opened 2023-08-15
fix(refs T34313): Set active Tab

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34313

### How to review/test

As Institution Coordinator goto
`/verfahren/<prcedure-id>/stellungnahmen/endfassungenGruppe`
- You need voted statements. 
If You use the upper Navigation (not the Tabs on the right hand side) then the tab nd content should change too when switching between "my statements" and "voted statements"

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


elmasdemos (approved):
lgtm, code review and it works

PR #1858, opened 2023-08-16
fix(refs T34314): Re-Add list styles for proseMirror / fragments

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34314

While moving to tailwind we lost the default styles for ul/ol. Since we only want to add them in placecs, we need them and not globally and then reset them, we have to find all the places where they are lost now. here are two...

--> statement_fragments-list
--> "split statement"

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


sakutademos (approved):
code review, lgtm

PR #1862, opened 2023-08-17
fix(refs T29585): improves the default value handling for addon controller_paths

AlexMenschoid:

In this version you can either just ignore the value for 'controller_paths' in your addon manifest (like this: `controller_paths:`) or completely remove the key 'controller_paths' altogether. In both cases the system will default back to '/src/Controller' as per Symfony Best Practice. If you do want to specify one or more deviating paths, you need to add them as an array.

Sadly, I did not find any reasonable way to include the fallback for a single path as a string (i.e. `controller_paths: '/path/to/controllers'`) as that would require casting it as an array before normalization. But that messes with a default fallback for an empty key.

### How to review/test
This can be tested by unpacking an addon of your choice and then changing the `controller_paths` key in the 'demosplan-addon.yml' in the unpacked addon. After either leaving it empty or removing it or adding one or more paths as an array, you can create a zip from it again. Now install that addon-zip as usual and check out the `controller_paths` values in the resulting 'addons.yml'.

### Linked PRs (optional)
PR introducing this feature #1490


MoritzMandler (approved):
lgtm - tested - works well

PR #1866, opened 2023-08-18
fix(refs T34144): Make Request syncronouse again

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34144

we tried to remove jQuery here and replaced ajax with (under the hood) axios.
(@see 43ad166fab54583ac4147b19071b3ab815381925)

that didn't work out because axios doesn't have a syncronous flag like ajax. Without that, we run into a stacked trap of race conditions which can't be resolved easily without rebuilding huge parts of this file.

Since I don't have a better solution, I roleback the changes from once.

### How to review/test

https://diplanrog-dev.demos-europe.eu/verfahren/f96129bb-56c2-4b62-96e4-f52628adefc6/public/detail -> "test layer" should load. (there should appear getMap-Requests from https://geodienstelandesplanungstage.bob-sh.de/mapproxy/lep)

### PR Checklist

### DIFF
--- a/client/js/components/map/publicdetail/Map.vue
+++ b/client/js/components/map/publicdetail/Map.vue
@@ -602,17 +602,21 @@ export default {
 
       if (serviceType === 'wmts') {
         const layerArray = Array.isArray(layer.attributes.layers) ? layer.attributes.layers : layer.attributes.layers.split(',')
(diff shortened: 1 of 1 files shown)


elmasdemos (approved):
lgtm, code review and it works

PR #1874, opened 2023-08-18
fix(refs T33908):  Prevent map from panning/zooming out of extent

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T33908

use set extent to stay within the allowed area of the map
The delete flag has the effect, that the panning and zoomout in the map handles the mapExtent diffferently than our expectations.

### How to review/test

as FP-A set the "Kartenbegrenzung" to a narrow space. (/verfahren/<procedure-id>/verwalten/globaleGisEinstellungen)
then goto the public Detail and try to zoom out / pan out that area. That should not work anymore

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/map/publicdetail/Map.vue
+++ b/client/js/components/map/publicdetail/Map.vue
@@ -1767,7 +1767,6 @@ export default {
     },
 
     redrawMap () {
-      const map = this.map
       this.map.updateSize()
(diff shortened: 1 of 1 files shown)


hwiem (dismissed):
While I can't say for what reason the `constrainOnlyCenter` option was introduced, this change seems legit to me

PR #1876, opened 2023-08-18
chore: Refactor `createLayerSource()`

salisdemos:
split the method createLayerSource  into 3 methods. this should not affect anything. Layers (WMS and WMTS) should still work.

This is a side outcome while bugfixing [T34144](https://yaits.demos-deutschland.de/T34144)

### DIFF
--- a/client/js/components/map/publicdetail/Map.vue
+++ b/client/js/components/map/publicdetail/Map.vue
@@ -36,7 +36,7 @@ import { addProjection, Projection, transform } from 'ol/proj'
 import { Attribution, FullScreen, MousePosition, OverviewMap, ScaleLine } from 'ol/control'
 import { Circle, Fill, Stroke, Style } from 'ol/style'
 import { defaults as defaultInteractions, DragZoom, Draw } from 'ol/interaction'
-import { dpApi, DpAutocomplete, externalApi, formatDate, hasOwnProp, prefixClassMixin } from '@demos-europe/demosplan-ui'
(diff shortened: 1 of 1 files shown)


client/js/components/map/publicdetail/Map.vue
```diff
@@ -1587,6 +1539,60 @@ export default {
       return layers
     },
 
+    getWMTSSource (layer) {
+      const layerArray = Array.isArray(layer.attributes.layers) ? layer.attributes.layers : layer.attributes.layers.split(',')
+      const url = this.addGetCapabilityParamToUrl(layer.attributes.url)
+
+      return $.ajax({
```
gruenbergerdemos, line 1546: Why did you change this back to ajax?

salisdemos: because ajax has this tiny "async: false" (what ever that does in real - because technically its always async) Option, which is nor provided by axios. And the code is so messy, that I wasn't able to refactor it, that it works asynchronously.

salisdemos: this is a bit mixced up. I initially wanted to cherrypick https://github.com/demos-europe/demosplan-core/pull/1866. I thought the cherrypick-PR would be faster that this one. Then I waited for the approval for 1866 and created this PR in the meanwhile which lead to the mess...

gruenbergerdemos (approved):
lgtm

PR #1886, opened 2023-08-21
Remove default margins for ul, ol elements

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T27595

While inuitcss followed the (opinionated) approach of applying a default margin to ul/ol elements, tailwind goes the other way by removing margin for all elements by default. As we're designing a webapp here, where ul elements are mostly used for semantic purposes (not as part of content), the tailwind approach is the way to go. Wherever we need margins on ol or ul, it will be applied manually by using a spacing utility, or, if the intention is to structure content for better readability, use a wrapper element with the class `c-styled-html`.

So, the class `list-style-none` is removed from ul/ol elements, because this is the new default, as the default margins are removed from inuitcss, too.

### How to review/test
Things should be spaced as before.

### PR Checklist


salisdemos (approved):
Just one question in the comments.
Apart from that legit. If we want to go this way

PR #1888, opened 2023-08-21
chore: remove unneeded tooltip lib

spiess-demos:
As we're using v-tooltip in almost every tooltip instance now, the legacy tooltip can be removed. The last occurrences are refactored to also use v-tooltip.

### How to review/test
Tooltips should work as before, also in the places that have been changed with this PR.

### PR Checklist


templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_member_list_entry.html.twig
```diff
@@ -23,9 +23,8 @@
                 <br>
                 {% if orga.masterToeb is defined %}
                     {% for key, district in districts %}
-                        <span class="o-tooltip inline-block u-mr-0_25">
+                        <span class="inline-block u-mr-0_25" v-tooltip="{{ district.name }}">
```
salisdemos, line 26:
```suggestion
                        <span
                          class="inline-block u-mr-0_25"
                          v-tooltip="{{ district.name }}">
```

salisdemos (dismissed):
Just some none-blocing line-brake nitpicks

PR #1890, opened 2023-08-22
Rename component to use agnostic name

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T24040

Tiptap is our current implementation for DpEditor, however it makes no sense to reflect that in the naming of an application specific component.

### PR Checklist


salisdemos (approved):
Thank You. I wanted to do this long time ago...

PR #1892, opened 2023-08-23
fix: correct some rather small padding issues

spiess-demos:
- as the icon in DpResettableInput has changed height, this needs to be adapted in the corresponding css
- DpFormRow applies a padding on each child, which led to a weird button spacing on the login button. this is corrected by simply removing DpFormRow which is not needed here anyway, as the form fields are each on their own line.

### How to review/test
- The X icon in the search field in ewm procedurelist should be aligned correctly
- The "Anmelden" button on the login page should have equal padding to the left and right

### PR Checklist


ahmad-demos (approved):
lgtm, codeReview

PR #1893, opened 2023-08-23
fix(refs T33908): use set extent to stay within the allowed area of the map

salisdemos:
cherrypick from https://github.com/demos-europe/demosplan-core/pull/1874

**Ticket:** https://yaits.demos-deutschland.de/T33908


elmasdemos (approved):
lgtm, code review

PR #1897, opened 2023-08-24
refactor: avoid entity wrapping for twig

dresslerdemos:
A previous attempt to pass objects into
twig files that use the same properties
as the resources returned by the generic
web API did not establish itself. The
attempt is thus rolled back,
simplifying code logic.

As the `ProcedureTypeResourceType` does
not apply any aliases or special
property permissions, no adjustment of
the twig files is necessary.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm

PR #1909, opened 2023-08-28
chore: Remove unused prosemirror

salisdemos:
To Use TipTap v2 with ease we have to move our prosemirror-dependencies to demosplan-ui to make sure the versions of prosemirror don't conflict.

And while moving the files I recognized, that they are not in use anyway here. Se we can just delete them


spiess-demos (commented):
As i understand it, these prosemirror libs are used in SplitStatementView, which has been moved to demosplan-addon-demospipes. However this was not well thought out, as the base functionality of splitting statements into segments should be available in demosplan-core. However we can move SplitStatementView back to core, along with all needed dependencies, in the future.

salisdemos:
I am not sure if it has to live in core. Maybe it needs a dedicated addon...
But as long as the component is not in core we should remove the dependencies too. Moving them back in is easy and we don't know when or even if a movement back will happen

spiess-demos (approved):
lgtm (code review)

PR #1923, opened 2023-08-30
chore: External orga id from identity provider can be longer than 36 chars

mussbach:
The id of an organization is saved as gw_id to be able to cope with orgachanges. When logging in from external identity providers this id might be longer than 36 chars.

### How to review/test
Login via an idp with an id more than 36 Chars

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, code review only

PR #1940, opened 2023-09-01
feature (T32026) : remove unused attribute from handler

jalalifardemos:
**Tickets**: https://yaits.demos-deutschland.de/T32026

**Description**: Remove 'email2' from handler 

### Linked PRs (optional)

- https://github.com/demos-europe/demosplan-project-ewm/pull/86

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/User/OrgaHandler.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/User/OrgaHandler.php
@@ -159,7 +159,6 @@ class OrgaHandler extends CoreHandler
             'ccEmail2',
             'copySpec',
             'cssvars',
-            'email2',
             'emailNotificationEndingPhase',
             'emailNotificationNewStatement',
             'participationEmail',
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Logic/User/OrgaHandler.php
```diff
@@ -159,7 +159,6 @@ public function getWritableAttributes(): array
             'ccEmail2',
             'copySpec',
             'cssvars',
-            'email2',
```
AlexMenschoid, line 162: Why do you think, this is no longer used? I still see the property on the Orga entity which to me suggests, that the attribute should also still be writable.

AlexMenschoid (changes requested):

demosplan/DemosPlanCoreBundle/Logic/User/OrgaHandler.php
jalalifardemos: In `DpOrganisationFormFields.vue:407-584` used **email2** as a condition and i should remove it to not show email field. (base on Florian's Comment )

jalalifardemos: what about [3250e47](https://github.com/demos-europe/demosplan-core/pull/1940/commits/3250e47382beef4c3f55e0e9e394c0196418d0e3)

AlexMenschoid (approved):
lgtm, code review only

PR #1941, opened 2023-09-01
feature : refacor Permissions base on EWM disabled method

jalalifardemos:
Description: Remove all permissions, that in EWM used in disabled() method


jalalifardemos:
Because reviewing the code is complicated and difficult. I remove the method step by step. In the first step, I have removed only for EWM.
https://github.com/demos-europe/demosplan-core/pull/1983

PR #1942, opened 2023-09-01
fix: commit the yarn.lock changes manually

spiess-demos:
Something got mixed up, so yarn.lock is not in sync with the actual module resolution anymore. if `yarn install` does not result in the current yarn.lock resolutions, it will try to change yarn.lock which fails if `--frozen-lockfile` is set.


mussbach (approved):
lgtm, thx

PR #1956, opened 2023-09-05
fix: make icons appear centered within buttons

spiess-demos:
This is a minor ui fix. Before, icons within the two upmost buttons were misaligned:
image

This is fixed now:
image


spiess-demos:
wrong branch

PR #1957, opened 2023-09-05
fix: make icons appear centered within buttons

spiess-demos:
This is a minor ui fix. Before, icons within the two upmost buttons were misaligned:
image

This is fixed now:
image

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
@@ -229,6 +229,7 @@
           }"
           @click="isFullscreen = !isFullscreen">
           <dp-icon
+            class="inline-block"
             :icon="isFullscreen ? 'compress' : 'expand'"
             aria-hidden="true" />
         </button>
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -259,7 +260,9 @@
           }"
           @click.prevent="showSegmentVersionHistory"
           data-cy="segmentVersionHistory">
-          <dp-icon icon="history" />
+          <dp-icon
+            class="inline-block"
```
salisdemos, line 264: should this be added in dpIcon? Or are there cases, whee its not usefull to set it to inline-block?

salisdemos (approved):

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
spiess-demos: this was my first thought as well, but then, i realised that the "new standard" way of doing micro ui would rather involve `inline-flex` which did not work for this special case. Iguess more research is needed to set a reasonable default for the `display` property in `DpIcon`.

PR #1958, opened 2023-09-05
Delete procedure and all related things command (via SQL)

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T32997

This is a rather risky and hacky solution for deleting a procedure with everything related to it. It is meant as an interim solution while we are working on a more stable way of deleting all related parts in the system preferably with Doctrine.

We need this solution to clean everything pertaining to customers that had to be deleted. Since we are not able to cleanly delete statements or procedures or some other related things via our repositories, this direct approach in SQL has been chosen to accomplish the task in a somewhat timely manner.

I feel somewhat confident that I got most if not all entries I have to delete. But I might have missed something.

I tested it with multiple procedures in an older Prod-DB-Dump and it seemed to work so far.

### How to review/test
If you have a procedure with statements and annotated statement pdfs and everything else, then try to delete it with this command.
During code review: please pay attention to what entries in what tables get deleted and especially if you think I forgot to include something that also has a relationship to the procedure!

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
+++ b/demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
@@ -161,8 +161,6 @@ class DeleteProcedureCommand extends CoreCommand
     {
         $statementIds = array_column($this->fetchFromTableByProcedure(['_st_id'], '_statement', '_p_id'), '_st_id');
 
-        print_r($statementIds);
-
         // delete statement meta
         $this->output->writeln("Deleting Statement-Meta");
         $this->deleteStatementMeta($statementIds);
@@ -185,6 +183,12 @@ class DeleteProcedureCommand extends CoreCommand
 
     private function processStatementAttachments(array $statementIds): void
     {
+        if (!$this->doesTableExist('statement_attachment')) {
+            $this->output->writeln("No table with the name 'statement_attachment' exists in this database. Data could not be fetched.");
+
+            return;
+        }
+
         $statementAttachmentQueryBuilder = $this->dbConnection->createQueryBuilder();
         $statementAttachmentQueryBuilder
             ->select('id', 'file_id')
@@ -247,6 +251,12 @@ class DeleteProcedureCommand extends CoreCommand
 
     private function processImportEmailAttachments(array $importEmailIds): void
     {
+        if (!$this->doesTableExist('statement_import_email_attachments')) {
+            $this->output->writeln("No table with the name 'statement_import_email_attachments' exists in this database. Data could not be fetched.");
+
+            return;
+        }
+
         $importEmailAttachmentQueryBuilder = $this->dbConnection->createQueryBuilder();
         $importEmailAttachmentQueryBuilder
             ->select('file_id')
@@ -418,6 +428,12 @@ class DeleteProcedureCommand extends CoreCommand
 
     private function deleteReportEntriesByIdentifierAndType(string $identifierType, array $identifierArray): void
     {
+        if (!$this->doesTableExist('_report_entries')) {
+            $this->output->writeln("No table with the name '_report_entries' exists in this database. Data could not be deleted.");
+
+            return;
+        }
+
         $deletionQueryBuilder = $this->dbConnection->createQueryBuilder();
         $deletionQueryBuilder
             ->delete('_report_entries')
@@ -438,6 +454,12 @@ class DeleteProcedureCommand extends CoreCommand
 
     private function deleteFromTableByIdentifierArray(string $tableName, string $identifier, array $ids): void
     {
+        if (!$this->doesTableExist($tableName)) {
+            $this->output->writeln("No table with the name $tableName exists in this database. Data could not be deleted.");
+
+            return;
+        }
+
         $deletionQueryBuilder = $this->dbConnection->createQueryBuilder();
         $deletionQueryBuilder
             ->delete($tableName)
@@ -456,6 +478,12 @@ class DeleteProcedureCommand extends CoreCommand
 
     private function fetchFromTableByProcedure(array $targetColumns, string $tableName, string $identifier): array
     {
+        if (!$this->doesTableExist($tableName)) {
+            $this->output->writeln("No table with the name $tableName exists in this database. Data could not be fetched.");
+
+            return [];
+        }
+
         $fetchQueryBuilder = $this->dbConnection->createQueryBuilder();
         $fetchQueryBuilder
             ->select(...$targetColumns)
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
```diff
@@ -0,0 +1,479 @@
...
+     * Find all statements + segments
+     * Iterate over all Statements and delete related stuff like meta, attachments, etc.
+     *
+     * @throws Exception
+     */
+    private function processAllStatements(): void
+    {
+        $statementIds = array_column($this->fetchFromTableByProcedure(['_st_id'], '_statement', '_p_id'), '_st_id');
+
+        print_r($statementIds);
```
mussbach, line 164: debug statement should be removed or $output used

demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
```diff
@@ -0,0 +1,479 @@
...
+            $this->output->writeln('Deleting Procedure');
+            $this->deleteProcedure();
+
+            // reactivate foreign key checks
+            $this->output->writeln('Activate FK Checks');
+            $this->activateForeignKeyChecks();
+
+            $this->output->writeln("Procedure $this->procedureId was purged successfully!");
+
+            return Command::SUCCESS;
```
mussbach, line 145: When not in dry-run you should repopulate elasticsearch

demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
```diff
@@ -0,0 +1,479 @@
...
+        $this->deleteReportEntriesByIdentifierAndType('statement', $statementIds);
+        // delete statements
+        $this->output->writeln('Deleting Statements');
+        $this->deleteFromTableByIdentifierArray('_statement', '_p_id', [$this->procedureId]);
+    }
+
+    private function processStatementAttachments(array $statementIds): void
+    {
+        $statementAttachmentQueryBuilder = $this->dbConnection->createQueryBuilder();
+        $statementAttachmentQueryBuilder
```
mussbach, line 189: Perhaps you should always check whether a table exists to make the command more robust. Scheme may change and it would be more safe regarding addons

mussbach (commented):
thanks a lot for this! I did not check every table name, key and relation, but it looks quite good to me. We could start with this and in case we find some more relations improve. Some remarks

AlexMenschoid:
> First question is: https://yaits.demos-deutschland.de/T32997 is about deleting customer. This command deleting procedures. Why?

2 reasons: 
1. Deleting procedures is going to be part of deleting customers but also useful as its own thing. In the future, deleting statements should also become its own command/service that is only used in deleting procedures.
2. Since the customers and some related stuff has already been deleted in the 2 referenced tickets, we need a way to handle remaining things, like procedures, without having the customer present.

Regarding the missing deletion from the tables you mentioned: 
I have some of those tables already in my script and I noticed some irregularities regarding that as well. I'm already investigating. But some others I very much might have missed.

demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
AlexMenschoid: done in [86915ab](https://github.com/demos-europe/demosplan-core/pull/1958/commits/86915ab1e3d524cbe7f84dde906ed09fd0352f95)

AlexMenschoid:
I also wrapped all database interactions in a transaction to provide more stability in [f70200c](https://github.com/demos-europe/demosplan-core/pull/1958/commits/f70200cf476ca3c3596fe9a999e3d41823f1605b). Thanks for the idea @martendemos

demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
mussbach: you should probably pass the environment, as the command will be performed at prod servers, where no dev environment is available

AlexMenschoid: Done in [5160eff](https://github.com/demos-europe/demosplan-core/pull/1958/commits/5160effd139a87571a03d6aac785795342c4f215)

demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
```diff
@@ -540,8 +540,12 @@ private function repopulateElasticsearch(): void
             return;
         }
 
+        $env = $this->parameterBag->get('kernel.environment');
+        $this->output->writeln("Repopulating ES with env: $env");
+
+        $repopulateEsCommand = 'dev' === $env ? 'dplan:elasticsearch:populate' : 'dplan:elasticsearch:populate -e prod';
```
mussbach, line 546: Nitpick: you could even add `--no-debug` to add some more performance

AlexMenschoid: Done in [1e10b10](https://github.com/demos-europe/demosplan-core/pull/1958/commits/1e10b1063a5e74cc76a420fd31aa1e1c8a12234b)

AlexMenschoid:
I added more things to be deleted in [972227c](https://github.com/demos-europe/demosplan-core/pull/1958/commits/972227c9de4752a420a6d5dec524ecaa7595c1e6) and [f989ed5](https://github.com/demos-europe/demosplan-core/pull/1958/commits/f989ed502c9863de4d5403f926fd18f605531a90). The second commit also fixes string handling when setting an array as a parameter for the SQL query.

I consider this a good wrap up point. Maybe I'm still missing some things, but nothing major anymore, I hope.

demosplan/DemosPlanCoreBundle/Command/Data/DeleteProcedureCommand.php
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    private function deleteStatementAttachment(array $statementIds): void
+    {
+        $this->deleteFromTableByIdentifierArray('statement_attachment', 'statement_id', $statementIds);
+    }
+
+    private function deleteFiles(array $fileIds): void
+    {
+        $this->deleteFromTableByIdentifierArray('_files', '_f_ident', $fileIds);
```
MoritzMandler, line 490: Shouldn't these files also be removed physically?
```DemosPlanCoreBundle/Logic/FileService::deleteFile($hash)```

AlexMenschoid: Yes, I could do that. But if I remember correctly, we also have an orphaned files removal in our maintenance service/command/whatever. So I decided to leave the cleanup to that.

mussbach: Indeed and I would also go with the existing orphan removal in MaintenanceTask

mussbach (approved):
I think this is a very good starting point. I vote for merge and fix if some relations are missing

AlexMenschoid:
Even if incomplete, this is going to leave us in a way better position for the current use cases. This can be improved in the future to be more complete or ideally be replaced very soon.

PR #1960, opened 2023-09-05
bug (refs T34564): enable recommendation icon in editor

muellerdemos:
https://yaits.demos-deutschland.de/T34564

We shall not check for segment relationships here anymore but if we have an addon enabled. The data prop async components is representing this state hence the check was adjusted accordingly.

(I could not test this locally properly since the segmentation does not work locally atm, however if i checked several existing segments, the icon appears for all of them)


gruenbergerdemos (approved):
lgtm, code review only

PR #1961, opened 2023-09-05
fix(T34304): issue with numbering in the ListStatements component

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33650

Description:  Add a class called `table-text` to the `ListStatements` component to apply styling to text containing the `ul` and `ol` tags within the table.

### DIFF
--- a/client/js/components/statement/listStatements/ListStatements.vue
+++ b/client/js/components/statement/listStatements/ListStatements.vue
@@ -140,7 +140,7 @@
         </template>
         <template v-slot:text="{ text }">
           <div
-            class="line-clamp-3 break-words"
+            class="line-clamp-3 c-styled-html"
             v-cleanhtml="text" />
         </template>
         <template v-slot:flyout="{ assignee, id, originalPdf, segmentsCount, synchronized }">
--- a/client/js/components/statement/listStatements/ListStatements.vue
+++ b/client/js/components/statement/listStatements/ListStatements.vue
@@ -231,7 +231,7 @@
           </statement-meta-data>
 
           <!-- Statement text -->
-          <div class="u-pt-0_5 break-words">
+          <div class="u-pt-0_5 break-words table-text">
             <strong>{{ Translator.trans('statement.text.short') }}:</strong>
             <template v-if="typeof fullText === 'undefined'">
               <div v-cleanhtml="text" />
--- a/client/js/components/statement/listStatements/ListStatements.vue
+++ b/client/js/components/statement/listStatements/ListStatements.vue
@@ -231,7 +231,7 @@
           </statement-meta-data>
 
           <!-- Statement text -->
-          <div class="u-pt-0_5 break-words table-text">
+          <div class="u-pt-0_5 break-words c-styled-html">
             <strong>{{ Translator.trans('statement.text.short') }}:</strong>
             <template v-if="typeof fullText === 'undefined'">
               <div v-cleanhtml="text" />
--- a/client/js/components/statement/listStatements/ListStatements.vue
+++ b/client/js/components/statement/listStatements/ListStatements.vue
@@ -231,7 +231,7 @@
           </statement-meta-data>
 
           <!-- Statement text -->
-          <div class="u-pt-0_5 break-words c-styled-html">
+          <div class="u-pt-0_5 c-styled-html">
             <strong>{{ Translator.trans('statement.text.short') }}:</strong>
             <template v-if="typeof fullText === 'undefined'">
               <div v-cleanhtml="text" />
--- a/demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_data-table.scss
+++ b/demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_data-table.scss
@@ -158,23 +158,4 @@
 
         background-color: $dp-color-white;
     }
-
-    .table-text {
-        & ul,
-        & ol {
-            padding-left: 1.3em;
-
-            & li > p {
-                margin: 0;
-            }
-        }
-
-        & ul {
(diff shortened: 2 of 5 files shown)


demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_data-table.scss
```diff
@@ -158,4 +158,23 @@
...
+
+    .table-text {
+        & ul,
+        & ol {
+            padding-left: 1.3em;
+
+            & li > p {
+                margin: 0;
+            }
+        }
```
salisdemos, line 170: I am not sure, but if I did't forget everything about scss we can omit the `&` all along.

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -231,7 +231,7 @@
           </statement-meta-data>
 
           <!-- Statement text -->
-          <div class="u-pt-0_5 break-words">
+          <div class="u-pt-0_5 break-words table-text">
```
salisdemos, line 234: @spiess-demos do we have or want to have a generic reset-class?

spiess-demos: As i understand the code changes, it is not about a reset, but about the same issue that has arisen in different places since we employed the tailwind reset for ul and ol elements.

for this, i'd argue to use `c-styled-html`  which is defined here: DemosPlanCoreBundle/Resources/client/scss/components/_styled-html.scss:86 

If you like, you could extend it with the p reset:

``` 
           & li > p {
                margin: 0;
            }
```

However i tend to believe that the actual bug originates in the splitStatement logic which does not wrap li elements within subsequent segments with the needed `ul` elements.

sakutademos: The problem is that the `ul` and `ol` elements have a global `list-style: none` declaration. To resolve this, we need to specify the list style where it's required...

If you suggest that `c-styled-html` is the appropriate class for these purposes, I have to make some adjustments 
(like add the list style) --> https://github.com/demos-europe/demosplan-core/pull/1961/commits/f9f46c2452acb7b8bdb95d564b4f98095eae3358 Is it ok?

Does the `li` tag really require the `margin-bottom: 12px` as declared in the `c-styled-html` class?
image

or can it be removed? 

image

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -231,7 +231,7 @@
           </statement-meta-data>
 
           <!-- Statement text -->
-          <div class="u-pt-0_5 break-words table-text">
+          <div class="u-pt-0_5 break-words c-styled-html">
```
spiess-demos, line 234: I think you can remove the `break-words` utility then, too.

client/js/components/statement/listStatements/ListStatements.vue
spiess-demos: It all goes back to [this commit of mine](https://github.com/demos-europe/demosplan/commit/8c1f5b5a55b8e221e19b9b182b72d5aa5c939877). We can see the importance of good commit messages here :) However i remember that _paragraph.scss is the precursor of _styled-html.scss, and it was first bootstrapped to style the content in templates/bundles/DemosPlanCoreBundle/DemosPlanDocument/public_paragraph_document.html.twig:183. However this was just as opinionated as it would probably be to remove the margin-bottom, i guess it is ok to remove it!

demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_styled-html.scss
```diff
@@ -89,10 +89,20 @@
         min-width: 50px;
 
         > li {
-            margin-bottom: $inuit-base-spacing-unit--small;
+            p {
+                margin: 0;
```
spiess-demos, line 93: Why do you set the p margin to 0, here? Apart from a p inside a li is not valid html, we should only "reset the reset" here (as done with the list style), but not remove spacing that user agents would also apply.

Which brings me to the point that we should check all elements for reasonable spacing within `styled-html`.

sakutademos: This type of structure, where the `<p>` tag is inside the `<li>` tag, is created in the editor and passed to the `ListStatement` component as text:
`<ol><li><p>eins</p></li><li><p>zwei</p></li><li><p>drei</p><ul><li><p>punkt</p></li><li><p>punkt</p></li><li><p>punkt</p><ol><li><p>eins</p></li></ol></li></ul></li></ol><p></p>`

The `<p>` tag has a margin-bottom. If I remove the margin only from the `<li>` element, it remains due to the `<p>` tag. That's why I made this adjustment. To resolve this (`<p>` tag is inside the `<li>` tag) , it should be fixed in the editor (DpEditor).

sakutademos: remove styling for `<p>` tag inside the `<li>` here https://github.com/demos-europe/demosplan-core/pull/1961/commits/dca72886a18ef5b6488c87f80760270224b3de34

spiess-demos: when thinking about it for a second time, i guess it is ok to remove the margin on the p element.

sakutademos: done here https://github.com/demos-europe/demosplan-core/pull/1961/commits/fd87632472427ec7cd1a10a1cf039224852cf413

PR #1968, opened 2023-09-06
bug(refs: T34572): fix missing notification after inserting recommendation

muellerdemos:
https://yaits.demos-deutschland.de/T34572

This PR adds a notification with its translation key, when a recommendation has been pasted in to the editor.


elmasdemos (approved):
lgtm, code review

gruenbergerdemos (approved):
lgtm and works

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -608,6 +608,7 @@ export default {
 
     toggleRecommendationModal () {
       this.$refs.recommendationModal.toggle()
+      dplan.notify.notify('confirm', Translator.trans('recommendation.pasted'))
```
gruenbergerdemos, line 611: I just realized that this is not the optimal place to call the notification, because the toggle method is also called when the modal is opened (so there's a notification when opening).

gruenbergerdemos (changes requested):
Sorry, just saw this one

salisdemos (approved):

gruenbergerdemos (approved):

PR #1972, opened 2023-09-07
chore (refs T34581): Define Entity to ensure field length

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T3458

The data in the session table is no longer stored as a medium blob but as a blob by symfony default. This entity is defined to help doctrine to create the correct field size during database setup.

As a side effect this PR reduces the number of entries that are generated during doctrine migrations diff

### How to review/test
create a new database (be sure to change the database name before!) with `bin/<project> dplan:db:init --create-database` and find a sessions table. 

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, code review only

PR #1976, opened 2023-09-07
chore: Adjust namespace

mussbach:
Project namespace needs to be adjusted to match organization name

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
-    "name": "demos/demosplan",
-    "description": "demosplan",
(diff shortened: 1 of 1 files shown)


graupnerdemos (approved):
I'm kinda wondering if we should go with demosplan or with demosplan-core. For one it makes sense in my head to have repo name and package name synchronous. On the other hand, we still don't have a reasonably good definition of "the core".

PR #1990, opened 2023-09-11
feat (refs T34644): remove subdomains_allowed

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T34644

subdomains_allowed was an attempt not to take over all malicious subdomains. The approach is unfortunately too static and can be replaced by searching for the subdomain in the customer table. If the entry exists, the subdomain can be used.

### PR Checklist

- [x] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1290,16 +1290,16 @@
         },
         {
             "name": "demos-europe/demosplan-addon",
-            "version": "v0.8",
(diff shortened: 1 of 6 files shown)


AlexMenschoid (approved):
lgtm, code review only

PR #1999, opened 2023-09-13
fix(T34655): add tooltip on the Beta-Version text

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34655

**Description:** This PR adds a tooltip to the Beta-Version text. 

- change the text from _Beta Version_ to _Beta-Version_ , as it was recommended in the ticket.


client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -128,11 +128,17 @@
                   icon="ai"
                   size="large"
                   :text="Translator.trans('segment.oracle.tooltip')" />
-                <dp-badge
-                  v-if="activeId === 'oracleRec'"
-                  class="absolute u-right-0 u-mr"
-                  size="smaller"
-                  :text="Translator.trans('segment.oracle.beta')" />
+                <span
```
spiess-demos, line 131: Out of curiosity: why didn't it work to put the tooltip directly onto the DpBadge component?

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -370,7 +377,8 @@ export default {
   },
 
   directives: {
-    cleanhtml: CleanHtml
+    cleanhtml: CleanHtml,
+    tootlip: Tooltip
```
spiess-demos, line 381:
```suggestion
    tooltip: Tooltip
```

spiess-demos (changes requested):

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
sakutademos: I was not sure whether the badge element should include a tooltip inside. I initially considered it as more of a specific case.  If you think that it would be better to put the tooltip inside the badge, I can do it. Now, I believe it's a better idea...

The implementation could be similar to [this one](https://github.com/demos-europe/demosplan-core/blob/22e6a44496bd491cab280d8578bc83b3f33a164d/client/js/components/map/map/DpOlMapEditFeature.vue#L210-L217)

spiess-demos (approved):
Lgtm (code review only).

spiess-demos (commented):
Do we want to upgrade demosplan-ui to v0.1.15 within this PR?

sakutademos:
> Do we want to upgrade demosplan-ui to v0.1.15 within this PR?

yes, it should be done in parallel

spiess-demos (approved):
Lgtm!

PR #2005, opened 2023-09-14
feat(refs T34626): adds addon wrapper to core template

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T34626

This is needed to allow addons to hook into this spot in the procedure settings in all projects. The explanation text is now also in the addons, so I only copied over the addon wrapper itself. I also changed the div class to match the ones around it. Lastly, the import of the AddonWrapper in the DpBasicSettings.vue was outdated and is now fixed.

### How to review/test
I think the addon-permissions are not properly working atm, at least not for me. So just a code review might be easiest.

### DIFF
--- a/client/js/components/procedure/basicSettings/DpBasicSettings.vue
+++ b/client/js/components/procedure/basicSettings/DpBasicSettings.vue
@@ -10,7 +10,6 @@
 <script>
 import {
   dpApi,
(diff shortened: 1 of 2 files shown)


gruenbergerdemos (dismissed):
lgtm

PR #2006, opened 2023-09-14
feature: update addon

jalalifardemos:
Description: Composer update for new method in globalconfig


jalalifardemos:
I have fatal Error. In globalConfig missed one method, that resolve with composer update demosplanAddon

jalalifardemos:
fixed. Thanks

mussbach:
Just for reference: composer install would have done it

PR #2007, opened 2023-09-14
bug(refs T34969): Fix addon hook for maillaine in basic settings

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34969

- Hook name changed
- We don't need permission request here, because it is handled in the addon
- Label and tooltip was moved to the addon

### How to review/test
- Install maillaine Addon
- Go to Verfahren -> Grundeinstellungen. You should see "Berechtigte Absenderadressen für den E-Mail-Import" and the button for adding addresses.

### DIFF
--- a/client/js/components/procedure/basicSettings/DpBasicSettings.vue
+++ b/client/js/components/procedure/basicSettings/DpBasicSettings.vue
@@ -10,7 +10,6 @@
 <script>
 import {
   dpApi,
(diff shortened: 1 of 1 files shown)


muellerdemos (approved):
oh thanks, it seems this was not reverted back by me after we decided not to add the addon wrapper to demosplan-ui

PR #2008, opened 2023-09-14
fix (refs T34642): On create a new tag, check for already existing tag in procedure

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T34642

Tag(titles) should be unique per procedure.

- [x] Tests updated/created

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/TagService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/TagService.php
@@ -86,7 +86,7 @@ class TagService extends CoreService
             throw new InvalidArgumentException('Tag title may not be empty.');
         }
 
-        if ($this->tagRepository->isTagTitleFree($procedureId, $title)) {
+        if (!$this->tagRepository->isTagTitleFree($procedureId, $title)) {
             throw DuplicatedTagTitleException::createFromTitleAndProcedureId($title, $procedureId);
         }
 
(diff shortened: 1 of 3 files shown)


demosplan/DemosPlanCoreBundle/Logic/Statement/TagService.php
```diff
@@ -77,20 +77,16 @@ public function getTag($id)
...
+    public function createTag(string $title, TagTopic $topic, bool $persistAndFlush = true): Tag
     {
         $procedureId = $topic->getProcedure()->getId();
         if ('' === $title) {
             throw new InvalidArgumentException('Tag title may not be empty.');
         }
 
-        $titleCount = $this->tagRepository->count(['id' => $procedureId, 'title' => $title]);
-        if (0 !== $titleCount) {
+        if ($this->tagRepository->isTagTitleFree($procedureId, $title)) {
```
bakridemos, line 89:
```suggestion
        if (!$this->tagRepository->isTagTitleFree($procedureId, $title)) {
```

bakridemos: only tiny thing here, if the tag exists, isTagTitleFree will return false and the exception will not be thrown otherwise seems good, i've tested too and it works good with the tiny fix

bakridemos (changes requested):

dresslerdemos (dismissed):
Review in pair session.

dresslerdemos (approved):

bakridemos (approved):

PR #2017, opened 2023-09-15
fix(refs T34626): rolls back maennchen/zipStream

AlexMenschoid:
**Ticket:** https://yaits.demos-deutschland.de/T34626

The new version 3.0 changed the public surface of the library which would force us to rebuild parts of our zip exports. For the moment we roll the dependency back as we don't have currently the time to do so.

### How to review/test
Without this rollback, procedure exports should not work and instead throw an error. With this rollback, they should work again.

### Linked PRs (optional)
- Update-PR #1582


mussbach (approved):
lgtm, thank you

PR #2019, opened 2023-09-15
update demospla-ui to a specific state - release_ewm

sakutademos:
**Description:** This PR updates demosplan-ui to a specific state and enables testing bug fixes in this project without including significant changes from the demosplan-ui main branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "lint:scss": "stylelint {demosplan/DemosPlanCoreBundle/Resources/client/scss,projects}/**/*.scss"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


spiess-demos:
I doubt that this is installable in a CI environment - when publishing to npm, yarn prepack is run which builds the needed scss and js files. If requiring the package via git hash, the needed files are not there and the frontend build of demosplan-core will simply fail.

i guess we'll have to release a `v0.1.15` (from your feature branch? i dunno) to get this running.

mussbach:
I second this. I think we will need a distinct tag

PR #2025, opened 2023-09-18
bug(refs T30049): Add import center for project

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T30049

- Only show tabs for import center if there is content for the tabs

### How to review/test
Disable all import functions (pdf, excel, direkteingabe, email). There should be no tabs in the import center and no error in the console.

### DIFF
--- a/client/js/components/procedure/AdministrationImport/AdministrationImport.vue
+++ b/client/js/components/procedure/AdministrationImport/AdministrationImport.vue
@@ -8,33 +8,35 @@
 </license>
 
 <template>
-  <dp-tabs
-    v-if="allComponentsLoaded"
-    :active-id="activeTabId"
-    use-url-fragment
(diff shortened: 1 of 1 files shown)


sakutademos (approved):
lgtm

hwiem (commented):
While this works, I think the ideal solution would be to not display the tabs if there is only one item, but instead show a normal heading. If there's time to implement it, I'd suggest going with that solution, otherwise this one is also fine.

gruenbergerdemos:
> While this works, I think the ideal solution would be to not display the tabs if there is only one item, but instead show a normal heading. If there's time to implement it, I'd suggest going with that solution, otherwise this one is also fine.

It is a little more complicated than I thought, so I'm putting it off for the moment.

PR #2027, opened 2023-09-18
bug(refs T34708): Fix moving list items

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34708

We need to set the value as `newOrder` here to be able to set the children in `client/js/store/map/Layers.js:142`.

### How to review/test
Go to Verfahren -> Planungsdokumente und Planzeichnung -> Kartenebenen definieren and create some folders and layers. You should be able to move the layer and folders and nest them.


ahmad-demos (approved):
lgtm, only codeReview!

elmasdemos (approved):
lgtm, code review

PR #2028, opened 2023-09-18
fix: removes gdpr email addresses from orphan email removal

AlexMenschoid:
The table does not exist anymore and having it still in the delete query here produces an error that prevents the orphan removal to correctly work.

### How to review/test
Without this fix, the orphan email removal in the MaintenanceTask should fail with an error.


mussbach (approved):
lgtm, code review only

PR #2030, opened 2023-09-18
feat: logout from keycloak during local logout

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T34685

Symfony security system is used to log out user instead of building custom logout. Code is mostly only moved to listener, the part with `oauth_keycloak_logout_route` is new. A bug is fixed that caused the part of `saml_idp_slo_url` to always be true.

### How to review/test
code review might be best, normal login should still work

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


elmasdemos (approved):
lgtm, code review only

PR #2034, opened 2023-09-20
bug(refs T34738): Fix user filter sets

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34738

Since refactoring DpMultiselect option is now a property of props. 

### How to review/test
Verfahren -> Abwägungstabelle -> Filter. You should see the list with user specific filters (first drop down).


muellerdemos (approved):
looks legit

PR #2040, opened 2023-09-22
chore: set user and group for generated cache and log files to current user

mussbach:
As phpstorm runs tests via docker (compose) as root user file permissions are screwed up that leads to error when visiting the browser after performing a project test

### How to review/test
core review or visit the browser after performing a project test

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "ext-mbstring": "*",
         "ext-openssl": "*",
         "ext-pdo": "*",
+        "ext-posix": "*",
         "ext-simplexml": "*",
         "ext-soap": "*",
         "ext-sockets": "*",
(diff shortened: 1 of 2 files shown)


AlexMenschoid:
I honestly don't fully understand the scope of the problem. But my intuition does not like that we're adapting our testing setup to the tool instead of trying to change the tool (in this case PHPStorm) to work correctly as we want it. But that might not be possible, I'm not sure, I just would prefer that.

Otherwise I would not consider the SonarCloud issue in this case relevant as it is imo a clear case of a runtime exception and making it more precise is not really adding much.

mussbach:
@AlexMenschoid  I am basically with you, but as this is a [feature request](https://youtrack.jetbrains.com/issue/PY-46794/Add-option-to-choose-the-user-with-which-to-run-in-a-docker-compose-environment.) that was opened 6 Years ago I do not think that this will happen soon in Phpstorm

AlexMenschoid (approved):
lgtm after seeing the 6 year old feature request XD

PR #2041, opened 2023-09-22
Fix(T34759): adjust text for an empty statement list

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34759

**Description:** This PR adjusts the text for an empty statement list.


spiess-demos (changes requested):
`statement.list.empty` is mistakenly used there. It is intentionally used in DpPublicStatementList in demosplan-core. Instead of changing the value here (which would lead to it being the wrong value in the other place) i'd suggest to craft a new key `segments.recommendations.oracle.none` which would also play nice with the other keys around this feature.

spiess-demos (approved):
lgtm!

PR #2045, opened 2023-09-22
Fix(T34715): add missing hints to the register form

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34715

**Description:** Add missing hints to the register form and adjust layout.


client/js/components/user/orgaRegisterForm/OrgaRegisterForm.vue
```diff
@@ -49,37 +49,52 @@
...
-                name="r_orgatype[]"
-                value-to-send="OPAUTH" />
+            <div :class="hasPermission('feature_identity_broker_login') ? prefixClass('space-stack-xs') : prefixClass('o-form__group flex-col')">
+              <div :class="prefixClass('u-mb-0_75')">
+                <dp-checkbox
+                  id="orgatype_invitable_institution"
+                  data-cy="orgatype_institution"
+                  :class="prefixClass('o-form__group-item')"
+                  :label="{
+                    text: Translator.trans('invitable_institution')
```
spiess-demos, line 62: Does it also work when placing the hints directly within the labels? this way they'd be clickable as well.

```suggestion
                    text: Translator.trans('invitable_institution'),
                    hint: Translator.trans('register.institution.hint')
```

sakutademos: thx for the hint. It works fine. done here https://github.com/demos-europe/demosplan-core/pull/2045/commits/1b12e7682333daa0429f0039368fe118123c6e39

client/js/components/user/orgaRegisterForm/OrgaRegisterForm.vue
```diff
@@ -52,31 +52,37 @@
             <legend :class="prefixClass('font-size-medium is-label u-mb-0_25')">
               {{ Translator.trans('organisation.type') }}
             </legend>
-            <div :class="hasPermission('feature_identity_broker_login') ? prefixClass('space-stack-xs') : prefixClass('o-form__group')">
+            <p :class="prefixClass('u-mb')">
+              {{ Translator.trans('organisation.kind.explanation') }}
+            </p>
+            <div :class="hasPermission('feature_identity_broker_login') ? prefixClass('space-stack-xs') : prefixClass('o-form__group flex-col')">
```
spiess-demos, line 58: I think when removing all classes from the checkbox elements, you could use `space-stack-s` here.

```suggestion
            <div :class="prefixClass('space-stack-s')">
```

client/js/components/user/orgaRegisterForm/OrgaRegisterForm.vue
```diff
@@ -52,31 +52,37 @@
...
+                  text: Translator.trans('invitable_institution'),
+                  hint: Translator.trans('register.institution.hint')
                 }"
                 name="r_orgatype[]"
                 value-to-send="OPSORG" />
               <dp-checkbox
                 id="orgatype_municipality"
                 data-cy="orgatype_municipality"
-                :class="prefixClass('o-form__group-item')"
+                :class="prefixClass('o-form__group-item u-mb-0_75 u-ml-0')"
```
spiess-demos, line 72: Why do we need to set `u-ml-0` here? The `o-form__group-item` class is (somewhat beta-style) used for horizontally layed out items, it can be removed here i guess, alongside the margin.

client/js/components/user/orgaRegisterForm/OrgaRegisterForm.vue
sakutademos: hmm, it doesn't have any influence when I use the class `space-stack-s`. But you are right, if I remove those classes, it doesn't break the layout

client/js/components/user/orgaRegisterForm/OrgaRegisterForm.vue
sakutademos: thx, done here https://github.com/demos-europe/demosplan-core/pull/2045/commits/7b10dcc10ab897b309f3438f1a7a055f3eb65660

client/js/components/user/orgaRegisterForm/OrgaRegisterForm.vue
spiess-demos: `space-stack-s` is overruled by the `u-mb-0_75` on the child elements.

sakutademos: hhm, no.. it doesn't have any impact even when I remove the margin. It seems like the class is not being recognized

sakutademos: ok, got it. I've removed margins from the checkboxes and added the `space-stack-s` class instead of all these classes

spiess-demos (approved):
lgtm!

PR #2048, opened 2023-09-25
chore (refs T32796): fix faulty unique index on procedure_message.procedure_id

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T32796

procedure_id was modeled as a unique foreign key which does not make sense as we need to be able to save multiple messages to one procedure

### How to review/test
code review or modify procedure twice

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
code review und tested too

PR #2052, opened 2023-09-26
Fix(T34713): move missing translations from blp to the core

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34713

**Description:** Move missing translations from BLP to the core to make them accessible from other projects.


spiess-demos (changes requested):

spiess-demos (approved):
lgtm!

PR #2059, opened 2023-09-26
chore: lock ol to same version that masterportal uses

spiess-demos:
To not include two versions of Openlayers within our build, we have to lock it to the version used by masterportalapi.

The [Openlayers Upgrade notes](https://github.com/openlayers/openlayers/blob/v7.5.0/changelog/upgrade-notes.md) imply that we can roll back without losing important stuff... the other question is, do we really want to lock to the masterportalapi here?

<details>
   <summary>Ergebnisse des Prod Builds mit --analyze</summary>
Before:

image_480

After:

image_480
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "leaflet": "^1.9.4",
     "leaflet.markercluster": "^1.5.3",
     "lscache": "^1.3.2",
-    "ol": "~7.5.1",
+    "ol": "7.3.0",
     "plyr": "^3.7.8",
     "portal-vue": "^2.1.7",
     "proj4": "^2.9.0",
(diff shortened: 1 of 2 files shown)


hwiem (commented):
If we update the masterportalapi to the newest version (2.27.0), we can set ol to 7.5.2.
I'm not entirely sure what this means though. Will we have to change the ol version whenever it changes in the masterportalapi? How do we make sure we don't forget about it?

salisdemos (approved):
C>

spiess-demos:
> If we update the masterportalapi to the newest version (2.27.0), we can set ol to 7.5.2. 

I'd say it would need a closer look to update masterportalapi, [according to its changelog](https://bitbucket.org/geowerkstatt-hamburg/masterportalapi/src/master/CHANGELOG.md). However i'd suggest to 

> I'm not entirely sure what this means though. Will we have to change the ol version whenever it changes in the masterportalapi? How do we make sure we don't forget about it?

You're right, it would be better to resolve the problem by updating masterportalapi.

PR #2077, opened 2023-09-29
F adjust methods parameters und returntypes

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T43766

Description: 
- add new method 'addSimilarStatementSubmitters' : it to add a new section which contain the  Similar Statement Submitters section
- adjust methods return types and remove uneeded parameters


dresslerdemos (approved):
Code review only

PR #2079, opened 2023-09-29
Fix(T34841): issue with displaying the icons

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34841

**Description:** This PR fixes an issue with two icons in the modal window. They were not being displayed because the `activeId` was only updated after clicking on the tab. Therefore, set the `activeId` value when clicking on the `toggleRecommendationModal` button.

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
@@ -737,9 +737,6 @@ export default {
     },
 
     toggleRecommendationModal () {
-      if (this.asyncComponents[0]) {
-        this.activeId = this.asyncComponents[0].options.id
-      }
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -737,6 +737,9 @@ export default {
     },
 
     toggleRecommendationModal () {
+      if (this.asyncComponents[0]) {
+        this.activeId = this.asyncComponents[0].options.id
+      }
```
salisdemos, line 742: I suggest setting the ID in the then from `loadAddonComponents` . Or is it intended to be reset every time the modal gets reopended

salisdemos (approved):

PR #2106, opened 2023-10-04
F validate draft json

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T34807

Description: 

In the case where the segments draft jsons are not correct (Ranges divided one more time or are splitted in two parts), and despite the fact that they are already checked in frontend, they are saved in the database which causes an error afterwards. to avoid this issue, the draft jsons have to be valided in backend too:

- add new 'SegmentDraftJsonConstraintValidator' and 'SegmentDraftJsonConstraint'
- add property constraint in 'Statement' entity
- adjust the update method in the 'StatementResourceTypeService'
- add new translation key 'error.segmentation.invalid_draft_json'
- add tests:  in the 'StatementResourceTypeServiceTest' there are two version of the the method 'testUpdate' to test if the statements draft jsons are valid or not (Ranges can not be split in two parts). The first method use the 'StatementResourceTypeService' as system under test, which doesn't return the desired result due to other technical problems ( it seems that the validator injections is not working while testing allthough it works normally ). The second method create and use a new Validator Builder and return the desired result. Otherwise it's been tested manually.

- [X] Tests updated/created


MoritzMandler (approved):
lgtm

PR #2119, opened 2023-10-06
test coverage for supportContact of Custsomer

martendemos:
Testcoverage for create, update and delete in context of not yet existing SupportContact Entity

**Ticket:** https://yaits.demos-deutschland.de/T34940

### How to review/test
This test are written without an existing entity, therefore there are will failing until merged with the corresponding entity and CURD logic.

So code review is the only thing to do.
- [x] Tests updated/created


martendemos:
Within this PR there is code which shouldn't be in the target branch.
This PR will be closed and a new one will be opened: https://github.com/demos-europe/demosplan-core/pull/2130

PR #2123, opened 2023-10-09
chore: add option to enable addon during installation

mussbach:
Addons may be enabled during installation by using the option `--enable`.
Along the way some improvements in the installation process where introduced, you may review commit wise.

I voted against tests in this specific case as the commands main task is to cope with infrastructure and configuration.

### How to review/test
code review or install an addon and enable it directly

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


jalalifardemos (approved):
Installation is successfully. Thanks

bakridemos (approved):

PR #2132, opened 2023-10-11
chore (refs T34979): add missing props, when editor used dpUploadModal.

ahmad-demos:
https://yaits.demos-deutschland.de/T34979
https://github.com/demos-europe/demosplan-ui/pull/567
https://github.com/demos-europe/demosplan-ui/pull/567/commits/2fc10b706e8b427d0a2e8ff420e209799380f9b7

- add missing props, when editor used dpUploadModal.


gruenbergerdemos (changes requested):
Since you're already here, could you sort the props alphabetically? Except `id`, `ref` and `v-model`, these should stay on top in this order: `id`, `ref`, `v-model` (for correct order also see https://eslint.vuejs.org/rules/attributes-order.html).

spiess-demos:
From the PR description, i do not understand why these changes are needed.

ahmad-demos:
> From the PR description, i do not understand why these changes are needed.

because when Editor used `imageButton`, then we need `DpUploadModal & DpUploadFiles`, like `getFileByHash`
https://github.com/demos-europe/demosplan-ui/pull/567/commits/2fc10b706e8b427d0a2e8ff420e209799380f9b7

salisdemos (approved):

gruenbergerdemos (approved):

PR #2142, opened 2023-10-12
fix(refs: T34961): Pass ID as Tooltip container param

salisdemos:
Pass Id to Tooltipcontainer parameter instead of DOM-Object.
The Tooltip-Directive only can handle querySelectables

**Ticket:** https://yaits.demos-deutschland.de/T34961


elmasdemos (approved):
lgtm, code review

PR #2144, opened 2023-10-13
feature (refs T15052): remove faulty ORM-Annotations

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T15052

Description: 
remove faulty ORM-Annotations
BLOB and TEXT columns cannot have DEFAULT values.
https://demosdeutschland.slack.com/archives/CEDGU47DX/p1697120716335369

All properties are set with an empty string initially - the default value is set this way.
`private string $property = '';` 

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-core/pull/2145

- [ ] Tests updated/created
- [ ] Update documentation
- [x] Link all relevant tickets
- [ ] Move the tickets on the board accordingly

PR #2145, opened 2023-10-13
F t15052 tidy up docreine migrations diff

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T15052

Description:
set default value for DATETIME properties where the the default is already
set in the dataBase and after refactorings - making use of the @gedmo\Timestampable(on="create"),
the default got removed from the ORM-Annotations. Added the default back as they do no harm and there is no need to change the schema.

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-core/pull/2144

- [ ] Tests updated/created
- [ ] Update documentation


AlexMenschoid (approved):
lgtm, code review only

PR #2163, opened 2023-10-18
fix(refs T35059): Catch ununique Titles before sending them to the BE

salisdemos:
They have to be unique by design. however we can't validate them with a propper error message
 in the Backend, so we check before saving. As a safety net the Be will throw a generic error
 if this check here is insufficient.

**Ticket:** https://yaits.demos-deutschland.de/T35059

As mandatenAdmin Goto `/einstellungen/plattform` and try to create two customerContacts with the same title.  


elmasdemos (approved):
lgtm, it works

PR #2166, opened 2023-10-18
chore: add data user role to fixtures

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T30599

The prod fixtures need to provide a role for the data user so that a user could be created. Moreover, the prod fixtures are not deprecated, this is cleaned up as well.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (commented):
I think the deprecated mark was set to raise awareness for a refactoring with foundry factories.
see here: https://yaits.demos-deutschland.de/T34038
Should a separate ticket be created for the desired replacement by Foundry?

mussbach:
The prod fixtures will not be replaced by foundry as we need distinct values for the database baseline

breierDemos (approved):

PR #2174, opened 2023-10-20
chore: Group Babel dependencies for Dependabot

salisdemos:


.github/dependabot.yml
```diff
@@ -39,6 +39,13 @@ updates:
       - dependency-name: chalk
         versions:
           - ">= 5"
+    groups:
+        babel:
+            patterns:
+                - "@babel/*"
+            ingnore-patterns:
+                - "@babel/plugin-*"
+                - "@babel/eslint-*"
```
spiess-demos, line 48: Why do we want to exclude those? And, should we eventually add the reasoning as a comment here?

salisdemos: because they follow other update-cycles which means they don't have the same latest version then the others.
686e5d49ea30c9109f16f7f3da5308a63eed2f6d

spiess-demos (dismissed):
lgtm (code review)

PR #2175, opened 2023-10-20
chore(refs T35059): Optimize Translations (keys and String)

salisdemos:
Find a better Wording for the errormessage, where a name has to be unique.

And even though the transkey is in use only in the customer area, The key itself should be generic enough to be used in other contexts.

**Ticket:** https://yaits.demos-deutschland.de/T35059


translations/messages+intl-icu.de.yml
```diff
@@ -680,10 +680,24 @@ consultation.notification.mail.on.token.creation.confirmation: "Ein Zugangscode
...
 consultation.notification.mail.subject.on.token.creation: "Ihre Stellungnahme zum Verfahren {procedureName} ist eingegangen"
 
+contact: "Kontakt"
 contact.data: "Kontaktdaten"
+contact.add: "Kontakt hinzufügen"
+contact.change: "Kontakt bearbeiten"
+contact.delete: "Kontakt löschen"
+contact.deleted: "Der Kontakt wurde gelöscht."
+contact.new: "Neuen Kontakt anlegen"
+contact.no: "Keine Kontakte angelegt"
```
muellerdemos, line 690: Could we tweak this translation key a little bit. Imho I think the key `contact.no` is not really transporting the translation message. Maybe something like `contact.no_entries` or `contact.no_new_entries`?

muellerdemos (changes requested):
just a tiny remark

translations/messages+intl-icu.de.yml
salisdemos: good catch b9dbaffc0a16edd4bd35e58abdaac2bac8a3f44b

muellerdemos (approved):
thanks - lgtm

PR #2191, opened 2023-10-24
bug(refs T34955): Add required prop

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34955

ProcedureId is now a required prop.


hwiem (approved):
lgtm, code review only

PR #2210, opened 2023-10-25
Feat(refs:35155): add the new greeting formel

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35155

Description: 

- replace the old greeting form with the new one

### PR Checklist

- [x] Tests updated/created
- [x] Update documentation

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/notify_consultation_token_creation.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/notify_consultation_token_creation.html.twig
@@ -17,7 +17,7 @@ Hierfür benötigen Sie einen Zugangscode für die Teilnahme. Ihr Zugangscode la
 Dies ist eine automatisch erzeugte Benachrichtigung. Bitte antworten Sie nicht darauf! Für Rückfragen kontaktieren Sie
 bitte die verfahrensführende Behörde.
 
-Mit freundlichen Grüßen
+Mit freundlichem Gruß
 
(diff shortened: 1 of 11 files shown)


hwiem (commented):
Should this be changed in all of these templates?

muellerdemos (changes requested):
These changes shall be committed to the core templates? The tickets reads it is only for diplanrog, or am I mistaken here?

elmasdemos:
> Should this be changed in all of these templates?

yes, but i can do it another PR, but than it doesnt have a ticket. it says in the main ticket, that this should have been changed in core

elmasdemos:
> These changes shall be committed to the core templates? The ticket https://yaits.demos-deutschland.de/T33939 reads it is only for diplanrog, or am I mistaken here?

when you look into the main ticket, it says it shall be changed in the core. I wrote with @po but if desired I cfan do another PR without any reference or ask for a new ticket

elmasdemos:
anyway, my idea was to make the work later easier, but i reverted the changes and just changed only the one place

hwiem:
> anyway, my idea was to make the work later easier, but i reverted the changes and just changed only the one place

I think it was fine to include the other templates, I just didn't find the info that they need to be changed, hence my question

muellerdemos (approved):
I did not meant to remove the changes if they are needed, I just saw in the ticket the checkbox for a single project and wanted to be sure this is correct. However this lgtm - sorry for the hassle

hwiem (approved):

PR #2221, opened 2023-10-27
refactor: Declare fields when fetching orgas and use the proper rels

salisdemos:
- we don't want to access the customer directly from the orga. Its deprecated
- and we should define explicitly which fields we want to get from the Backend
- even if that means listing all fields.

As Mandaten-Admin / Support The Page `/organisation/list` should work like before but without the purple warnings

### DIFF
--- a/client/js/bundles/user/listOrgas.js
+++ b/client/js/bundles/user/listOrgas.js
@@ -38,7 +38,7 @@ const presetStoreModules = {
           }
         }
       },
-      include: ['customer.branding', 'currentSlug', 'customers', 'statusInCustomers'].join(),
+      include: ['customer.branding', 'currentSlug', 'statusInCustomers'].join(),
       group: 'pending'
     }
   }]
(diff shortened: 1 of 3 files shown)


client/js/bundles/user/listOrgas.js
```diff
@@ -38,7 +38,7 @@ const presetStoreModules = {
           }
         }
       },
-      include: ['branding', 'currentSlug', 'customers', 'statusInCustomers'].join(),
+      include: ['customer.branding', 'currentSlug', 'customers', 'statusInCustomers'].join(),
```
salisdemos, line 41:
```suggestion
      include: ['customer.branding', 'currentSlug', 'statusInCustomers'].join(),
```

martendemos (approved):
BE: lgtm

dresslerdemos (approved):
backend review only

elmasdemos (approved):
code review, lgtm and it works. thanks

PR #2222, opened 2023-10-27
fix(refs T35173): Skip own Title when checking for Uniqueness

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35173

as mandanten-admin (support) goto `/einstellungen/plattform` and try to edit a support-contact. The error, that You use an already used title should not appear and saving should be possible

### PR Checklist


muellerdemos (approved):
lgtm

PR #2212, opened 2023-10-25
refactor(refs T34504): Use json api instead of form

salisdemos:
***Test: https://yaits.demos-deutschland.de/T35204***

### Description: 

Use the API to update parts of the form one by one, so we don't have to update all at once which improves the ux and makes the data handling easier

### How to review/test
- login as mandant admin and go to `Plattform-Einstellungen`, do some changes and save them

PR #2239, opened 2023-10-31
feat (refs T35204): Adjust input validation 

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T35204
Only checking for min string length / valid URL only in case of incoming user input to avoid errors on already existing empty strings.

todo: draft because not tested locally

### How to review/test
change and save something in /einstellungen/plattform as role.cmu

### Linked PRs (optional)
- #2212


dresslerdemos (approved):
Code review only

PR #2254, opened 2023-11-06
chore refs T35309: add csrf protection listener

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T35309

All HTML post forms must be provided with csrf protection. 
protection. On the one hand this is best practice, on the other hand it is required by the BSI and by customers.

After this PR all forms must include `{{ include('@DemosPlanCore/DemosPlanCore/includes/csrf.html.twig') }}`. Initially only a dev message is triggered, later on the check could be activated

### How to review/test
run tests, post any form and see the dev message. Adding or editing a procedure news should not trigger the message

### PR Checklist

- [x] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/EventSubscriber/CsrfSubscriber.php
+++ b/demosplan/DemosPlanCoreBundle/EventSubscriber/CsrfSubscriber.php
@@ -1,0 +1,64 @@
+<?php
+
+/**
+ * This file is part of the package demosplan.
+ *
+ * (c) 2010-present DEMOS plan GmbH, for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+namespace demosplan\DemosPlanCoreBundle\EventSubscriber;
+
+use DemosEurope\DemosplanAddon\Contracts\MessageBagInterface;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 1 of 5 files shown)


demosplan/DemosPlanCoreBundle/EventSubscriber/CsrfSubscriber.php
```diff
@@ -0,0 +1,61 @@
...
+            return;
+        }
+
+        $token = $this->csrfTokenManager->getToken($tokenId);
+        if ($token instanceof CsrfToken && $this->csrfTokenManager->isTokenValid($token)) {
+            // all clear, token is set and valid
+            return;
+        }
+
+        $this->logger->info('CSRF token invalid', ['uri' => $request->getRequestUri(), 'token' => $tokenId]);
```
breierDemos, line 52: Is $token or $tokenId wanted here?

breierDemos (changes requested):

demosplan/DemosPlanCoreBundle/EventSubscriber/CsrfSubscriber.php
mussbach: $tokenId because it might be that token could not be Found and I want to know what tokenId was not found

breierDemos (approved):

PR #2257, opened 2023-11-06
B t35312 adjust styling for support card

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T35312

**Description:** This PR adjusts styling for the support card as well as fixes the issue where all cards were displayed instead of only the visible ones.

### DIFF
--- a/client/js/components/support/DpSupport.vue
+++ b/client/js/components/support/DpSupport.vue
@@ -77,7 +77,7 @@ export default {
     }),
 
     visibleContacts () {
-      return Object.values(this.contacts).filter(el => el.attributes.visible)
+      return Object.values(this.contacts).filter(contact => contact.attributes.visible)
     }
   },
 
--- a/client/js/components/support/DpSupport.vue
+++ b/client/js/components/support/DpSupport.vue
@@ -9,7 +9,7 @@ All rights reserved
 
 <template>
   <div
-    class="c-support space-inset-m color--black bg-color--blue-light-3">
+    class="space-inset-m bg-color--blue-light-3">
     <h2 class="font-normal color--black">
       {{ Translator.trans('support.heading') }}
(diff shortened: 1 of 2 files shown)


client/js/components/support/DpSupport.vue
```diff
@@ -24,12 +24,12 @@ All rights reserved
     </h3>
     <ul
       class="u-mb-0_75"
-      :class="contactCount === 1 ? '' : 'grid lg:grid-cols-3 gap-3'">
+      :class="{ 'grid lg:grid-cols-3 gap-3': visibleContacts.length !== 1 }">
```
gruenbergerdemos, line 27: How does this work? The classes are only added if the condition is true?

client/js/components/support/DpSupportCard.vue
```diff
@@ -26,8 +26,7 @@ All rights reserved
       v-if="email"
       v-text="email" />
     <div
-      v-if="reachability.service"
-      class="u-pb-0_75">
+      v-if="reachability.service">
```
gruenbergerdemos, line 29: v-if should be ordered first

gruenbergerdemos (changes requested):

client/js/components/support/DpSupport.vue
sakutademos: yes, right

client/js/components/support/DpSupportCard.vue
sakutademos: the class attribute is not needed any more, it was removed

gruenbergerdemos: Oh, right, my mistake

gruenbergerdemos (approved):
lgtm

PR #2268, opened 2023-11-07
Fix(T33788): disabled scrolling after closing the modal window

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33788

**Description:** This PR addresses the issue where scrolling becomes impossible after closing the modal window. Manually invoke the preventScroll function within the handleClose function.

### DIFF
--- a/client/js/components/statement/assessmentTable/AssignEntityModal.vue
+++ b/client/js/components/statement/assessmentTable/AssignEntityModal.vue
@@ -140,6 +140,7 @@ export default {
 
     handleClose (isOpen) {
       if (!isOpen) {
+        this.$refs.assignModal.preventScroll(false)
         this.setModalProperty({ prop: 'assignEntityModal', val: { ...this.assignEntityModal, show: false } })
       }
     },


client/js/components/statement/assessmentTable/AssignEntityModal.vue
```diff
@@ -140,6 +140,7 @@ export default {
 
     handleClose (isOpen) {
       if (!isOpen) {
+        this.$refs.assignModal.preventScroll(false)
```
spiess-demos, line 143: While this will surely fix the bug, it uses an internal API of DpModal, thereby making it harder and more error prone to refactor DpModal later. Also, the same fix would have to be applied to every place `this.setModalProperty` is used to set the visibility of the modal to `false`. The modal should just properly do its job by itself: removing the scroll-related classes onClose. 

The underlying problem i see here is that DpModal is toggled programatically by the parent component (which is a specialized modal as AssignEntityModal, ConsolidateModal or CopyStatementModal) by accessing the `toggle()` method of DpModal, which in this context leads to the transition not to be calling `preventScroll`. However i do not see the root cause of this behavior.

spiess-demos (changes requested):

client/js/components/statement/assessmentTable/AssignEntityModal.vue
sakutademos: It should, but the `@after-leave` transition does not work in this case and could not find the root...

Is it ok to put the `preventScroll` function into the `toggle` function of the `DpModal`? This way, it will depend on the `isOpenModal` value and not only on the `@after-leave` transition?

spiess-demos: > It should, but the `@after-leave` transition does not work in this case and could not find the root...

what do you men by "cannot find the root"?

> Is it ok to put the `preventScroll` function into the `toggle` function of the `DpModal`? This way, it will depend on the `isOpenModal` value and not only on the `@after-leave` transition?

it would be ok, i guess.

sakutademos: sorry, I mean _'the reason'_

PR #2270, opened 2023-11-07
fix(refs T35139): pass slots to dp-multiselect and use them

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35139

The used slot from vue-multiselect wasn't supported by DpMultiselect.
(https://github.com/demos-europe/demosplan-ui/pull/629)
That was fixed with a change in demosplan-ui that needs now the changes mad here.

### How to review/test

As FPA selecting "Angaben übernehmen (Privatperson)" should display the data without any HTML.
(page: /statement/new/manual/<procedure-id>)

:warning: Needs a demosplan-ui update

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


sakutademos (approved):
there are no more HTML tags in the multiselect component; instead, there is a text: _Daten eingefügt!_
 Is that the intended behavior? You are saying in the description that there should be **data** without any HTML.
image

salisdemos:
> there are no more HTML tags in the multiselect component; instead, there is a text: _Daten eingefügt!_ Is that the intended behavior? You are saying in the description that there should be **data** without any HTML. image

Sorry for the misleading description. Yes this is intended. It states out, that the selected contact is filled into the form below.

PR #2274, opened 2023-11-08
bug(refs T35342): Use the correct property for submitter city

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35342

The property for submitter's city was wrong. We get `ort` from the backend.

### DIFF
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -139,7 +139,7 @@
                     :size="5" />
                   <dp-input
                     id="r_orga_city"
-                    v-model="values.submitter.city"
+                    v-model="values.submitter.ort"
                     class="o-form__group-item"
                     name="r_orga_city"
                     :label="{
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -139,7 +139,7 @@
                     :size="5" />
                   <dp-input
                     id="r_orga_city"
-                    v-model="values.submitter.city"
+                    v-model="values.submitter.ort"
```
muellerdemos, line 142: Is there a possibility that `data` can adjust this? I am not a huge fan of mixing english and german variable names in the code. Also I don't understand why this was changed to `ort`. 

Besides from that code lgtm

muellerdemos (commented):
just wondering, if data can not adjust this this lgtm

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
gruenbergerdemos: Yes, I agree. I asked data if they can do it fast, otherwise I would change it here, because the sprint is already over.

muellerdemos (approved):
In case data can not adjust this fast, we can use this as a fix

PR #2276, opened 2023-11-08
fix: Type-o in dependabot.yml

salisdemos:
its tiny

for interested Persons https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file


spiess-demos (approved):
good catch!

PR #2288, opened 2023-11-09
fix(refs T35369) Display contact cards for none admins

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35369

If You don't have the Permission to edit customer contacts,
the api doesn't return `visible` as attribute since you only
receive visible contacts. That means it doesn't work to filter
by that state and it's not only unnecessary but results in empty lists.

### How to review/test

as "Mandanten admin" create some entities in "inhaltlicher support" (`/einstellungen/plattform`). set some to visible and other to not-visible.
Log out. check if You only see the visible ones on the page `/informationen`

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/support/DpSupport.vue
+++ b/client/js/components/support/DpSupport.vue
@@ -8,8 +8,7 @@ All rights reserved
 </license>
 
 <template>
-  <div
-    class="space-inset-m bg-color--blue-light-3">
+  <div class="space-inset-m bg-color--blue-light-3">
     <h2 class="font-normal color--black">
       {{ Translator.trans('support.heading') }}
     </h2>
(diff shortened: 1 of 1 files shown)


gruenbergerdemos:
Is `main` the correct target branch?

salisdemos:
> Is `main` the correct target branch?

absolutely not. Thanks

gruenbergerdemos (commented):
But now the Mandanten Admin sees all contacts, visible or not, on the "Über.." page. Is this intended? I think it could be confusing, if the admin is checking how it looks after setting it up and seeing also contacts, that should be invisible.

salisdemos:
> But now the Mandanten Admin sees all contacts, visible or not, on the "Über.." page. Is this intended? I think it could be confusing, if the admin is checking how it looks after setting it up and seeing also contacts, that should be invisible.

It's neither nor. It's a trade of. Because otherwise the logic there would be complicated. 
Do You think we should add it anyway?

gruenbergerdemos (approved):
Maybe we could add a little hint for each contact "Nur für die Rolle Mandanten-Administration sichtbar" or something like that. But since it is so late in the sprint, it is also ok to leave it like that, I think.

PR #2290, opened 2023-11-09
bug: Only show office hours if there are office hours

gruenbergerdemos:
Show office hours only if there are office hours to avoid console errors. Appeared if you create a contact without office hours (the big text field). 

check


hwiem (approved):
lgtm, code review only

PR #2291, opened 2023-11-09
feat (refs T35314): fix invalid form on update/create ProcedureType

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T35314

I wondering why this was not happening earlier but on create or update a ProcedureType additional fields will be added which invalidates the form. This fix keep the logic but avoid creating a additional field.

Additional cleanup will be done in a separate PR.

### How to review/test
Update and Create a ProcedureType.

- [ ] Tests updated/created

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
@@ -282,8 +282,7 @@ class ProcedureTypeService extends CoreService implements ProcedureTypeServiceIn
      * Uses all names of each path to retrieve nested values from the given data. The last name of
      * each path will be used as key.
      *
-     * @param array<string,mixed>   $data
-     * @param PropertyPathInterface ...$keyValuePairs
+     * @param array<string,mixed> $data
      *
      * @return array<string,mixed>
      */
--- a/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
@@ -334,8 +334,9 @@ class ProcedureTypeService extends CoreService implements ProcedureTypeServiceIn
     }
 
     /**
-     * Adds the fieldDefinitions and participationGuestOnly (it is never send in request since it can't be manually changed) of the original procedure type to the request. Also adds the original ID for easier
-     * error handling in case of a redirect to the form.
+     * Adds the fieldDefinitions and participationGuestOnly (it is never send in request since it can't be
+     * manually changed) of the original procedure type to the request. Also adds the original ID for easier error
+     * handling in case of a redirect to the form.
      */
     public function addMissingRequestData(
         string $formName,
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
```diff
@@ -363,16 +374,10 @@ public function addMissingRequestData(
 
                 return $return;
             })->toArray();
-            $params[$formName]['statementFormDefinition']['fieldDefinitions'] = $fieldDefinitions;
+            $statementFormDefinition['fieldDefinitions'] = $fieldDefinitions;
+            $request->request->set('statementFormDefinition', $statementFormDefinition);
         }
 
-        // Adds id to the form if missing
```
mussbach, line 369: Do i understand it correctly that this triggers the error? Otherwise the id is not set anymore. This was also handled in https://yaits.demos-deutschland.de/T21643 and initially in https://yaits.demos-deutschland.de/T19137

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
```diff
@@ -363,16 +374,10 @@ public function addMissingRequestData(
...
-        if (!isset($params[$formName]['id'])) {
-            $params[$formName]['id'] = $params['id'];
-        }
-
-        $request->request->add([$formName => $params[$formName]]);
```
mussbach, line 374: Is it correct that you completely delete this general add?

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
```diff
@@ -334,8 +333,9 @@ public function calculateStatementFieldDefinitionChanges(
...
 
     /**
-     * Adds the fieldDefinitions and participationGuestOnly (it is never send in request since it can't be manually changed) of the original procedure type to the request. Also adds the original ID for easier
-     * error handling in case of a redirect to the form.
+     * Adds the fieldDefinitions and participationGuestOnly (it is never send in request since it can't be
+     * manually changed) of the original procedure type to the request. Also adds the original ID for easier error
+     * handling in case of a redirect to the form.
      */
     public function addMissingRequestData(
         string $formName,
```
mussbach, line 341: $formName is not used anymore. When this is correct you should delete it

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
martendemos: yes, this is intended, because we are adding/setting the needed values directly to the request instead of creating and filling '$params'

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
martendemos: To keep this PR small I've shift the cleanup to #2292

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureTypeService.php
martendemos: No, the error was triggered by the removed line 374:
`$request->request->add([$formName => $params[$formName]]);`

mussbach: But shouldn't then the Id still be set?

martendemos: The issue (in line 374) was, that a key-value pair was add to the request.
This includes the $formName as key. The $formName was already deprecated (see BaseController.php line 345) and seems to me(!) not necessary for the logic anymore.
Without the $formName as key, the whole logic of setting the id does not make any sense anymore.
But maybe I have missed something and the $formName  is already necessary at some point?!

mussbach (approved):
lgtm, code review only

PR #2301, opened 2023-11-10
Fix(refs: t34474): incorrect welcomepage

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34474#834922

### Description: 

- the welcome page shall have more common formulation/content

### PR Checklist


hwiem (approved):
lgtm, code review only

templates/bundles/DemosPlanCoreBundle/DemosPlanUser/gateway_newUser.html.twig
```diff
@@ -1,5 +1,7 @@
 {% extends '@DemosPlanCore/DemosPlanCore/base_oeb.html.twig' %}
+
 {% if templateVars is defined %}
+
```
muellerdemos, line 4: I'd only use a blank line here if this section would be dense codewise.

templates/bundles/DemosPlanCoreBundle/DemosPlanUser/gateway_newUser.html.twig
```diff
@@ -38,26 +39,42 @@
 
                 {% if feedback.success_message is defined %}
                     <p class="flash flash-confirm">
-                        <i class="fa fa-check-circle" aria-hidden="true"></i>
+                        <i
+                            class="fa fa-check-circle"
+                            aria-hidden="true"></i>
```
muellerdemos, line 44:
```suggestion
                        <i
                            class="fa fa-check-circle"
                            aria-hidden="true">
                        </i>
```

muellerdemos: This way we can see easier where the tag ends - and other occurences too

templates/bundles/DemosPlanCoreBundle/DemosPlanUser/gateway_newUser.html.twig
```diff
@@ -38,26 +39,42 @@
...
+                    </h3>
 
                     <label class="u-mb-0_5">
                         {{ "email"|trans }}
-                        <input type="text" class="layout__item" value="{% if user is defined %}{{- user.email|default("-") -}}{% endif %}" disabled>
+                        <input
+                            type="text"
+                            class="layout__item"
+                            value="{% if user is defined %}{{- user.email|default("-") -}}{% endif %}"
+                            disabled>
```
muellerdemos, line 77: why is this input field always disabled, I wonder if we need this if it is always disabled.

templates/bundles/DemosPlanCoreBundle/DemosPlanUser/gateway_newUser.html.twig
```diff
@@ -71,9 +88,19 @@
...
+                                        rel="noopener noreferrer">
+                                        {{ 'newsletter.subscribe'|trans }}
+                                    </a>
+
                                 {% else %}
-                                    <input type="checkbox" name="newsletter" {{ user.wantsNewsletter is defined and user.wantsNewsletter == true? 'checked="true"' : '' }}>
+                                    <input
+                                        type="checkbox"
+                                        name="newsletter"
+                                        {{ user.wantsNewsletter is defined and user.wantsNewsletter == true? 'checked="true"' : '' }}>
```
muellerdemos, line 103:
```suggestion
                                        {{ user.wantsNewsletter is defined and user.wantsNewsletter == true ? 'checked="true"' : '' }}>
```

muellerdemos (commented):
I was too slow ;(

templates/bundles/DemosPlanCoreBundle/DemosPlanUser/gateway_newUser.html.twig
elmasdemos: I learned that our convention is to put a blink line before every if condition

salisdemos: before, but not after ;-)
```
const some  = code

if(some) {
  someCode()

  return foo
} else {
  return bar
}
```

hwiem: ...unless the after is also a before:

```
{% if templateVars is defined %}

  {% if templateVars.title is defined %}
	 // something here
  {% endif %}

{% endif %}
```

salisdemos: true

PR #2303, opened 2023-11-10
f_T35396 fix assignee filter

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T35396

Description: 
- the departement resource type is not available while filter segments by assignee:
- a new permission was added to make the departement resource type available while segment by assignee filter
- Department resource type was adjusted : add relevant permission to make the resource type available.


bakridemos:
Wrong branch was used here that's why this pr will be redone in https://github.com/demos-europe/demosplan-core/pull/2325

PR #2304, opened 2023-11-10
F t35367 adjust authored date export value format

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T35367

Description:
The exported xls file contained a unix timestamp instead of a readable
formatted date string. This change formats this value so it is the same as the other date value
present within the export (d.m.Y)

### How to review/test
get an up to date regio db, check out branch and download the xls statement(s) as described within the ticket.
The date value should look like `01.11.2023` instead of the unix timeStamp - lokal exampleExport:
[Abwägungstabelle-10-11-2023-14 07.xlsx](https://github.com/demos-europe/demosplan-core/files/13320402/Abwagungstabelle-10-11-2023-14.07.xlsx)

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -139,7 +139,7 @@
                     :size="5" />
                   <dp-input
                     id="r_orga_city"
-                    v-model="values.submitter.city"
+                    v-model="values.submitter.ort"
                     class="o-form__group-item"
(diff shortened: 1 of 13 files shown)


MoritzMandler:
I changed the root branch from release to main, but updated the branch beforehand - so I have these into release merged changes within my diff... https://github.com/demos-europe/demosplan-core/pull/2325 
This should not be a problem I think.

PR #2306, opened 2023-11-13
chore: Add required prop to test

gruenbergerdemos:
Required prop `procedureId` was missing in jest test for DpPublicStatement.


muellerdemos (approved):
thanks for fixing the test, code lgtm

PR #2321, opened 2023-11-14
F t35348 error message when adding text while editing sections

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T35348

Description: 
The error message when adding text while editing sections suggests that there's a syntax error in your DQL (Doctrine Query Language) query. It expects a closing parenthesis but got an other character instead. This usually occurs due to malformed DQL. The $recommendationText variable might contain a value that is breaking the query syntax that's why we have to bind it to make sure that if the content of this variable is supposed to be a string literal in the query, it is properly escaped or handled.


mussbach (approved):
I am not very familiar with this syntax, but using params is definitely a good idea. Code review only

PR #2329, opened 2023-11-15
Fix sentry pattern for dependabot grouping

spiess-demos:
Maybe dependabot is confused by the slash in @sentry package names.


salisdemos (approved):
lets give it a crack

PR #2330, opened 2023-11-15
feat(refs T33536): tweak import menu entry in sidemenu

muellerdemos:
https://yaits.demos-deutschland.de/T33536

- The menu entry for import should be renamed to `Import(Stellungnahmen)`
- The menu entry should be in the analysis section and not in the configuration section anymore

Screenshot 2023-11-15 at 14-47-44 Übersicht EIN TEST Raumordnung Online

### How to review/test

- checkout branch
- build either `blp`or the `regio` project
- login as FPA user
- confirm the entries of the side menu are in the correct spot

### PR Checklist

related PR to override the translation for ewm: https://github.com/demos-europe/demosplan-project-ewm/pull/97


translations/messages+intl-icu.de.yml
```diff
@@ -1828,6 +1828,7 @@ import.process.ready.for.review: "Bereit zur Überprüfung"
 import.process.generating.text: "In Verarbeitung (Text-Erkennung)"
 import.process.ready.for.approval: "Bereit für Bestätigung"
 import.process.finished: "Import abgeschlossen"
+import.statements: "Import (Stellungnahmen)"
```
spiess-demos, line 1831: This is true for all projects but EWM. In EWM, segments can be imported as well.
I'd suggest to override the translation key in EWM to just say "Import" as it is now. To have a translation key that still makes sense in both situations, we could rename it to `import.menu` or something like that.

spiess-demos (changes requested):

translations/messages+intl-icu.de.yml
muellerdemos: Thanks, this makes sense to me and was done here https://github.com/demos-europe/demosplan-core/pull/2330/commits/ab92bb44a5c8f85f7894520e4d9b9a57334c9c7a 

and the ewm translation override PR can be found here: 
https://github.com/demos-europe/demosplan-project-ewm/pull/97

spiess-demos (approved):
lgtm! (code review only)

PR #2335, opened 2023-11-17
feature (ref 35482): statement interface instead of segment interface

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T35482

Description: 

- use statementInterface instead of segmentInterface: 
  the parameter 'segmentOrStatement' could be a segment or a statement.  The Segment entity is already a child of StatementInterface otherwise the Statement entity implement not the Segment Interface that's why we have to use the parent class here to make both entities valid.  
- adjust name of method: method name was false

### How to review/test
test with export file or code review

### Linked PRs (optional)

- https://github.com/demos-europe/demosplan-project-ewm/pull/98

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm, code review and tested too

PR #2338, opened 2023-11-17
f_T35396 fix assignee filter

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T35396

Description: 
The departement resource type is not available while filter segments by assignee. A new permission was added to make the departement resource type available while segment by assignee filter


jalalifardemos (approved):
thx, code review only

PR #2341, opened 2023-11-20
F t35309 add csrf attr for post form

ahmad-demos:
https://yaits.demos-deutschland.de/T35309
https://github.com/demos-europe/demosplan-core/pull/2254

- add csrf protection attr for Post forms.


mussbach (approved):
lgtm, thx

PR #2356, opened 2023-11-22
fix(refs: T35510): add flex class to wrapper to align button properly

muellerdemos:
https://yaits.demos-deutschland.de/T35510

- Easy fix to place the button on the same horizontal axis as the icons

### How to review/test

- checkout
- select a procedure and  navigate to assessment table detail view
- check if the button is properly aligned on the same axis as the icons

image

### PR Checklist

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/statement_detail_subheader.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/statement_detail_subheader.html.twig
@@ -39,7 +39,7 @@
 {% endif %}
 
 
-<div class="u-pt-0_125 layout--flush">
+<div class="u-pt-0_125 layout--flush flex">
     <div class="layout__item u-2-of-4">
 
         {# The css class "c-actionmenuat__trigger" is used here only lazily to style the single links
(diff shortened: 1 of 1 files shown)


gruenbergerdemos (approved):
lgtm

templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/statement_detail_subheader.html.twig
```diff
@@ -39,7 +39,7 @@
 {% endif %}
 
 
-<div class="u-pt-0_125 layout--flush">
+<div class="u-pt-0_125 layout--flush flex">
```
spiess-demos, line 42: could you try to remove the `flow-root` class in line 102 instead of adding flex here? i guess that introduced the issue in the first place (shame on me).

spiess-demos (changes requested):

templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/statement_detail_subheader.html.twig
muellerdemos: thank you, this is indeed more elegant than adding another class. done here https://github.com/demos-europe/demosplan-core/pull/2356/commits/ad17afe38a7475976616044d763b5b5ea6bd5dd4

spiess-demos (approved):
lgtm! code review only

PR #2357, opened 2023-11-22
fix(refs: T34709): adjust button placement on public detail document view

muellerdemos:
https://yaits.demos-deutschland.de/T34709

- The button was not properly included in the corresponding warpper component
- The class `elements__grid-grow` prevent a proper adjustment of the button hence was replaced with `flow-root` to easily adjust the download button at the end of the container
- The download button class list was added float-right to align it

### How to review/test

- checkout
- find a plan with planning documents
- navigate to the procedure's public detail view and make sure the download buttons are consistently aligned in relation to each other:
image

### PR Checklist

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanDocument/public_elements_list.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanDocument/public_elements_list.html.twig
@@ -83,7 +83,7 @@
                                         </dp-video-player>
                                     {% else %}
                                         {# Simple download link that is also read out by screen readers #}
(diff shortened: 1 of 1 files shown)


gruenbergerdemos (commented):
The buttons are not completely aligned if that was the intention. The blue button is a little more right.

muellerdemos:
@gruenbergerdemos sorry, I forgot to link the ticket, which I have done now.

The issue this PR wants to solve is visible in this screenshot:
image

chapter related categories had a download button which just seemed off compared to the download button in document related categories. I think it is fine that the content has a little padding within

gruenbergerdemos (approved):
lgtm

PR #2361, opened 2023-11-23
bug (refs T28095): Change width of label modal

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T28095

The modal should not stretch over the whole site's width.

### DIFF
--- a/client/js/components/procedure/imageAnnotator/DpLabelModal.vue
+++ b/client/js/components/procedure/imageAnnotator/DpLabelModal.vue
@@ -10,7 +10,7 @@
 <template>
   <dp-modal
     ref="labelModal"
-    content-classes="width-auto">
+    content-classes="u-1-of-3">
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/imageAnnotator/DpLabelModal.vue
```diff
@@ -10,7 +10,7 @@
 <template>
   <dp-modal
     ref="labelModal"
-    content-classes="width-auto">
+    content-classes="u-1-of-3">
```
spiess-demos, line 13: This way it looks ok "in most cases", excluding extremely large and very narrow devices. We could also set a fixed width here, as we already know the possible values for the select (maybe `width-300` would be fine?).

spiess-demos (approved):
lgtm! code review only.

PR #2362, opened 2023-11-23
F T35519 fix deleting orphaned emails

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T35519

Description:  
add new event 'GetToDeleteEmailIdsEvent' which allow to retrieve used emails ids in addons which have to be not removed from core while deleting orphaned emails.


MoritzMandler (dismissed):
lgtm

PR #2364, opened 2023-11-24
F T35528 make support contact work the same email

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T35528

Description: 
- Using the EmailAddress causes many issues which do not meet the users expectations ( use the same emails for many support contact ). that's why the email address property have to be of type a string, not any more a relation to EmailAddress and with no unique constraint.
- The entity 'SupportContact' and the table 'support_contact' are adjusted.
- The resource Types a re adjusted.

### DIFF
--- a/client/js/components/user/CustomerSettings/CustomerSettingsSupport.vue
+++ b/client/js/components/user/CustomerSettings/CustomerSettingsSupport.vue
@@ -197,7 +197,14 @@ export default {
     },
 
     resetForm () {
-      this.customerContact = emptyCustomer
+      const {title, phoneNumber, eMailAddress, text, visible } = emptyCustomer
+      this.customerContact = {
+        title,
+        phoneNumber,
+        eMailAddress,
+        text,
+        visible
+      }
       this.updating = false
     },
 
--- a/client/js/components/user/CustomerSettings/CustomerSettingsSupport.vue
+++ b/client/js/components/user/CustomerSettings/CustomerSettingsSupport.vue
@@ -197,14 +197,7 @@ export default {
     },
 
     resetForm () {
-      const {title, phoneNumber, eMailAddress, text, visible } = emptyCustomer
-      this.customerContact = {
-        title,
-        phoneNumber,
-        eMailAddress,
-        text,
-        visible
-      }
+      this.customerContact = { ...emptyCustomer }
       this.updating = false
     },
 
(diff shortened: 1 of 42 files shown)


client/js/components/user/CustomerSettings/CustomerSettingsSupport.vue
```diff
@@ -197,7 +197,14 @@ export default {
     },
 
     resetForm () {
-      this.customerContact = emptyCustomer
+      const {title, phoneNumber, eMailAddress, text, visible } = emptyCustomer
```
salisdemos, line 200:
```suggestion
      const { title, phoneNumber, eMailAddress, text, visible } = emptyCustomer
```

client/js/components/user/CustomerSettings/CustomerSettingsSupport.vue
```diff
@@ -197,7 +197,14 @@ export default {
...
+      this.customerContact = {
+        title,
+        phoneNumber,
+        eMailAddress,
+        text,
+        visible
+      }
```
salisdemos, line 207: what do we want to archive with that?
Isn't it the same as
```suggestion
      this.customerContact = { ...emptyCustomer }
```

salisdemos (changes requested):

client/js/components/user/CustomerSettings/CustomerSettingsSupport.vue
muellerdemos: this was removed in favor of your suggestion in the comment below.

client/js/components/user/CustomerSettings/CustomerSettingsSupport.vue
muellerdemos: Yes this does basically the same what I wanted to achieve. Since the empty customer object was not defined in the export default scope I had in mind that vue could not establish the needed reactivity links. Since the spread operator also creates a copy of the emptyCustomer object within the export default scope, your suggestion is leading to less lines of code. This was adjusted here: https://github.com/demos-europe/demosplan-core/pull/2364/commits/6cf2739d3a13f2e971d3802781e78f5e56c6b07e

salisdemos (dismissed):

jalalifardemos (dismissed):
Backend side is approved

bakridemos:
this pr is already redone and merged

PR #2366, opened 2023-11-24
fix(refs: T35563) add register button to public detail view

muellerdemos:
https://yaits.demos-deutschland.de/T35563

This PR is the core part for this bug fix PR in regio:
https://github.com/demos-europe/demosplan-project-regio/pull/90 and simply adds the register flyout component to the entry points and adds two utility css classes which were needed for the fix.

### PR Checklist

### DIFF
--- a/client/js/InitVue.js
+++ b/client/js/InitVue.js
@@ -29,6 +29,7 @@ import loadSentry from './loadSentry'
 import NotificationStoreAdapter from '@DpJs/store/core/NotificationStoreAdapter'
 import NotifyContainer from '@DpJs/components/shared/NotifyContainer'
 import PortalVue from 'portal-vue'
+import RegisterFlyout from '@DpJs/components/user/RegisterFlyout'
 import Vue from 'vue'
 import Vuex from 'vuex'
 
@@ -40,6 +41,7 @@ Vue.use(DPVueCorePlugin)
 
 // Register components that are used globally
 Vue.component('DpObscure', DpObscure)
+Vue.component('RegisterFlyout', RegisterFlyout)
 
 function initialize (components = {}, storeModules = {}, apiStoreModules = [], presetStoreModules = {}) {
   bootstrap()
--- a/client/js/InitVue.js
+++ b/client/js/InitVue.js
@@ -41,7 +41,6 @@ Vue.use(DPVueCorePlugin)
 
 // Register components that are used globally
 Vue.component('DpObscure', DpObscure)
-Vue.component('RegisterFlyout', RegisterFlyout)
 
 function initialize (components = {}, storeModules = {}, apiStoreModules = [], presetStoreModules = {}) {
   bootstrap()
@@ -67,7 +66,8 @@ function initialize (components = {}, storeModules = {}, apiStoreModules = [], p
         DpAccordion,
         DpFlyout,
         HamburgerMenuButton,
-        NotifyContainer
+        NotifyContainer,
+        RegisterFlyout
       },
       store: store,
       mounted () {
--- a/client/js/bundles/procedure/publicIndexMapAndList.js
+++ b/client/js/bundles/procedure/publicIndexMapAndList.js
@@ -14,9 +14,13 @@
 import { dpApi, prefixClass } from '@demos-europe/demosplan-ui'
 import DpSearchProcedureMap from '@DpJs/components/procedure/publicindex/DpSearchProcedureMap'
 import { initialize } from '@DpJs/InitVue'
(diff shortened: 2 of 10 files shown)


client/js/bundles/procedure/publicIndexMapAndList.js
```diff
@@ -14,9 +14,13 @@
 import { dpApi, prefixClass } from '@demos-europe/demosplan-ui'
 import DpSearchProcedureMap from '@DpJs/components/procedure/publicindex/DpSearchProcedureMap'
 import { initialize } from '@DpJs/InitVue'
+import RegisterFlyout from '@DpJs/components/user/RegisterFlyout'
 
 const stores = {}
-const components = { DpSearchProcedureMap }
+const components = {
+  DpSearchProcedureMap,
+  RegisterFlyout
```
spiess-demos, line 22: I'm wondering if this is another case for registering a component globally... like we already do it with [some other components](https://github.com/demos-europe/demosplan-core/blob/21da7fe683e43c2151fadd59ef6a2d06cbeb0b4e/client/js/InitVue.js#L64-L68) that are required basically on every page.

demosplan/DemosPlanCoreBundle/Resources/client/scss/utility/_font-size.scss
```diff
@@ -73,3 +73,7 @@
 .font-size-0 {
     font-size: 0;
 }
+
+.font-size-1_2 {
```
spiess-demos, line 77: As we are soon moving those into yaml/tailwind, i do not recommend to add it here. Also, why do we need a "one off" font size?

demosplan/DemosPlanCoreBundle/Resources/client/scss/utility/_height.scss
```diff
@@ -48,6 +48,10 @@ $height-map-px: (
         height: 100vh; // .height-100vh
     }
 
+    &-44px {
```
spiess-demos, line 51: If you really need that "random" px value, try Tailwind with `h-[44px]`.

spiess-demos (changes requested):

demosplan/DemosPlanCoreBundle/Resources/client/scss/utility/_height.scss
muellerdemos: I tried to use an [arbitrary](https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values) value here but for some reason it did not work (which I couldn't find out why)

client/js/bundles/procedure/publicIndexMapAndList.js
muellerdemos: I think adding this component globally is less troublesome than adding it on multiple entry points for the same result, hence added it globally here: https://github.com/demos-europe/demosplan-core/pull/2366/commits/e0fa4600232ad27e204250d7a239759864279120

demosplan/DemosPlanCoreBundle/Resources/client/scss/utility/_font-size.scss
muellerdemos: I was struggling with style overrides in the flyout component, this way appeared as simple solution to me with just minor drawbacks. However what would your proposal be how to handle this case?

spiess-demos: i see. in that case, could you move the css declaration into the project override?

demosplan/DemosPlanCoreBundle/Resources/client/scss/utility/_height.scss
gruenbergerdemos: I removed the class here and used tailwind class `h-11` which fits exactly [5b6ce57](https://github.com/demos-europe/demosplan-core/pull/2366/commits/5b6ce57a2d6e31276391eb15750fbf31b1fa0f21)

client/js/InitVue.js
```diff
@@ -40,6 +41,7 @@ Vue.use(DPVueCorePlugin)
 
 // Register components that are used globally
 Vue.component('DpObscure', DpObscure)
+Vue.component('RegisterFlyout', RegisterFlyout)
```
spiess-demos, line 44: We could also register it in line 70.

gruenbergerdemos: Done in [b700e9c](https://github.com/demos-europe/demosplan-core/pull/2366/commits/b700e9c45352bf30fcfb77b3dd5a4cd9b361e297)

spiess-demos (approved):
lgtm (code review)

PR #2368, opened 2023-11-27
chore: add version to addon list command

mussbach:
This PR enhances the addon list command to show versions. These are derived from the composer.json file in the addon and saved in addons.yml file during installation

### How to review/test
run `bin/<project> dplan:addon:list` 

### PR Checklist

- [x] Tests updated/created
- [ ] Update documentation


mussbach:
> There is not a version of the addon in all composer.json files. In this case, version 0.0.0.0 would be output. Should it be added there with every new tag (new version) in the future? Where there is a version, only the initial version is there, for example: 0.1.0. Should these composer.json files be adjusted?
> 
> test works

You are right. I built a release script that sets the version automatically to the composer.json and package.json and updates the Changelog. So in a hopefully new future (when you agree to my suggestion of the script) the versions will be set. At the moment this PR does not help yet but does no harm

breierDemos (approved):
Okay, that sounds good :)

PR #2375, opened 2023-11-27
feature (ref 28716): fix key name for user

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T28716

Description: fix key name for name column in xlsx export file

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
not too sure about this one but it may well work

PR #2377, opened 2023-11-27
bug (refs T34935): Move tooltip outside of label

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34935

Move tooltip, so it can float right outside of label.

### How to review/test
Go to Verfahren -> Blaupause -> Grundeinstellungen. Tooltip of "Email Verfahrensträger" should align right with the other tooltips.


salisdemos (approved):

spiess-demos:
in my opinion, the bug is rather that tooltips are that far away from the elements they are referring to. As it is now, users have a hard time spotting that there might be helpful texts.

gruenbergerdemos:
> in my opinion, the bug is rather that tooltips are that far away from the elements they are referring to. As it is now, users have a hard time spotting that there might be helpful texts.

I suggest to open a new ticket for this and look trough the whole application to change it everywhere for consistency.

PR #2378, opened 2023-11-27
bug (refs T34593): Add more margin to beta badge

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34593

Add more margin so there is more space between the badge and the close icon (I used tailwind class for this).

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
@@ -131,7 +131,7 @@
                   :text="Translator.trans('segment.oracle.tooltip')" />
                 <dp-badge
                   v-if="activeId === 'oracleRec'"
-                  class="absolute u-right-0 u-mr-0_75"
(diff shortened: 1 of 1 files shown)


salisdemos (approved):

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -131,7 +131,7 @@
                   :text="Translator.trans('segment.oracle.tooltip')" />
                 <dp-badge
                   v-if="activeId === 'oracleRec'"
-                  class="absolute u-right-0 u-mr-0_75"
+                  class="absolute u-right-0 mr-12"
```
salisdemos, line 134: wait. If its positioned absolute. Should'nt we use `r-12` instead and remove `u-right-0`?

gruenbergerdemos: Yes, you're right, it's the better solution. I open a new PR.

PR #2386, opened 2023-11-28
fix: Use correct Option to exclude pacakges from pattern

salisdemos:
I used a not existing option... 
see: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups


spiess-demos (approved):
lgtm!

PR #2388, opened 2023-11-28
bug (refs T34593): Remove margin and adjust positioning right

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34593

Since it is positioned absolute, we can change the position to the right and don't need any margin.

### Linked PRs (optional)

(https://github.com/demos-europe/demosplan-core/pull/2378)


salisdemos (approved):
Thanks!

PR #2373, opened 2023-11-27
fix(refs: T34613): add orga status filter

muellerdemos:
### Ticket:
https://yaits.demos-deutschland.de/T34613

### Description:
This PR fixes:
- Rejected orga statuses should be filtered out if the corresponding checkbox has been selected
image

### How to review/test

- checkout
- login as support user
- navigate to `/organisation/list`
- select a filter checkbox 
image
- make sure the selected checkbox results in a list where none of the statuses (of the selection) shows a `zurückgewiesen` status here: 
image

### PR Checklist

Delete the checkbox if it doesn't apply/isn't necessary.
- [x] Link all relevant tickets
- [x] Move the tickets on the board accordingly

PR #2390, opened 2023-11-29
bug (refs T34613): Fix filter orga status

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T34613

Only filter for orga status if the filter is active. Otherwise we want all orgas. 

### How to review/test
- login as support user
- navigate to `/organisation/list`
- select a filter checkbox 
- make sure the selected checkbox results in a list where none of the statuses (of the selection) shows a `zurückgewiesen` status
- remove filter -> make sure all orgas are shown, especially the ones that were filtered out before with the status "zurückgewiesen"

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-core/pull/2373

### DIFF
--- a/client/js/components/user/DpOrganisationList/DpOrganisationList.vue
+++ b/client/js/components/user/DpOrganisationList/DpOrganisationList.vue
@@ -420,7 +420,6 @@ export default {
 
     fetchAllOrganisations (page) {
       this.isLoading = true
-      console.log('here')
 
       this.list({
         page: {
--- a/client/js/components/user/DpOrganisationList/DpOrganisationList.vue
+++ b/client/js/components/user/DpOrganisationList/DpOrganisationList.vue
@@ -382,19 +382,19 @@ export default {
               memberOf: 'orgaType'
             }
           }
+
+          filterObject.orgaStatus = {
+            condition: {
+              path: 'statusInCustomers.status',
+              operator: '<>',
+              value: 'rejected'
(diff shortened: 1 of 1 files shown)


client/js/components/user/DpOrganisationList/DpOrganisationList.vue
```diff
@@ -420,6 +420,7 @@ export default {
 
     fetchAllOrganisations (page) {
       this.isLoading = true
+      console.log('here')
```
spiess-demos, line 423: Did you forgot to remove this or is it intentionally left here?

spiess-demos (commented):
I do not understand what is the difference between this PR and [the other PR](https://github.com/demos-europe/demosplan-core/pull/2373/) linked in the description.

gruenbergerdemos:
> I do not understand what is the difference between this PR and [the other PR](https://github.com/demos-europe/demosplan-core/pull/2373/) linked in the description.

The difference is, that now the filter condition for orga status is only applied if a filter is selected

spiess-demos:
If we understand the checkboxes as "Show only orgas which have an `accepted` status for orga type a, b, or c" (which is ok), then we would also need to exclude all orgas with `pending` status whenever a checkbox is checked. This could easily be accomplished.

spiess-demos (approved):
works and lgtm!

PR #2393, opened 2023-11-29
feat (refs T35632): duplicate csrf token usage leads to error

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T35632

In some cases csrf token is already set in BE via symfony forms.
For this cases the additional csrf in template is removed with this PR.

### How to review/test
Creation of Boilerplatetext and procedure template should work with this PR and should not without this one.


mussbach (approved):
lgtm, thx

PR #2409, opened 2023-12-01
feature (ref 31520, 32997): delete Orga and all dependency by command…

jalalifardemos:
**Ticket:** 

- https://yaits.demos-deutschland.de/T32997
- https://yaits.demos-deutschland.de/T31520

**Description**: create a class to delete an organization and all dependencies (addresses, departments, users, procedures, settings, statement_meta and slug). Instead of use pure queries to find one organization and dependencies, we can use this command. I am almost sure, that I have found all entries that must have been deleted. I tested it with multiple organizations in local DB, and it seemed to work so far.

### How to review/test
bin/{project} dplan:orga:delete [-wrp | --without-repopulate] [--] <orgaId>
During code review: please pay attention to what entries in what tables get deleted and especially if you think I forgot to include something that also has a relationship to the organization!

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Command/Data/DeleteOrgaCommand.php
+++ b/demosplan/DemosPlanCoreBundle/Command/Data/DeleteOrgaCommand.php
@@ -1,0 +1,342 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of the package demosplan.
+ *
+ * (c) 2010-present DEMOS plan GmbH, for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+namespace demosplan\DemosPlanCoreBundle\Command\Data;
+
+use demosplan\DemosPlanCoreBundle\Command\CoreCommand;
+use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception;
+use Doctrine\ORM\EntityManagerInterface;
+use EFrane\ConsoleAdditions\Batch\Batch;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Command/Data/DeleteOrgaCommand.php
```diff
@@ -0,0 +1,341 @@
...
+    {
+        $this->deleteFromTableByIdentifierArray('_orga_departments_doctrine', '_o_id', [$this->orgaId]);
+    }
+
+    /**
+     * @throws Exception
+     */
+    private function deleteUsers(array $usersId): void
+    {
+        $this->deleteFromTableByIdentifierArray('_user', '_u_id', $usersId);
```
martendemos, line 166: Afaik your system can not handle deleted users at all.

demosplan/DemosPlanCoreBundle/Command/Data/DeleteOrgaCommand.php
```diff
@@ -0,0 +1,341 @@
...
+    {
+        $this->deleteFromTableByIdentifierArray('orga_slug', 'o_id', [$this->orgaId]);
+    }
+
+    /**
+     * @throws Exception
+     */
+    private function deleteOrga(): void
+    {
+        $this->deleteFromTableByIdentifierArray('_orga', '_o_id', [$this->orgaId]);
```
martendemos, line 222: What about the relations to orgas which will be deleted by this?

demosplan/DemosPlanCoreBundle/Command/Data/DeleteOrgaCommand.php
martendemos: Additionally this and also the other relations of an orga should be delete correctly by doctrine via cascade deletion. If this does not happend yet, why not fixing this instead of creating new logic?

martendemos (changes requested):
Not finished reviewed, but I think there are many dependencies which needs to be cleared (technically and functionally) before we can implement the deletion of central entites like orgas, users, statements or even procedures.

mussbach:
ping @jalalifardemos What happens with this PR?

jalalifardemos:
@MoritzMandler and @bakridemos  created another PR , Then we can close it

PR #2431, opened 2023-12-06
bug (T35722): Show hint before export

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35722

Show a hint for data protection reasons before exporting statements as docx or zip.

### How to review/test
Go to Verfahren -> Stellungnahmen and try to export as docx or zip (the first two options). Check hint and both options (Cancel, Ok).


ahmad-demos (approved):
lgtm, only codeReview!

PR #2442, opened 2023-12-11
f text field is not required

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T35794

Description: text field is not required.


MoritzMandler (approved):
lgtm

PR #2445, opened 2023-12-11
chore (refs T35631): Filter out segments with less than 10 characters

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T35631

Those may lead the frontend to crash and should probably not be needed in a real world scenario

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/store/procedure/SplitStatementStore.js
+++ b/client/js/store/procedure/SplitStatementStore.js
@@ -220,6 +220,7 @@ const SplitStatementStore = {
           const initialData = data.data.attributes.segmentDraftList.data
           const segments = initialData.attributes.segments
             // Filter out segments with less than 10 characters as those may lead the frontend to crash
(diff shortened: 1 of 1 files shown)


client/js/store/procedure/SplitStatementStore.js
```diff
@@ -219,6 +219,9 @@ const SplitStatementStore = {
           }
           const initialData = data.data.attributes.segmentDraftList.data
           const segments = initialData.attributes.segments
+            // Filter out segments with less than 10 characters as those may lead the frontend to crash
```
salisdemos, line 222:
```suggestion
            // Filter out segments with less than 10 characters as those may lead the frontend to crash
            // (because often that are closing or opening tags)
```

salisdemos (approved):

PR #2447, opened 2023-12-12
Allow dots in selectors for purgeCss

spiess-demos:
As tailwind uses the dot for selector names, we have to tell that to purgeCss.

### DIFF
--- a/client/fe/webpack/moduleRules.js
+++ b/client/fe/webpack/moduleRules.js
@@ -80,7 +80,7 @@ if (config.cssPurge.enabled) {
       content: config.cssPurge.paths,
       defaultExtractor (content) {
         const contentWithoutStyleBlocks = content.replace(/<style[^]+?<\/style>/gi, '')
-        return contentWithoutStyleBlocks.match(/[A-Za-z0-9-_/:]*[A-Za-z0-9-_/.]+/g) || []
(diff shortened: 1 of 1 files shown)


graupnerdemos (commented):
What about adding `[`, and `]` while we're at this. Those can also be part of Tailwind selectors.

client/fe/webpack/moduleRules.js
```diff
@@ -80,7 +80,7 @@ if (config.cssPurge.enabled) {
       content: config.cssPurge.paths,
       defaultExtractor (content) {
         const contentWithoutStyleBlocks = content.replace(/<style[^]+?<\/style>/gi, '')
-        return contentWithoutStyleBlocks.match(/[A-Za-z0-9-_/:]*[A-Za-z0-9-_/]+/g) || []
+        return contentWithoutStyleBlocks.match(/[A-Za-z0-9-_/:]*[A-Za-z0-9-_/.]+/g) || []
```
salisdemos, line 83: shouldn't the `.` be escaped?

graupnerdemos: Not when it's in a character group.

spiess-demos: within square brackets, the dot loses its special meaning and is treated as a literal dot.

spiess-demos:
> What about adding `[`, and `]` while we're at this. Those can also be part of Tailwind selectors.

eab1caefbb6eb13a09eb9ae2c20cfeaa207e5258

PR #2455, opened 2023-12-13
feat (refs T33578): delete related original statement on delete statement.

martendemos:
Ticket: [T35868](https://yaits.demos-deutschland.de/T35868)

Remaining original statement on deletion of a statement leads to several issues.
One of them is, that a re-import of a statement ([T33578](https://yaits.demos-deutschland.de/Txxyyzz)) is not possible but the reason is not visible for the user at all.
To solve this as a whole, on deletion of a statement, the related original statement will now be deleted in case of no other children are related to the original statement.

Delete a statement and check in the database, that the related original statement is also deleted.

This PR inherit the major changes of https://github.com/demos-europe/demosplan/pull/7803 (outdated)

- [x] Tests updated/created
- [ ] Update documentation


jalalifardemos (approved):
lgtm, Thanks

PR #2458, opened 2023-12-14
feature (ref 35620): ignore token from post request form to use array…

jalalifardemos:
**Ticket:** : https://yaits.demos-deutschland.de/T35620

Description: Token is part of post request and because of that, the type of first variable in foreach is string. I used an IF condition to ignore '_token' variable.

- [x] Tests updated/created
- [ ] Update documentation


mussbach (approved):
lgtm, code review only

PR #2467, opened 2023-12-15
f T35837 remove draft json validation

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T35837

Description:  the SegmentDraftJsonConstraintValidator cover not the case when text proposals come from DATA, the validator in this case will not allow to save the draft info json which is correct but handle this rest of action not anymore : Json is not saved and DATA is not able to see or handle that. This will be removed now and redone in a way to cover all cases .


mussbach (approved):
thx

PR #2468, opened 2023-12-15
chore: automatically enable addon during installation

mussbach:
This quality of life PR takes the required action in 99% of the installation cases as default and enables the addon immediately after installation. The new option `no-enable` only installs the addon without enabling it

### How to review/test
install an addon via `bin/<project> dplan:addon:install` and choose the addon to install. Then check addons/addons.yaml, whether the addon is activated

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (changes requested):
Was this discussed again? I thought we didn't want addons enabled by default?

mussbach:
Some loud single voices said so, yes. I got so many grey hair and lost so much time from this weird behavior that I need to install an addon twice (sorry, install only once, manually enable it then...) that I made this PR. When this is only my personal problem and the team votes against it so be it, but I do not understand or like the current behavior at all.

breierDemos (approved):
So for me that would be fine :)

PR #2444, opened 2023-12-11
feature (ref 33679): create a column for original statement in xlx fi…

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T33579

Description: add original statement as a xlsx column in zip export

### Linked PRs (optional)
List other PRs that are somehow connected to this and explain the connection.

- https://github.com/demos-europe/demosplan-project-regio/pull/91

Delete the checkbox if it doesn't apply/isn't necessary.

- [ ] Tests updated/created
- [ ] Update documentation
- [x] Link all relevant tickets
- [x] Move the tickets on the board accordingly

PR #2472, opened 2023-12-19
feature (ref 33579): test for assessmentTableZipExporter

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T33579

Description: Test for AssessmentTableZipExporter

### Linked PRs (optional)

- https://github.com/demos-europe/demosplan-core/pull/2444
- https://github.com/demos-europe/demosplan-core/pull/2363


martendemos (approved):
Only __invoke is covered by a tests, which seems not enough for me on the first look. 
At a second look I see, the complication here. To test more, multiple non-empty mocks would be necessary, which seems to much for know. 
So I approve and appreciate the base, which is created with this PR.

PR #2481, opened 2023-12-20
fix: T35881 fix path to fix reordering of elements

breierDemos:
**Ticket:** https://yaits.demos-deutschland.de/T35881

If you reordering elements in "Planungsdokumente und Planzeichnung", this will not be saved.

This is the case because: getConfigPath builds a path for the config directory, so 'config' is twice in the path, was caused a failure to get the content of a schema json file.

I removed the config in the relative path string. Now the rearrangement should be saved again.

### How to review/test
reorder elements in "Planungsdokumente und Planzeichnung"

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
lgtm, thx. Code review only

martendemos (approved):
lgtm
(Causing change was 24c205c084d649f925e9c62b749b00677dba954d.)

PR #2485, opened 2023-12-21
Fix validation of unique statement.internId per procedure

martendemos:
**Ticket:** [T35878](https://yaits.demos-deutschland.de/T35878)
Bug of Story:[T33577](https://yaits.demos-deutschland.de/T33577)

statement->getInternId() was always return the internId of the related original statement, which is "correct for the application-logic but leads to wrong results on validate data of a specific statement obviously because the data of a different statement was returned.

### How to review/test
Using the "Beteiligungs-Import" (ZipImport) to import a Zip wich statements also has internIds("Eingangsnummern").
In [T35878](https://yaits.demos-deutschland.de/T35878) is a Zip, which covering this case.

Please merge in case of approve.


jalalifardemos (approved):
lgtm, Code review only

PR #2486, opened 2023-12-22
feature (ref 35885): use statement parent id except statement id

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T35885 
https://yaits.demos-deutschland.de/T35896

Description: Because attachment files stored in parent of statement, we should call it except statement. In file container entity, files stored with statement parent ID as an entity_id.

### How to review/test
Browser or code review

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm

PR #2488, opened 2023-12-22
Fix(T35897): no layer available after fetch capabilities

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T35897

**Description:** This PR fixes the issue with the fetch request that occurs after refactoring externalApi method from axios to fetch. All processing that depends on the text being read from the URL should be in the final 'then' clause of the fetch.

### DIFF
--- a/client/js/components/map/admin/LayerSettings.vue
+++ b/client/js/components/map/admin/LayerSettings.vue
@@ -413,9 +413,12 @@ export default {
 
       externalApi(url)
         .then(response => {
+          return response.text()
+        })
+        .then(text => {
           this.serviceType = hasWMTSType ? 'wmts' : 'wms'
           parser = this.serviceType === 'wmts' ? new WMTSCapabilities() : new WMSCapabilities()
-          this.currentCapabilities = parser.read(response.data)
+          this.currentCapabilities = parser.read(text)
 
           if (this.currentCapabilities !== null) {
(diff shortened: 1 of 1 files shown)


client/js/components/map/admin/LayerSettings.vue
```diff
@@ -413,9 +413,12 @@ export default {
 
       externalApi(url)
         .then(response => {
+          return response.text()
+        })
+        .then(text => {
```
muellerdemos, line 418:
```suggestion
        .then(capabilities => {
```
Please use a more descriptive variable name. I think text is too generic and makes it harder to understand.

client/js/components/map/admin/LayerSettings.vue
```diff
@@ -413,9 +413,12 @@ export default {
...
           this.serviceType = hasWMTSType ? 'wmts' : 'wms'
           parser = this.serviceType === 'wmts' ? new WMTSCapabilities() : new WMSCapabilities()
-          this.currentCapabilities = parser.read(response.data)
+          this.currentCapabilities = parser.read(text)
```
muellerdemos, line 421:
```suggestion
          this.currentCapabilities = parser.read(capabilities)
```

muellerdemos (changes requested):

PR #2497, opened 2024-01-02
feat: allow addon installation via GitHub directly

mussbach:
With this PR we are able to install tagged addon versions directly without the need to download the zip file first. Env var GITHUB_TOKEN needs to be set to be able to connect to GitHub repository. Alternatively a custom addon repository may be used by passing `--repos`. 

### How to review/test
run `bin/<project> dplan:addon:install --github` 

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


graupnerdemos (changes requested):
I like the code but this can't be in demosplan-core, since it lists our non-free addons in the code. I thought the idea was to put this into dp?

mussbach:
I see your point but I o not know whether dp is the right place, as the core needs a possibility to install addons as well. I do not see much harm when the names of the addons are public, as they are not accessible.

mussbach:
@graupnerdemos you are right. I built an addon repository 28a8a69e

mussbach:
And call github for the repositories so no names are leaked [6150a68](https://github.com/demos-europe/demosplan-core/pull/2497/commits/6150a68fd3670d6ae769027962c6fd06bf32e5a7)

muellerdemos (approved):
the FE part lgtm but I can't say much about the php part. Thanks  a lot for putting in the effort to make addons more simple to work with.

martendemos (changes requested):
Does not work in each case. 
Still try to find out how to reproduce ....

martendemos (changes requested):
Some Repos are not selectable for example '0' leads to 

> Choice question must have at least 1 choice available.

mussbach:
> Some Repos are not selectable for example '0' leads to
> 
> > Choice question must have at least 1 choice available.

@martendemos this was the case when repository did not have any tags. Caught this case in 0b2e9d38

martendemos (approved):
Still some issues on installation of some of the addons, but seems to be this has nothing to do with the logic implemented with this PR.

PR #2498, opened 2024-01-02
Use Tailwind width + height utility classes

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T27593

The `width` and `height` utility classes are moved to tailwind with this PR. The new spacing values have been applied, where appropriate. For edge cases, the arbitrary value syntax is used.

New spacing scale, for reference:

image

Addditional changes:

- `%` is allowed as part of selector name in moduleRules.js (to enable percentual arbitrary values)
-  Jest snapshots are updated to match the new classnames
- the (now) obsolete scss files are removed

### How to review/test
Things should look roughly the same as before (in terms of height and width).

### PR Checklist

- [X] Tests updated/created
- [x] Update documentation


ahmad-demos (approved):
lgtm, only codeReview!
Thanks for many changes!

PR #2508, opened 2024-01-04
Fix(refs: T35879): incorrect parameter in getRequest

elmasdemos:
***Ticket: https://yaits.demos-deutschland.de/T35879***

- the getRequest should have only a resourceType as parameter

### PR Checklist


mussbach (approved):
lgtm, thx

PR #2511, opened 2024-01-05
F t35495 stn shall not be overwritten when segmentdraftlist exist

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35495

Description: 
- if statement is started to get segmented and the user would like change the statement, he has to finish the segmentation and to edit the segments otherwise the statement is readable only and there is a waning box to warn and inform the user
- so if the user would like change the statement he should do it before starting the segmentation

### PR Checklist


spiess-demos (approved):
lgtm! (code review only)

PR #2521, opened 2024-01-08
fix(refs T35941): enable editor to update after save

muellerdemos:
### Ticket:
https://yaits.demos-deutschland.de/T35941

### Description:
The reactivity was an issue when updating a statement or recommendation text, the change was only visible when re-loading the page hence the component will be forced to remount with the updated data. It was not possible to establish a proper reactivity with the statement or recommendation store objects due to a lot of internal processing of the texts. Since this problem disappears when the component remounts (e.g. on page reload) I decided it would be a valid solution 

- Enhances the behavior of the component after statement text or recommendation text has been adjusted
- Adds a loading spinner when updating one of the editors
- Adds styling to loading spinner to keep the design and not break the view while loading

### How to review/test

- checkout branch
- navigate to A-table in regio project as FPA user
- change statement or recommendation text via editor, save and make sure everything works without a page reload
image

### DIFF
--- a/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
+++ b/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
@@ -563,7 +563,11 @@
                     With tiptap we can set obscure as prop always when the obscure button should be visible in the field,
                     because the permission check (featureObscureText) takes place in tiptap
                     -->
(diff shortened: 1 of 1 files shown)


elmasdemos (approved):
lgtm, code review

PR #2524, opened 2024-01-09
Fix(refs: T35961): incorrect redirect by click to procedure in nav menu

elmasdemos:
***Ticket: https://yaits.demos-deutschland.de/T35961***

### Description:

- there are 2 different cases due to the role:

### 1. case: the case with the role as data capturer you see a list of the captured statements of the procedure with a different path and permission
\
screnshot24

### 2. case: the other cases you see the overview of the procedure
\
screenshot25

### How to review/test
- login as data capturer, click to a procedure and than e.g. to statement list. than click back to the procedure from the nav menu, you will see the statement list
- login as fpa and do the same and repeat the same way like described

### PR Checklist

### DIFF
--- a/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
+++ b/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
@@ -563,7 +563,11 @@
                     With tiptap we can set obscure as prop always when the obscure button should be visible in the field,
                     because the permission check (featureObscureText) takes place in tiptap
                     -->
+                <dp-loading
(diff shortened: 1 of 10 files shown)


jalalifardemos (approved):
lgtm, code review and test. Thx

mussbach:
While this works it is not really good style. It opens the route to anybody and relies on the permission check on the targeting route. I'd rather prefer to make the link dependent on the permission if possible but no blocker

mussbach (approved):
lgtm, code review only

PR #2535, opened 2024-01-10
bug(refs T35838): Check for overlapping segments

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35838

If we get overlapping segments from data, delete all segments to prevent breaking the site.


client/js/store/procedure/SplitStatementStore.js
```diff
@@ -218,12 +218,29 @@ const SplitStatementStore = {
...
-          const segments = initialData.attributes.segments
+          let segments = initialData.attributes.segments
             // Filter out segments with less than 10 characters as those may lead the frontend to crash
             // (because often that are closing or opening tags)
             // and should probably not be needed in a real world scenario.
             .filter(segment => (segment.charEnd - segment.charStart) > 10)
 
+          // Check if we are getting overlapping segments from pipeline that would cause errors
+          if (doUpdate) {
+            for (let i = 0; i < segments.length; i++) {
```
hwiem, line 229: It doesn't hurt to check this in the FE, but shouldn't overlapping segments be prevented on the pipeline side?

gruenbergerdemos: Yeah, it should. It is just a double check, because we got faulty data in the past and the site was not usable anymore.

hwiem (approved):
code lgtm

elmasdemos (approved):
lgtm. code review only

PR #2543, opened 2024-01-10
chore: update headers to match http2 standard

mussbach:
http2 Standard needs headers to be lowercase. We should adopt this standard. https://httpwg.org/specs/rfc7540.html#HttpHeaders

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


dresslerdemos (dismissed):
A link in the docblock to the "write in lower case" spec would have been nice, but it is absolutely fine without that too.

mussbach:
Good idea, thx, done

PR #2549, opened 2024-01-11
fix: some fields are incorrect or there were missing check

elmasdemos:
### Description: 

- remove lila mesages by removing incorrect fields or using checks


client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -715,10 +717,6 @@ export default {
           Statement: statementFields.join(),
           File: [
             'hash'
-          ].join(),
-          // Segments are only counted, no data needed here.
```
mussbach, line 719: Segments are no longer counted?

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_statement_segments_list.html.twig
```diff
@@ -13,6 +13,7 @@
 {% block component_part %}
     <statement-segments-list
         :current-user="JSON.parse('{{ user|json_encode|e('js', 'utf-8') }}')"
+        :is-source-and-coupled-procedure="{{ templateVars.isSourceAndCoupledProcedure == 1 ? 'true' : 'false' }}"
```
mussbach, line 16: Doesn't `{{ templateVars.isSourceAndCoupledProcedure }}` do the trick as well?

elmasdemos: thanks done https://github.com/demos-europe/demosplan-core/pull/2549/commits/c4fc0363d3264411df7471d54a2d797781273aa5

client/js/components/statement/listStatements/ListStatements.vue
elmasdemos: thanks for the advice. done https://github.com/demos-europe/demosplan-core/pull/2549/commits/c4fc0363d3264411df7471d54a2d797781273aa5

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_statement_segments_list.html.twig
mussbach: Did you try it? Only `{{ templateVars.isSourceAndCoupledProcedure }}` did not work? Perhaps then as `is-source-and-coupled-procedure="{{ templateVars.isSourceAndCoupledProcedure }}"` (without `:`) But I am fine with it as is

client/js/components/statement/listStatements/ListStatements.vue
mussbach: That was more a review question :-). I won't comment this as this may only lead to confusion why this is explicitly stated

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_statement_segments_list.html.twig
elmasdemos: ah yes, I see. sorry didnt recognized it here

elmasdemos: okay, I tried it like your suggestion but it didnt work tthis way

client/js/components/statement/listStatements/ListStatements.vue
elmasdemos: oh, I see. the topic is here that the id of segments cannot be a relationship or attribute. it would be in a higher hierarchy I got explained from backend. so it has to be removed here

elmasdemos: i removed the commit that i put in cause I missunderstood you before https://github.com/demos-europe/demosplan-core/pull/2549/commits/cb0eb76df72eb99a390bf4169b26057299223612

mussbach (approved):
lgtm, thx

PR #2567, opened 2024-01-15
bug(refs T35883): Remove unnecessary title prop

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35883

We don't need the title in the addon. It is only used as page title. If we include it as a prop, the title is added automatically as an attribute to the outer div of the addon component and displayed as tooltip on hover, because the component has no title prop.

### How to review/test
Verfahren -> STN aufteilen -> The statement text shouldn't show a tooltip on hover


muellerdemos (approved):
lgtm

PR #2569, opened 2024-01-15
fix(T36027): issue with elements in the OriginalStatementsTableItem component

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T36027

**Description:** Refactor the code to follow the same logic as in the `DpAssessmentTableCard` component (`line: 469 - 470`):
 utilize `statement.elementId` to identify the corresponding element in the elements list from the `assessmentTable` store.

`statement.elements` often lacks a value, either being an empty array or containing only one item. This adjustment is made to minimize the use of IF checks as well.

### DIFF
--- a/client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
+++ b/client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
@@ -241,16 +241,15 @@ export default {
 
     element () {
       let elementTitle = ''
+      const element = this.statement.elementId ? this.elements.find((el) => el.id === this.statement.elementId) : null
 
-      if (hasOwnProp(this.statement, 'elements') && this.statement.elements.title !== '') {
-        elementTitle = this.statement.elements.title
+      if (element) {
+        elementTitle = element.title
         if (hasOwnProp(this.statement, 'document') && this.statement.document.title !== '') {
           elementTitle += ` / ${this.statement.document.title}`
         }
       } else {
-        const element = this.statement.elementId ? this.elements.find((el) => el.id === this.statement.elementId) : null
-
-        elementTitle = element ? element.title : Translator.trans('notspecified')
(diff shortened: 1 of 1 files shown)


client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
```diff
@@ -241,16 +241,15 @@ export default {
 
     element () {
       let elementTitle = ''
+      const element = this.statement.elementId ? this.elements.find((el) => el.id === this.statement.elementId) : null
 
-      if (hasOwnProp(this.statement, 'elements') && this.statement.elements.title !== '') {
-        elementTitle = this.statement.elements.title
+      if (element) {
+        elementTitle = element.title
```
gruenbergerdemos, line 247: Is there always a title for an element or do we need to check for the title, too?

sakutademos: As I see it, elements have parameters such as category, id, ident, and title. Therefore, if the element exists, it must have the title parameter

sakutademos: added here https://github.com/demos-europe/demosplan-core/pull/2569/commits/813f451342773422bf0ec88cc85df090e5cb210d

gruenbergerdemos: Sorry, that was misunderstanding. I meant if we should check for an empty title like we did before `this.statement.elements.title !== ''`

sakutademos: ah, sorry done here https://github.com/demos-europe/demosplan-core/pull/2569/commits/cf1574cbc71cca4e2908ab1aebb849f0addf1c97

gruenbergerdemos (approved):
lgtm

PR #2570, opened 2024-01-15
F t36053 files on statement copy

martendemos:
**Ticket:** https://yaits.demos-deutschland.de/T36053

On coping a original statement (due to submit a draftStatement), the related files, were missing on coping the statement from the original statement.

### How to review/test
Create and submit a new statement within attachements/files.
All of them should be appearing on the created original statement as well as on the "normal" statement.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
@@ -560,7 +560,11 @@ class StatementService extends CoreService implements StatementServiceInterface
             $this->statementAttributeRepository->copyStatementAttributes($draftStatement, $originalStatement);
 
             // Create a statement copy for the assessment table
-            $assessableStatement = $this->statementCopier->copyStatementObjectWithinProcedure($originalStatement, false);
+            $assessableStatement = $this->statementCopier->copyStatementObjectWithinProcedureWithRelatedFiles(
+                $originalStatement,
+                false,
+                true
+            );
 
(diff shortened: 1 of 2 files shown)


demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
```diff
@@ -560,7 +560,11 @@ public function submitDraftStatement(
             $this->statementAttributeRepository->copyStatementAttributes($draftStatement, $originalStatement);
 
             // Create a statement copy for the assessment table
-            $assessableStatement = $this->statementCopier->copyStatementObjectWithinProcedure($originalStatement, false);
+            $assessableStatement = $this->statementCopier->copyStatementObjectWithinProcedureWithRelatedFiles(
```
jalalifardemos, line 563: What happens If draftstatement has file?
It's not better we have a condition for this situation. Like:
`if ($draftStatement->getFiles()){$assessableStatement = $this->statementCopier->copyStatementObjectWithinProcedureWithRelatedFiles(...)
}else{$assessableStatement = $this->statementCopier->copyStatementObjectWithinProcedure(...)}`

martendemos: I do not see any reason to use `copyStatementObjectWithinProcedureWithRelatedFiles()` for both cases (with file and without file) in this case of creation a new statement and his original.
If there is no file the loop over the files will just skipped.

jalalifardemos (approved):
Lgtm, Code Review. Thx

dresslerdemos (approved):

PR #2574, opened 2024-01-16
feature: enable resource property sortability by default

dresslerdemos:
feature: enable resource property sortability by default except for to-many relationships

Untested

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
lgtm, code review only. I think we could try it

PR #2575, opened 2024-01-16
feature (ref 36036): set method internIdInUsed to call in Importer an…

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T36036

Description: If user import some statements with same intern Id then should just ignore new ones. because of that I created a new method in repo to can find intern ID in used and then in spreadsheet importer called it. I should remove too, the method as an intern ID validation 

- [x] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementSpreadsheetImporter.php
```diff
@@ -176,15 +177,24 @@ public function process(SplFileInfo $workbook): void
             $originalStatementOrViolations = $builder->buildStatementAndReset();
```
martendemos, line 177: original Statement will be created nevertheless.
This will lead to OSTNs (including internId) in the DB without any way to access/delete these by the user.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementSpreadsheetImporter.php
```diff
@@ -176,15 +177,24 @@ public function process(SplFileInfo $workbook): void
...
             if ($originalStatementOrViolations instanceof Statement) {
                 $externId = $originalStatementOrViolations->getExternId();
+                $internId = $originalStatementOrViolations->getInternId(false);
                 if (array_key_exists($externId, $usedExternIds)) {
                     // skip statements with existing extern IDs
                     $this->skippedStatements[$externId] = ($this->skippedStatements[$externId] ?? 0) + 1;
 
                     continue;
                 }
+                if (array_key_exists($internId, $usedInternIds)) {
```
martendemos, line 187: internId can be null, therefore imports with more than one STN without (null) an internId, only the first STN without an internId will be imported.

martendemos (changes requested):
Two issues found.
I will try to find a solution ... :factory_worker:

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementSpreadsheetImporter.php
jalalifardemos: fix in [4e52b43](https://github.com/demos-europe/demosplan-core/pull/2575/commits/8586a664b1fd40f8f48c3106c1f375ee1eb30ef8)

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementSpreadsheetImporter.php
jalalifardemos: fix in [4e52b43](https://github.com/demos-europe/demosplan-core/pull/2575/commits/8586a664b1fd40f8f48c3106c1f375ee1eb30ef8)

martendemos:
Tests still missing. I will try to implement some.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/AbstractStatementFromRowBuilder.php
```diff
@@ -71,4 +72,8 @@ abstract public function setNumberOfAnonymVotes(Cell $cell): ?ConstraintViolatio
      */
     abstract public function buildStatementAndReset(): StatementInterface|ConstraintViolationListInterface;
 
+    public function reset(): void
+    {
+        $this->statement = new Statement();
+    }
```
dresslerdemos, line 78: I proposed making this non-abstract, but now it think it's wrong. One child class needs to implement this and the other one needs to call `$this->baseStatementFromRowBuilder->reset()`, or am I wrong?

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
```diff
@@ -19,6 +26,10 @@
 
 class StatementFromRowBuilderWithZipSupport extends AbstractStatementFromRowBuilder
 {
+    protected Cell $fileReferences;
+
+    protected Cell $originalFileReferences;
```
dresslerdemos, line 31: I think this can be `null`. Thus, `null` needs to be handled when accessing this property.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
```diff
@@ -69,18 +80,25 @@ public function setOriginalFileReferences(Cell $cell): ?ConstraintViolationListI
...
+        $this->originalFileReferences = $cell;
+
+        return null;
+    }
+
+    protected function handleFileReferences(): ?ConstraintViolationListInterface
     {
         // early return in case no file-reference is found
-        $cellValue = $cell->getValue();
+        $cellValue = $this->fileReferences->getValue();
```
dresslerdemos, line 93: `$cellValue` needs to be validated if it is actually a string. Otherwise we may trigger exceptions instead of collecting violations as intended.

demosplan/DemosPlanCoreBundle/Repository/StatementRepository.php
```diff
@@ -511,6 +511,25 @@ public function getExternIdsInUse(string $procedureId): array
...
+        $query = $this->getEntityManager()
+            ->createQueryBuilder()
+            ->select('statement.internId')
+            ->from(Statement::class, 'statement')
+            ->where('statement.procedure = :procedureId')
+            ->setParameter('procedureId', $procedureId)
+            ->getQuery();
+        $internIds = array_column($query->getScalarResult(), 'internId');
+
+        return array_combine($internIds, $internIds);
```
dresslerdemos, line 530: May the result contain `null`?

demosplan/DemosPlanCoreBundle/Validator/PrePersistUniqueInternIdConstraintValidator.php
```diff
@@ -31,11 +31,9 @@ public function validate($value, Constraint $constraint): void
...
         $value = $this->validateType($value, $constraint);
 
-        $occupiedInSavedEntities = $this->isIdOccupiedInSavedEntities($value);
         $occupiedByLoadedEntities = $this->isIdOccupiedByLoadedEntities($value->getInternId(false), $value);
-        if ($occupiedInSavedEntities || $occupiedByLoadedEntities) {
+        if ($occupiedByLoadedEntities) {
             $this->context->buildViolation($constraint->message)
-                ->setParameter('{{ internId }}', $value->getInternId())
                 ->addViolation();
         }
```
dresslerdemos, line 38: Why was this validation logic simplified.

dresslerdemos (changes requested):
Please be sure to run phpstan at level 1 on the core and additionally run phpstan level 2 and look for any of the changed classes.

demosplan/DemosPlanCoreBundle/Repository/StatementRepository.php
jalalifardemos: done in [5b62cbe](https://github.com/demos-europe/demosplan-core/pull/2575/commits/5b62cbe1599777796d3d460b18a8775cb86c82a6)

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/AbstractStatementFromRowBuilder.php
jalalifardemos: done in [e4c767a](https://github.com/demos-europe/demosplan-core/pull/2575/commits/e4c767ad2a943ab771c3fd9a37ca27accb65955e)

demosplan/DemosPlanCoreBundle/Validator/PrePersistUniqueInternIdConstraintValidator.php
jalalifardemos: What should it look like?

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
dresslerdemos: This simply skips the value. Please use [Symfony Validation](https://symfony.com/doc/current/validation.html).

demosplan/DemosPlanCoreBundle/Validator/PrePersistUniqueInternIdConstraintValidator.php
dresslerdemos: I'd prefer we keep the `$occupiedInSavedEntities` check.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
dresslerdemos: The accesses to these properties still need to handle `null` gracefully.

dresslerdemos (changes requested):
Commented in the threads above.

demosplan/DemosPlanCoreBundle/Validator/PrePersistUniqueInternIdConstraintValidator.php
jalalifardemos: done in [ac92fca](https://github.com/demos-europe/demosplan-core/pull/2575/commits/ac92f3afaa6cb926ce7e77f44f24df9dc6dc82af)

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
jalalifardemos: done in [a960b74](https://github.com/demos-europe/demosplan-core/pull/2575/commits/a960b749db2f368ae8a416e655bcd47b1e9240b1)

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
jalalifardemos: done in [5018d06](https://github.com/demos-europe/demosplan-core/pull/2575/commits/5018d06ad508b8dcd5f0812c9e3319b038ee892c)

martendemos (changes requested):
Writing on the last change I noticed that there are other cases, where an original statement will be created without the related statement.
Actually all these cases should be covered by the doctrine connection transaction, but for some reasons the statement will be persisted into the DB.
Find the reason why the transaction does not roll back to solve this issue.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/AbstractStatementFromRowBuilder.php
MoritzMandler: done in 8ae3fff923d885cba0c253fa0f60141c08909b34

martendemos (dismissed):

bakridemos:
> Writing on the last change I noticed that there are other cases, where an original statement will be created without the related statement. Actually all these cases should be covered by the doctrine connection transaction, but for some reasons the statement will be persisted into the DB. Find the reason why the transaction does not roll back to solve this issue.

we tried to reproduce this case without success, when the transaction is roll backed, no original statements or statements will be persisted

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
```diff
@@ -69,28 +90,48 @@ public function setOriginalFileReferences(Cell $cell): ?ConstraintViolationListI
         return null;
     }
 
-    public function setFileReferences(Cell $cell): ?ConstraintViolationListInterface
+    public function setOriginalFileReferences(?Cell $cell): ?ConstraintViolationListInterface
```
dresslerdemos, line 93: Like all other setter methods, there is no reason for this method to allow `null` as parameter.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
```diff
@@ -209,11 +260,33 @@ public function setFeedback(Cell $cell): ?ConstraintViolationListInterface
 
     public function setNumberOfAnonymVotes(Cell $cell): ?ConstraintViolationListInterface
     {
-       return $this->baseStatementFromRowBuilder->setNumberOfAnonymVotes($cell);
+        return $this->baseStatementFromRowBuilder->setNumberOfAnonymVotes($cell);
+    }
+
+    public function setFileReferences(?Cell $cell): ?ConstraintViolationListInterface
```
dresslerdemos, line 266: Again, as to my knowledge this method will never be called with a `null` parameter.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
```diff
@@ -209,11 +260,33 @@ public function setFeedback(Cell $cell): ?ConstraintViolationListInterface
...
     {
+        $violations1 = $this->handleFileReferences();
+        if (null !== $violations1 && 0 !== $violations1->count()) {
+            return $violations1;
+        }
+
+        $violations2 = $this->handleOriginalFileReferences();
+        if (null !== $violations2 && 0 !== $violations2->count()) {
+            return $violations2;
+        }
```
dresslerdemos, line 283: We want to show as many violations on the error page as possible. If there is no interdependence between these two methods, I suggest to execute them both and return all resulting violations.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
dresslerdemos: for history: we did need the validation

dresslerdemos (changes requested):
Beside the code concerns:

If local testing is possible I highly recommend to temporarily add a synthetic violation where the `FileContainer` is validated for testing purposes. Looking back on the previous behavior I suspect the original statement still being added to the database and thus triggering the original bug.

demosplan/DemosPlanCoreBundle/Logic/Import/Statement/StatementFromRowBuilderWithZipSupport.php
MoritzMandler: done in ae76ffe5a4f3439ed179fcd37d6cbb974151a2c2

dresslerdemos (approved):
Code review only.

PR #2582, opened 2024-01-17
feat (refs T27469): Filter by unassigned segments.

sierrazamudiodemos:
**Ticket:** https://yaits.demos-deutschland.de/T27469

Description:
It is possible to filter by unassigned segments. This needs adjustments in the FE and in the BE.
The BE uses Elastic Search with Facets to group by assignee ids. Given that filtering segments by assignee = NULL is not possible using Facets, there is the need of the FE to send a particular filter using "IS NULL".
The BE is supposed to tell FE what filter was selected based on the assignee id. Given that there is not assigneeId, a new meta field was introduced in the RPC request to let the FE know that unassigned filter was selected.

### How to review/test
- Go to Regio > select a Prcoedure (Verfahren), and then Abschnitten
- Select Bearbeiter*in filter (in each filter should be the number of segmenter corresponding to the filter)
-- Select an assignee
-- Select Nicht zugewiesen (this is the new feature)
- Click on Anwenden and see if the segments are displayed properly

- [x] Update documentation

### DIFF
--- a/client/js/components/procedure/SegmentsList/FilterFlyout.vue
+++ b/client/js/components/procedure/SegmentsList/FilterFlyout.vue
@@ -239,6 +239,8 @@ export default {
 
     /**
      * Construct the query object to be passed into vuex-json-api call.
+     * if the id is unassigned, then set the operator to IS NULL so that only segments with assignee null are taking.
+     * That is because the BE uses Elastic Search to make the filtering
      */
     filter () {
       const filter = {}
@@ -247,7 +249,7 @@ export default {
           condition: {
             path: this.path,
             value: id,
-            operator: this.operator
+            operator: id === "unassigned"? "IS NULL" : this.operator
           }
         }
       })
@@ -399,6 +401,11 @@ export default {
                 this.$set(this.itemsObject, el.id, el)
               }
             })
+
+            // If the current filter is assignee, display amount of Segments that have assignee as null. That is given by the field missingResourcesSum
+            if (result.data[0].attributes.path === 'assignee') {
+              this.$set(this.itemsObject, "unassigned", { attributes: { count: result.data[0].attributes.missingResourcesSum , label: Translator.trans('not.assigned'), ungrouped : true,  selected: result.meta['unassigned_selected']} , id: "unassigned", type: "AggregationFilterItem", ungrouped : true})
+            }
           }
         })
         .catch(err => console.log(err))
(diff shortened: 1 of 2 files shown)


client/js/components/procedure/SegmentsList/FilterFlyout.vue
```diff
@@ -247,7 +249,7 @@ export default {
           condition: {
             path: this.path,
             value: id,
-            operator: this.operator
+            operator: id === "unassigned"? "IS NULL" : this.operator
```
gruenbergerdemos, line 252:
```suggestion
            operator: id === 'unassigned' ? 'IS NULL' : this.operator
```
Please use single quotes instead of double ones.

client/js/components/procedure/SegmentsList/FilterFlyout.vue
```diff
@@ -399,6 +401,11 @@ export default {
                 this.$set(this.itemsObject, el.id, el)
               }
             })
+
+            // If the current filter is assignee, display amount of Segments that have assignee as null. That is given by the field missingResourcesSum
+            if (result.data[0].attributes.path === 'assignee') {
+              this.$set(this.itemsObject, "unassigned", { attributes: { count: result.data[0].attributes.missingResourcesSum , label: Translator.trans('not.assigned'), ungrouped : true,  selected: result.meta['unassigned_selected']} , id: "unassigned", type: "AggregationFilterItem", ungrouped : true})
+            }
```
gruenbergerdemos, line 408:
```suggestion
             if (result.data[0].attributes.path === 'assignee') {
              this.$set(this.itemsObject, 'unassigned', {
                attributes: {
                  count: result.data[0].attributes.missingResourcesSum,
                  label: Translator.trans('not.assigned'),
                  ungrouped: true,
                  selected: result.meta.unassigned_selected
                },
                id: 'unassigned',
                type: 'AggregationFilterItem',
                ungrouped: true
              })
            }
```
- Please add some line breaks for better readability
- Use single quotes instead of double ones
- You can also use dot notation for `result.meta.unassigned_selected`
- Delete spaces before comma

gruenbergerdemos (changes requested):
Just some styling issue, otherwise FE looks good and works

MoritzMandler (approved):
lgtm - and tested - works fine

gruenbergerdemos (approved):
lgtm

PR #2584, opened 2024-01-17
fix(refs T36079): fix segment can not be updated

muellerdemos:
https://yaits.demos-deutschland.de/T36079

- the api call contains the non-conform JSON API spec property `group` which will result in an invalid api call hence has to be removed

### How to review/test

- navigate to /abschnitte
- claim a segment and change data
- save

### PR Checklist


gruenbergerdemos (approved):
lgtm, code review only

PR #2585, opened 2024-01-17
fix (refs T36081): remove unnecessary quoting

dresslerdemos:
Pre-T35348 the text was passed directly into the DQL query, which required quoting it. After the corresponding fix that passes the text value as DQL parameter, this quoting is no longer necessary and in fact results in quotes added to the text itself.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


elmasdemos (approved):
lgtm, code review

jalalifardemos (approved):
lgtm, thx, Test and Code review

mussbach (approved):
lgtm, code review only

bakridemos (approved):
lgtm, code review only

PR #2587, opened 2024-01-18
chore (refs T34287): Add missing Hooks/Selectors.

ahmad-demos:
https://yaits.demos-deutschland.de/T34287

- Add missing Hooks/Selectors.


client/js/components/procedure/SegmentsList/CustomSearch.vue
```diff
@@ -41,6 +43,7 @@
             v-if="isLoading === false">
             <dp-checkbox
               v-for="({label, value}, i) in fields"
+              :data-cy="'customSearch:' + label"
```
spiess-demos, line 46: `label` may contain whitespace, so i think it will be safer to use `value` instead.

client/js/components/procedure/SegmentsList/CustomSearch.vue
```diff
@@ -61,14 +64,16 @@
           <dp-details
             v-for="explanation in explanations"
             :key="explanation.title"
-            :summary="explanation.title">
+            :summary="explanation.title"
+            :data-cy="'customSearch:' + explanation.title">
```
spiess-demos, line 68: - i think we can use es6 string literal instead of concatenation
- `customSearch` is too vague, i suggest `customSearchExplanation`
- title contains spaces, so it would be better to just add `(explanations, i)` in line 65 and append the index instead of the title here

client/js/components/procedure/SegmentsList/FilterFlyoutCheckbox.vue
```diff
@@ -10,6 +10,7 @@
 <template>
   <li>
     <input
+      :data-cy="option.attributes.label"
```
spiess-demos, line 13: may contain spaces, better use `flyoutOption:${instance}:${option.id}`.

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -31,6 +31,7 @@
           </span>
           <filter-flyout
             v-for="filter in filters"
+            :data-cy="'segmentsList:' + filter.labelTranslationKey"
```
spiess-demos, line 34:
```suggestion
            :data-cy="`segmentsListFilter:${filter.labelTranslationKey}`"
```

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -43,6 +44,7 @@
         </div>
         <dp-button
           class="ml-auto"
+          data-cy="segmentsList:reset"
```
spiess-demos, line 47:
```suggestion
          data-cy="segmentsList:resetFilter"
```

spiess-demos (changes requested):

client/js/components/procedure/SegmentsList/CustomSearch.vue
ahmad-demos: https://github.com/demos-europe/demosplan-core/pull/2587/commits/26e1e9ee847358fab7a3e1b2bab8a82f0523deb5

client/js/components/procedure/SegmentsList/CustomSearch.vue
ahmad-demos: I think we don't need hooks here!
https://github.com/demos-europe/demosplan-core/pull/2587/commits/0ac8f7ce8160a1a27e21e25300e770c0f569e60e

ahmad-demos: because here no actions, only data and infos!

ahmad-demos: I think we don't need that in E-2-E!

spiess-demos: ok, that's right, i understand. maybe we want a jest test of DpDetails in demosplan-ui, but there is no need for an e2e test here.

spiess-demos (approved):
lgtm! (code review only)

PR #2606, opened 2024-01-22
Harden file downloads

mussbach:
For improved security we should use `core_file_procedure` instead of `core_file` whenever possible. I checked the remaining calls of `core_file` and adjusted them if they are in procedure context. I tested all the cases in the interface

### How to review/test
code review or check the cases in the interface

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


salisdemos (approved):
looks legit. Thanks

PR #2607, opened 2024-01-22
chore: ensure that field is nullable as this is not the case in all projects

mussbach:
The adjusted migration failed in an existing project. We need to ensure that field is nullable.

### How to review/test
code review

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm

PR #2616, opened 2024-01-23
chore: sort addon tags during install via GitHub

mussbach:
Newest tag should be show last as the list will grow over time, and it might be inconvenient to scroll up a lot

### How to review/test
beginn to install an addon via `bin/<project> dplan:addon:install --github` and find the most recent tag in the second step below

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm - tested

PR #2617, opened 2024-01-23
bug(refs T29693): Add margin for buttons

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T29693

- Add some margin, use flex instead of floating and inline for positioning
- Use optional chaining for better robustness


sakutademos (approved):
lgtm, code review + works

PR #2620, opened 2024-01-24
Fix(refs:36153): use the correct access for the dates

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36153

### Description: 
- the dates were incorrect
- they showed only the same current date always 

### PR Checklist


ahmad-demos (approved):
lgtm, only codeReview!

PR #2621, opened 2024-01-24
Fix(refs:T36142): no column title shown

elmasdemos:
***Ticket***: https://yaits.demos-deutschland.de/T36142

### Description: 

- the title of the columns are not shown
- the headerData is always empty and doesnt has to be used here. the headerFields have field and value, which we need for the header of the columns
- some code beautify due to attribute and methods order

### PR Checklist


client/js/components/user/DpMasterToebList.vue
```diff
@@ -593,6 +532,67 @@ export default {
...
+      })
+
+      // Apply filters and search term
+      sortedList = sortedList.filter(item => this.searchedItems[item.ident] === 'in' && this.filteredItems[item.ident] === 'in')
+
+      this.rowItems = sortedList
+      this.$nextTick(() => {
+        if (this.isEditable) this.addCellIdsAndFields(this.dataTableElement.getElementsByTagName('tbody')[0], this.onPageItems)
+      })
+    },
```
muellerdemos, line 589: I would reduce the cognitive load here a little bit by extracting some parts of this function into own functions. This separates the different logical steps in this operation and improves the readability. Also we can use the `toSorted` array method which also returns a new array and that way we do not need to re-declare the `sortedList` variable.

```suggestion
 sortCurrentItems (items) {
      items.toSorted((a, b) => {
        if (a[this.sortOrder.key] === null || typeof a[this.sortOrder.key] === 'undefined') {
          return this.sortOrder.direction * 1
        }
        
        if (b[this.sortOrder.key] === null || typeof b[this.sortOrder.key] === 'undefined') {
          return this.sortOrder.direction * -1
        }

        return String(a[this.sortOrder.key])
          .localeCompare(String(b[this.sortOrder.key]), 'de', { sensitivity: 'base' }) * this.sortOrder.direction
      })   
    },
    
    updateFields () {
      const sortedItems = this.sortCurrentItems(this.currentItems)
      const filteredSortedItems = sortedItems.filter(item => this.searchedItems[item.ident] === 'in' && this.filteredItems[item.ident] === 'in')

      this.rowItems = filteredSortedItems
      this.$nextTick(() => {
        if (this.isEditable) this.addCellIdsAndFields(this.dataTableElement.getElementsByTagName('tbody')[0], this.onPageItems)
      })
    },
```

client/js/components/user/DpMasterToebList.vue
```diff
@@ -355,37 +404,20 @@ export default {
       this.updateFields()
     },
 
+    possiblyInsertDeletion (items) {
+      let returnItems = items
+      if (this.isEditable) {
+        returnItems = returnItems.map(item => ({ ...{ deletion: 'delete' }, ...item }))
+      }
+      return returnItems
```
muellerdemos, line 412: this can be simplified:
```suggestion
      if (this.isEditable) {
        return items.map(item => ({ ...{ deletion: 'delete' }, ...item }))
      }
```

muellerdemos (changes requested):
I'd simplify certain areas of this PR

client/js/components/user/DpMasterToebList.vue
```diff
@@ -453,95 +450,37 @@ export default {
...
-      })
+    transformValue (value, field) {
+      let returnVal = value
+      if (value === null || typeof value === 'undefined' || value === false) {
+        returnVal = '-'
+      }
+      if (this.boolToStringFields.includes(field)) {
+        returnVal = value ? 'x' : ''
+      }
+      return returnVal
```
muellerdemos, line 483: this can also be simplified:
```suggestion
      if (value === null || typeof value === 'undefined' || value === false) {
        return '-'
      }
      if (this.boolToStringFields.includes(field)) {
        return value ? 'x' : ''
      }
```
And one logical question:
Why does value be checked for `null`, `undefined` and `false`? Why can this variable have so many states?

client/js/components/user/DpMasterToebList.vue
```diff
@@ -593,6 +532,67 @@ export default {
...
       }
+    },
+
+    updateOrga (ident, updatedField, status) {
+      this.updatedItems = this.updatedItems.map(item => {
+        let newItem = item
+        if (item.ident === ident) {
+          newItem = { ...newItem, ...updatedField }
+        }
+        return newItem
```
muellerdemos, line 543: This can be simplified:
```suggestion
        if (item.ident === ident) {
          return { ...newItem, ...updatedField }
        }
        return item
```

client/js/components/user/DpMasterToebList.vue
```diff
@@ -593,6 +532,67 @@ export default {
...
+      })
+
+      this.rowItems = this.rowItems.map(item => {
+        let newItem = item
+        if (item.ident === ident) {
+          newItem = { ...newItem, ...updatedField }
+        }
+        return newItem
+      })
```
muellerdemos, line 552: this function is nearly the same as the one above, have you considered extracting this in its own function and use the return of the new function on both instances?

muellerdemos (changes requested):

client/js/components/user/DpMasterToebList.vue
```diff
@@ -87,18 +87,18 @@
...
                   checked
+                  type="checkbox"
                   @change="toggleAllCols">
                 {{ Translator.trans('aria.select.all') }}
               </label>
               <label
                 v-for="(filterField, idx) in filteredFields"
-                class="lbl--text u-pl-0_5 u-mb-0_25 w-10"
-                :key="idx">
+                :key="idx"
```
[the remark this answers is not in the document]
gruenbergerdemos: Done in [e394069](https://github.com/demos-europe/demosplan-core/pull/2621/commits/e3940695951045a3926ebfaebec1770b137d7c73)

client/js/components/user/DpMasterToebList.vue
gruenbergerdemos: Thanks for the suggestion. Done in [5a95554](https://github.com/demos-europe/demosplan-core/pull/2621/commits/5a95554434ea9f4b918f1cd4be78f63a9c2312bc). 
I changed some things:
- `sortItems` instead of `sortCurrentItems`: The function just sorts some items it is given. In our case they are the current items, but it could be any items.
- We don't need the `toSorted` function in the `sortItems` function, because we do nothing with the param items. The `this.currentItems` are not altered. But thanks for the reminder that this function exists now :)
- Removed the `filteredSortedItems`, because it was an unnecessary step

client/js/components/user/DpMasterToebList.vue
gruenbergerdemos: Thanks, done in [bd09859](https://github.com/demos-europe/demosplan-core/pull/2621/commits/bd09859369b585d08b71b93f350cb84aeb0534ba)

client/js/components/user/DpMasterToebList.vue
gruenbergerdemos: Done in [3ee5419](https://github.com/demos-europe/demosplan-core/pull/2621/commits/3ee541977317297a5722adfe38cd0dd1d04fbea0)
I don't know exactly why there are so many states, we get it like that from the backend. I guess it is `null` if it is not set in the backend and it can be `false` for boolean values. `undefined` maybe if something went wrong in the frontend?

gruenbergerdemos: I realized we have to check for `boolToStringFields` first, because there should be displayed ' ' for `false` and otherwise there is '-' displayed. Done in [1b1fe82](https://github.com/demos-europe/demosplan-core/pull/2621/commits/1b1fe82b0f98518dfe1243d3bb039d040fa488e8)

client/js/components/user/DpMasterToebList.vue
gruenbergerdemos: Done with [c706ee7](https://github.com/demos-europe/demosplan-core/pull/2621/commits/c706ee785dffafea8563a6127954102775662818)

client/js/components/user/DpMasterToebList.vue
gruenbergerdemos: I extracted it in a new function `updateItems` in [c706ee7](https://github.com/demos-europe/demosplan-core/pull/2621/commits/c706ee785dffafea8563a6127954102775662818)

client/js/components/user/DpMasterToebList.vue
muellerdemos: thank you for your efforts here, this looks really clean now :heart:

client/js/components/user/DpMasterToebList.vue
muellerdemos: thank you for investigating and improving this !

muellerdemos (approved):
lgtm - thanks a lot for picking up this PR and also putting in so much effort. Nice work imho

PR #2625, opened 2024-01-24
Select checkbox by default if only 1 boilerplate category is present

spiess-demos:
To improve usability, the checkbox to chose which categories to display a boilerplate in is checked by default now, if only one category exists.

### How to review/test
In projects with only one boilerplate category enabled, create a new boilerplate entry. The checkbox should be preselected.
When editing boilerplates, the checkbox should reflect the state saved before.

### PR Checklist


mussbach (approved):
lgtm, code review only

PR #2626, opened 2024-01-24
chore: modernize frontcontrollers to match symfony 5.x

mussbach:
Reimplement changes of https://github.com/symfony/recipes/pull/724/files to our customized frontcontrollers

### How to review/test
Browsercalls as well as commands should work as before

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Application/FrontController.php
+++ b/demosplan/DemosPlanCoreBundle/Application/FrontController.php
@@ -16,6 +16,7 @@ use demosplan\DemosPlanCoreBundle\Addon\AddonAutoloading;
 use demosplan\DemosPlanCoreBundle\Logic\HttpCache;
 use demosplan\DemosPlanCoreBundle\Utilities\DemosPlanPath;
 use Exception;
+use LogicException;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\ErrorHandler\Debug;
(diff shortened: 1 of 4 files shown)


demosplan/DemosPlanCoreBundle/Application/FrontController.php
```diff
@@ -76,21 +58,41 @@ public static function bootstrap(): void
...
+            echo 'Warning: The console should be invoked via the CLI version of PHP, not the '.PHP_SAPI.' SAPI'.PHP_EOL;
+        }
+
         set_time_limit(0);
 
+        require DemosPlanPath::getRootPath('vendor/autoload.php');
+
+        if (!class_exists(ConsoleApplication::class) || !class_exists(Dotenv::class)) {
+            throw new LogicException('You need to add "symfony/framework-bundle" and "symfony/dotenv" as Composer dependencies.');
+        }
```
graupnerdemos, line 71: We technically don't need this as this FrontController should not be available outside of a demosplan-core with all dependencies installed. However, we should keep it to make diffing against the Symfony baseline easier.

graupnerdemos (commented):
Is there any specific reason why you opted to manually load DotEnv and the Addon registry in `console` but used `bootstrap` in `web`? 

I can't quite decide what the best solution might be here. Maybe, we rename `bootstrap` to `dplanInit` and – for now – only leave the Addon registration in there. That way, the front controller functions could remain more or less identical to the official ones except for this one extra method to be called.

demosplan/DemosPlanCoreBundle/Application/FrontController.php
mussbach: this is exactly my intention to keep it here

mussbach:
Bootstrap is used twice and I did not want to duplicate the logic

graupnerdemos:
Ah, I see, it's also used in the PHPUnit bootstrap. In that case, can we not also use it in `web()`? Doesn't change logic, but is a nice repetition reduction refactoring.

graupnerdemos (approved):
lgtm. thanks.

PR #2631, opened 2024-01-25
chore: remove useless and misleading Interface

mussbach:
`DraftStatementListFilters` should not implement `ParameterBagInterface` and only use get and has as this is misleading and might even confuse dependency injection

### How to review/test
call to draft statement list should work as before

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm, code review only

PR #2637, opened 2024-01-26
F t35807 no feature for sorting planning documents

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T35807

### Description: 

- the planning categories and their subgroups are not sorted
- actually its not a bug, its a feature, because sorting was not implemented for planning categories and their subgroups

### PR Checklist

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
@@ -2243,6 +2243,10 @@ class StatementService extends CoreService implements StatementServiceInterface
 
         $noManualSortElementsIds = $this->serviceElements->getHiddenElementsIdsForProcedureId($procedureId);
 
+        $this->statementEntityGrouper->sortSubgroupsAtAllLayers(
+            $group,
+            new TitleGroupsSorter()
+        );
         // sorting only needed if there are elements to be moved to the end
         if (0 < count($noManualSortElementsIds)) {
             // sort hidden elements to end: sort elements not manually sortable in the admin list (because hidden) at the end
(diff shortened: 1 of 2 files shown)


demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
```diff
@@ -2243,6 +2243,10 @@ public function createElementsGroupStructureBobHH(string $procedureId, array $st
 
         $noManualSortElementsIds = $this->serviceElements->getHiddenElementsIdsForProcedureId($procedureId);
 
+        $this->statementEntityGrouper->sortSubgroupsAtAllLayers(
+            $group,
+            new TitleGroupsSorter()
+        );
```
dresslerdemos, line 2249: I was under the impression that at least the "Kapitel" are to be sorted by title, which is not the case if we look at the bug ticket closely. They are to be displayed with the order defined inside their corresponding planning category detail view.

dresslerdemos (changes requested):

dresslerdemos:
With https://github.com/demos-europe/demosplan-core/pull/2637/commits/2cccf6babc1b8744d7128b29c84957a6ed65d0a8 I've added an approach that should be able to carry the sorting of paragraphs over into the group structure. Please test if this solves the bug.

dresslerdemos:
If the fix works, please have an eye on the performance. If Doctrine accesses the database for each `$this->paragraphService->getParaDocumentObject($paragraphId)` this may be quite slow.

breierDemos (approved):
I tested with the related procedure from the ticket. Seems to work!

jalalifardemos (approved):
lgtm, Code review only, Thanks

dresslerdemos (approved):
This approve is only to not block the merge. No review was done by me, because too much code is by myself.

PR #2638, opened 2024-01-26
feature (ref 35807): change class KeysAtEndSorter to find excatly pos…

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T35807

Description: reset KeysAtEndSorter to create right sort of elements. That problem appeared after 
`if ($noManualSortElementsIds) {
            // sort hidden elements to end: sort elements not manually sortable in the admin list (because hidden) at the end
            // the sorting is applied to all layers but only the elements layer will be affected as
            // the other two layers do not contain groups with element IDs
            $this->statementEntityGrouper->sortSubgroupsAtAllLayers(
                $group,
                new KeysAtEndSorter($noManualSortElementsIds)
            );
        }`
because of that, we should change the structure of this class. 

### How to review/test
In UI or Code review

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/AssessmentTable/KeysAtEndSorter.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/AssessmentTable/KeysAtEndSorter.php
@@ -46,13 +46,15 @@ class KeysAtEndSorter implements ArraySorterInterface
         if (0 === count($array)) {
             return $array;
         }
-        foreach ($this->keys as $key) {
-            if (array_key_exists($key, $array)) {
-                $toBePlacedLast = $array[$key];
-                unset($array[$key]);
-                $array[$key] = $toBePlacedLast;
-            }
-        }
+        // Sort array keys based on their position in $this->keys
+        uksort($array, function ($a, $b) {
+            $posA = array_search($a, $this->keys);
+            $posB = array_search($b, $this->keys);
+
+            // If both keys are in $this->keys, compare their positions
+            // If only one key is in $this->keys, prioritize moving it to the end
+            return ($posA && $posB) ? $posA - $posB : ($posB ? -1 : 0);
+        });
 
         return $array;
     }
--- a/demosplan/DemosPlanCoreBundle/Logic/AssessmentTable/KeysAtEndSorter.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/AssessmentTable/KeysAtEndSorter.php
@@ -12,7 +12,6 @@ namespace demosplan\DemosPlanCoreBundle\Logic\AssessmentTable;
 
 use demosplan\DemosPlanCoreBundle\Exception\InvalidArgumentException;
 use TypeError;
-use function array_key_exists;
 
 /**
  * Class KeysAtEndSorter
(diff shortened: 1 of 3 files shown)


demosplan/DemosPlanCoreBundle/Logic/AssessmentTable/KeysAtEndSorter.php
```diff
@@ -46,13 +45,15 @@ public function sortArray(array $array): array
...
-        }
+        // Sort array keys based on their position in $this->keys
+        uksort($array, function ($a, $b) {
+            $posA = array_search($a, $this->keys);
+            $posB = array_search($b, $this->keys);
+
+            // If both keys are in $this->keys, compare their positions
+            // If only one key is in $this->keys, prioritize moving it to the end
+            return ($posA && $posB) ? $posA - $posB : ($posB ? -1 : 0);
+        });
```
dresslerdemos, line 56: It seems like this will ensure the order of keys is carried over into the array, while breaking the intended use to sort all matching keys to the very end of the array.

demosplan/DemosPlanCoreBundle/Logic/Document/ElementsService.php
```diff
@@ -190,11 +190,13 @@ public function getHiddenElementsIdsForProcedureId(string $procedureId): array
...
 
         // return IDs only:
-        return collect(array_merge($mapCategories, $hiddenByConfigCategories))->map(
-            fn ($element) =>
-                /* @var Elements $element */
+        return collect(array_merge($mapCategories, $hiddenByConfigCategories))
+            ->map(
+                fn ($element) =>
+                    /* @var Elements $element */
                 $element->getId()
```
dresslerdemos, line 197: While pairing we determined the need to sort the element keys by their corresponding element titles, which was done [here](https://github.com/demos-europe/demosplan-core/pull/2637/files). Is that no longer needed? This PR now conflicts with the linked one. My understanding is that this ensured that hidden elements are sorted alphabetically at the very end of their group.

dresslerdemos (changes requested):
Could you please describe what behavior you're trying to achieve? I can't dig out all the requirements that accumulated over the years, but [here](https://yaits.demos-deutschland.de/T10646) is an example where the ticket specifies that the "Planzeichnung" is to be sorted at the end (like the other hidden planning categories). [Here](https://yaits.demos-deutschland.de/T10346#210482) is the requirement that beside the special requirements for specific items like the hidden ones, the manual sorting is decisive for the items, not their title.

> Sortierung auf einer Ebene ergibt sich nicht aus den Namen (alphabetisch) sondern aus der manuellen Reihenfolge der jeweilgen Gruppe (Schlagworte/Planunterlagen/Kapitel)

jalalifardemos:
I didn't know which kind of sort, we should use. I thought they should be sorted with titles. PR works good with this goal, but we have another structure for that, therefore I don't need to change the structure of KeysAtEndSorter.

jalalifardemos:
> Sortierung auf einer Ebene ergibt sich nicht aus den Namen (alphabetisch) sondern aus der manuellen Reihenfolge der jeweilgen Gruppe (Schlagworte/Planunterlagen/Kapitel)

I didn't know which kind of sort, we should use. I thought they should be sorted with titles. PR works good with this goal, but we have another structure for that, therefore I don't need to change the structure of KeysAtEndSorter. Then I close this PR and we continue with https://github.com/demos-europe/demosplan-core/pull/2637

PR #2641, opened 2024-01-29
fix(T36199): display notification when a network error occurs

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T36199

**Description:** The above message was not displayed [due to a refactoring of externalApi](https://github.com/demos-europe/demosplan-ui/commit/3a2b55a0d00c29d76fe26bcac2fddfd34c5f3ea1). Previously, the Axios library was used, but now Fetch is used. Fetch has different error handling. The error object does not have additional parameters like 'code' or 'type'.

If the promise is rejected, the error occurs in the catch() block. The error object has a message property that says: _TypeError: Failed to fetch_. This means that the Fetch request failed, but it could have various reasons, including CORS issues.

This means that we cannot directly determine whether it is CORS or other network problems. So display the notification every time an error occurs in the catch block.


muellerdemos (approved):
lgtm

PR #2642, opened 2024-01-29
bug(refs T18750): Add min height for editor p tags

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T18750

We want empty p tags inside the editor to have the same height as non empty ones for the sake of presentation.

### How to review/test
Verfahren -> Abwägungstabelle -> edit statemente text -> Add some empty lines and save -> the empty lines should be displayed


muellerdemos (approved):
lgtm

PR #2644, opened 2024-01-29
fix (refs T34551): implement customer delete command.

MoritzMandler:
**Ticket:** https://yaits.demos-deutschland.de/T34551

Description: 
- add customer delete command: 
- store logic inside the CustomerDeleter

### How to review/test

Try to remove a customer using the following command:

- Procedure deletion: 
bin/<projectname> dplan:customer:delete

Note: you can use:
- the 'dry-run' option to initiates a dry run with verbose output to see what would happen.
- the 'without-repopulate' to ignores repopulating the ES. This should only be used for debugging purposes.


bakridemos (approved):
lgtm, code review und tested too

PR #2648, opened 2024-01-30
bug(refs T36230): check response api call

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36230

Add a check response for DpApi call, so the user does not get a validation message when the api call was not succesfull.

### How to review/test
Verfahren -> Aktuelles -> Toggle Status -> There should be an error message if the api call was not successfull


muellerdemos (approved):
lgtm

breierDemos (commented):
target branch not release?

PR #2651, opened 2024-01-30
feature (ref 36221): create relation between entitis to can remove or…

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T36221

Description: Set Relation between 

- Department and DraftStatement and DraftStatementVersion
- Orga and Statement

and update masterToebRepo to can remove connected entity

### How to review/test
in UI or Code Review

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (changes requested):
I see your intention, but I would solve this differently. We should not silently delete the (Draft)Statements but show an error message in the interface during merge instead. We should not give too much time in this one though, as this is no real case and can only occur on test instances

jalalifardemos:
this only occurs during testing, it can't happen on Prod

PR #2652, opened 2024-01-30
fix(T36196): procedure phase with write permission should have auto-switch activated by default

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T36196

**Description:** Procedure phases with the write permission should have auto-switch activated by default. Here is a list of the [procedure phases](https://github.com/demos-europe/demosplan-project-blp/blob/main/app/Resources/DemosPlanCoreBundle/config/procedure/procedurephases.yml) in blp project.

[Thread](https://demosdeutschland.slack.com/archives/C8M7GFXTP/p1706551936339009) in Slack

### DIFF
--- a/client/js/components/procedure/basicSettings/AutoSwitchProcedurePhaseForm.vue
+++ b/client/js/components/procedure/basicSettings/AutoSwitchProcedurePhaseForm.vue
@@ -196,7 +196,16 @@ export default {
      * @return {boolean}
      */
     isParticipationPhaseSelected () {
-      const participationPhases = ['participation', 'earlyparticipation', 'anotherparticipation']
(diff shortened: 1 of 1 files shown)


spiess-demos (approved):
in an ideal world we would pass all possible procedurePhases into the component via a twig variable like [here](https://github.com/demos-europe/demosplan-core/blob/6d11afd0661f8c7f5d72338e31efdf9ac7364595/templates/bundles/DemosPlanCoreBundle/DemosPlanAssessmentTable/DemosPlan/dhtml/v1/assessment_table_new_statement.html.twig#L33-L34), or we'd get it via an api call. However as this is a fix, it looks good enough.

PR #2653, opened 2024-01-30
feat: allow local addon development

mussbach:
To improve the DX when developing addons this PR adds the possibility to install an addon directly from the repository sources.

To allow that the git repository needs to be checked out in a folder named `addonDev` in the root of dplan. It could be installed then via `bin/<project> dplan:addon:install --develop`. Any repositories that exist in the folder are installable.

You could change the files directly in your dplan ide. To build the addon assets you may use `bin/<project> dplan:addon:build-frontend`. The build is much faster than during "normal" addon installation, as the assets are build in dev mode.

### How to review/test
see description

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Command/Addon/AddonInstallFromZipCommand.php
```diff
@@ -503,4 +517,32 @@ function ($tag) {
 
         return $this->getHelper('question')->ask($input, $output, $question);
     }
+
+    private function getPathFromLocalDevelopment(InputInterface $input, SymfonyStyle $output, mixed $folder): string
+    {
+        $fs = new Filesystem();
+        $addonDevFolder = DemosPlanPath::getRootPath('addonDev');
+        if (!file_exists($addonDevFolder)) {
+            throw new RuntimeException("No folder {$folder} found. To develop addons locally, create a folder {$folder} and put your addons in there.");
```
breierDemos, line 526: Maybe stranger things on my system happen, but I understand the code right that if I don't have a addonDev folder then this message will be printed to my console with addonDev instead of $folder right? During testing I had only the addonZips folder with four zips in it. But no addons and addonDev folder. So I expected this message: "No folder addonDev found. To develop addons locally, create a folder addonDev and put your addons in there.", Instead I had this message: "No folder addonZips found. To develop addons locally, create a folder addonZips and put your addons in there."

demosplan/DemosPlanCoreBundle/Command/Addon/AddonInstallFromZipCommand.php
```diff
@@ -503,4 +517,32 @@ function ($tag) {
...
+        }
+
+        $localAddons = glob($addonDevFolder.'/*');
+        if (!is_array($localAddons) || 0 === count($localAddons)) {
+            throw new RuntimeException("No local addons found in folder {$folder}. Please check out the demosplan-addon-* repositories into this folder.");
```
breierDemos, line 531: The same happens after I created a empty addonDev folder. Have you an Idea why the addonZips folder name is used?

breierDemos (changes requested):
lgtm but a strange thing happened to me

demosplan/DemosPlanCoreBundle/Command/Addon/AddonInstallFromZipCommand.php
mussbach: did you provide `--develop` to the command @breierDemos? Looks as if you where in the "wrong" if branch

spiess-demos (approved):
Lgtm & works, from a non-php users perspective.

demosplan/DemosPlanCoreBundle/Command/Addon/AddonInstallFromZipCommand.php
breierDemos: @mussbach, yes I used bin/ewm dp:addon:install --develop. I think I know why this happen. With `$this->addOption('folder', '', InputOption::VALUE_REQUIRED, 'Folder to read addon zips from', 'addonZips');`  the default of `$folder` is "addonZips". So the default will be used.

breierDemos (approved):
lgtm

PR #2655, opened 2024-01-30
bug(refs T36246): Allow char start and char end to be the same

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36246

While checking for overlapping segments we need to exclude the start and end char, because they can be the same for different segments.


ahmad-demos (approved):
lgtm, only codeReview!

PR #2656, opened 2024-01-30
fix(refs T36221): throw error if statements present

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T36221

Orgas should only be merged if no statement has been submitted from them. As this will anyway not happen in prod instances (only when testing), a simple error message is shown in that case, instead of the "Ein Fehler ist aufgetreten"-Page.

image

### How to review/test
- When trying to merge an organisation, which has already submitted statements (or has saved draftStatements), the above error message should show up. 
- When merging orgas, which do not have statements or draftStatements attached, everything should work as before.

### PR Checklist


jalalifardemos (approved):
lgtm, thx, Code review only

PR #2662, opened 2024-01-31
fix (refs T30474): reduce inconsistent code duplication

dresslerdemos:
Enhance the `OwnsProcedureConditionFactory` implementation, so that it can not only create conditions to match unknown users against a known procedure, but also to match unknown procedures against a known user. Logically the comparison is the same in both cases, but their implementation differs significantly.

Keeping the implementations as close together as possible avoids divergences, like were already present in
`ProcedureService::getAdminProcedureConditions`. The matching logic in this method behaved very similar to the one executed via `ProcedureAccessEvaluator::isOwningProcedure` (which uses `OwnsProcedureConditionFactory`), but was not exactly the same, causing a bug. Hence, `isOwningProcedure` was selected as "truth" and the logic in `getAdminProcedureConditions` was dropped.

**Ticket:** https://yaits.demos-deutschland.de/T30474

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
@@ -31,13 +31,13 @@ use Webmozart\Assert\Assert;
  *
  * This class can be either used to create conditions based on a **given user, to evaluate arbitrary procedure
  * instances**, or to create conditions based on a **given procedure, to evaluate arbitrary user instance**. Logically
- * these two cases require comparison of the same properties, but their implementation differs significantly. To
+ * these two cases require comparisons of the same properties, but their implementation differs significantly. To
  * avoid divergence in their logic, the implementations of both cases are kept as close together as possible in this
  * class and its methods.
  *
- * If a {@link Procedure} instance is given in the constructor, then the returned conditions must be used to
+ * If a {@link Procedure} instance is given in the constructor, then the returned conditions must only be used to
  * fetch/evaluate {@link User} instances. If a {@link User} instance is given in the constructor, then the returned
- * conditions must be used to fetch/evaluate {@link Procedure} instances.
+ * conditions must only be used to fetch/evaluate {@link Procedure} instances.
  */
 class OwnsProcedureConditionFactory
 {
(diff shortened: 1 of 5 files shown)


demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
```diff
@@ -13,120 +13,154 @@
...
-                $userIsAuthorized,
-                $planningAgencyIsAuthorized
-            );
-        }
-
-        return $orgaOwnsProcedure;
+        // T8427: allow access by manually configured users if the config is set to `true`,
+        // overwriting the organisation-based access
+        return $this->globalConfig->hasProcedureUserRestrictedAccess()
+            ? $this->userIsExplicitlyAuthorized()
```
mussbach, line 97: Isnt in the true case here now the permissions for planningAgencies ditched? Now only authorized users are checked, not anymore whether planning agency is permitted to procedure, if I see it correctly

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
```diff
@@ -13,120 +13,154 @@
...
+            : $this->userOwnsProcedureViaOrgaOfUserThatCreatedTheProcedure();
     }
 
     /**
-     * Returns a condition to match users having the roles to theoretically own a procedure.
+     * Returns a condition to match users having the roles in the given customer to theoretically own a procedure.
      *
-     * @return FunctionInterface<bool>
+     * @return list<FunctionInterface<bool>>
```
mussbach, line 104: is this correct?

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
```diff
@@ -13,120 +13,154 @@
...
-                    Role::HEARING_AUTHORITY_ADMIN,
-                    Role::HEARING_AUTHORITY_WORKER,
-                ],
-                ['roleInCustomers', 'role', 'code']
-            );
+            $ownsOrgaRoleCondition = [
+                $this->conditionFactory->propertyHasAnyOfValues($relevantRoles, ['roleInCustomers', 'role', 'code']),
+                $this->isUserInCustomer($customer),
+            ];
+        } else {
```
mussbach, line 130: You can avoid this else by setting this as default value of the variable

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
```diff
@@ -13,120 +13,154 @@
...
             // ist es ein PLanungsbüro?
-            $planningAgencyOwnsProcedure = $this->conditionFactory->propertyHasValue(
-                Role::PRIVATE_PLANNING_AGENCY,
-                ['roleInCustomers', 'role', 'code']
-            );
+            $planningAgencyOwnsProcedure = [
+                $this->conditionFactory->propertyHasValue($relevantRole, ['roleInCustomers', 'role', 'code']),
+                $this->isUserInCustomer($customer),
+            ];
+        } else {
```
mussbach, line 162: you can avoid this else

demosplan/DemosPlanCoreBundle/Logic/ProcedureAccessEvaluator.php
```diff
@@ -87,51 +100,61 @@ public function isOwningProcedure(User $user, Procedure $procedure): bool
     }
 
     /**
+     * Compiles a list of conditions to check if a procedure or procedure template is owned by the given user.
+     *
+     * A procedure is owned a user if all conditions in the returned list match the procedure/user respectively.
```
mussbach, line 105:
```suggestion
     * A procedure is owned by a user if all conditions in the returned list match the procedure/user respectively.
```

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
```diff
@@ -135,11 +169,91 @@ public function hasPlanningAgencyRole(): FunctionInterface
...
+
+        return $this->conditionFactory->propertyHasValue($procedure->getOrgaId(), ['orga', 'id']);
+    }
+
+    /**
+     * @return ClauseFunctionInterface<bool>
+     *
+     * @throws PathException
+     */
+    public function isUndeletedProcedure(): ClauseFunctionInterface
```
mussbach, line 247: Undeleted sounds a bit strange here. Perhaps `isNotDeleted`?

mussbach (changes requested):
Zwei Herzen ach.. I absolutely see where you come from and this might well be a good Idea but it is quite uncommon to pass two such different Objects into the check. In this case this might be a very good possibility and let's try it, but I do not dare to predict how we look at this in one or two years.
Some remarks, but basically fine though it is very hard to review

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
dresslerdemos: Yes that was one of the weird inconsistencies that was noticed some time (about a year?) ago but not dared to be touched. I tried to confirm the intended behavior with Andrea via Slack. The "new" behavior (it was already implemented in a different place) will still authorize users via their planning agency (`$this->isAuthorizedViaPlanningAgency();`), but
1. to be authorized, users must have the `PRIVATE_PLANNING_AGENCY` role
2. the authorization via planning agency is active, regardless of the `hasProcedureUserRestrictedAccess` global config setting.

Again this is not really new, but consolidated behavior and so far makes sense to me logically.

How planning agency are authorized can be found by searching for the calls of `isAuthorizedViaPlanningAgency`.

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
dresslerdemos: `list` is a type introduced by phpstan and as far as I know accepted by PHPStorm. It denotes an array of values whose keys are integers starting with `0` that increase without holes.

`FunctionInterface<bool>` is "good enough". There are inconsistencies when to use `FunctionInterface` vs `ClauseFunctionInterface` (the latter extending the former). There is a clear solution for that, but it requires some refactoring which execution needs to be planned and approved first.

dresslerdemos: `list` is a type introduced by phpstan and as far as I know accepted by PHPStorm. It denotes an array of values whose keys are integers starting with `0` that increase without holes.

`FunctionInterface<bool>` is "good enough". There are inconsistencies when to use `FunctionInterface` vs `ClauseFunctionInterface` (the latter extending the former). There is a clear solution for that, but it requires some refactoring which execution needs to be planned and approved first.

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
dresslerdemos: I'm fine with that for simple cases like `null` or primitive types as default, but here a method is called and an object would be created, which is instantly thrown away if the `if` branch is entered. This does not seem like a good design choice, as a programmer not familiar with this code may feel inclined to check for side effects in the logic of `$this->conditionFactory->false()` even though he is interested in the `if` case only.

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
dresslerdemos: https://github.com/demos-europe/demosplan-core/pull/2662/files#r1478637156

demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
dresslerdemos: I renamed it to `isNotDeletedProcedure`, to keep the clarity that this checks the deletion of the procedure, not the user. e54af829944d655083c2a208ead4fcc931ff6bbd

dresslerdemos:
> quite uncommon to pass two such different Objects into the check

My initial idea was to use inheritance from an common `OwnsProcedureConditionFactoryInterface` and have one implementation taking a `User` and another implementation taking a `Procedure`. But even though this would have added a connection via the same parent methods in the interface, I reckoned the likelyhood for changes in only one class was too great (leave alone the complexity of three instead of one class).

One thing that could happen in the future would be that both procedure and user are variable, to fetch all users from the database that own *any* procedure in the current customer. This would introduce a third case in each method, `$userOrProcedure` being `null`.

mussbach (approved):
concerns are resolved, lgtm, code review only

PR #2668, opened 2024-01-31
Adapt xhr call to changed lib

spiess-demos:
Ticket: https://yaits.demos-deutschland.de/T36261

As we changed dpApi to use fetch under the hood, usage must be adapted.

### How to review/test
The search should work again on the public index page.


sakutademos (approved):
lgtm, code review + works

PR #2674, opened 2024-02-01
bug(refs T36258): Add 'br' tag to allowed nodes for prosemirror

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36258

We also need to transform 'br' tags to prosemirror positions. Additionally exclude 'br' from markRegex, because otherwise it will be included due to 'b' tag and it will be transformed double (once as node and once as mark).

### How to review/test
Cannot be tested locally


gruenbergerdemos:
Already done with https://github.com/demos-europe/demosplan-core/pull/2675/

PR #2675, opened 2024-02-01
bug(refs T36258): fix slicing tagging

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36258 
**Ticket:** https://yaits.demos-deutschland.de/T36246

- While checking for overlapping segments we need to exclude the start and end char, because they can be the same for different segments.
- Add 'br' tag to allowed nodes for prosemirror. We also need to transform 'br' tags to prosemirror positions. 
- Additionally exclude 'br' from markRegex, because otherwise it will be included due to 'b' tag and it will be transformed double (once as node and once as mark).

### How to review/test
Cannot be tested locally

### DIFF
--- a/client/js/store/procedure/SplitStatementStore.js
+++ b/client/js/store/procedure/SplitStatementStore.js
@@ -230,8 +230,8 @@ const SplitStatementStore = {
               for (let j = i + 1; j < segments.length; j++) {
                 // Check for overlap
                 if (
-                  (segments[i].charStart >= segments[j].charStart && segments[i].charStart <= segments[j].charEnd) ||
-                  (segments[j].charStart >= segments[i].charStart && segments[j].charStart <= segments[i].charEnd)
+                  (segments[i].charStart > segments[j].charStart && segments[i].charStart < segments[j].charEnd) ||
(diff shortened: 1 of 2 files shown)


client/js/store/procedure/SplitStatementStore.js
```diff
@@ -230,8 +230,8 @@ const SplitStatementStore = {
               for (let j = i + 1; j < segments.length; j++) {
                 // Check for overlap
                 if (
-                  (segments[i].charStart >= segments[j].charStart && segments[i].charStart <= segments[j].charEnd) ||
-                  (segments[j].charStart >= segments[i].charStart && segments[j].charStart <= segments[i].charEnd)
+                  (segments[i].charStart > segments[j].charStart && segments[i].charStart < segments[j].charEnd) ||
```
spiess-demos, line 233: Curious question: Why do we loop twice here?

spiess-demos (approved):
lgtm (code review only)

client/js/store/procedure/SplitStatementStore.js
gruenbergerdemos: Because we want to compare all segments in case they are not ordered. Without a nested loop you can only compare one segment with the next one.

PR #2679, opened 2024-02-02
fix: fix phpunit test after refactoring

dresslerdemos:
The concept of `PropertyDefinition`s and `IncludeDefinition`s was previously improved on in the corresponding dependency and these types deleted. This warrants a couple of adjustments in the `DynamicTransformerTest`.

As it is untested, this PR may only go the famous first 80% currently. But the adjustments may already be sufficient and the test work now.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (changes requested):
The CI does its duty. The initial error is fixed, tests still fail https://conti.demos-europe.eu/blue/organizations/jenkins/demosplan-core/detail/PR-2679/2/pipeline Could this be handled in this PR?

dresslerdemos:
Current state: multiple errors in the test were fixed. For further fixing I need to enable debugging in my local test execution again.

mussbach:
@dresslerdemos what is the expected result when `''` is passed into `testProcedureIncludesWithWarning`? How should it fail?As the DynamicTransformer uses `Assert::stringNotEmpty($requestedInclude)` I think the mock is invalid, wdyt? I would really like to get this merged to fix the ci

sierrazamudiodemos:
@mussbach I removed the empty string on the data provider. 12 tests passed and 7 are ignored. what about merging like this? I will check the ignored tests in another PR

PR #2684, opened 2024-02-02
F t29878 disconnect addon data storage

sierrazamudiodemos:
**Ticket:** https://yaits.demos-deutschland.de/T29878

Description:
- Link CoreEntities to AddOn Interfaces on core
- Instead of linking every entity manually to its corresponding Interface in doctrine.yaml, the DemosPlanKernel does it using using AddonResolveTargetEntity (which is a CompilerPass)
- AddonResolveTargetEntity goes through the folder "Entity" in core and detects which Entities implement AddOn interfaces. If they do, then link them using addResolveTargetEntity.

### How to review/test
- Everytime that we create a new class in core (which implements an Addon Interfaces) we should clear cache e.g. bin/ewm dp:c:c , so that the DemosPlanKernel and consequently the new entities are linked with its interfaces.
- You can install demospipe:
-- in DemosEurope\DemosplanAddon\DemosPipes\Entity\PdfImport\AnnotatedStatementPdf update demosplan\DemosPlanCoreBundle\Entity\Procedure\Procedure for DemosEurope\DemosplanAddon\Contracts\Entities\ProcedureInterface
- Test DemosPipes


demosplan/DemosPlanCoreBundle/Addon/AddonResolveTargetEntity.php
```diff
@@ -0,0 +1,69 @@
...
+
+                    }
+                }
+            }
+        }
+    }
+
+    private function addResolveTargetEntityMethodCalls($definition, $interface, $entity): void
+    {
+        $definition->addMethodCall('addResolveTargetEntity', array(
```
mussbach, line 63: I do not understand how this `resolveTargetEntity` works. Could you please add some documentation here?

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/Addon/AddonResolveTargetEntity.php
```diff
@@ -0,0 +1,80 @@
...
+    private const CORE_ENTITY_DIRECTORY = 'demosplan/DemosPlanCoreBundle/Entity';
+    private const ADDON_INTERFACE_DIRECTORY = 'DemosEurope\DemosplanAddon\Contracts\Entities';
+
+    public function process(ContainerBuilder $container): void
+    {
+        $definition = $container->findDefinition('doctrine.orm.listeners.resolve_target_entity');
+        $corePath = DemosPlanPath::getRootPath(self::CORE_ENTITY_DIRECTORY);
+
+
+        $iterator = new RecursiveDirectoryIterator($corePath);
```
mussbach, line 37: I like to use the `Filesystem` Component from Symfony for such tasks, but this is just as I like it

mussbach (approved):
lgtm, thx

demosplan/DemosPlanCoreBundle/Addon/AddonResolveTargetEntity.php
sierrazamudiodemos: Thx for the feedback @mussbach I will use it next time

PR #2694, opened 2024-02-07
bug(refs T36341): Show title for icon only if there is an icon

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36341

Show title for icon only if there is an icon. 

### How to review/test
Verfahren -> Abwägungstabelle -> The spaces should look like the screenshot in the ticket


muellerdemos (approved):
looks legit

PR #2699, opened 2024-02-07
feat (refs T29878): Add doctrine event so that when entityFlush happe…

sierrazamudiodemos:
https://yaits.demos-deutschland.de/T29878

Add doctrine event so mapping is preserved when calling $entity->flush


breierDemos (approved):
lgtm thanks :)

PR #2700, opened 2024-02-07
fix error while adding new steps : configure cascade persist this ass…

bakridemos:
Description: fix error while adding new steps : configure cascade persist this association


MoritzMandler (approved):
lgtm - tested

PR #2701, opened 2024-02-07
F-remove manual statement condition while statement updaten 

bakridemos:
Description: remove manual statement condition while statement updaten


MoritzMandler (approved):
lgtm

PR #2514, opened 2024-01-05
feature (refs T34551): refactor procedure customer relation

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T34551

Description:
alters the property customer for the procedure table
This property was used to indicate a blueprint that is just for a specific customer
Now this property shall indicate that the procedure is owned by an organisation within this customer
This is necessary to determine all customer related activity.
The logic to filter a customer-blueprint had to be adjusted in order to restore the previously expected behavior.

### How to review/test
check for instance diplanbau -> login as a 'Mandanten-Administration' -> goTo 'Blaupause' 
and check the results and behavior stays the same  as compared to main... (on diplanbau you have the option to set a default-customer-blueprint)
check ewm -> do the same (ewm does not have the option to edit a default-customer-blueprint)

you can check the migrations up/down via:
```bin/ewm doc:mig:exe Application\\Migrations\\Version20240105113406 --up```
```bin/ewm doc:mig:exe Application\\Migrations\\Version20240108110342 --up```
the second one gives you an echo output for organisations that are Kommune or Anhörungsbehörde in multiple customers
or both types within the same customer. These are expceted to be only our own test-organisations!!!  example output:
 ```
Hannes1 für AG - is "Kommune or Anhörungsbehörde" in customers: sh, sh
Amt Nordwest - is "Kommune or Anhörungsbehörde" in customers: sh, by
```
The first example row indicates the organisation is Komnuna as well as Anhörungsbehörde in sh - which is completely fine as the customer to set is the same.
The second row would not be fine if it wouldn't be our own testorganisation - the customer to set is unclear and so the first one will be chosen. 

### Linked PRs (optional)
demosplan-addon: https://github.com/demos-europe/demosplan-addon/pull/85

checked for usages of changed methods using the github searchBar with ```"isCustomerMasterBlueprint" AND (repo:demos-europe/demosplan-addon-xbeteiligung-async OR repo:demos-europe/demosplan-addon-segment-oracle OR repo:demos-europe/demosplan-addon-demospipes OR repo:demos-europe/demosplan-addon-maillane OR repo:demos-europe/demosplan-addon-datasheets-sh OR repo:demos-europe/demosplan-addon-bimschg-antrag)``` for ```isCustomerMasterBlueprint and setCustomer``` - no usages found.

- [ ] Tests updated/created
- [ ] Update documentation
- [x] Link all relevant tickets
- [ ] Move the tickets on the board accordingly

PR #2711, opened 2024-02-08
fix (refs T34551): fix output warnig for organisations In Multiple Customers

MoritzMandler:
Ticket:
https://yaits.demos-deutschland.de/T34551

Description:
When transforming the gathered data into a more human readable form - the OrgaId got replaced by the OrgaName if present.
Some organisations have the same name even-though their ids are different.
Do not treat those organisations as same when calculating the warning output.

The logic did not change - so altering this migration should be fine.

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-core/pull/2514

- [ ] Tests updated/created
- [ ] Update documentation


mussbach (approved):
lgtm, code review only

PR #2718, opened 2024-02-12
feature (ref 32892): set permission to required contact person

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T32892

Description: created new procedure, should have required field of contact person, because of that we need a new permission to set default value of that

### How to review/test
set new procdure and if not set contact person under information of procedure in procedure setting, user can not save it.

- [ ] Tests updated/created
- [ ] Update documentation


sakutademos (approved):
lgtm + works

PR #2720, opened 2024-02-12
bug (refs t30978) license notice in mapview missing

sierrazamudiodemos:
**Ticket:** https://yaits.demos-deutschland.de/T30978

Description:
- The copyright (or also called license) saved in procedure_settings table was not displayed in the map (on the right bottom corner) for the corresponding procedure
- Now it is displayed. To do so:
-- MapService was adjusted to get the copyright.
-- FE makes a call to DemosPlanMapApiController -> dplan_api_map_options_public where MapService was adjusted
-- FE was also modified so that the new field copyright is detected and displayed

### How to review/test
1. Make a copy of bobsh dev
2. Go to this particular http://bobsh.dplan.local/app_dev.php/verfahren/da1074e4-c2c1-4e56-971c-411484cde6af/stellungnahmen/entwuerfe
3. Click on "Ansehen"
4. You will see the copyright saved in procedure_settings table (and not the default one)

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/map/map/DpOlMap.vue
+++ b/client/js/components/map/map/DpOlMap.vue
@@ -440,6 +440,8 @@ export default {
     this.centerX = (this.maxExtent[0] + this.maxExtent[2]) / 2
     this.centerY = (this.maxExtent[1] + this.maxExtent[3]) / 2
 
+    this.options.defaultAttribution =  mapOptions.copyright
+
     // Initial view values that can be defined in options object
     if (this._options.initialExtent.length > 0) {
       this.initialExtent = this._options.initialExtent
--- a/demosplan/DemosPlanCoreBundle/Logic/Map/MapService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Map/MapService.php
@@ -772,6 +772,9 @@ class MapService extends CoreService
         $mapOptions->setPublicSearchAutoZoom($config->getMapPublicSearchAutozoom());
         $mapOptions->setAvailableProjections($config->getMapAvailableProjections());
         $mapOptions->setDefaultProjection($config->getMapDefaultProjection());
+
+        $mapOptions->setCopyright($procedureSettings->getCopyright());
+
         $mapOptions->setId($procedureId);
 
         $mapOptions->lock();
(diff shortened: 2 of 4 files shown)


client/js/components/map/map/DpOlMap.vue
```diff
@@ -440,6 +440,8 @@ export default {
     this.centerX = (this.maxExtent[0] + this.maxExtent[2]) / 2
     this.centerY = (this.maxExtent[1] + this.maxExtent[3]) / 2
 
+    this.options.defaultAttribution =  mapOptions.copyright
```
spiess-demos, line 443: In some places, `options.defaultAttribution` is set via prop from the parent component. While we need to preserve the ability to change that prop client side dynamically, the approach to set a default via api is a handy improvement.

I'd suggest to use `this._options` instead of `this.options` (just for consistency, because this._options already [contains the local defaults](https://github.com/demos-europe/demosplan-core/blob/9268027b23f06bf302c1195276ea321e8f757d46/client/js/components/map/map/DpOlMap.vue#L416). [This line](https://github.com/demos-europe/demosplan-core/blob/9268027b23f06bf302c1195276ea321e8f757d46/client/js/components/map/map/DpOlMap.vue#L98) has to be adapted accordingly.

Untested but the suggestion should lead to the prop value being respected while using the default of mapOptions.copyright:

```suggestion
    this._options.defaultAttribution = this.options?.defaultAttribution || mapOptions.copyright
```

spiess-demos (changes requested):
nice improvement! there are several places we could remove the defaultAttribution being passed as a prop now.

just one concern regarding the prop being overridden.

client/js/components/map/map/DpOlMap.vue
sierrazamudiodemos: hey @spiess-demos thank you for the feedback. I will do it. I realized what you said:

1. DpMapMpdel.vue passes "mapOptionsRoute" parameter to DpOlMap.  "mapOptionsRout contains the the api url to call the BE
2. Then the DpOlMap.vue calls the BE using that api url (mapOptionsRoute) and gets several info to adjust the map (including the copyright of the map)
3.  Then what I did in DpOlMap.vue is to set the copyright into "options.defaultAttribution", because DpOlMap.vue passes "options" as parameter to DpOlMapLayer.vue, and ultimaltey DpOlMapLayer.vue uses "options.defaultAttribution" value to set the copyright text on the right corner of the map.

sierrazamudiodemos: @spiess-demos huhu Hans, I adjusted the changes and it does not work, the copyright coming from BE is not taken

spiess-demos (approved):
Lgtm + works, approved from a FE perspective. 

Cleaning up "all the places" where we do not need to pass defaultAttribution via twig/Vue props can be done in a followup PR.

elmasdemos (approved):
lgtm, code review

PR #2723, opened 2024-02-13
f T36454 convert paramater to string

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T36454

Description:  convert paramater to string: the value of $formattedStatement[$attributeKey] can be type int that's why it have to be converted


MoritzMandler (approved):
lgtm - tested together :)

PR #2735, opened 2024-02-15
f T36440 Additional submitters should only be displayed if they have been specified

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T36440

Description: check whether similar statement submitters even exist before adding text


jalalifardemos (approved):
lgtm, code review only. thx

PR #2736, opened 2024-02-15
Fix(T35464): dp-support layout issue 

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T35464

**Description:** The .length property works on arrays, not on objects, so .length returns undefined in this case. To obtain the length of an object, it needs to be converted into an array first using Object.keys(), which allows accessing the length property.


gruenbergerdemos (approved):
lgtm

PR #2747, opened 2024-02-19
Fix(T34304): adjust layout on the 'Stellungnahme mitzeichnen' page

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34304

**Description:** This PR fixes the issue with the layout on the _Stellungnahme mitzeichnen_ page (Part 2). 
I suggest to test the layout in the browser.

It now has the same layout as the _Meine Daten_ block above.
image

Part 1 in diplafest --> [Fix(T34304): adjust layout on the public index page](https://github.com/demos-europe/demosplan-core/pull/2747#issue-2142732766)


templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/new_public_participation_statement_confirm_entry.html.twig
```diff
@@ -1,64 +1,66 @@
...
+                    {% else %}
+                        {{ "notspecified"|trans }}
+                    {% endif %}
+                </dd>
             {% endif %}
-        </dd>
-    </dl>
+        </dl>
 
+        <dl class="{{ 'grid grid-cols-8 w-full md:w-1/2 mb-0'|prefixClass }}">
```
hwiem, line 38: Can't we apply the same layout here as in the "Meine Daten" section ("Datei" below "Dokument"? I think that would be more consistent and it would prevent the 'Datei' field from appearing detached. 
Also, while we're at it, we could make the spacing below the headings more consistent, so that "Stellungnahme mitzeichnen" has the same space below as the "Meine Daten" heading.

sakutademos: thx for the idea, done here https://github.com/demos-europe/demosplan-core/pull/2747/commits/3e984a79f52dc83711d398bc3e20d9f6abdced01

hwiem (dismissed):
lgtm, thanks! :heart:

PR #2748, opened 2024-02-20
bug(refs T36509): Remove empty comment which is breaking the site

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36509

- Remove empty html comment which is breaking the site if the permission is not enabled
- Add twig whitespace control to trim any whitespace before or after the tag (`{%- ... -%}`)

### How to review/test
Verfahren -> Grundeinstellungen -> Informationen zum Verfahren -> Menu should be on the right side and more topics and save button should appear beneath "Informationen zum Verfahren"


sakutademos (approved):
ahh, right! thank you, ltgm

PR #2752, opened 2024-02-20
chore (refs T33654): use variable to portal in trans key

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T33654

Portalurl needs to be adjusted according to the current environment. We can reuse the twig variable used in the header.

### How to review/test
URL in information text like stated in the ticket should be adjusted as variable is set. Otherwise code review

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


sakutademos (approved):
lgtm

PR #2768, opened 2024-02-22
feature (ref 34305): use method to export umlaut alphabet

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T31585

Description: use method `u()` to support umlaut alphabet in export

### How to review/test
export protocol of procedures as a PDF, there should be user in correct format

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Report/ExportReportService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Report/ExportReportService.php
@@ -21,6 +21,7 @@ use PhpOffice\PhpWord\Exception\Exception;
 use PhpOffice\PhpWord\IOFactory;
 use PhpOffice\PhpWord\Writer\WriterInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
+use function Symfony\Component\String\u;
 
 class ExportReportService extends CoreService
 {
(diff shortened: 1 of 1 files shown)


mussbach (approved):
lgtm, code review only. Seems a bit weird though. Might the locale of the document be something else than utf8?

jalalifardemos:
I looked for it too much, can't find anything

demosplan/DemosPlanCoreBundle/Logic/Report/ExportReportService.php
```diff
@@ -177,7 +179,7 @@ private function writeReportEntry(ReportEntry $reportEntry, Table $table)
 
         $userName = $reportEntry->getUserName();
         $userCell = $table->addCell($this->styles['userCellWidth']);
-        $userCell->addText($userName, $this->styles['baseFont']);
+        $userCell->addText(u($userName)->normalize()->toString(), $this->styles['baseFont']);
```
MoritzMandler, line 182: The extra u seems strange.

jalalifardemos: I try with another methods to support umlaut alphabet, but it doesn't work, only with `u()` you can have a correct export

PR #2771, opened 2024-02-23
feature (ref 34305): use break class to handle big insert data

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T35058

Description: use class break-words to handle big inserted data in support fields

### How to review/test
as client administrator, the large inserted data in Content Support should remain in the frame
in the information page, the large inserted data should no longer be the size of the frame

- [ ] Tests updated/created
- [ ] Update documentation


sakutademos (approved):
thx, lgtm

PR #2776, opened 2024-02-23
fix(T36253): performance issue - ElementsList

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T36253

**Description:** `nodeSelectionChange` is an expensive function where a lot of interaction happens:

The parent component `ElementsList` receives a list/array of Objects with `nodeIds`:
1. it is filtered by _leaf_
2. a new array with Node-IDs is created
3. afterward `allFiles` are filtered accordingly to these nodeIds

There are a lot of interactions. Wouldn't it make more sense to not only pass nodeIds from child to parent, but also the node (with the file information, etc.)?

Changes here and in the [demosplan-ui](https://github.com/demos-europe/demosplan-ui/pull/761
) (DpTreeList components) helped optimize the loading time from 5-6 seconds to 2 (for a list of 800 elements).

But the main reason it freezes up is a reactive button that displays the size of the list.

### DIFF
--- a/client/js/components/document/ElementsList.vue
+++ b/client/js/components/document/ElementsList.vue
@@ -29,7 +29,8 @@
         @node-selection-change="nodeSelectionChange"
         :tree-data="recursiveElements"
         :branch-identifier="isBranch()"
-        :options="treeListOptions">
+        :options="treeListOptions"
+        leaf-only>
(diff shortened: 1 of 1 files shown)


mussbach:
@sakutademos what hapens with this gem?

PR #2778, opened 2024-02-26
fix(refs T36457): prevent corrupt layers from breaking draw fn

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T36457

Although this is an edge case we have to handle it: if a layer that is set as print layer becomes disfunctional (maybe because the service is discontinued), it must not break the frontend.

### How to review/test
Set a non-working layer as print layer, try to draw in the public detail map -> it should work.

### PR Checklist


hwiem (approved):
code lgtm

PR #2780, opened 2024-02-26
bug(refs T23768): Disable opacity for base map

gruenbergerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T23768

- Fix regex pattern for opacity. Period is a special character in regex and needs to be escaped (double escape because backslash is a special character in twig and needs to be escaped). Remove redundant pipe that causes error
- Disable opacity settings for base maps, because the opacity should always be 100%. 

### How to review/test
Verfahren -> Antragsunterlagen -> Kartenebenen definieren -> Neue Kartenebene oder Kartenebene bearbeiten -> If Grundkarte is selected opacity should be disabled and 100%. Also check if regex for opacity is applied correctly when selecting Overlay (you should be able to put 0-100 in with two decimals (`.` or `,`).

### DIFF
--- a/client/js/lib/map/GisLayerEdit.js
+++ b/client/js/lib/map/GisLayerEdit.js
@@ -44,7 +44,7 @@ export default () => {
     }
   })
 
-  const elementsHiddenForBase = [
+  const elementsHiddenForBaseMap = [
     {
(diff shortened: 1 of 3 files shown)


hwiem (approved):
lgtm, code review only

PR #2791, opened 2024-02-28
Allow for opting out of default filter of procedures on public_index 

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T36464

Some projects may prefer not to filter procedures
when loading the public index page. This is now supported via dropping a `data-default-selected="all"` on the select element. In future iterations, other keys may be used to set another default filters.

### How to review/test
When loading the public index page in a project that uses `publicIndexNoMap.js` for its view, applying `data-default-selected="all"` to the select element with the id `filterPhases` should lead to no option being selected automagically.

### PR Checklist


hwiem (approved):
lgtm, code review only

PR #2793, opened 2024-02-28
F t19759 t19774 restructured global internal ap is

ahmad-demos:
https://yaits.demos-deutschland.de/T19759
https://yaits.demos-deutschland.de/T19786

- restructured Global APIs (nextTick, delete).
- add & renaming emits options as String. `fix`


hwiem (approved):
lgtm

PR #2794, opened 2024-02-28
bug (ref T36592) Adjust "element" in include for API request so EDT can understand and match the entity

sierrazamudiodemos:
**Ticket:** https://yaits.demos-deutschland.de/T36582z

Description:
- Set ToOneRelationshipConfigBuilderInterface in the $element. It was configured to ToManyRelationshipConfigBuilderInterface. When the $element is null, it breaks. Now it does not.

Context:
- Abwägungstabelle does not load because the Procedure has Statements with no PlanungsDokumenten
- The PlanungDokumenten are this _st_element_id FK in the statement table. When _st_element_id is null, the Abwägungstabelle does not load
- This happens because element is defined as ToManyRelationshipConfigBuilderInterface but in reality it should be  ToOneRelationshipConfigBuilderInterface.

### How to review/test
- Go to Bebauungsplan Nr. 3 Supermarkt Kleinstadt-West Verfahren
- Click on "new"
- The Abwägungstabelle should load now

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/store/statement/Statement.js
+++ b/client/js/store/statement/Statement.js
@@ -473,7 +473,7 @@ export default {
      */
     getStatementAction ({ commit, state, rootState }, data) {
       const includes = [
-        'elements',
+        'element',
         'paragraph',
         'document',
         'assignee',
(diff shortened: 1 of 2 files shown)


MoritzMandler (approved):
nice one :+1: 
(many)TO_ONE relation it is

PR #2806, opened 2024-03-01
Feat(refs:T36493): show the default copyright note in the correct maps

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36493

Description: 

- show the default copyright in the map in basic settings for procedure, in startmap outcut and in plandrawing settings
- due to the projects there are different defaults

### PR Checklist

### DIFF
--- a/client/js/components/map/map/DpOlMapLayer.vue
+++ b/client/js/components/map/map/DpOlMapLayer.vue
@@ -62,19 +62,6 @@ export default {
   },
 
   computed: {
-    defaultAttributions () {
-      const currentYear = formatDate(new Date(), 'YYYY')
-      // If a value is currently given, replace the {currentYear} placeholder within that value
-      if (this?.attributions) {
-        return this.attributions.replaceAll('{currentYear}', currentYear)
-      }
-      // If not, default to the default message
-      return Translator.trans('map.attribution.default', {
-        linkImprint: Routing.generate('DemosPlan_misccontent_static_imprint'),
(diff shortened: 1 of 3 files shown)


client/js/components/map/map/DpOlMapLayer.vue
```diff
@@ -87,12 +74,27 @@ export default {
...
+     */
+    defaultAttributions () {
+      const currentYear = formatDate(new Date(), 'YYYY')
+      return this.attributions
+        ? this.attributions.replaceAll('{currentYear}', currentYear)
+        : Translator.trans('map.attribution.default', {
+          linkImprint: Routing.generate('DemosPlan_misccontent_static_imprint'),
+          currentYear
+        })
+    },
```
spiess-demos, line 90: I d# not understand in which way this change alters the behavior. It seems like the code is reformatted to use a ternary, but the logic remains the same. Moving from a computed property to a method should not change, either - so i don't get it.

I would expect the change in the places where default attribution is set. In DpOlMap, the `defaultAttribution` option is passed to DpOlMapLayer as `attributions`. So i would remove this option whereever the default attribution should show up. Admittedly, naming something `defaultAttribution` which, if ommitted, leads to showing the default attribution, is quite confusing but hey, naming is hard :)

spiess-demos (changes requested):

spiess-demos (approved):
lgtm! code review only..

PR #2812, opened 2024-03-04
feature (ref 36588): create a permission for profile edit

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T36588

Description: Institution_Sachbearbeitung and Fachplaner-Sachbearbeiter (PLANNING_AGENCY_WORKER and PUBLIC_AGENCY_WORKER) should not have a save/cancel option in my data form, because of that I created a new permission and use that in every project

### How to review/test
go to mein Daten and check these roles have those options or not

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm

PR #2817, opened 2024-03-05
fix: return extent from function

spiess-demos:


mussbach (approved):
lgtm, code review only

PR #2821, opened 2024-03-06
fix(refs T27465): display only documents that should be visible

hwiem:
**Ticket:** https://yaits.demos-deutschland.de/T27465

- only documents with the attribute `visible` set to `true` should be displayed in the list
- normally, this is filtered in the backend, but if the user `hasPermission('area_admin_single_document')`, documents are not
filtered, so all documents are displayed
- since we can't currently filter the relationships of a resourceType via json api request, we decided to add a new relationship
`visibleDocuments` that is now used instead of `documents`

### PR Checklist

### DIFF
--- a/client/js/components/document/ElementsList.vue
+++ b/client/js/components/document/ElementsList.vue
@@ -189,8 +189,8 @@ export default {
         const isTopLevel = node.attributes.parentId === null
 
         // Make documents direct children of node, if there are any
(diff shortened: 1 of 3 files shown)


muellerdemos (approved):
fe lgtm

elmasdemos (approved):
lgtm, code review

PR #2823, opened 2024-03-07
fix(T36253): improve TreeList performance

sakutademos:
**Bug:** https://yaits.demos-deutschland.de/T36253

When selecting a folder with many subfolders and approximately 800 files, each element in the list starts the process (points 1, 2, 3, 4) one after the other, and the page freezes because it takes about 1 minute to perform the calculations:

1. `DpTreeListNode.vue`: **An object is created** with the `nodeId` and **sent to parent** via emit.
2. `DpTreeList.vue`: The object is added to the list `selectedNodesObject`. Then this list is **transformed into an array** and forwarded to the parent via emit.
3. `ElementsList.vue`: The **array is filtered** by _leaf_ and a **new array** with IDs is **created**. Then **allFiles are filtered** by these IDs, and **a new array** with selected files is **created**.
4. ` ElementsList.vue`: Reactive Button Label responds to changes in the list, and the size of the selected files is calculated and displayed.

**Main problems:**

- the list elements are in the DOM even when they are not visually visible. This was implemented so that the parent component (Folder) could have access to the child component in the DOM and clicked `nodeIds` could be collected.
- the data transmission from the child to the parent component with various data transformations are also time consuming, especially when there are many elements.
- the page freezes, because the computed variable _buttonLabel_ is being calculated 800 times (Point 4), causing performance issues.

**Changes in these PRs:**

My suggestion would be to adjust the logic slightly and directly add properties `nodeIsSelected`, `nodeType`, and `nodeId` to each node object. If the node has children, the value will also be recursively changed for the children. Using this, we could remove points 1, 2 and 3, and the `buttonLabel` would be calculated only once. It **improves the performance from 1 minute to 1-2 seconds** for the list of 800 elements.

Currently, I cannot replace `v-show` with `v-if` because then we have an issue with the download action. The `formAction` is triggered, and the `name` attributes of inputs are sent to the BE. If a checkbox does not exist in the DOM, its `name` cannot be sent to the BE.

[PR in demosplan-ui](https://github.com/demos-europe/demosplan-ui/pull/774)

**How to test:**

To test the DptreeList you need a large ZIP file like [this one](https://planfeststellung.bob-sh.de/verfahren/schienenanbindung-fbq-pfa-4/public/detail). 

- `ElementsList`: public detail page 
- `ElementsAdminList`: Fachplanung-Admin > Verfahren > Grundeinstellungen > Dokumente
- `DpFaqList`: Mandanten-Administration > Häufige Fragen


spiess-demos (approved):
lgtm & works!

PR #2824, opened 2024-03-07
f(refs:T36655): use v-popover instead of v-tooltip wip

elmasdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36655

This PR introduces VPopover in a place where v-tooltip was used before in an unintended way. Also, it "repairs" all VTooltip instances within demosplan-core, by re-adding the trigger config and styles needed to display the popovers like tooptips.

### How to review/test
Visit each of the places changed in the PR, hover over the elements, a popover should appear.

### PR Checklist

### DIFF
--- a/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
+++ b/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
@@ -66,7 +66,9 @@
           <label
             :for="`checkStatement:${displayedCheckboxId}`"
             class="layout__item u-1-of-6 u-mb-0 u-pb-0_25">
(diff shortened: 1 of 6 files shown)


elmasdemos:
the tooltip is in down direction, but I think it looks better when its over the word and not under the word, which it belongs to

spiess-demos:
> the tooltip is in down direction, but I think it looks better when its over the word and not under the word, which it belongs to

hopefully fixed with 761fb13ed06b26c355c2ec92daf857f9b1971cf6

elmasdemos:
lgtm, it works, thanks

PR #2825, opened 2024-03-07
fix(refs T36638): remove obsolete negative margin

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T36638

The surrounding container seems to not apply a margin anymore, so there is no need for the negative margin anymore.

### How to review/test
should look like this:
image

### PR Checklist


gruenbergerdemos (approved):
lgtm

PR #2831, opened 2024-03-11
chore (refs T36726): strip invalid a tags without href in export

mussbach:
**Ticket:** https://yaits.demos-deutschland.de/T36726

Docx creation failed for a tags without href defined like `<a rel="noopener noreferrer nofollow">nolink</a>`. We need to strip all those occurrences without loosing `<a rel="noopener noreferrer nofollow" href="foo">mylink</a>`

### How to review/test
code review or tamper with the recommendation data and perform an docx export

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm code review and tested locally,
but it seems that one file could be missing?

PR #2843, opened 2024-03-13
feature (ref 15290): remove session between pages.

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T15290

Description: the method (`remmemberFilters`) used just in two methods  `getStatementFragmentListAction` and `getStatementFragmentListArchiveAction` in the `DemosPlanAssessmentStatementFragmentController`, this `else` in the method save session and used it in another page. For example, if we set filter in Datensätze and then go to Datensätze-Archive, the filter applies the wrong value in the new page.

### How to review/test
login as an FP-FB and set filter in Datensätze then go to Datensätze-Archive, filter should not run in new page.

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm, code review and tested too

PR #2844, opened 2024-03-13
f T26867 remove chapter (kapitel) from statement filter

bakridemos:
**Ticket:** https://yaits.demos-deutschland.de/T26867

Description: the permission 'feature_documents_category_use_paragraph' is used while creating a new document category to check if it's chapter related which can not be the case in planfest (there are nor chapter in this project). The same permission can be used too while getting the available filters to make sure that's chapter are not included.


jalalifardemos (approved):
Igtm Thx, Code review and test together

PR #2845, opened 2024-03-13
feature (ref 15938): use right key to show variable in pdf export

jalalifardemos:
**Ticket:** https://yaits.demos-deutschland.de/T15938

Description: Export PDF from data records archive (Datensätze-Archive) has some missed Variable like (Tag, County, …). I used just right key for right value.

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm - tested together

PR #2846, opened 2024-03-13
fix(refs T36689): document attribute name has to be without s

salisdemos:
getting the proper include fixes finishing the loading process because now the field-name matches the include name. With that the code enters the relevant condition.

**Ticket:** https://yaits.demos-deutschland.de/T36689

See Ticket to reproduce the bug(fix)

### DIFF
--- a/client/js/store/statement/Statement.js
+++ b/client/js/store/statement/Statement.js
@@ -857,18 +857,18 @@ export default {
           statementId: data.data.id,
           procedureId: state.procedureId,
           include: [
-            'elements',
-            'paragraph',
-            'documents',
-            'counties',
-            'municipalities',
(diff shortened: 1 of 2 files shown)


client/js/store/statement/Statement.js
```diff
@@ -857,18 +857,18 @@ export default {
...
-            'documents',
-            'counties',
-            'municipalities',
-            'priorityAreas',
-            'tags',
             'assignee',
             'attachments',
             'attachments.file',
+            'counties',
+            'document',
```
salisdemos, line 864: changing this from 'documents' to 'document' is the only real change

gruenbergerdemos (approved):
lgtm

PR #2847, opened 2024-03-13
fix(refs T35941): rollback partial fix, serialize params

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T36652

The fix applied in #2521 introduced the side effect that the ref="recommendation" was undefined while the loading state was applied to the EditableText instance, and so, the `field` property was not correctly filled within the payload of the entityTextSaved event. As the bug originates in a change in dpApi, #2521 is partially rolled back. Instead, the params in dpApi.get() are now serialized, with the result that the `shortened` attribute is present in the response and the code in client/js/components/statement/assessmentTable/EditableText.vue:355 runs so that and everything is updated as intended.

### How to review/test
In the assessment table, statement text and recommendation should both be editable and display the correct values after saving. also, the "Mehr anzeigen" button should be present - or absent, with text shorter than 500 chars - as needed.

### PR Checklist

### DIFF
--- a/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
+++ b/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
@@ -558,11 +558,7 @@
                     With tiptap we can set obscure as prop always when the obscure button should be visible in the field,
                     because the permission check (featureObscureText) takes place in tiptap
                     -->
-                <dp-loading
-                  v-if="reloadStatementEditor"
-                  class="u-pb-0_5 u-pr-0_5 u-pt-0_25 u-1-of-2 border--right" />
                 <editable-text
-                  v-else
                   class="u-pb-0_5 u-pr-0_5 u-pt-0_25 u-1-of-2 border--right"
                   title="statement"
                   :procedure-id="procedureId"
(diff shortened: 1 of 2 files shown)


client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
```diff
@@ -558,11 +558,7 @@
                     With tiptap we can set obscure as prop always when the obscure button should be visible in the field,
                     because the permission check (featureObscureText) takes place in tiptap
                     -->
-                <dp-loading
-                  v-if="reloadStatementEditor"
-                  class="u-pb-0_5 u-pr-0_5 u-pt-0_25 u-1-of-2 border--right" />
```
salisdemos, line 563: why did You remove the loading state?

spiess-demos: as pointed out in the PR description, the loading state was introduced in https://github.com/demos-europe/demosplan-core/pull/2521 to overcome reactivity issues but those were no real issues as the only bug was the absent of the "shortened" attribute in the api response.

salisdemos (approved):

PR #2848, opened 2024-03-13
fix(refs T36753): use full width to display the stn text

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36753

As not logged in user, write a statement a proceed to the "recheck"-view in the statement modal. There, the statement text should be displayed and readable.

### PR Checklist

### DIFF
--- a/client/js/components/statement/publicStatementModal/StatementModalRecheck.vue
+++ b/client/js/components/statement/publicStatementModal/StatementModalRecheck.vue
@@ -218,7 +218,7 @@
       </span>
 
       <div
-        :class="prefixClass('sm:h-9 overflow-auto w-2')"
+        :class="prefixClass('sm:h-9 overflow-auto w-full')"
         v-cleanhtml="statement.r_text" />
     </div>
   </fieldset>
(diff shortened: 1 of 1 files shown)


spiess-demos (commented):
doesn't the div take the full width by default if not told otherwise when being set to `display: block;`?

salisdemos:
> doesn't the div take the full width by default if not told otherwise when being set to `display: block;`?

true

client/js/components/statement/publicStatementModal/StatementModalRecheck.vue
```diff
@@ -218,7 +218,7 @@
       </span>
 
       <div
-        :class="prefixClass('sm:h-9 overflow-auto w-2')"
+        :class="prefixClass('sm:h-9 overflow-auto w-full')"
```
salisdemos, line 221:
```suggestion
        :class="prefixClass('sm:h-9 overflow-auto')"
```

spiess-demos (approved):
lgtm!

PR #2849, opened 2024-03-13
feat(refs T36528): add message to confirm export.

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36528

To be consistent, we want to have confirm messages for all stn-docx-exports. There was one missing on the stn-detail page.

this here is implemented like here https://yaits.demos-deutschland.de/T35722 // https://github.com/demos-europe/demosplan-core/pull/2433

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
@@ -69,7 +69,7 @@
           <li>
             <dp-button
               class="u-ph-0_25"
-              @click="showHintAndDoExport('dplan_segments_export')"
+              @click="showHintAndDoExport()"
               :text="Translator.trans('export.verb')"
               variant="subtle" />
           </li>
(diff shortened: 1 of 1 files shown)


gruenbergerdemos (approved):
lgtm

client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
```diff
@@ -69,7 +69,7 @@
           <li>
             <dp-button
               class="u-ph-0_25"
-              :href="Routing.generate('dplan_segments_export', { procedureId: procedureId, statementId: statementId })"
+              @click="showHintAndDoExport('dplan_segments_export')"
```
spiess-demos, line 72: Why do you pass the route name here? it is not used within the fn body.

nitpicky question: while `showHintAndDoExport` pretty much sums up what the fn does, i'd rather chose the more handy `exportStatements` as fn name

salisdemos: I just messed it up. Changed in d588a994814e4dae4257378d82e38773058b4566

salisdemos: > i'd rather chose the more handy exportStatements as fn name
I think I would stick to the name hence its the same as the other place. And If we want to move the Code out to some external fn its easier to find all occurrences

spiess-demos (approved):
lgtm!

PR #2874, opened 2024-03-20
fix(refs T36849): display link inline

hwiem:
**Ticket:**  https://yaits.demos-deutschland.de/T36849

- `flow-root` unexpectedly overwrote the display property (to `block`) which is desired in most projects, but not in all of them
- it is replaced with 'block' and passed to the pageheader as a variable, so in projects that display the link inline, we can just use the pageheader template as is
- vertical margin of the link has been adjusted so it aligns properly with the procedure title in projects that display link and procedure title inline

### PR Checklist


spiess-demos (approved):
lgtm!

PR #2875, opened 2024-03-20
fix: remove setupFiles in config because it is not used any more

elmasdemos:
### Description: 

due to https://github.com/demos-europe/demosplan-core/pull/2870 the config has to be adjusted, so that the tests doesnt fail


muellerdemos (approved):
lgtm

PR #2878, opened 2024-03-21
Cleanup calls of dpApi

spiess-demos:
Due to the removal of axios within dpApi there were some leftovers that are tackled.

- responseType: 'json' is not used anywhere, can be removed
- for clarity, properties within dpApi param are always sorted method, url, data, headers
- method always uppercase
- the "x-www-form-urlencoded" header was actually never applied to the request after dropping axios, and as it turns out, it is not needed anyway
- serialize: true is removed, as all GET calls have serialized params now
- "data" carrying an empty string was removed in the assumption that it is the same as omitting it
- within the call to dplan_api_statement_fragment_edit the params have been moved to Routing.generate (which, in fact, also takes care of url params), because atm dpApi only appends params to GET requests

### How to review/test
Api calls should work as before...

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-ui/pull/803

### PR Checklist

### DIFF
--- a/client/js/bundles/procedure/publicIndexMapAndList.js
+++ b/client/js/bundles/procedure/publicIndexMapAndList.js
@@ -42,9 +42,8 @@ initialize(components, stores).then(() => {
 
       dpApi({
         method: 'POST',
-        data: { procedureId: procedureId },
-        responseType: 'json',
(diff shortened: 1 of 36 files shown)


salisdemos (approved):
Thanks for cleaning this. That looks good

PR #2879, opened 2024-03-21
fix(refs T36875): Pass ProcedureId as string instead of Param

salisdemos:
**Ticket:** https://yaits.demos-deutschland.de/T36875

Pass the current ProcedureID to the getFileHash-method. This has to be done "static" so demosplan-ui doesn't need to know about the procedure-id. That information is to specific to be handled in a lib that tries to be agnostic.

This is an alternative approach to https://github.com/demos-europe/demosplan-ui/pull/804


spiess-demos (changes requested):
i prefer this solution over the other one. however in templates/bundles/DemosPlanCoreBundle/DemosPlanDocument/paragraph_admin_edit.html.twig:79 it could be fixed as well.

salisdemos:
Thanks for spotting. Added it with [a1c4d1e](https://github.com/demos-europe/demosplan-core/pull/2879/commits/a1c4d1e39e82f6108978c8f77fb263b5d7e07475)

spiess-demos (approved):
lgtm (code review)

PR #2903, opened 2024-04-05
chore: Bump Vue@2.7.16

salisdemos:
Since vue and vue-template-compiler hs to be dumped at the same time, her we go


salisdemos:
first read, then write

PR #2907, opened 2024-04-08
feature: change method from private to protected

jalalifardemos:
Description: Phpstan level 1 :

```
Line   DemosPlanCoreBundle/Logic/Segment/SegmentsByStatementsExporter.php  
 ------ -------------------------------------------------------------------- 
  143    Call to private method addSimilarStatementSubmitters() of class     
         demosplan\DemosPlanCoreBundle\Logic\Segment\SegmentsExporter.       
 ------ --------------------------------------------------------------------                                                                                
 [ERROR] Found 1 error
```

### How to review/test
bin/projectname dp:phpstan -l 1


dresslerdemos (approved):
Code review only. Verified that it is indeed used in a child class.

PR #2921, opened 2024-04-08
fix(refs DPLAN-11392): Add Pattern to iterator input

salisdemos:
allow only numbers and not `0`

https://demoseurope.youtrack.cloud/issue/DPLAN-11392/Falsche-Reaktion-beim-Eintragen-ungultiger-Daten-im-Feld-fur-Durchgangsnummer

In "Grundeinstellungen" try to change the iterator (Durchgangsnummer). For internal and external only numbers should be valid and `0` should be unvalid


gruenbergerdemos (dismissed):
lgtm

PR #2933, opened 2024-04-09
Fix DPLAN-11393 11393 reportentry on change phaseiteration

martendemos:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11393/Durchgangsnummer-und-Anderungen-werden-nicht-im-Protokoll-berucksichtigt

- detect change of iteration (as phase change)
- get iteration numbers and enhance created report message by them
- adjust message on reading the reports by adding the additional data
- do all of these for external- as well as for internal- phases
- improve readability by adjustments of variable naming

### How to review/test
Change the iteration of a phase of a procedure and check the report entries (protocoll) of the procedure.

- [ ] Tests updated/created


demosplan/DemosPlanCoreBundle/Logic/Procedure/ServiceStorage.php
```diff
@@ -1073,4 +1078,26 @@ private function handleTokenField($token): ?string
 
         return $token;
     }
+
+    private function validatePhaseIteration(array $procedure): array
```
breierDemos, line 1082: You could reduce the amount of returns. You could save the if conditions in variables. So you have two booleans -> one if condition, inside the if condition you set the error, and as default $mandatoryErrors is empty []. What do you think?

mussbach (dismissed):
lgtm, code review only

breierDemos (changes requested):

breierDemos (approved):
code review only

demosplan/DemosPlanCoreBundle/Logic/Procedure/ServiceStorage.php
martendemos: I choose to use early return to be performant on this check, but you are right, this does'nt matter at this point.
Readability is more valuable here.
90368a8ef34c54b8bf90bcdb84a6e7fbed2b977d

breierDemos (dismissed):

PR #2939, opened 2024-04-10
fix(refs dplan-11402, dplan-11403):  Refactor logic to display Permissions for selected Phase

salisdemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-11402/Hinweis-Grundeinstellungen-Angaben-der-Rechte-zu-Verfahrensschritten-fehlen

https://demoseurope.youtrack.cloud/issue/DPLAN-11403/Layout-in-Grundeinstellungen-nicht-einheitlich

To fix the Bug, I hat to move the logic from twig and wrapper to a dedicated Component.

### How to review/test

Goto Grundeinstellungen and change the phases. See if the Permission text is displayed and makes sense.

### DIFF
--- a/client/js/components/procedure/basicSettings/ParticipationPhases.vue
+++ b/client/js/components/procedure/basicSettings/ParticipationPhases.vue
@@ -93,7 +93,7 @@ export default {
       type: Object,
       required: false,
       default: () => ({}),
-      validate: val => {
+      validator: val => {
         const requiredKeys = ['label', 'name', 'tooltip', 'value']
         let keyCounter = 0
 
--- a/client/js/components/procedure/basicSettings/ParticipationPhases.vue
+++ b/client/js/components/procedure/basicSettings/ParticipationPhases.vue
@@ -1,31 +1,28 @@
 <template>
   <div>
     <div class="flex gap-2">
-      <div class="w-8-12">
-        <dp-select
-          :label="{
-            text: labelText,
-            tooltip: helpText
-          }"
-          :name="fieldName"
-          :options="phaseOptions"
-          required
-          v-model="selectedPhase" />
-      </div>
+      <dp-select
+        class="w-8/12"
+        :label="{
+          text: labelText,
+          tooltip: helpText
+        }"
+        :name="fieldName"
+        :options="phaseOptions"
+        required
+        v-model="selectedPhase" />
 
-      <div
-        v-if="true || hasPermission('field_phase_iterator')"
-        class="w-4/12 inline-block relative">
-        <dp-label
-          for="r_public_participation_phase_iteration"
-          :text="iterator.label"
-          :tooltip="iterator.tooltip"
-          required />
-        <dp-input
-          name="r_public_participation_phase_iteration"
-          :value="iterator.value"
-          pattern="^[1-9][0-9]*$" />
-      </div>
+      <dp-input
+        v-if="hasPermission('field_phase_iterator')"
+        width="w-4/12"
+        :label="{
+          text: iterator.label,
+          tooltip: iterator.tooltip
+        }"
+        :name="iterator.name"
+        :value="iterator.value"
+        pattern="^[1-9][0-9]*$"
+        required />
     </div>
 
     <dp-inline-notification
@@ -44,7 +41,6 @@
 import {
   DpInlineNotification,
   DpInput,
-  DpLabel,
   DpSelect
 } from '@demos-europe/demosplan-ui'
 
@@ -54,7 +50,6 @@ export default {
   components: {
     DpInlineNotification,
     DpInput,
-    DpLabel,
     DpSelect
   },
 
--- a/client/js/components/procedure/basicSettings/ParticipationPhases.vue
+++ b/client/js/components/procedure/basicSettings/ParticipationPhases.vue
@@ -135,14 +135,14 @@ export default {
   computed: {
     permissionMessageText () {
       const currentPhase = this.phaseOptions.find(option => option.value === this.selectedPhase)
-       /*
-       * Generated Trans-Keyes:
+      /*
+       * Generated Trans-Keys:
        *
        * 'permissionset.hidden'
        * 'permissionset.read'
        * 'permissionset.write'
        */
-      const permissionsetMessage =  Translator.trans(`permissionset.${currentPhase.permissionset}`)
+      const permissionsetMessage = Translator.trans(`permissionset.${currentPhase.permissionset}`)
 
       return `${this.permissionMessage} ${permissionsetMessage}`
     },
--- a/client/js/components/procedure/basicSettings/ParticipationPhases.vue
+++ b/client/js/components/procedure/basicSettings/ParticipationPhases.vue
@@ -118,14 +118,16 @@ export default {
   computed: {
     permissionMessageText () {
       const currentPhase = this.phaseOptions.find(option => option.value === this.selectedPhase)
-      /*
+       /*
        * Generated Trans-Keyes:
        *
        * 'permissionset.hidden'
        * 'permissionset.read'
        * 'permissionset.write'
        */
-      return this.permissionMessage + Translator.trans(`permissionset.${currentPhase.permissionset}`)
+      const permissionsetMessage =  Translator.trans(`permissionset.${currentPhase.permissionset}`)
+     
+      return `${this.permissionMessage} ${permissionsetMessage}`
     },
 
     isInParticipation () {
--- a/client/js/components/procedure/basicSettings/ParticipationPhases.vue
+++ b/client/js/components/procedure/basicSettings/ParticipationPhases.vue
@@ -134,7 +134,7 @@ export default {
 
   computed: {
     permissionMessageText () {
-      const currentPhase = this.phaseOptions.find(option => option.value === this.selectedPhase)
+      const currentPhase = this.phaseOptions.find(option => option.value === this.selectedPhase) || this.phaseOptions[0]
       /*
        * Generated Trans-Keys:
        *
--- a/client/js/components/procedure/basicSettings/ParticipationPhases.vue
+++ b/client/js/components/procedure/basicSettings/ParticipationPhases.vue
@@ -14,6 +14,7 @@
 
       <dp-input
         v-if="hasPermission('field_phase_iterator')"
+        :id="iterator.name"
         width="w-4/12"
         :label="{
           text: iterator.label,
@@ -142,7 +143,7 @@ export default {
        * 'permissionset.write'
        */
       const permissionsetMessage =  Translator.trans(`permissionset.${currentPhase.permissionset}`)
-     
+
       return `${this.permissionMessage} ${permissionsetMessage}`
     },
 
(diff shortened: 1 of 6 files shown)


client/js/components/procedure/basicSettings/ParticipationPhases.vue
```diff
@@ -0,0 +1,141 @@
...
+            tooltip: helpText
+          }"
+          :name="fieldName"
+          :options="phaseOptions"
+          required
+          v-model="selectedPhase" />
+      </div>
+
+      <div
+        v-if="true || hasPermission('field_phase_iterator')"
```
hwiem, line 17: Is this correct?

client/js/components/procedure/basicSettings/ParticipationPhases.vue
```diff
@@ -0,0 +1,141 @@
...
+        class="w-4/12 inline-block relative">
+        <dp-label
```
hwiem, line 19: Shouldn't there be an input with the id `r_public_participation_phase_iteration`?

client/js/components/procedure/basicSettings/ParticipationPhases.vue
```diff
@@ -0,0 +1,141 @@
...
+  },
+
+  props: {
+    autoswitchHint: {
+      type: String,
+      required: false,
+      default: ''
+    },
+
+    labelText: {
```
hwiem, line 68: nitpick: please order the props alphabetically

client/js/components/procedure/basicSettings/ParticipationPhases.vue
```diff
@@ -0,0 +1,141 @@
...
+      default: ''
+    },
+
+    initSelectedPhase: {
+      type: String,
+      required: false,
+      default: ''
+    },
+
+    iterator: {
```
hwiem, line 92: Why did you call this prop `iterator`? (I can't really imagine what this is, but I see that iterator is used in other places, too. Maybe a doc block would help)
I think we should always add a validator or a doc block to explain the properties an object prop should have, so it's easy to see which data we need to pass into a component

client/js/components/procedure/basicSettings/ParticipationPhases.vue
```diff
@@ -0,0 +1,141 @@
...
+    permissionMessageText () {
+      const currentPhase = this.phaseOptions.find(option => option.value === this.selectedPhase)
+      /*
+       * Generated Trans-Keyes:
+       *
+       * 'permissionset.hidden'
+       * 'permissionset.read'
+       * 'permissionset.write'
+       */
+      return this.permissionMessage + Translator.trans(`permissionset.${currentPhase.permissionset}`)
```
hwiem, line 133: nitpick: this way it's clearer what the doc block refers to

```suggestion
      const currentPhase = this.phaseOptions.find(option => option.value === this.selectedPhase)
       /*
       * Generated Trans-Keyes:
       *
       * 'permissionset.hidden'
       * 'permissionset.read'
       * 'permissionset.write'
       */
      const permissionsetMessage =  Translator.trans(`permissionset.${currentPhase.permissionset}`)
     
      return `${this.permissionMessage} ${permissionsetMessage}`
```

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -488,7 +451,10 @@
 
                                     <dp-date-range-picker
                                         class="u-2-of-4"
-                                        data-cy="procedureSettingsDate"
+                                        :data-cy="{
```
hwiem, line 454: Why this change? Doesn't `dataCy` have to be a string?

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -564,64 +530,29 @@
...
+                                            label: phase.name,
+                                            permissionset: phase.permissionset,
+                                        }]) %}
+                                    {% endfor %}
+                                    <participation-phases
+                                        autoswitch-hint="{{ 'period.autoswitch.hint'|trans({ phase: 'procedure.phases.external.evaluating'|trans }) }}"
+                                        field-name="r_publicParticipationPhase"
+                                        help-text="{{ 'text.procedure.edit.external.change_phase'|trans }}"
+                                        label-text="{{ "procedure.public.phase"|trans }}"
+                                        :participation-phases="['participation', 'earlyparticipation', 'anotherparticipation']"
```
hwiem, line 546: Is it correct to hard code the phases here?

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -642,7 +573,10 @@
                                     } %}
                                     <dp-date-range-picker
                                         class="u-2-of-4"
-                                        data-cy="procedureSettingsParticipationDate"
+                                        :data-cy="{
```
hwiem, line 576: Shouldn't this prop be a string?

hwiem (changes requested):

client/js/components/procedure/basicSettings/ParticipationPhases.vue
salisdemos: obviousely not.
```suggestion
        v-if="hasPermission('field_phase_iterator')"
```

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
salisdemos: At least as correct as it was before.
See removed line templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig:112

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
salisdemos: Means. It has to be changed with the next demosplan-ui release. But for now it has to be an object. The core-PR that changed this got merged before the new demosplan-ui was released

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
salisdemos: see below

client/js/components/procedure/basicSettings/ParticipationPhases.vue
salisdemos: Now as You question: I could "hardcode" everything but the iterator.value. Would You prefer that?
Otherwise, I would follow Your suggestion and add a doc-Block and a validator.

hwiem: If it is unlikely that we want to pass in other values for these properties, I think we should "hard code" them. But a short explanation for the `iterator.value` might still be useful. :)

salisdemos: I missed, that we still need the unique name. Here we go 9808008566ea3c69576b2dd86890ce78d9b8714b

tests/frontend/unit/ParticipationPhases.spec.js
```diff
@@ -0,0 +1,89 @@
...
+    }, {
+      label: 'My Label B',
+      permissionset: 'hidden',
+      value: 'fifth'
+    }],
+    initSelectedPhase: 'first',
+    iterator: {
+      label: 'Iterator Label',
+      name: 'iterator_input_name',
+      tooltip: 'Tooltip text for the itereator',
```
hwiem, line 30:
```suggestion
      tooltip: 'Tooltip text for the iterator',
```

tests/frontend/unit/ParticipationPhases.spec.js
```diff
@@ -0,0 +1,89 @@
...
+
+  it('should be an object', () => {
+    expect(typeof ParticipationPhases).toBe('object')
+  })
+
+  it('should be named ParticipationPhases', () => {
+    expect(ParticipationPhases.name).toBe('ParticipationPhases')
+  })
+
+  it('should set the "inParticipation"-State as expected', async () => {
```
hwiem, line 44: I think tests should also document how a component is supposed to work, so I find it useful to say it explicitly (rather than 'as expected'. what is expected?), e.g. 'sets `inParticipation` to true if the selected Phase is a participation phase'. I would also test only one case in a single test, so I'd add separate tests for when `inParticipation` should be `false`.

tests/frontend/unit/ParticipationPhases.spec.js
```diff
@@ -0,0 +1,89 @@
...
+    expect(wrapper.vm.isInParticipation).toBe(false)
+
+    await wrapper.setData({
+      selectedPhase: 'noneAtall',
+    })
+
+    expect(wrapper.vm.isInParticipation).toBe(false)
+  })
+
+  it('should set the Permission-Message-Text as expected', async () => {
```
hwiem, line 67: see above

tests/frontend/unit/ParticipationPhases.spec.js
salisdemos, line 44:
```suggestion
  it('should set the "inParticipation"-State to true if the selected Phase occures in the participationPhases Array', async () => {
```

tests/frontend/unit/ParticipationPhases.spec.js
salisdemos:
```suggestion
  it('should set the Permission-Message-Text depending on the selected participation phase permissionset ', async () => {
```

tests/frontend/unit/ParticipationPhases.spec.js
```diff
@@ -0,0 +1,89 @@
...
+
+  it('should be an object', () => {
+    expect(typeof ParticipationPhases).toBe('object')
+  })
+
+  it('should be named ParticipationPhases', () => {
+    expect(ParticipationPhases.name).toBe('ParticipationPhases')
+  })
+
+  it('should set the "inParticipation"-State to true if the selected Phase occures in the participationPhases Array', async () => {
```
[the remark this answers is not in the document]
salisdemos: Maybe you have an Idea for a fitting description? I am quite new to this field so have to get used to write proper descriptions still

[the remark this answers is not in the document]
hwiem: I don't really have a good idea, but how about

`it('sets the 'inParticipation' state correctly depending on whether the selected Phase is a participation phase',`

? I'd like to keep implementation details out of the description as much as possible.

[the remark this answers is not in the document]
salisdemos:
```suggestion
  it('sets the 'inParticipation' state correctly depending on whether the selected Phase is a participation phase', async () => {
```

hwiem (approved):
lgtm, code review only

PR #2942, opened 2024-04-10
feature (ref dplan-11397): remove return type in phpdoc when return i…

jalalifardemos:
**Ticket**: https://demoseurope.youtrack.cloud/issue/DPLAN-11397/Phpstan-level-2-Error-beheben

**Fehler**: 
`PHPDoc tag @return with type bool is incompatible with native type never`

**How Test**:
`bin/{projectName} dp:phpstan -l 2`


bakridemos (approved):
lgtm

PR #2947, opened 2024-04-11
chore: remove vue-multiselect as direct dependency

spiess-demos:
When reworking TagSelect to use demosplan-ui instead of directly using vue-multiselect in https://github.com/demos-europe/demosplan-addon-demospipes/pull/88, removing the dependency was forgotten as it seems. This is done here.

### PR Checklist


hwiem (approved):
lgtm, thanks!

PR #2962, opened 2024-04-15
fix: elevate slidebar above z-fixed

spiess-demos:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11431

As the slidebar is taking the whole available page height, it should be elevated above z-fixed which is applied to fixed headers and other fixed iu elements.

### How to review/test
In the page view mentioned in the bug report, the sidebar should display above the sticky header (for the bug to be reproducable, the browser window should be not too wide).

### PR Checklist


hwiem (approved):
lgtm + works!

PR #2974, opened 2024-04-17
fix(refs DPLAN-11457): Remove "Dead" Code

salisdemos:
The Backend Code was already moved to an addon.
This was supposed to happen with that logic too. But since its not in use anymore and would be really complicated to migrate to an addon, we decided to delete the FE Part.

Along I found some corresponding Templates without any usage.

**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11457/Verfahrensdetailseite-des-Verfahrenstyps-Teilaufstellung-Regionalplan-Windenergie-fuhrt-zu-Fehlerseite

### PR Checklist

### DIFF
--- a/client/js/components/map/publicdetail/Map.vue
+++ b/client/js/components/map/publicdetail/Map.vue
@@ -943,38 +943,37 @@ export default {
                   url: this.getFeatureInfoUrlPlanningArea
                 })
                   .then(responsePr => {
-                    // If we can't check the procedure we want to get the featureInfos anyway
-                    if (responsePr.data.code === 100 && responsePr.data.success && responsePr.data.body !== null && responsePr.data.body !== '') {
-                      /* Check if coordinates are in the area of the current procedure, but only if planningarea is not set to 'all' */
-                      if (responsePr.data.body.id !== this.procedureId && this.procedureSettings.planningArea !== 'all' && hasPermission('feature_map_new_statement')) {
-                        /*
-                         * Roll back to this one when we can handle procedure versions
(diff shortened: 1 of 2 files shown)


client/js/components/map/publicdetail/Map.vue
```diff
@@ -943,38 +943,37 @@ export default {
...
-                        this.getFeatureInfoAndShowPriorityAreaPopup(remappedUrl, coordinate)
-                      }
+                    /*
+                     * If we can't check the procedure we want to get the featureInfos anyway and
+                     * if coordinates are in the area of the current procedure, but only if planningarea is not set to 'all'
+                     */
+                    if (responsePr.data.code === 100 &&
+                      responsePr.data.success &&
+                      responsePr.data.body?.id !== this.procedureId &&
+                      this.procedureSettings.planningArea !== 'all' &&
```
mussbach, line 953: I think you need to remove `this.procedureSettings.planningArea` at all from this if

mussbach (changes requested):

client/js/components/map/publicdetail/Map.vue
salisdemos: IT depends what we want to archive. IF I do so, that would change the logic and eventually display the wrong popup IF the planningArea is set to "all"

mussbach: hm, ok. I just saw `all` and empty string in the database

mussbach (approved):
lgtm, code review only

client/js/components/map/publicdetail/Map.vue
salisdemos: cant say. especially not for Prod

PR #2986, opened 2024-04-22
chore: remove menu highlighting that is not used any more

mussbach:
With the usage of the KnpLabs menu bundle the former mechanism is not used any more and could be removed

How to test:
navigate the app, highlighting should still work

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
+++ b/demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
@@ -134,13 +134,11 @@ class Permissions implements PermissionsInterface, PermissionEvaluatorInterface
     /**
      * Initialisiere die Permissions.
      */
-    public function initPermissions(UserInterface $user, ?array $context = null): PermissionsInterface
+    public function initPermissions(UserInterface $user): PermissionsInterface
     {
         $this->user = $user;
 
         $this->setInitialPermissions();
-
-        $this->initMenuhightlighting($context);
 
         // set Permissions which are user independent
         $this->setPlatformPermissions();
@@ -526,7 +524,7 @@ class Permissions implements PermissionsInterface, PermissionEvaluatorInterface
     /**
      * Setze die Rechte, die ein Verfahren betreffen.
      */
-    protected function setProcedurePermissions(): void
+    public function setProcedurePermissions(): void
     {
         // Ist Inhaberin des Verfahrens. Nur FP*-Rollen
         if ($this->ownsProcedure()) {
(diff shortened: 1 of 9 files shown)


demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
```diff
@@ -526,7 +524,7 @@ static function (Permission $permission) {
     /**
      * Setze die Rechte, die ein Verfahren betreffen.
      */
-    protected function setProcedurePermissions(): void
+    public function setProcedurePermissions(): void
```
martendemos, line 527: leads to error in \demosplan\DemosPlanCoreBundle\Permissions\ProjectPermissions::setProcedurePermissions

martendemos (changes requested):

martendemos (changes requested):
Maybe I'm missing something, but when I try to access a procedure as admin, I got an "permission denied".

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
mussbach: Thank you, unfortunately this is part of https://github.com/demos-europe/demosplan-core/pull/2991/files#diff-5e713faa2c7b521d13b127049eebd3f676f2ae97460924b7a75c9f100c033561 due to an error in my commits.

mussbach: Done in daa39454

mussbach:
> Maybe I'm missing something, but when I try to access a procedure as admin, I got an "permission denied".

I cannot reproduce that, do you receive the error with any user? @martendemos

martendemos:
> > Maybe I'm missing something, but when I try to access a procedure as admin, I got an "permission denied".
> 
> I cannot reproduce that, do you receive the error with any user? @martendemos

I tried with "Walter West" ("Fachplanung-Admin" and "Institutions-Koordination")

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
```diff
@@ -986,8 +957,6 @@ public function checkProcedurePermission(): void
     {
         // Prüfe, ob der User ins Verfahren darf
         if (null !== $this->procedure) {
-            $this->setProcedurePermissions();
```
martendemos, line 989: Shouldn't it be called somewhere else?
This may explain why access to a procedure isn't possible anymore?
(Admin hasn't the "area_admin_dashboard" permission enabled)

martendemos (changes requested):

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
mussbach: this was also accidentally pushed to #2991 thank you. reverted in [4483601](https://github.com/demos-europe/demosplan-core/pull/2986/commits/4483601256e62c20151af602c2ec269f5521ac50) but I wonder why this branch worked locally

PR #2991, opened 2024-04-22
chore: use listeners for permission setup

mussbach:
Permissions are initially set in ConfigurePermissionsListener, Procedure permissions are enhanced in AccessProcedureListener, general procedure access check is also done in AccessProcedureListener.

Nothing new is introduced in this PR, only shifted to listeners.

### How to review/test
Browse the app, permissions within and outside procedures should work as before

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/EventListener/AccessProcedureListener.php
+++ b/demosplan/DemosPlanCoreBundle/EventListener/AccessProcedureListener.php
@@ -1,0 +1,38 @@
+<?php
+declare(strict_types=1);
+
+namespace demosplan\DemosPlanCoreBundle\EventListener;
+
+use DemosEurope\DemosplanAddon\Contracts\PermissionsInterface;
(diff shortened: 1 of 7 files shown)


martendemos (commented):
Review and quick test via UI but this needs to be smoketested.
Can we install this to let QA test it?

mussbach:
We could explicitly ask QA to smoketest it after it is merged. We could also install this branch on dev but this will infer with normal testing and I think this is a "works-entirely-or-not-at-all" kind of PR as all code is only moved. But I am with you that this is a kind of risky place as it handles with permission checking. what do you think?

martendemos (approved):
> "I think this is a "works-entirely-or-not-at-all" kind of PR as all code is only moved."
> "Lets explicitly ask QA to smoketest it after it is merged"

I agree :+1:

PR #3010, opened 2024-04-24
fix(refs DPLAN-11435): Fetch Lost Hash to request userFilterSets

salisdemos:
It looks like the response from the API changed, so we have to adjust the request to match the json:api specs more strict.

**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11435/gespeicherte-Filterkombinationen-laden-nicht-A-Tabelle

With this fix, saved filterSets should be loadable again. (A-table as FPA)

### PR Checklist

- [x] Tests updated/created

### DIFF
--- a/client/js/components/procedure/admin/AdministrationProceduresList.vue
+++ b/client/js/components/procedure/admin/AdministrationProceduresList.vue
@@ -321,6 +321,11 @@ export default {
             originalStatementsCount: el.attributes.originalStatementsCount,
             statementsCount: el.attributes.statementsCount
           }))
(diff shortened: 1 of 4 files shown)


salisdemos:
> lgtm, code review only, but shouldn't this be done on release_robobsh?

I think the moment I created the PR it was somehow unclear where it has to be published first
https://github.com/demos-europe/demosplan-core/pull/3074

spiess-demos (approved):
lgtm! (code review)

PR #3020, opened 2024-04-26
use ado_sync_workflow per ssh

shumyachkin:
this PR switches to new ado workflow that is using ssh for git operations


mussbach (approved):
good luck, lgtm, thx

PR #3022, opened 2024-04-29
F dplan 11494 update pi proposals with excel imported tags

bakridemos:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11494/Nach-Excel-Import-von-verschlagworteten-Abschnitten-werden-trotzdem-keine-Vorschlage-bei-Slicing-Tagging-angezeigt

Description: 

PI proposals have to be updated with the new Excel imported tags. The 'StatementCreatedViaExcelEvent' will be dispatched with every new already segmented generated statement and a segmentation request will be send to DATA.
The only purpose to do that here is to get a pi segments proposal resource url from Data which is needed while sending segmentation confirmation request and passing imported tagged sections with Excel to DATA to make them available in the next segmentation proposals.

- [X] Update documentation


bakridemos:
redone in https://github.com/demos-europe/demosplan-core/pull/3057, target branch has to be adjusted

PR #3026, opened 2024-04-29
feat: Set permissions in Http Tests

mussbach:
With this PR it is possible to set permissions directly in integration tests like API or symfony client tests.

You need to extend the `HttpTestCase` and just call ` $this->enablePermissions(['my_permission']);` just like in other tests. You can find an example in `DemosPlanUserControllerTest`

### How to review/test
code review and run tests

### PR Checklist

- [x] Tests updated/created
- [ ] Update documentation


tests/backend/base/HttpTestCase.php
sierrazamudiodemos, line 1: Looks good, thanks for this PR!
What about adding these 2 methods here? They are needed when accessing API requests that need authorization. I did successfully an API test for reaching ProcedureResourceType, and I need the authorization to work. They are here: _tests/backend/base/AbstractApiTest.php:57_ and _tests/backend/base/AbstractApiTest.php:82_
What do you think? 

    protected function initializeUser(User $user): string
    {
        $token = $this->tokenManager->create($user);
        $userToken = new JWTUserToken($user->getDplanRolesArray(), $user, $token);
        $this->tokenStorage->setToken($userToken);

        return $token;
    }

    protected function getAdditionalHeaders(string $jwtToken, ?Procedure $procedure): array
    {
        $headers = [
            'HTTP_X-JWT-Authorization' => "Bearer $jwtToken",
        ];
        if (null !== $procedure) {
            $headers['HTTP_X_DEMOSPLAN_PROCEDURE_ID'] = $procedure->getId();
        }

        return $headers;
    }

sierrazamudiodemos (approved):
huhu @mussbach we can merge the PR, I can make the adjustments on other PRs to support authentication

PR #3032, opened 2024-04-30
chore: improve list performance

mussbach:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11522/blp-Liste-der-Verfahren-auf-der-Startseite-ladt-sehr-lang

Loading the list of procedures has a very poor performance, taking almost 10 Seconds for a list of only 388 procedures. Two main bottlenecks could be avoided for guest users, as they can neither have DraftStatements nor can they own procedures. Logged in users won't have such a long list, so the performance does (usually) not bite. 
This PR reduces the time to fetch the List for guest users from 9.57s to 1.48s which is still slow but feasible. More performance optimizations can be achieved later on, to keep this PR small I only focused on guest users.

image

### How to review/test
code review and visit blp homepage

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm

PR #3034, opened 2024-04-30
bug (ref DPLAN-1970) Remove duplicated breadcrumbs

sierrazamudiodemos:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-1970/Breadcrumbs-durcheinander-kaputt

**Description:**
- The bug  was introduced by exchanging the `match` function instead of the `switch` function (see here 92d24a2899d895fadc1cb057d994330113b922ef)
- The match function does not break inside a foreach, the switch function can break (using break (2)) to stop the switch function and the foreach
- Because the match function is used, then when a user has 2 roles, the breadcrumb is added twice.
- I wanted to keep the match function, that is why I added the if else. I want to keep the match function in case more user roles are added in the future with a custom breadcrumb

### How to review/test
- Go to ewm and login with a user that has 2 or more roles
- Then you should see the duplicated breadcrumbs are gone
- Also test with a user  that has 1 role to be safe

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Services/Breadcrumb/Breadcrumb.php
+++ b/demosplan/DemosPlanCoreBundle/Services/Breadcrumb/Breadcrumb.php
@@ -72,7 +72,7 @@ class Breadcrumb
      * @param array|null  $procedure
      * @param bool        $isOwner
      */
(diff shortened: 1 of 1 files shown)


mussbach (approved):
nice catch, thx! lgtm. Automatic code refactoring is obviously no silver bullet

PR #3035, opened 2024-05-02
pass secret to reusable workflow

shumyachkin:
pass the necessary secret to the reusable workflow, ensuring it is accessible for use within that context.


mussbach (approved):
lgtm, lets try it

PR #3036, opened 2024-05-02
Chore: simplify components

spiess-demos:
- As the tokenLength is never changed, it can also be hardcoded in the respective component
- comparison expression is simplified for better readability
- maxlength + minlength must be strings according to the validator of DpInput

### How to review/test
Nothing should be different in the newProcedure view.

### PR Checklist

### DIFF
--- a/client/js/components/procedure/admin/DpNewProcedure/CoupleTokenInput.vue
+++ b/client/js/components/procedure/admin/DpNewProcedure/CoupleTokenInput.vue
@@ -30,7 +30,7 @@
   </div>
 </template>
 <script>
-import { DpInlineNotification, DpInput, dpRpc } from '@demos-europe/demosplan-ui'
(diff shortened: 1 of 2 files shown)


gruenbergerdemos (commented):
Why do you want to hardcode it here? I think it is better to have it stored in one place (demosplan\\DemosPlanCoreBundle\\Entity\\Procedure\\ProcedureCoupleToken::TOKEN_LENGTH) rather than in two.

spiess-demos:
> Why do you want to hardcode it here? I think it is better to have it stored in one place (demosplan\DemosPlanCoreBundle\Entity\Procedure\ProcedureCoupleToken::TOKEN_LENGTH) rather than in two.

i have overseen that it is passed from backend... reverted in 0ee6b1032a6cdb049c5da41eab87b83390f6b387

PR #3041, opened 2024-05-06
chore (refs DPLAN-11547): Tus path needs to be relative

mussbach:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11547/Dateien-auf-bauleitplanung.muenchen.de-konnen-nicht-hochgeladen-werden 

When specific organisations have special domains that are mapped to a certain orga page in a customer the absolute url path may show to the customer, which will lead to cors errors. Due to them a file upload will not be possible

### How to review/test
upload any file. The upload methods are all same

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


graupnerdemos (approved):
This makes a lot of sense. Lgtm.

PR #3045, opened 2024-05-06
Fix firefox scrollbars in SegmentsList

spiess-demos:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11516/Scroll-Handle-soll-unten-festgehalten-werden

In Firefox, scrollbars appear only onHover by default. To mitigate this, the logic for displaying the fixed custom scrollbar has changed a little bit:

- in non-fullscreen mode, the custom scrollbar is shown while the native scrollbar is completely hidden via a new utility class `scrollbar-none`
- in fullscreen mode, the native scrollbar is shown and the custom scrollbar is hidden (we cannot use v-if as the element needs to be present within the dom)
- The compare function is adjusted to always compare the number values, not the strings returned by `window.getComputedStyle(el).width`
- the custom scrollbar container must have an initial height (`h-3`) to create a "hoverable" element for Firefox
- `dataTableContainerElement` is also observed to account for browser resizing

Apart from this, empty states have been tweaked in segments- and statementslist by always using the correct check (just checking for `items` always returns true even if there are no items, as this is a reactive variable)

### Related PRS

- https://github.com/demos-europe/demosplan-ui/pull/847

### How to review/test
Custom scrollbar should work also on firefox.

### PR Checklist


spiess-demos:
may be merged after demosplan-ui v0.3.13 has arrived.

sakutademos (approved):
lgtm

PR #3046, opened 2024-05-06
feat(refs DPLAN-11484): Change naming for planning agencies in settings

gruenbergerdemos:
**Ticket:** DPLAN-11484

- Translation for `planningagency.participating` changed, so we can use it at other places, too. Colon needs to be added in template directly
- Change translations to better naming


mussbach (approved):
lgtm, code review only

PR #3050, opened 2024-05-06
fix(refs dplan-11557): Only send set extents on patch

salisdemos:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11557/Karte-nicht-sichtbar-wenn-der-Benutzer-nur-Startkartenausschnitt-wahlt

Now it shouldn't matter if only one of the extents is set. It should work under all constellations.

### How to review/test

To test goto "Karteneinstellungen" and only set the boundingbox.

### PR Checklist

- [x] Tests updated/created
- [x] Update documentation


gruenbergerdemos (approved):
lgtm, thanks

PR #3059, opened 2024-05-07
feat(refs DPLAN-3570): add permissions for map settings

gruenbergerdemos:
### Ticket
DPLAN-3570

- Add missing permissions for fields
- Remove redundant input fields for territory and coordinate
- Add updating for territory and coordinates
- Show translation if there is no (default) map extent/boundingbox

### How to review/test
Verfahren -> Planunterlagen und Planzeichnung -> Startkartenausschnitt -> Geltungsbereich und Ortsbezug zeichnen -> Edit and save should work as expected

### DIFF
--- a/client/js/components/map/map/MapView.vue
+++ b/client/js/components/map/map/MapView.vue
@@ -240,7 +240,7 @@ export default {
     },
 
     center () {
-      return this.procedureCoordinates ? this.procedureCoordinates : false
+      return this.procedureCoordinates?.length > 0 ? this.procedureCoordinates : false
     }
   },
 
--- a/client/js/components/map/map/MapView.vue
+++ b/client/js/components/map/map/MapView.vue
@@ -51,12 +51,12 @@
           <dp-ol-map-set-extent
             data-cy="mapDefaultBounds"
             translation-key="map.default.bounds"
-            @extentSet="data => emitFieldUpdate({ field: 'boundingBox', extent: data })" />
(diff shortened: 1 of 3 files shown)


client/js/components/map/map/MapView.vue
```diff
@@ -240,15 +240,7 @@ export default {
...
-      if (this.procedureCoordinates) {
-        const array = this.procedureCoordinates.split(',')
-        return [
-          Number(array[0]),
-          Number(array[1])
-        ]
-      } else {
-        return false
-      }
+      return this.procedureCoordinates ? this.procedureCoordinates : false
```
salisdemos, line 243:
```suggestion
      return this.procedureCoordinates ?? false
```

salisdemos: may this can be an empty array too? then wouldn't mix types

salisdemos (changes requested):
just tiny stuff

client/js/components/map/map/MapView.vue
gruenbergerdemos: If we return an empty array, we need to change code in DpOlMap and various other places I think. Wouldn't do this now, because of time. But I added the check for empty array here [2b5514e](https://github.com/demos-europe/demosplan-core/pull/3059/commits/2b5514e07618c138ec193957c1eb5a70c3102090)

salisdemos (approved):

PR #3068, opened 2024-05-08
chore (refs PlanF-58): Add missing Hooks/Selectors for CookieBanner e…

ahmad-demos:
### Ticket
https://demoseurope.youtrack.cloud/issue/PlanF-58

- Add missing Hooks/Selectors for CookieBanner elements.


salisdemos:
I think its the wrong approach to add Hooks to the lib. The lib itself already contain similar hooks. I propose to use them.

PR #3086, opened 2024-05-13
bug(refs DPLAN-11472): Add event for phase selecting

gruenbergerdemos:
### Ticket
DPLAN-11472

- Set internal and public phase dynamically based on select
- Give data-cy as string, because component expects a string
- Add data-cy for participation phases
- Update test for participation phases
   - Mount wrapper before each test to avoid duplications in tests
   - Remove test for naming, because it is deprecated (testing implementation details is considered bad practice)
   - Add test for emitting event

### How to review/test
Verfahren -> Grundeinstellungen -> Verfahrensschritt Institiution/Verfahrensschritt Öffentlichkeit -> Based on which phase you select, "Verfahrensschritt automatisch umstellen" should be disabled or enabled (disabled for participation phases)

### PR Checklist

- [x] Tests updated/created


templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -450,10 +453,10 @@
 
                                     <dp-date-range-picker
                                         class="u-2-of-4"
-                                        :data-cy="{
+                                        :data-cy="JSON.stringify({
```
salisdemos, line 456: shouldn't this be something like `institutionPeriod` ?
The comeonent then appends `:endDate` and `startDate` 
https://github.com/demos-europe/demosplan-ui/blob/0c690b83c39d396b24c8ff2320537605e72a99b0/src/components/DpDateRangePicker/DpDateRangePicker.vue#L14

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -573,10 +579,10 @@
                                     } %}
                                     <dp-date-range-picker
                                         class="u-2-of-4"
-                                        :data-cy="{
+                                        :data-cy="JSON.stringify({
```
salisdemos, line 582: see above

tests/frontend/unit/ParticipationPhases.spec.js
```diff
@@ -1,53 +1,47 @@
...
+          value: '1'
+        }
+      }
+    })
   })
 
-  it('should be named ParticipationPhases', () => {
-    expect(ParticipationPhases.name).toBe('ParticipationPhases')
+  it('should be instantiated', () => {
+    expect(wrapper).toBeTruthy()
```
salisdemos, line 41: why this change?

tests/frontend/unit/ParticipationPhases.spec.js
```diff
@@ -1,53 +1,47 @@
...
 
-  it('sets the `inParticipatin` state correctly depending on whether the selected Phase is a participation phase', async () => {
-    const localVue = createLocalVue()
-
-    const wrapper = shallowMountWithGlobalMocks(ParticipationPhases, {
-      propsData: componentsPropsData,
-      localVue
-    })
-
+  it('should set the `inParticipatin` state correctly depending on whether the selected Phase is a participation phase', async () => {
```
salisdemos, line 44:
```suggestion
  it('should set the `inParticipation` state correctly depending on whether the selected Phase is a participation phase or not', async () => {
```

salisdemos (changes requested):

client/js/components/procedure/basicSettings/AutoSwitchProcedurePhaseForm.vue
```diff
@@ -64,10 +64,10 @@
             enforce-plausible-dates
             :min-date="startDate"
             required
-            :data-cy="{
+            :data-cy="JSON.stringify({
                 endDate: dataCyEndDate,
                 startDate: dataCyStartDate
-            }"
+            })"
```
sakutademos, line 70: I think it would be better to either directly change the type to an object in the DateRangePicker or to only pass a string, such as `data-cy="newProcedureForm"`. This is because in DpDateRangePicker at [lines 14 and 28]([url](https://github.com/demos-europe/demosplan-ui/blob/main/src/components/DpDateRangePicker/DpDateRangePicker.vue#L14)), we are already passing `${dataCy}:startDate` and `${dataCy}:endDate`.

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
gruenbergerdemos: Good to know, thanks for spotting

client/js/components/procedure/basicSettings/AutoSwitchProcedurePhaseForm.vue
gruenbergerdemos: Thanks, changed in [38972aa](https://github.com/demos-europe/demosplan-core/pull/3086/commits/38972aa27cd8f569f68b315b090a6107fc08b0a9)

tests/frontend/unit/ParticipationPhases.spec.js
gruenbergerdemos: Good point. I now removed testing for object, too (like the test for name), because testing for implementation details is considered bad practice and provides little to no value. Name() is also deprecated in vue test utils and will be removed soon (see https://v1.test-utils.vuejs.org/upgrading-to-v1/#name). 
But I also saw that we test name and object very often. Should we discuss this with the team?

salisdemos (approved):

PR #3088, opened 2024-05-14
fix(refs DPLAN-3570): Add resetbutton to map settings

salisdemos:
### Ticket
DPLAN-3570

on the mapSettingspage there now is a "reset" Button. It should reset the settings  (not the map zoom) to the data stored before saved. 

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/map/admin/MapAdmin.vue
+++ b/client/js/components/map/admin/MapAdmin.vue
@@ -98,15 +98,16 @@
         type="submit"
         @click="() => save(true)" />
       <dp-button
+        v-if="hasPermission('area_admin_single_document')"
         color="secondary"
         data-cy="abort"
         :href="Routing.generate('DemosPlan_element_administration', { procedure: procedureId })"
         :text="Translator.trans('abort')" />
       <dp-button
+        v-else
         color="secondary"
         data-cy="reset"
         :text="Translator.trans('reset')"
-        variant="outline"
         @click="reset" />
     </div>
   </div>
--- a/client/js/components/map/admin/MapAdmin.vue
+++ b/client/js/components/map/admin/MapAdmin.vue
@@ -82,35 +82,41 @@
       }" />
 
     <div class="layout__item u-1-of-1 text-right u-mt-0_5 space-inline-s">
-      <input
-        class="btn btn--primary"
+      <dp-button
+        data-cy="save"
         :disabled="!areScalesSuitable"
-        type="submit"
         name="saveConfig"
-        :value="Translator.trans('save')"
-        @click="() => save()">
-      <input
-        v-if="hasPermission('area_admin_single_document')"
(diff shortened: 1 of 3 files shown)


spiess-demos (changes requested):
i think some users will be confused by the (not-so-clear) difference between the abort- and reset buttons. why not implementing it [next to the buttons](https://github.com/demos-europe/demosplan-core/blob/4a1b2f2388e0c0606bfefc3bb602330bc774f4ea/client/js/components/map/map/MapView.vue#L60) above the map?

client/js/components/map/admin/MapAdmin.vue
```diff
@@ -82,42 +82,49 @@
...
+        color="secondary"
+        data-cy="abort"
+        :href="Routing.generate('DemosPlan_element_administration', { procedure: procedureId })"
+        :text="Translator.trans('abort')" />
+      <dp-button
+        color="secondary"
+        data-cy="reset"
+        :text="Translator.trans('reset')"
+        variant="outline"
+        @click="reset" />
```
spiess-demos, line 110: I find it somewhat unusual to find both buttons next to each other. Usually i find either "reset", or "abort".

So we got those...
- "Speichern"
- "Speichern und zurück" (only in projects where there is a location to navigate back to)
- "Abbrechen" (which should also only show up in projects where there is a location to navigate back to, but it is currently always shown)
- "Zurücksetzen" (which has the same effect on data entry as "Abbrechen" but stays on the same page)

I'd suggest to offer this combination of buttons for projects "with an overview page" (leaving out the "zurücksetzen" because users can afford nearly the same thing with "abbrechen"):
**[Speichern] [Speichern und zurück] [Abbrechen]**

And for projects without the overview we are left with
**[Speichern] [Zurücksetzen]**

client/js/components/map/admin/MapAdmin.vue
```diff
@@ -82,42 +82,49 @@
...
+      <dp-button
+        color="secondary"
+        data-cy="abort"
+        :href="Routing.generate('DemosPlan_element_administration', { procedure: procedureId })"
+        :text="Translator.trans('abort')" />
+      <dp-button
+        color="secondary"
+        data-cy="reset"
+        :text="Translator.trans('reset')"
+        variant="outline"
```
spiess-demos, line 109: I never documented it but within a button group, buttons should only differenciate by color, not by variant.

spiess-demos (changes requested):

client/js/components/map/admin/MapAdmin.vue
salisdemos: all right 46713c14e5376654601d7d0c02b140fccbf25677

spiess-demos (approved):
lgtm (code review)

PR #3089, opened 2024-05-14
fix(DPLAN-2231): location validation issue on the basic settings page

sakutademos:
### Ticket
[DPLAN-2231](https://demoseurope.youtrack.cloud/issue/DPLAN-2231/Fehlermeldung-das-Verfahren-nicht-verortet-ist-in-jeden-Assistent-schritt-angezeigt)

**Description:** This PR resolves the error validation issue on the _Grundeinstellungen_ page. The error messag: _"Bitte geben Sie unter "Verortung des Verfahrens" einen Ortsbezug in der Karte ein."_ was appearing on every assistant step. This occurred due to the hardcoded `dplan.notify.error(Translator.trans('public.participation.relation.map.missing'))` in the `setProcedureCoordinateValidState` method. It was triggered whenever the `currentProcedureCoordinate` value was empty. It should only validate if `#r_publicParticipationPhase!==configuration and #r_phase!==configuration`

- adjust `setProcedureCoordinateValidState` function and check the `requiredMapIsValid` only if the` r_publicParticipationPhase `and `r_phase` phases are not set to `configuration`
- add the validation attribute `data-dp-validate-error-fieldname` to enhance validation
- add a `name` attribute to the <fieldset> tag to prevent certain warnings from the `removeFormHiddenField` function (which requires a name or id attribute)

### How to review/test
anmelden ->verfahren ->Grundeinstellungen ->Assistant starten ->Speichern und zum nächsten Schritt

### Linked PRs
- demosplan-ui #{[PR-850](https://github.com/demos-europe/demosplan-ui/pull/850)}

- [ ] Tests updated/created
- [ ] Update documentation


client/js/components/procedure/basicSettings/AutoSwitchProcedurePhaseForm.vue
```diff
@@ -64,10 +64,7 @@
             enforce-plausible-dates
             :min-date="startDate"
             required
-            :data-cy="{
-                endDate: dataCyEndDate,
-                startDate: dataCyStartDate
```
spiess-demos, line 69: As we're using `AutoSwitchPrcedurePhaseForm` in two places, we must preserve the possibility to pass a unique value here. The differentiation between start- and enddate is handled by DpDateRangePicker and can be removed here, so i'd suggest to just have one prop `dataCy` which could then receive the values of `administrationEditForm:procedurePhasePeriod` and `administrationEditForm:procedurePhasePeriodPublic` (or something similar) from where the component is used.

spiess-demos (changes requested):

hwiem (dismissed):
code lgtm

client/js/components/procedure/basicSettings/AutoSwitchProcedurePhaseForm.vue
sakutademos: so you mean smth. like that: https://github.com/demos-europe/demosplan-core/pull/3089/commits/122220aad2f1df711b48358ec63645467443a77a ?

the result in DpDatepicker will be `administrationEditForm:procedurePhasePeriod:startDate` and `administrationEditForm:procedurePhasePeriod:endDate`. Is it okay to have such a long name, or would it be better to use smth. like `procedurePhasePeriod:startDate` and `procedurePhasePeriod:endDate `

sakutademos: after merging the main branch into mine I realized that it was done already, so my comment is outdated....

spiess-demos (approved):
now it all makes sense to me... i did not even know that fieldset may have a name attr. crazy html :) 

lgtm (code review)

PR #3090, opened 2024-05-14
f DPLAN-11596 copy blueprint legend while procedure creation

bakridemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-11596

Description: 
we have till now a universal legend which references only the to blueprint procedure and can not be applied anymore to all other procedures. To make the legend available, the method 'copyLegend' copy the blueprint legend file and make it refer to the created procedure.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
@@ -734,7 +734,7 @@ class ProcedureService extends CoreService implements ProcedureServiceInterface
      */
     public function getAuthorizedUsers(
         $procedureId,
-        User $user = null,
+        ?User $user = null,
         $excludeUser = false,
         $excludeProcedureAuthorizedUsers = true
     ): Collection {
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
```diff
@@ -2704,6 +2710,36 @@ private function copyFromBlueprint(array $data, string $blueprintId, Procedure $
...
+
+            if ('' === $legendStringFromBlueprint) {
+                return;
+            }
+
+            // copy legend
+            $copiedFile = $this->fileService->copyByFileString($legendStringFromBlueprint, $newProcedure->getId());
+
+            // set planPDF with the referenced legends file and update procedure setting
+            $newProcedurePlanPdf = 'kopie_von'.$copiedFile->getName().':'.$copiedFile->getId().':'.$copiedFile->getMimetype();
```
mussbach, line 2732: I would not call it 'kopie_von'. I would just use `$newProcedurePlanPdf = $copiedFile->getFileString()`

mussbach (changes requested):
one remark, otherwise fine

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/3090/commits/5777f37e617b6f10ee5c2c5d71cbd81579cafeaf , thx

mussbach (approved):
lgtm, thx

PR #3091, opened 2024-05-14
F dplan 11483 additional permissions paid orga

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-11483/8-2-Als-Planungsburo-brauche-ich-eine-Admin-Rolle-um-Verfahren-selbst-anlegen-und-durchfuhren-zu-konnen.

### Description
**Goal** Allow for this case and for the future to add permissions based on certain conditions. In this case, add the permissions 'feature_admin_new_procedure' to be allowed for certain orga, role and customer

**Step 1** With SUPPORT role, update orga: click on checkbox "Darf Verfahren anlegen": 
- If checked, then a new permission (feature_admin_new_procedure) is created with the following criteria: user role PRIVATE_PLANNING_AGENCY, the current orga id and the current customer id
- The SUPPORT role also has a new permission 'feature_orga_new_procedure_edit' that enables this checkbox "Darf Verfahren anlegen" to be displayed.

**Technical details:** 
a) New Entity demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionService.php was created with index: permission, role,customer,orga so that it is not possible to add repeated permissions
b) New Service demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionService.php was created. These are the ground work for permissions. Some examples:
- If permission has orga, role, customer not null (like the case of this pr) then the permission is given for the orga, role, customer combination
- if permission has role and customer not null, and orga null: then the permission is given for all orgas on that particular customer and role
- and so on. That said: null is a wildcard
- c) Only users with Support Role and permission 'feature_orga_new_procedure_edit' are able to click this chechbox. See demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php and demosplan/DemosPlanCoreBundle/ResourceTypes/OrgaResourceType.php

**Step 2** With  PRIVATE_PLANNING_AGENCY role in a Planing Office is logged-in on the updated orga/customer (of step 1), the user will be able to create a Procedure and edit it.

**Technical details:**
a) When a procedure is created, by default the PlaningOffices and AuthorizedUsers are copied from the Blueprint. We cannot do that in this case, as the orga that is creating the procedure should be the only one that can access it. This pr sets the PlaningOffices as the current orga as long as the user has the permission feature_admin_new_procedure. See this file demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php that 
b) If the Support Role of step 1 removes the permission of the orga to creare Procedures, the Orga should still be able to keep editing the created Procedures (that is why it is important to do a) point.

### How to review/test

1. Pick a user of an orga that has Fachplanung-Planungsbüro
2. . Logged in and check that you cannot create Procedures
3.  Logged In as support role
4. Update any field of the orga that is Planungs buro
5. Then logged in as the same user as step 1
6. Create a procedure
7. Go back to support role and remove the permission
8. Go back to the user role Fachplanung-Planungsbüro. You should be able to still edit your procedures and not create new ones

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * (c) 2010-present DEMOS plan GmbH, for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+namespace demosplan\DemosPlanCoreBundle\Entity\Permission;
+
+use DateTime;
+use DemosEurope\DemosplanAddon\Contracts\Entities\AccessControlPermissionInterface;
```
martendemos, line 16: Unable to find `AccessControlPermissionInterface`

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2024/05/Version20240513094920.php
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+class Version20240513094920 extends AbstractMigration
```
martendemos, line 20: Error on execution:
`An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1005 Can't create table `blp_release`.`access_control_permissions` (errno: 150 "Foreign key constraint is incorrectly formed")`
Not sure at all if this may be my bad, need

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+    private $modificationDate;
+
+    /**
+     * @var DateTime
+     *
+     * @Gedmo\Timestampable(on="update")
+     *
+     * @ORM\Column(type="datetime", nullable=false)
+     */
+    private $deletionDate;
```
martendemos, line 74: regarding to you question via slack: this can be removed, if there is no specialized need for that.

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+use DemosEurope\DemosplanAddon\Contracts\Entities\CustomerInterface;
+use DemosEurope\DemosplanAddon\Contracts\Entities\OrgaInterface;
+use DemosEurope\DemosplanAddon\Contracts\Entities\RoleInterface;
+use DemosEurope\DemosplanAddon\Contracts\Entities\UuidEntityInterface;
+use demosplan\DemosPlanCoreBundle\Entity\CoreEntity;
+use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Mapping\Annotation as Gedmo;
+
+/**
+ * Stores the information about the current phase of a procedure.
```
martendemos, line 26: Naming:
In the context of a procedure, a phase is one of the listed in procedure_phases.yml.
So this naming is -at least to me- confusing. 
What is meant with "phases" in this context?
But Im not fully understand, what this Entity

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+     * @ORM\JoinColumn(name="orga_id", referencedColumnName="_o_id", nullable=true, onDelete="SET NULL")
+     */
+    protected $organisation;
+
+    /**
+     * @var RoleInterface|null
+     *
+     * @ORM\ManyToOne(targetEntity="demosplan\DemosPlanCoreBundle\Entity\User\Role")
+     *
+     * @ORM\JoinColumn(referencedColumnName="_r_id", nullable=true, onDelete="SET NULL")
```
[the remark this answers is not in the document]
mussbach: It may be related to any user in a customer (with any role) or any user within a procedure

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+     */
+    protected $role;
+
+    /**
+     * @var CustomerInterface|null
+     *
+     * @ORM\ManyToOne(targetEntity="demosplan\DemosPlanCoreBundle\Entity\User\Customer")
+     *
+     * @ORM\JoinColumn(referencedColumnName="_c_id", nullable=true, onDelete="SET NULL")
```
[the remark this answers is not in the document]
mussbach: It may be the case that a dynamic permission is independent of a customer, indeed

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
```diff
@@ -145,9 +147,34 @@ public function initPermissions(UserInterface $user): PermissionsInterface
         // set Permissions which are dependent on role but independent of procedure
         $this->setGlobalPermissions();
 
+        // set Permissions which are dependent on orga
+        $this->setOrgaPermissions();
```
mussbach, line 151: couldn't this be more generic like `loadDynamicPermissions`?

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionPerOrgaService.php
```diff
@@ -0,0 +1,92 @@
...
+use demosplan\DemosPlanCoreBundle\Repository\AccessControlPermissionRepository;
+
+/**
+ * This file is part of the package demosplan.
+ *
+ * (c) 2010-present DEMOS plan GmbH, for more information see the license file.
+ *
+ * All rights reserved
+ */
+class AccessControlPermissionPerOrgaService extends CoreService
```
mussbach, line 27: Why did you build this specifically for Orgas?

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+    private DateTime $creationDate;
+
+    /**
+     * @var DateTime
+     *
+     * @Gedmo\Timestampable(on="update")
+     *
+     * @ORM\Column(type="datetime", nullable=false)
+     */
+    private $modificationDate;
```
MoritzMandler, line 65: could be DateTime type hinted

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+     * @ORM\Column(type="datetime", nullable=false)
+     */
+    private $deletionDate;
+
+    /**
+     * @var OrgaInterface|null
+     *
+     * @ORM\ManyToOne(targetEntity="demosplan\DemosPlanCoreBundle\Entity\User\Orga")
+     *
+     * @ORM\JoinColumn(name="orga_id", referencedColumnName="_o_id", nullable=true, onDelete="SET NULL")
```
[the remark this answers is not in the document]
MoritzMandler: This could be type strict 'protected ?OrgaInterface $organisation;

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+    protected $organisation;
+
+    /**
+     * @var RoleInterface|null
+     *
+     * @ORM\ManyToOne(targetEntity="demosplan\DemosPlanCoreBundle\Entity\User\Role")
+     *
+     * @ORM\JoinColumn(referencedColumnName="_r_id", nullable=true, onDelete="SET NULL")
+     */
+    protected $role;
```
MoritzMandler, line 92: strict type possible?

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+
+    public function setPermission(string $permission): void
+    {
+        $this->permission = $permission;
+    }
+
+    /**
+     * @param OrgaInterface $orga
+     */
+    public function setOrga($orga)
```
MoritzMandler, line 121: strict types instead of @param hints would be nice.

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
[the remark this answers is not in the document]
martendemos: This would mean there is a special (instransparent) logic: in case of null here, it means all.
Do wo rather create an entry for each orga in this case?
In case of a new orga in the DB, this one would automatically have the related permission. Is this intended?

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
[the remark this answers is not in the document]
martendemos: same as above:
A new user would be automatically/initially have the related permission, but this is intended?

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionPerOrgaService.php
mussbach: The big picture I could imagine is that the dynamic permissions can be applied at any application level like global, customer, organization, procedure (sometimes in the future perhaps even user) in combination with any (or specific or none) roles. Permissions could then be granted in the interface when and how needed. 
I think we will need a list of permissions, that are available for this dynamic mechanism. It probably would be best if you introduce it now from the beginning of this feature.

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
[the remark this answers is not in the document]
mussbach: I would favor against an entry for any organization. This does not scale and you would need to keep it in sync with any changes of the organizations and new dynamic permissions. I would think that `null` of any of these fields mean "irrelevant for permission calculation". By default the dynamic permission is deactivated and only if a row that matches the current context is found it is enabled. When a row exists with a permission that has only nulls in the "calculation" fields (orga, customer, role, procedure) it would always be granted.

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionPerOrgaService.php
mussbach: Not sure whether we would want to migrate existing permissions into the database. I would rather stress test this new concept and decide then

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
```diff
@@ -360,6 +365,18 @@ public function updateOrgaAction(
                 // explicitly set that show list may be updated
                 $userHandler->setCanUpdateShowList(true);
             }
+            $role = $roleHandler->getUserRolesByCodes([RoleInterface::PRIVATE_PLANNING_AGENCY])[0];
+            if (is_array($orgaDataArray['attributes'])
+                && array_key_exists('has_paid', $orgaDataArray['attributes'])) {
```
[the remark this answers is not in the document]
salisdemos: shouldn't it be called something like "can_create_procedures" ?

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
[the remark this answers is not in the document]
sierrazamudiodemos: hey @mussbach @martendemos I feel a bit lost here, how we would realize that an orga has paid and can create procedures? how would a row in this table would look like?

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,141 @@
...
+     */
+    protected string $id;
+
+    /**
+     * @ORM\Column(name="permission", type="string", nullable=false)
+     */
+    protected string $permission = '';
+
+    /**
+     * @Gedmo\Timestampable(on="create")
```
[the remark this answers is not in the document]
sierrazamudiodemos: Check new version, done

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: Check new version, done

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: @MoritzMandler what do you mean by stric types?

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: Updated in new version, also migration updated demosplan/DemosPlanCoreBundle/DoctrineMigrations/2024/05/Version20240521084035.php

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2024/05/Version20240513094920.php
sierrazamudiodemos: Updated in new version, also migration updated demosplan/DemosPlanCoreBundle/DoctrineMigrations/2024/05/Version20240521084035.php

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: Removed already

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: Updated in new version

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
sierrazamudiodemos: Updated in new version

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
```diff
@@ -360,6 +365,17 @@ public function updateOrgaAction(
                 // explicitly set that show list may be updated
                 $userHandler->setCanUpdateShowList(true);
             }
+
+            if ($permissions->hasPermission('feature_orga_new_procedure_edit') && is_array($orgaDataArray['attributes'])
```
mussbach, line 369: Why did you put it here? Feels wrong to me. This should live in Permissions imho

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
```diff
@@ -0,0 +1,130 @@
...
+     * @ORM\JoinColumn(referencedColumnName="_c_id", nullable=true, onDelete="CASCADE")
+     */
+    protected $customer;
+
+    public function getId(): string
+    {
+        return $this->id;
+    }
+
+    public function getPermission(): string
```
mussbach, line 97: perhaps this should be something like getPermissionString, as this is no Permission object?

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionService.php
```diff
@@ -0,0 +1,108 @@
...
+     * @param OrgaInterface     $orga
+     * @param CustomerInterface $orga
+     * @param string            $roles
+     */
+    public function canCreateProcedure($orga, $customer, $roles): bool
+    {
+        // Check if the user has the permission to create a procedure
+        $permissions = $this->getPermission($orga, $customer, $roles);
+
+        return in_array('feature_admin_new_procedure', $permissions);
```
mussbach, line 106: we should provide an enum, constant whatever to avoid string comparison here

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
```diff
@@ -834,6 +836,8 @@ public function addProcedureEntity(array $data, string $currentUserId): Procedur
             $blueprintId = $data['copymaster'] ?? null;
             $blueprintId = $blueprintId instanceof Procedure ? $blueprintId->getId() : $blueprintId;
             $newProcedure = $this->setAuthorizedUsersToProcedure($newProcedure, $blueprintId, $currentUserId);
+            $newProcedure = $this->setPlanningOffices($newProcedure, $blueprintId, $currentUserId);
```
mussbach, line 839: this looks weird. Is it correct?

demosplan/DemosPlanCoreBundle/ResourceTypes/OrgaResourceType.php
```diff
@@ -216,6 +219,14 @@ protected function getProperties(): array
                 ->readable(false, $this->getAllowedRoles(...));
         }
 
+        if ($this->currentUser->hasPermission('feature_orga_new_procedure_edit')) {
+            $properties[] = $this->createAttribute($this->canCreateProcedures)->readable(true, function (Orga $orga): bool {
+                $currentCustomer = $this->currentCustomerService->getCurrentCustomer();
+
+                return $this->accessControlPermissionService->canCreateProcedure($orga, $currentCustomer, RoleInterface::PRIVATE_PLANNING_AGENCY);
```
mussbach, line 226: can we avoid the role check here somehow? Roles should only be used in Permissions.php

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: sry Mortiz I just saw your comment, I will apply it **This could be type strict 'protected ?OrgaInterface $organisation;**

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
sierrazamudiodemos: I agree, it looks weird and maybe it can be solved differently. Let me explain the context:

a) When a procedure is created, by default the PlaningOffices and AuthorizedUsers are copied from the Blueprint. We cannot do that in this case, as the orga that is creating the procedure should be the only one that can access it. This pr sets the PlaningOffices as the current orga as long as the user has the permission feature_admin_new_procedure.
b) If the Support Role of step 1 removes the permission of the orga to creare Procedures, the Orga should still be able to keep editing the created Procedures (that is why it is important to do a) point.

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: done 4729c7e819bb26b566b5591394a88a3332ef8d97

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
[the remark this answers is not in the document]
martendemos: @sierrazamudiodemos I thought it would be a new row with an organisation an the new permission "allowed to create procedure bla bla". Not sure if the customer and/or the role would be filled but I guess so.
That would means, that the permission is enabled for the specific orga and imply that this permission is not for all orgas (of the customer) enabled by default.

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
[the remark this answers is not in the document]
martendemos: same kind of answer as above

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
martendemos: I do not see the change, is it pushed?

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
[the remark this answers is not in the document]
mussbach: In my mind `null` in the orga field would mean in this context for the calculation of the permissions "any organization", whereas an id would bound the permission to this specific orga. But this is open to discussion. I am with you @martendemos that both models have pros and cons

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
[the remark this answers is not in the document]
martendemos: minor: Im not sure if we want to define a DB-sited default value.
In case of inserting new rows by bypassing doctrine, like an direct SQL, the current timestamp would also be set.
We do not define this as best practice or something afaik, so this is more a question than a request.
```
    /**
     * @Gedmo\Timestampable(on="create")
     *
     * @ORM\Column(type="datetime", nullable=false,  options={"default":"CURRENT_TIMESTAMP"})
     */
    private DateTime $creationDate;
```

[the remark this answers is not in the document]
martendemos: additionally: I think the @var ist not needed in case of typed property. (@MoritzMandler)

config/permissions.yml
```diff
@@ -1237,6 +1237,13 @@ feature_orga_branding_edit:
     label: 'Permission to enable editing of orga branding (css styling).'
     loginRequired: true
     parent: [area_manage_orgas, area_manage_orgas_all]
+feature_orga_new_procedure_edit:
```
martendemos, line 1240: If it possible I would ask to rename the permission to something (in my mind) more speakable/readable like one of these:
```
feature_manage_procedure_creation_permission
feature_administrate_procedure_creation_permission
feature_set_procedure_creation_permission
feature_modify_procedure_creation_permission
```
What do you think?

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionService.php
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * This file is part of the package demosplan.
+ *
+ * (c) 2010-present DEMOS plan GmbH, for more information see the license file.
+ *
+ * All rights reserved
+ */
+class AccessControlPermissionService extends CoreService
+{
+    public const CAN_CREATE_PROCEDURES = 'feature_admin_new_procedure';
```
martendemos, line 31: minor naming:
`CREATE_PROCEDURES_PERMISSION` (my favor)
`PERMITTED_TO_CREATE_PROCEDURES`
 or even
`PERMITTED_TO_CREATE_PROCEDURES_PERMISSION`

This way it is maybe more clear that this is a string which holds one of your permissions?! :thinking:

config/permissions.yml
sierrazamudiodemos: Good idea, I will update it

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
MoritzMandler: setOrga(?OrgaInterface $orga)

MoritzMandler: but I think you need to change the property too:
protected ?OrgaInterface $organisation;

config/permissions.yml
sierrazamudiodemos: what about feature_orga_create_procedure ?

client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
```diff
@@ -82,7 +82,7 @@
           :id="organisation.id + 'addressHouseNumber'"
           class="w-full u-mb-0_5"
           style="height: 27px;"
-          data-cy="addressJouseNumber"
+          data-cy="orgaFormField:addressJouseNumber"
```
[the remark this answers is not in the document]
sierrazamudiodemos: Fixed here 26ea462e6af6efe435e86d1b1d016765ca87681e

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: done here 219857ca07df9886f4701178c08933285e6b2b9f

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
sierrazamudiodemos: done here 36ce1db346c5c2cc0057a1d44f6a60611e3b7cff

demosplan/DemosPlanCoreBundle/Permissions/Permissions.php
sierrazamudiodemos: done here https://github.com/demos-europe/demosplan-core/commit/219857ca07df9886f4701178c08933285e6b2b9f

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
sierrazamudiodemos: done here https://github.com/demos-europe/demosplan-core/commit/219857ca07df9886f4701178c08933285e6b2b9f

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
```diff
@@ -2114,6 +2118,26 @@ protected function setAuthorizedUsersToProcedure(Procedure $newProcedure, $bluep
...
+     * @return Procedure
+     *
+     * @throws Exception
+     */
+    protected function addCurrentOrgaToPlanningOffices(Procedure $newProcedure, $currentUserId)
+    {
+        $currentUser = $this->userService->getSingleUser($currentUserId);
+
+        if ($this->accessControlPermissionService->hasPermission(AccessControlPermissionService::CREATE_PROCEDURES_PERMISSION, $currentUser->getOrga(), $this->customerService->getCurrentCustomer(), $currentUser->getRoles())) {
+            $newProcedure->addPlanningOffice($currentUser->getOrga());
```
mussbach, line 2135: You need to ensure that orga has orgaType OPAUTH. Not sure what happens if there are orgas in the planningOffices property that are no planning offices

sierrazamudiodemos: gotcha, would you put that in the AccessPermissionService or in the ProcedureService? I guess in the AcessPermissionService?

sierrazamudiodemos: it is a business logic, but not sure where is the right place to put it

demosplan/DemosPlanCoreBundle/Entity/Permission/AccessControlPermission.php
martendemos: pls add the return type to all setters
`public function setRole(?RoleInterface $role): void`

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
sierrazamudiodemos: hey @mussbach 7b4032cbf94d4ae086424ff8e8271161e3a7456c I tried to make general this: Check permission if it is allowed for the orga - customer. What do you think? does it make sense? otherwise I can make the check only when adding the planing offices

salisdemos (commented):
Die we think about how to manage the "exposed" flag for the dynamic permissions? Coudln't find anything about that in this PR. Or is it not necessary?

sierrazamudiodemos:
> Die we think about how to manage the "exposed" flag for the dynamic permissions? Coudln't find anything about that in this PR. Or is it not necessary?

Hey @salisdemos could you elaborate for me a bit more? I dont understand what you mean by expose flag.
IF you mean the permission as a constant, I put it here: demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionService.php
If you mean a permission to allow a customer support role to see the checkbox, it is added here under this name feature_manage_procedure_creation_permission

salisdemos:
> > Die we think about how to manage the "exposed" flag for the dynamic permissions? Coudln't find anything about that in this PR. Or is it not necessary?
> 
> Hey @salisdemos could you elaborate for me a bit more? I dont understand what you mean by expose flag. IF you mean the permission as a constant, I put it here: demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlPermissionService.php If you mean a permission to allow a customer support role to see the checkbox, it is added here under this name feature_manage_procedure_creation_permission

Sorry. I think I have overseen the permissions.yml. They are still defined like before. There You can set a flag "exposed" to make it available for the vue/js context.
Thanks for the explanation.

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
martendemos: > You need to ensure that orga has orgaType OPAUTH. Not sure what happens if there are orgas in the planningOffices property that are no planning offices
+1 for that. Unfortunately it seems that the Type isn't checked in other cases :(
I suggest to solve this kind of typecheck by an actual strict use of a specific type.
Which means a own class for each orgatype (extending a kind of base-orga-class) But this isn't the PR to do this.

Anyway I do not see the check for the orga type, do I miss it?

sierrazamudiodemos: @mussbach @martendemos 
- here the commit for checking exactly only that orga type and permission combination: 
9ca77e869f5f9fcba11dea1d0116d185db0bb68f

In the commit you will see this line:  ` return in_array(OrgaTypeInterface::PLANNING_AGENCY, $orga->getTypes($this->globalConfig->getSubdomain(), true));`
I decided to check the subdomain and also if the orga is approved. as I understand an orga can be different types based on the customer

- here a commit for adding missed factories for the tests: 6bf7755f9e180fe6ca8c9d0330ad83da80b371de

mussbach (approved):
No blockers any more. Thanx a ton for this high quality PR for a really difficult problem

PR #3097, opened 2024-05-15
test: Remove tests for name and object

gruenbergerdemos:
- Remove tests for component name and object, because testing for implementation details is considered bad practice and has no real value most of the time.
- Add some tests for empty test suites


tests/frontend/unit/DpProcedureCoordinateInput.spec.js
```diff
@@ -11,13 +11,40 @@ import DpProcedureCoordinateInput from '@DpJs/components/procedure/basicSettings
...
-    const instance = shallowMountWithGlobalMocks(DpProcedureCoordinateInput, {})
-    expect(instance.html()).toEqual(expect.not.stringContaining('class="btn--disabled"'))
+  it('should update coordinates correctly', () => {
+    wrapper.vm.updateCoordinates(['789', '012'])
+    expect(wrapper.vm.latitudeValue).toBe('789')
+    expect(wrapper.vm.longitudeValue).toBe('012')
+  })
+
+  it('should convert string to float correctly', () => {
+    expect(wrapper.vm.convertToFloat('123.456')).toBe(123.456)
```
muellerdemos, line 39:
```suggestion
   it('should convert string to float correctly', async () => {
    await wrapper.setProps({coordinate: ['123', '456']})

    expect(typeof wrapper.vm.latitudeValue).toBe('string')
    expect(typeof wrapper.vm.longitudeValue).toBe('string')
    expect(typeof wrapper.vm.convertToFloat(wrapper.vm.latitudeValue)).toBe('number')
    expect(typeof wrapper.vm.convertToFloat(wrapper.vm.longitudeValue)).toBe('number')
  })
```

I think this tests more appropriate the conversion of the values. Also writing it like that makes sure the interaction between setting props and having valid lat/long values is correct (imho)

tests/frontend/unit/DpProcedureCoordinateInput.spec.js
```diff
@@ -11,13 +11,40 @@ import DpProcedureCoordinateInput from '@DpJs/components/procedure/basicSettings
...
+    expect(wrapper.vm.convertToFloat('123,456')).toBe(123.456)
+  })
+
+  it('should enable the button with valid and disable with invalid input from props', async () => {
+    await wrapper.setProps({ coordinate: ['123', '456'] })
+    expect(wrapper.html()).not.toContain('disabled="disabled"')
+
+    await wrapper.setProps({ coordinate: ['abc', '456'] })
+    expect(wrapper.html()).toContain('disabled="disabled"')
```
muellerdemos, line 48: I like these kinds of tests!

muellerdemos (dismissed):
Overall this lgtm and thanks for adding a quite good amount of test here. I have some discusions/improvement suggestions but these are non blocking from my perspective.

tests/frontend/unit/DpBulkEditStatement.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    wrapper.setData({ options: { newAssignee: { checked: true, value: '' } } })
+    await wrapper.vm.$nextTick()
+    expect(wrapper.vm.options.newAssignee.checked).toBe(true)
+  })
+
+  it('should disable the newAssignee option when unchecked', async () => {
+    wrapper.setData({ options: { newAssignee: { checked: false, value: '' } } })
+    await wrapper.vm.$nextTick()
+    expect(wrapper.vm.options.newAssignee.checked).toBe(false)
+  })
```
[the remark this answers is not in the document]
gruenbergerdemos: Good point! I agree it doesn't make to much sense. Changed it in [0e51694](https://github.com/demos-europe/demosplan-core/pull/3097/commits/0e516945253d0d31e89b2468214ca05e0dec584f)

tests/frontend/unit/DpProcedureCoordinateInput.spec.js
gruenbergerdemos: I think that's a valid test, but imo it does not replace the existing test, because it checks the method with a variety of inputs, including a string with a comma. I added your approach as a new test in [f9f5ca4](https://github.com/demos-europe/demosplan-core/pull/3097/commits/f9f5ca4617f14c86830b4d352931f46479e267f0)

PR #3103, opened 2024-05-16
chore: Bump demosplan-ui@0.3.13

salisdemos:


spiess-demos (approved):
lgtm (code review)
if this is the ewm release, we could as well target https://github.com/demos-europe/demosplan-core/pull/3045 onto release after merging #3103.

PR #3105, opened 2024-05-16
fix(refs dplan-11516): Make cusstom Scrollbar work in FF (cherry-pick)

salisdemos:
cherry pick from https://github.com/demos-europe/demosplan-core/pull/3045


spiess-demos (approved):
lgtm! (code review only)

PR #3121, opened 2024-05-17
fix(refs dplan-1929) Add Layer Fallback

salisdemos:
### Ticket
DPLAN-1929

### How to review/test

adding this Url in the layersettings should allow to select a Layer. 
https://cardo.cottbus.de/net3/public/ogc.ashx?NodeId=1636&SERVICE=WMS&Request=GetCapabilities
(it looks like the layer is wired configured. I coundnt see anythingg in the public detail and in qGis I only saw something in some zoome levels and only in one tiny region)

Every other layer, that worked before should still work

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/map/admin/LayerSettings.vue
+++ b/client/js/components/map/admin/LayerSettings.vue
@@ -66,7 +66,6 @@
       required />
     <dp-multiselect
       @input="filterMatrixSetByLayers"
-      required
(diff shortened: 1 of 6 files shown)


salisdemos:
wrong branching

PR #3130, opened 2024-05-21
chore: define return types and type hints to be compatible with symfony 6

mussbach:
### Ticket
DPLAN-11672

symfony 6 introduces several return types and type hints. I order to upgrade we should introduce them now to reduce deprecations

### How to review/test
phpstan should not cry and cache should be clearable

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Repository/UserRepository.php
```diff
@@ -533,7 +534,7 @@ public function getFirstUserInFhhnetByLogin(string $loginSuffix): User
     /**
      * @param User $user
      */
-    public function upgradePassword(UserInterface $user, string $newHashedPassword): void
+    public function upgradePassword(UserInterface|PasswordAuthenticatedUserInterface $user, string $newHashedPassword): void
```
sierrazamudiodemos, line 537: - what about using only PasswordAuthenticatedUserInterface. UserInterface is already extending from PasswordAuthenticatedUserInterface
- Maybe remove the doc comment?

demosplan/DemosPlanCoreBundle/Security/Authentication/Provider/SecurityUserProvider.php
```diff
@@ -63,7 +64,7 @@ public function getSecurityUser(string $identifier): SecurityUser
         return new SecurityUser($this->loadUserByLogin($identifier));
     }
 
-    public function upgradePassword(UserInterface $user, string $newHashedPassword): void
+    public function upgradePassword(UserInterface|PasswordAuthenticatedUserInterface $user, string $newHashedPassword): void
```
sierrazamudiodemos, line 67: It is similar to other case:
- what about using only PasswordAuthenticatedUserInterface. UserInterface is already extending from PasswordAuthenticatedUserInterface
- Maybe remove the doc comment?

demosplan/DemosPlanCoreBundle/Utilities/Pagination/DemosPlanArrayAdapter.php
```diff
@@ -37,7 +34,7 @@ public function getNbResults()
      *
      * @return array|Traversable
      */
-    public function getSlice($offset, $length)
+    public function getSlice($offset, $length): iterable
     {
```
sierrazamudiodemos, line 38: - what about adding int type on each parameter?
- we can remove the doc comment I feel

sierrazamudiodemos (changes requested):

demosplan/DemosPlanCoreBundle/Entity/User/User.php
```diff
@@ -540,7 +540,7 @@ public function getUsername()
     /**
      * Symfony > 6 needs getUserIdentifier() for auth system.
      */
-    public function getUserIdentifier(): ?string
+    public function getUserIdentifier(): string
     {
         return $this->getLogin();
```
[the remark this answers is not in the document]
mussbach: I see your point, symfony 6 enforces this method to return a string. So I'd rather return an empty string instead of null when nothing is set  4dbb3448 wdyt?

demosplan/DemosPlanCoreBundle/Repository/UserRepository.php
mussbach: This is the `Symfony\Component\Security\Core\User\UserInterface`, not ours. The type hints are taken from the implemented method

demosplan/DemosPlanCoreBundle/Security/Authentication/Provider/SecurityUserProvider.php
mussbach: same as above as it is the same implementation

demosplan/DemosPlanCoreBundle/Utilities/Pagination/DemosPlanArrayAdapter.php
mussbach: [1ad309d](https://github.com/demos-europe/demosplan-core/pull/3130/commits/1ad309d0027bf94d4572813cf92b056934a13ce4)

sierrazamudiodemos (approved):
Thanks for explaining and addressing the concerns :)

PR #3162, opened 2024-05-27
DPLAN-11647 ducktest firefox setting

spiess-demos:
### Ticket 
https://demoseurope.youtrack.cloud/issue/DPLAN-11647/Klick-Problem-auf-der-Seite-Importierte-Stellungnahmen-uberprufen-Schulung

The reason for this bug lies in the firefox setting "privacy.resistFingerprinting" when enabled leads to the openLayers method "forEachFeatureAtPixel" to fail. This cannot be fixed on our side. See discussions at openlayers and mozilla:

- https://support.mozilla.org/de/questions/1398931
- https://github.com/openlayers/openlayers/issues/14363

The best we can do (this is implemented here) is to inform users that something will not work with the current settings.

### How to review/test
When loading the view in Firefox with "privacy.resistFingerprinting" set to true, an error notification should appear that reads:
**Um alle Funktionen dieser Seite nutzen zu können, deaktivieren Sie in Ihrem Browser die Einstellung "privacy.resistFingerprinting" unter "about:config".**

### PR Checklist

### DIFF
--- a/client/js/components/procedure/imageAnnotator/DpImageAnnotator.vue
+++ b/client/js/components/procedure/imageAnnotator/DpImageAnnotator.vue
@@ -788,6 +788,50 @@ export default {
 
   mounted () {
     this.getInitialData()
+
+    /*
+     * Display a warning for firefox if "privacy.resistFingerprinting" is enabled,
+     * because openLayers' getFeaturesAtPixel() will not behave correctly then.
+     */
+    useResistFingerprintingDuckTest((isEnabled) => {
+      if (isEnabled) {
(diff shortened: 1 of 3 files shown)


salisdemos (changes requested):
Thanks for finding that nasty quirk. Anyhow can we put that method to somewhere more generic? I can assume, we may want to check that in other ol contexts too.

spiess-demos:
> Thanks for finding that nasty quirk. Anyhow can we put that method to somewhere more generic? I can assume, we may want to check that in other ol contexts too.

can we deliver this (= move to demosplan-ui) as an improvement in another PR?

salisdemos (approved):

PR #3165, opened 2024-05-27
fix(refs DPLAN-11646): Expose Permission to control setting the map_max_extent

salisdemos:
### Ticket
DPLAN-11646

The usage of the permission was before only in the twig context and is moved to vue now. while moving, it has been forgotten to expose the permission.

### How to review/test

in a project where setting the max-extent should be possible, goto the mapsettings and search for the possibility to set the max extent

### PR Checklist


gruenbergerdemos (approved):
lgtm

PR #3176, opened 2024-05-28
fix(refs DPLAN-11673): Expose feature_xplan_defaultlayers Permission

salisdemos:
### Ticket
DPLAN-11673

This was forgotten when moving the code from twig to vue


sakutademos (approved):
lgtm, thx

PR #3180, opened 2024-05-28
fix(refs DPLAN-11614): use XMLHttpRequest to getCapabilities

spiess-demos:
### Ticket
[DPLAN-11614](https://demoseurope.youtrack.cloud/issue/DPLAN-11614/ROBOB-Prod-und-Stage-Sobald-ein-WMTS-Layer-eingebunden-ist-werden-alle-Layer-nicht-mehr-dargestellt)

As we do not want to refactor Map.vue to use async all over the place, this is the simplest way of getting synchronous WMTS requests to work.

Prior art:
- https://github.com/demos-europe/demosplan-core/pull/1876
- https://github.com/demos-europe/demosplan-core/pull/1866

### How to review/test
WMTS Maps should load and not break the view.
ected to this and explain the connection.

### PR Checklist

### DIFF
--- a/client/js/components/map/publicdetail/Map.vue
+++ b/client/js/components/map/publicdetail/Map.vue
@@ -1521,20 +1521,25 @@ export default {
       const layerArray = Array.isArray(layer.attributes.layers) ? layer.attributes.layers : layer.attributes.layers.split(',')
       const url = this.addGetCapabilityParamToUrl(layer.attributes.url)
 
-      return $.ajax({
-        dataType: 'xml',
-        url: url || '',
-        async: false,
-        success: response => {
-          const result = this.parser.read(response)
-          const options = optionsFromCapabilities(result, {
-            layer: layerArray[0] || '',
-            matrixSet: layer.attributes.tileMatrixSet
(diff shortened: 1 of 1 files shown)


client/js/components/map/publicdetail/Map.vue
```diff
@@ -1521,20 +1521,25 @@ export default {
...
-          const result = this.parser.read(response)
-          const options = optionsFromCapabilities(result, {
-            layer: layerArray[0] || '',
-            matrixSet: layer.attributes.tileMatrixSet
-          })
+      let xml
 
-          return new WMTS({ ...options, layers: layerArray })
-        }
+      const xhr = new XMLHttpRequest()
```
salisdemos, line 1526: is there a reason, why You don't use `externalApi` from dpApi? (basically fetch)

salisdemos (changes requested):

salisdemos (approved):
as discussed here https://github.com/demos-europe/demosplan-core/pull/1876#discussion_r1299729587 this solution is legit

PR #3188, opened 2024-05-29
f-K3-70 in process

bakridemos:


bakridemos:
this will be closed, it's redone in https://github.com/demos-europe/demosplan-core/pull/3195

PR #3189, opened 2024-05-29
chore (refs DPLAN-11606): save segments even if multiple tags with the same name exists

mussbach:
### Ticket
DPLAN-11606

When a procedure has tags with the same name in multiple tagTopics creation of the segments failed. This PR tries to fetch the tag by a given id, if not possible the current behavior is kept with the difference that the segments are created and a message shown to the user

### How to review/test
As described above 

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm, thx

PR #3197, opened 2024-05-30
fix(refs DPLAN-11685): fix css in assessmentTable

spiess-demos:
### Ticket
[DPLAN-11685](https://demoseurope.youtrack.cloud/issue/DPLAN-11685/ROBOB-Anzeige-der-Datensatze-funktioniert-nicht)

`flow-root` does not play nice with `v-show` as it is set to `!important`.

### How to review/test
In the assessment table, toggling between statement and fragments should result in the statements (or fragments) being exclusively shown in the content area of each card. The "derzeit sind keine Datensätze an der Stellungnahme vorhanden" sentence should only show in the fragment view of the card, and only if there are really no fragments.

### PR Checklist

### DIFF
--- a/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
+++ b/client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
@@ -302,8 +302,8 @@
 
         <!--  statement tab  -->
         <div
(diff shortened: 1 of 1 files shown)


hwiem (approved):
Couldn't test it, but code lgtm

PR #3201, opened 2024-05-31
f-DPLAN-11656 include customer in query result 

bakridemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-11656/Verfahren-werden-nicht-mehr-sichtbar-nach-Phasenumstellung

Description: 
while phases change procedure will be fetched and updated. Customer is missing when a procedure is fetched which caused that customer will be always null after updating

see  'generateObjectValues' in 'ProcedureRepository' to have better Idea how the customer will be handled.


mussbach (approved):
lgtm, thx

PR #3203, opened 2024-05-31
chore: Run yarn upgrade jest

salisdemos:
Looks like `yarn upgrade` brings something, not every machine likes. `yarn upgrade jest` should be leaner but fix the problem with `string-width` while building the FE-assets


graupnerdemos (approved):
I've seen the magic happen. Thx.

PR #2413, opened 2023-12-01
fix(refs: T35449): fix logo saving

muellerdemos:
https://yaits.demos-deutschland.de/T35449

This PR fixes the customer support section to saves and delete logos. To achieve this the logic had to be adjusted at several spots. During the work I have refactored some sections to improve readability and reduce cognitive load for the developer.
Also removed some redundant props and data props

### How to review/test

- checkout
- login as demos support
- navigate to plattform tools
- upload a logo
- make sure the UI behaves as expected

### PR Checklist

Delete the checkbox if it doesn't apply/isn't necessary.
- [x] Link all relevant tickets
- [x] Move the tickets on the board accordingly

PR #3207, opened 2024-05-31
feat: further improve branding logo upload

spiess-demos:
Ticket: [DPLAN-11707](https://demoseurope.youtrack.cloud/issue/DPLAN-11707/Prod-DiPlanBeteiligung-BB-Konfiguration-nicht-moglich)

As a followup of https://github.com/demos-europe/demosplan-core/pull/2413 this PR improves UX in the branding logo upload. 

- use a button instead of a checkbox to delete the current logo. for this to work, a distinct method is introduced. To prevent unwanted deletion, a confirm dialog is implemented. the "isLogoDeletable" variable could be removed, then.
- the newly uploaded image is not set as branding.logoHash on setFile (because at this time, it is not yet saved) but only on after saveBranding. at this point, the method clearFilesList from DpUploadFiles is also called to remove the previously uploaded file.

### How to review/test
- already uploaded logo should be deletable via button, deletion has to be confirmed
- when selecting a new file to upload and saving it, the files list should be emptied. the newly uploaded image should only appear in the "Aktuelles Logo" spot after having saved, not after the upload itself has finished

### PR Checklist

- [X] Data-Cy attributes added/updated ([conventions](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/twig_html.html#guideline-for-naming-cypress-hooks))


muellerdemos (approved):
I like to have a button now instead of a checkbox which I also find more fluid in the workflow. 

I have no concerns about this PR but after having a look in the interface I kind of think the spacing in this area is a little awkward (But I also have no better idea for this and just wanted to mention this. I dont think this should be tackled in this PR or in this Sprint)

grafik

PR #3208, opened 2024-05-31
chore (refs DPLAN-609): file uploads itself should not be checked for procedure permissions

mussbach:
### Ticket
DPLAN-609

File uploads failed because permissions are checked more strictly when `X-Demosplan-Procedure-Id` header is set. File uploads should not check procedure permissions even if the file is uploaded within a procedure, as permissions are checked during access.

I tried really hard to write a test for this but was not able to mock the `$controllerEvent->getController()` to return an array.

### How to review/test
upload a file in a procedure.

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


spiess-demos (approved):
lgtm (code review).

PR #3211, opened 2024-06-03
bug(refs DPLAN-1909): Add whitespace normal for multiselect

gruenbergerdemos:
### Ticket
DPLAN-1909

We need whitespace normal for the multiselect, because otherwise the tags overflow the input.

### How to review/test
Login as support -> Nutzende -> tags for role should stay inside the input and not overflow

### DIFF
--- a/client/js/components/user/DpUserList/DpUserFormFields.vue
+++ b/client/js/components/user/DpUserList/DpUserFormFields.vue
@@ -107,7 +107,7 @@
       <dp-multiselect
         :id="userId + ':userRoles'"
         ref="rolesDropdown"
-        class="u-mb-0_5"
+        class="u-mb-0_5 whitespace-normal"
         :custom-label="option =>`${ roles[option.id].attributes.name }`"
         data-cy="role"
         label="name"


client/js/components/user/DpUserList/DpUserFormFields.vue
```diff
@@ -107,7 +107,7 @@
       <dp-multiselect
         :id="userId + ':userRoles'"
         ref="rolesDropdown"
-        class="u-mb-0_5"
+        class="u-mb-0_5 whitespace-normal"
```
salisdemos, line 110: shouldn't we fix that within demosplan-ui?

salisdemos (changes requested):

client/js/components/user/DpUserList/DpUserFormFields.vue
gruenbergerdemos: `whitespace: normal` is the default. We only need it in this component, because the outer div has `class="whitespace-nowrap"`.

PR #3214, opened 2024-06-03
chore: remove symlink to local addon during uninstall

mussbach:
When uninstalling a local addon during development the symlink was not correctly deleted from the cache directory

### How to review/test
you may install an addon via github/zip and one via addonDev local directory and uninstall them. Nothing should be left in the addons/cache folder

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


martendemos (approved):
tested manually

PR #3216, opened 2024-06-03
F dplan 11470 create command to add permissions

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-11470/3-0-BB-Als-Fachplanung-Administration-mochte-ich-keine-neuen-Verfahren-mehr-anlegen-konnen-um-falschliches-Anlegen-in-BLP-zu

Description:
1. We want a command that given customer id, role id, and permission, it enables this permission for all the orgas under the given customer and role.
2. The command has customerID roleID and permissionName parameters
3. The command also offers a dry-run, which does not execute the update, it only prints the impacted orgas
4. If customer or role or permission are not found, it throws an exception

### How to review/test
1. Typebin/blp dplan:permission:enable:customer-orga-role YOUR_CUSTOMER_ID YOUR_ROLE_ID CAN_CREATE_PROCEDURES --dry-run
2. Adjust param to insert ids based on your db
3. then check if the printed orgas are also inserted in the access_control DB

- [x ] Tests updated/created


graupnerdemos:
@sierrazamudiodemos to simplify the review, can you please update this against main?

sierrazamudiodemos:
> @sierrazamudiodemos to simplify the review, can you please update this against main?

@graupnerdemos I can change to another branch f_DPLAN-11483_additional_permissions_paid_orga so that you will only see the changes exactly.

sierrazamudiodemos:
@graupnerdemos I changed the base, you can review now

martendemos:
When I try to execute the command including the --dry-run option I got
  The "--dry-run" option does not exist.

sierrazamudiodemos:
> When I try to execute the command including the --dry-run option I got The "--dry-run" option does not exist.

Did you try `bin/blp dplan:permission:disable:customer-orga-role --dry-run`

martendemos:
> We want a command that given customer id, orga id, and role id, and permission, it enables this permission for all the orgas under the given customer and role, except for the given orga id
But what happens if I want to disable a second orga? Would the first dispabled orga enabled again?

sierrazamudiodemos:
> > > We want a command that given customer id, orga id, and role id, and permission, it enables this permission for all the orgas under the given customer and role, except for the given orga id
> > 
> > 
> > But what happens if I want to disable a second orga? Would the first dispabled orga enabled again?
> > @martendemos here are my thoughts:
> 
>     1. the command can accept an array of orga ids
> 
>     2. or the command runs, and then the user/dev has to go to the UI and disable the checkbox
>        I guess the 1 one might make sense to keep it flexible, what do you think?

@martendemos done: 2dc0a38607e4af77a2a408b59fce7390b6dad7ed

martendemos:
>the command can accept an array of orga ids
or the command runs, and then the user/dev has to go to the UI and disable the checkbox
I guess the 1 one might make sense to keep it flexible, what do you think?

My first thought is, that only the specific entries will be removed from the access_control_permission. This would means, that the given information (of the user)
will be checked for existing in the table and if found; removed.

martendemos:
I got an exception, in the case I try to execute the same command twice.
The reason seems to be, that the entries already existing.
```
62 Duplicate entry 'c7876461-4f1a-11ea-92dd-782bcb0d78b1-c9dc879b-9876-3568-bc67-...' for key 'unique_orga_customer_role_permission'              
```

martendemos (changes requested):
Found some issues. (see comments above)
Additionally: If I try to disable a specific role, orga, customer combination, the user I test with, still has access to the area and can create a new procedure.
So should this be denied with this PR or is something missing?

sierrazamudiodemos:
@martendemos I adjusted the command based on the criteria of the US, i updated the description as well

martendemos:
> @martendemos I adjusted the command based on the criteria of the US, i updated the description as well

This command only add permissions (to the DB-table and therefore) can not reduce permissions of a specific role.
As far as I understand, means this, that the related US is not fulfilled with this PR alone.

I recommend to rename the title (to something like "command to add permission) of this PR and create an additional one which
allows to remove permissions from the table and include the remove of the related permission from the permission.php from blp(!). (Remove the static way to enable the related permission, to make the table the source of truth)

martendemos:
> Type bin/blp dplan:permission:enable:customer-orga-role --customerId YOUR_CUSTOMER_ID --roleId YOUR_ROLE_ID --permission CAN_CREATE_PROCEDURES --dry-run

Does not work for me. Instead I'm using
`bin/blp dplan:permission:enable:customer-orga-role YOUR_CUSTOMER_ID YOUR_ROLE_ID CAN_CREATE_PROCEDURES`
, which seems to be working.

martendemos (dismissed):

PR #3220, opened 2024-06-04
bug(refs DPLAN-2168): Use size method to see if element exists

gruenbergerdemos:
### Ticket
DPLAN-2168

You can check if an element exists by using the size() method from D3.js. It returns the number of elements in the current selection.


elmasdemos (approved):
lgtm, code review

PR #3221, opened 2024-06-04
fix(refs DPLAN-11741): avoid loading/fetching addons multiple times

salisdemos:
### Ticket
DPLAN-11741

Addons have been loaded every time the component got mounted


graupnerdemos (approved):
Glad we found this one. Save requests! Save the planet!

PR #3233, opened 2024-06-07
chore: Bump node 18.18.0

salisdemos:
To Bump Node, we have to bump a bunch of dependencies too. So this PR has to be merged at the same time a new container gets rolled out with a Node-Version "18.18.0" 

The changes can be tested, by Bumping node in Your local container. (login as root (`dp shell --root`) -> `npm install -g n` --> `n 18.18.0` --> `exit` --> `dp shell` --> `yarn; yarn dev:blp` (or any other project) )
You may run into the well knows "string-with"-Problem. That can be fixed by running `yarn upgrade jest`

OR:
You set Your container-Version to 5.2 . That comes with node 18.20.0 (which also should work)


package.json
```diff
@@ -185,6 +185,6 @@
     "ff esr"
   ],
   "resolutions": {
-    "@mapbox/mapbox-gl-style-spec": "13.14.0"
+    "@mapbox/mapbox-gl-style-spec": "13.23.1"
```
spiess-demos, line 188: Do we still have to keep this around?

config/packages/nelmio_security.yaml
```diff
@@ -31,7 +31,7 @@ nelmio_security:
             base-uri: '%demosplan_csp_base_uri%'
             form-action: '%demosplan_csp_form_action%'
             worker-src:
-                - 'none'
+                - 'blob:'
```
spiess-demos, line 34: This change seems to be unrelated to the current PR.

salisdemos: actually no. Bumping the masterPortalApi made this change necessary. So its the last snippet of the change-reaction while "you not simply bump node"

package.json
salisdemos: what do You mean? If we can drop the whole resolution? Maybe. won't test that within that thing here. was ugly enough

spiess-demos: yes i meant to drop the resolution but i agree that's not in scope here.

spiess-demos (approved):
lgtm (code review only)

hwiem (approved):
Lgtm!

salisdemos:
Sorry. I had to merge main into the branch again (and recreate the yarn.lock-File), which dismissed the Reviews.

hwiem (approved):

PR #3234, opened 2024-06-07
chore: Bump node 20.12.2

salisdemos:
We may want to Bump to 18.18.0 first and then try this one. 
(https://github.com/demos-europe/demosplan-core/pull/3233)


package.json
```diff
@@ -51,14 +50,15 @@
     "@demos-europe/demosplan-ui": "^0.3.17",
     "@demos-europe/dp-consent": "^1.1.2",
     "@efrane/vuex-json-api": "0.0.38",
-    "@masterportal/masterportalapi": "2.22.0",
+    "@masterportal/masterportalapi": "2.35.0",
```
spiess-demos, line 53: Do we need to adapt further things when bumping this?

package.json
```diff
@@ -186,5 +186,6 @@
   ],
   "resolutions": {
     "@mapbox/mapbox-gl-style-spec": "13.23.1"
-  }
+  },
+  "packageManager": "yarn@4.2.2"
```
spiess-demos, line 190: the PR title does not reflect this change.

package.json
salisdemos: next to the blob-thing from the other PR, nothing I could find

package.json
salisdemos: true. I wan't to move that change out anyway and do that at very first step.

salisdemos, line 190:
```suggestion
  }
```

package.json
```diff
@@ -51,14 +50,15 @@
...
-    "@masterportal/masterportalapi": "2.22.0",
+    "@masterportal/masterportalapi": "2.35.0",
     "@sentry/browser": "^8.2.1",
     "@sentry/integrations": "^7.114.0",
     "@sentry/tracing": "^7.114.0",
     "@uppy/core": "^3.8.0",
     "@uppy/drag-drop": "^3.0.3",
     "@uppy/progress-bar": "^3.0.4",
     "a11y-datepicker": " ^0.9.0",
+    "cesium": "^1.118.2",
```
salisdemos, line 61: this is necessary as peerdependency for olcs wich is used in masterportalApi.

hwiem (dismissed):
lgtm

salisdemos:
This has to be done again. Some dependencies raised in the meantime, so we have to doublecheck and adjust all the changes anyway

PR #3245, opened 2024-06-11
bug(refs DPLAN-2759): Add h1 heading for public index

gruenbergerdemos:
### Ticket
DPLAN-2759

For accessibility reasons there always should be a h1.

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/public_index.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/public_index.html.twig
@@ -4,7 +4,7 @@
 
 {% block component_part %}
     {% apply spaceless -%}
-        <h1 class="hide-visually">{{ "public.index.heading"|trans({ projectName: projectName } }}</h1>
+        <h1 class="sr-only">{{ "public.index.heading"|trans({ projectName: projectName }) }}</h1>
 
         {# onboarding / welcome panel #}
         {% block panel %}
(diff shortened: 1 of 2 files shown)


translations/messages+intl-icu.de.yml
```diff
@@ -2764,6 +2764,7 @@ public.index.filter.gfk.location.results: |
         =0 {Es wurden zu der Kommune folgende Verfahren gefunden:}
         =1 {Es wurden zu der Kommune {locationName} folgende Verfahren gefunden:}
     }
+public.index.heading: "DiPlanBeteiligung: Plattform zur öffentlichen Mitwirkung bei räumlichen Planungen und Genehmigungsverfahren"
```
spiess-demos, line 2767: As this is project specific, it would be better placed as an override. Or better, you could pass the project name as variable:

```suggestion
public.index.heading: "{projectName}: Plattform zur öffentlichen Mitwirkung bei räumlichen Planungen und Genehmigungsverfahren"
```

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/public_index.html.twig
```diff
@@ -4,6 +4,7 @@
 
 {% block component_part %}
     {% apply spaceless -%}
+        <h1 class="hide-visually">{{ "public.index.heading"|trans }}</h1>
```
spiess-demos, line 7: While this is our "hand-rolled" class for visually hiding stuff, the [Tailwind class](https://tailwindcss.com/docs/screen-readers#screen-reader-only-elements) (which can already be used) is `sr-only`. 

However as we're using `hide-visually` all over the place, we could as well refacor it all at once.

Regarding the suggestion below, the `trans` call would change to:

```suggestion
        <h1 class="hide-visually">{{ "public.index.heading"|trans({ projectName: projectName } }}</h1>
```

spiess-demos (changes requested):

spiess-demos (approved):
lgtm!

PR #3246, opened 2024-06-11
feat (DPLAN-11785) Create NewOrgaCreated event and add listener to ca…

sierrazamudiodemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-11785/blp-additional-roles-should-not-have-hard-coded-permission

Description:

1. The permissions `feature_admin_new_procedure` should be dynamically set for every role only in the blp project.
2. Some roles have that permission hard-coded in the Permissions.php in core
3. The permission has been removed from core and migrated to each Permissions.php project (except BLP)
4. Below the roles that had this permission hard-coded:
5.  `HEARING_AUTHORITY_ADMIN` and `PLANNING_AGENCY_ADMIN` had the permission in Permissions.php in core
6. `CUSTOMER_MASTER_USER` had the permission in Permissions.php in blp
7. In order to make this permission dynamic, these 3 roles get the permission whenever a new orga is created. In order to do  that, there is an event `NewOrgaCreatedEvent` that is triggered everytime that a new orga is created.
8. The event has 2 parameters: orga and the flag `canCreateProcedures`
9. The flag can be set as true, false or null.
10. There is a listener in blp folder` OrganisationCreateSubscriber` that listens the event and decides based on the flag if the permission should be added for the `HEARING_AUTHORITY_ADMIN` and `PLANNING_AGENCY_ADMIN` and `CUSTOMER_MASTER_USER` roles.
11. Similar approach is happening when the `OrgaAdminEditedEvent` is triggered. There is a listener in blp folder` OrganisationCreateSubscriber` that listens the event and decides based on the flag if the permission should be added for the `HEARING_AUTHORITY_ADMIN` and `PLANNING_AGENCY_ADMIN` and `CUSTOMER_MASTER_USER`

### How to review/test
Case 1:
- using Support role Create an orga in BLP and checkbox the flag: Darf Verfahren anlegen
- then you can see in the table access_controll that there should be 3 entry permissions (it depends if the 3 roles are created in the DB)

Case 2:
- using Support role update an orga in BLP (where you can login later) and checkbox the flag: Darf Verfahren anlegen
- then you can see in the table access_controll that there should be 3 entry permissions (it depends if the 3 roles are created in the DB)
- Then login to that orga, you should be able to create Verfahren
- Then uncheck the box.
- Then login to that orga, you should not be able to create Verfahren

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-core/pull/3216

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
@@ -48,7 +48,6 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 use UnexpectedValueException;
-use Webmozart\Assert\Assert;
 
 class DemosPlanOrganisationAPIController extends APIController
 {
@@ -329,7 +328,6 @@ class DemosPlanOrganisationAPIController extends APIController
             $canCreateProcedures = null;
             if ($permissions->hasPermission('feature_manage_procedure_creation_permission')
                 && array_key_exists('canCreateProcedures', $orgaDataArray)) {
-
                 $role = $roleHandler->getRoleByCode(RoleInterface::PRIVATE_PLANNING_AGENCY);
 
                 try {
@@ -340,7 +338,7 @@ class DemosPlanOrganisationAPIController extends APIController
                 } catch (NullPointerException $e) {
                     $this->logger->warning('Role was not found in Customer. Permission is not created', [
                         'roleName'   => RoleInterface::PRIVATE_PLANNING_AGENCY,
-                        'permission' => AccessControlService::CREATE_PROCEDURES_PERMISSION
+                        'permission' => AccessControlService::CREATE_PROCEDURES_PERMISSION,
                     ]);
                 }
             }
@@ -408,7 +406,6 @@ class DemosPlanOrganisationAPIController extends APIController
             $canCreateProcedures = null;
             if ($permissions->hasPermission('feature_manage_procedure_creation_permission') && is_array($orgaDataArray['attributes'])
                 && array_key_exists('canCreateProcedures', $orgaDataArray['attributes'])) {
-
                 $role = $roleHandler->getRoleByCode(RoleInterface::PRIVATE_PLANNING_AGENCY);
 
                 try {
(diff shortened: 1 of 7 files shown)


demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
```diff
@@ -321,18 +325,26 @@ public function createOrgaAction(Request $request,
             }
 
             // Add new permission in case it is present in the request
-
+            $canCreateProcedures = null;
             if ($permissions->hasPermission('feature_manage_procedure_creation_permission')
                 && array_key_exists('canCreateProcedures', $orgaDataArray)) {
                 $roles = $roleHandler->getUserRolesByCodes([RoleInterface::PRIVATE_PLANNING_AGENCY]);
                 Assert::count($roles, 1);
```
mussbach, line 332: Has nothing really to do with this PR but as a side effect: Could you perhaps include the try/catch here as well?

mussbach (approved):
looks good, one remark which may be also done in another PR. Code review only

martendemos (commented):
What does it mean when $canCreateProcedures is null?

sierrazamudiodemos:
> What does it mean when $canCreateProcedures is null?

@martendemos when other projects do not have yet the dynamic permission set. maybe I can rename it to isPermissionset?

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
```diff
@@ -61,7 +59,11 @@ public function getPermissions(?OrgaInterface $orga, ?CustomerInterface $custome
         foreach ($roles as $roleName) {
             // Try to find an existing permission with the given parameters
 
-            $role = $this->getRoleByCode($roleName);
+            $role = $this->roleHandler->getRoleByCode($roleName);
+
+            if (null === $role) {
+                continue;
+            }
```
martendemos, line 66: minor personal note: I prefer to avoid jump actions like continue, because it increase the level of complexity (for me).
Instead you can wrap the whole logic into an if(null !== $role).
But I'm not sure if we as developers has this as best practice.

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
sierrazamudiodemos: @mussbach done 2b427400e9f616036b226cc6c46b410f54188df7

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
mussbach: I like this style more than nested ifs. For me this continue reduces cognitive complexity more deeply nested code

mussbach (approved):
without overruling @martendemos concerns, mine are done

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
sierrazamudiodemos: what about discussing it in the team? I am with Sami and I am open to discuss it as well :+1: thanks for the remarks

martendemos: Yes, we could, maybe we should discuss this. In this case I'm okay because it is a conscious decision.

PR #3249, opened 2024-06-11
refactor:  use tailwind class for hide visually

gruenbergerdemos:
Refactoring: Use tailwind class `sr-only` instead of custom class to hide elements visually.


gruenbergerdemos:
> Can we now also remove the hide-visually class in _display-visibility.scss ?
I'm not sure. I didn't look into the addons and if it is still used there and that's why I thought it would be best to leave it for now.

hwiem (approved):
Lgtm, code review only

spiess-demos (approved):
Nice refactoring, thanks!

I'd suggest to additionally change the used class in the [example in dplandocs](https://github.com/demos-europe/demosplan-documentation/blob/49db7abf4466f857151d9fe62cee51e9869613b4/design-system/guidelines/accessibility/README.md?plain=1#L220). Also, i made another [small PR within addon-maillane](https://github.com/demos-europe/demosplan-addon-maillane/pull/141).

spiess-demos:
> Can we now also remove the `hide-visually` class in `_display-visibility.scss` ?

We can remove the selector definition, however the scss mixin `%hide-visually` is still in use, it also needs to be replaced by `sr-only`.

gruenbergerdemos:
> I'd suggest to additionally change the used class in the [example in dplandocs](https://github.com/demos-europe/demosplan-documentation/blob/49db7abf4466f857151d9fe62cee51e9869613b4/design-system/guidelines/accessibility/README.md?plain=1#L220).

I already did, but I forgot to open a PR. Thanks!

PR #3250, opened 2024-06-11
bug(refs DPLAN-1879): Use the right attribute for label prop

gruenbergerdemos:
### Ticket
DPLAN-1879

Attribute was changed from hidden to hide, see https://github.com/demos-europe/demosplan-ui/pull/894


spiess-demos (approved):
lgtm!

PR #3269, opened 2024-06-13
feat (DPLAN-11470) adjust command to take orga type into account

sierrazamudiodemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-11470/3-0-BB-Als-Fachplanung-Administration-mochte-ich-keine-neuen-Verfahren-mehr-anlegen-konnen-um-falschliches-Anlegen-in-BLP-zu

Description:
- When the command is run to enable permission for all in a particular customer - role, we should check that which orga type has the orga and if it corresponds to the role given in the command.
- It is like that because we want to avoid cases like:
-- command enables permission CREATE_PROCEDURE for customer XXX and Role PLANNING_AGENCY_ADMIN, and there are orgas that have type Planingburo where actually the role PLANNING_AGENCY_ADMIN is not allowed.

### How to review/test
- Using customer support role, find and orga that only has the type PlanungBuro and is accepted in the customer
- Run command bin/blp dplan:permission:enable:customer-orga-role YOUR_CUSTOMER_ID YOUR_ROLE_ID CREATE_PROCEDURES_PERMISSION --dry-run
- You should not see the orga listed in the command
- Then change the type of the orga, you should see the orga

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
@@ -35,7 +35,6 @@ class AccessControlService extends CoreService
 {
     public const CREATE_PROCEDURES_PERMISSION = 'feature_admin_new_procedure';
 
-
     public function __construct(
         private readonly AccessControlRepository $accessControlPermissionRepository,
         private readonly RoleHandler $roleHandler,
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
```diff
@@ -209,15 +210,42 @@ public function enablePermissionCustomerOrgaRole(string $permissionToEnable, Cus
...
+            if (null !== $updatedOrga) {
+                $updatedOrgas[] = $updatedOrga;
+            }
+        }
+
+        return $updatedOrgas;
+    }
+
+    /**
+     * $orgatype = getorgatype.
```
mussbach, line 224: Wht do you mean by this dockblock? I do not understand

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
```diff
@@ -209,15 +210,42 @@ public function enablePermissionCustomerOrgaRole(string $permissionToEnable, Cus
...
+     *
+     * $orgatypeForRole = getOrgaTypebasedonRole
+     * if $orgatype === $orgatypeForRole --> add permission
+     */
+    private function addPermissionBasedOnOrgaType(string $permissionToEnable, RoleInterface $role, OrgaInterface $orgaInCustomer, CustomerInterface $customer, bool $dryRun): ?OrgaInterface
+    {
+        $orgaTypesInCustomer = $orgaInCustomer->getTypes($customer->getSubdomain(), true);
+        foreach ($orgaTypesInCustomer as $orgaTypeInCustomer) {
+            if (self::CREATE_PROCEDURES_PERMISSION === $permissionToEnable
```
mussbach, line 233: This is ok for now to get this done,  but it does not scale very well. We need to find an approach that this check could be done by configuration

mussbach (changes requested):
one remark regarding the dock block. When I run the command without dry-run nothing is written to the database

sierrazamudiodemos:
> one remark regarding the dock block. When I run the command without dry-run nothing is written to the database
@mussbach 
are you checking the access_control table? I was checking the old table access_control_permission the other day and that is why I also got confused

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
sierrazamudiodemos: @mussbach I am with you. What about disccusing in tehnix

mussbach:
Yes I was looking at the `access_control_permission`. `access_control` is filled with data. Is it old? Could it be deleted then?

demosplan/DemosPlanCoreBundle/Logic/Permission/AccessControlService.php
sierrazamudiodemos: @mussbach done ce8d4b56ccb5862b11ef0ea3875624a2c7b4d3bc

mussbach (approved):

PR #3272, opened 2024-06-14
Include Tailwind Preflight as separate file

spiess-demos:
The (slightly opinionated) tailwind browser reset, "preflight", sets some defaults that we want to
override with the project css. However, as we also want to include the tailwind utility classes last
(to get rid of making them !important), the preflight styles need to be separately included, not within the same file as the utility classes.

To illustrate the change, here is a before/after comparison of the order and contents of Styles included in demosplan:

**Before:**

- **demosplan-ui.css**
  - Tailwindcss Preflight (with opinionated defaults)
  - Tailwindcss Utilities (with !important, to win over styles included later in the document while having the same specificity)
- **style.css (or, style-public.css)**
  - Base styles (overriding the opinionated defaults of Tailwind, as configured with Scss variables)
  - All the other exciting stuff (see core_style.scss)

**After:**

- **preflight.css**
  - Tailwindcss Preflight (with opinionated defaults)
- **style.css (or, style-public.css)**
  - Base styles (overriding the opinionated defaults of Tailwind, as configured with Scss variables)
  - All the other exciting stuff (see core_style.scss)
- **demosplan-ui.css**
  - Tailwindcss Utilities (no more !important needed, because now included later in the document)

Other things changed when "driving by":
- PHPStorm seems to know @tailwind in current versions
- html.fixtinyfixed seems to no longer be used

### How to review/test
After running `yarn dev:<project>` locally,  things should look as before.

### DIFF
--- a/client/css/index.css
+++ b/client/css/index.css
@@ -1,7 +1,5 @@
-/**
- * @noinspection
- * PHPStorm complains about unset css vars in rendered css. This is irritating and can be suppressed.
- */
+/* stylelint-disable */
+@config '../../tailwind.config.js';
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
--- a/client/css/preflight.css
+++ b/client/css/preflight.css
@@ -1,0 +1,3 @@
+/* stylelint-disable */
+@config '../../tailwind.preflight.config.js';
+@tailwind base;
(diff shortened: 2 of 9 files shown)


elmasdemos (approved):
lgtm, code review

ahmad-demos (approved):
lgtm, codeReview!

PR #3273, opened 2024-06-14
fix(refs dplan-11814): Adjust pagerfanta-templates after upgrade

salisdemos:
### Ticket
DPLAN-11814

Check if the Public Detail -> Documents list works again and the pager looks like expected.


mussbach (approved):
lgtm, code review and tested in interface. Thanks for fixing this upgradebug

PR #3283, opened 2024-06-17
fix: optimize checkbox positioning + wording

spiess-demos:
- when placing the checkbox below the "Neue Berechtigung..." button, it is less confusing where it belongs
- explicitly say which user groups may create procedures
- template is not needed here, v-if can directly be placed on the element

### How to review/test
The checkbox/label should be placed below the button to add groups to an orga.
The label should show the updated content.
image

### DIFF
--- a/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
+++ b/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
@@ -300,7 +300,7 @@
 
         <dp-checkbox
           v-if="hasPermission('feature_manage_procedure_creation_permission')"
-          id="procedureCreatePermission"
+          :id="`${organisation.id}:procedureCreatePermission`"
           class="mt-2"
           data-cy="orgaFormField:procedureCreatePermission"
           :label="{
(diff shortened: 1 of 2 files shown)


salisdemos (approved):

salisdemos:
Thanks a lot

client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
```diff
@@ -309,6 +297,18 @@
             </div>
           </div>
         </template>
+
+        <dp-checkbox
+          v-if="hasPermission('feature_manage_procedure_creation_permission')"
+          id="procedureCreatePermission"
```
salisdemos, line 303: this needs an index or something, that it is unique in a list

salisdemos (changes requested):

client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
spiess-demos: indeed.. a5ab31e3e9d5916b94ec3c5c4a73b63e43f1962b

salisdemos (approved):

PR #3217, opened 2024-06-04
refacor: Create api2.0 Routes Dynamically

salisdemos:
With That, we don't have to register the routes manually in the VuexApiRoutes.js. Now they get build during the store initialization depending on the entrypoint

PR #3285, opened 2024-06-17
refacor: Create api2.0 Routes Dynamically

salisdemos:
This is a cherry-pick from https://github.com/demos-europe/demosplan-core/pull/3217 to allow targeting main straight away.

### DIFF
--- a/client/js/store/core/VuexApiRoutes.js
+++ b/client/js/store/core/VuexApiRoutes.js
@@ -10,7 +10,7 @@
 /*
  * Hardcoded API 1.0 Routes
  */
-const api1Routes = [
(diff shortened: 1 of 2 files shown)


spiess-demos (approved):
lgtm (code review)

graupnerdemos (approved):
Yes please!

hwiem (approved):
On the one hand, it's nice. :sparkles: On the other, I'm worried this will make it even more difficult to understand what's going on or to find where a certain route is used. (If I remember correctly, we wanted to make it easier to find where routes are used?) But it's maybe not a problem to solve in this PR.

PR #3287, opened 2024-06-17
fix(refs dplan 11843): Pass Procedure ID properly to documents editor

salisdemos:
### Ticket
DPLAN-11843

The ProcedureId is static in this place and has to be used via twig.


mussbach (approved):
lgtm, code review and tested, thx

PR #3291, opened 2024-06-18
chore (refs DPLAN-2758): Add ariaLabelledBy for input.

ahmad-demos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-2758

- Add ariaLabelledBy for input.
- For good reading instead of PLZ.

### DIFF
--- a/client/js/components/statement/publicStatementModal/formGroups/FormGroupPostalAndCity.vue
+++ b/client/js/components/statement/publicStatementModal/formGroups/FormGroupPostalAndCity.vue
@@ -12,6 +12,7 @@
     <div :class="prefixClass('layout')">
       <dp-input
         id="r_postalCode"
+        aria-labelledby="postalcode"
         autocomplete="postal-code"
         data-cy="postalCode"
         :class="prefixClass('layout__item')"


client/js/components/statement/publicStatementModal/formGroups/FormGroupPostalAndCity.vue
```diff
@@ -12,6 +12,7 @@
     <div :class="prefixClass('layout')">
       <dp-input
         id="r_postalCode"
+        aria-labelledby="postalcode"
```
salisdemos, line 15: thats not how it works. see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

salisdemos (changes requested):

client/js/components/statement/publicStatementModal/formGroups/FormGroupPostalAndCity.vue
ahmad-demos: then in this place we should use `aria-label` for screen renders and add a new prop in `DpInput`, right?
As I understood from the ticket that we add aria-label for all input fields, but I find all input fields have a label and are visible, but problem here only with PLZ field!

salisdemos: I think it should resolved in demosplan-ui with a general approach

salisdemos (changes requested):

mussbach:
@salisdemos could this PR be closed when it should be done generally?

salisdemos:
wrong approach

PR #3293, opened 2024-06-18
chore (refs DPLAN-11855): fix missing image size during import

mussbach:
### Ticket
DPLAN-11855

DocxImporter may not be able to calculate the correct image size. In this case the importer will set the width and height to 0. This method will fix the image size by reading the image file and setting the correct width and height.

### How to review/test
Import docx document. Any picture should be shown in the interface

### PR Checklist

- [x] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm and tested too

PR #3294, opened 2024-06-19
feat(refs DPLAN-4020): Add Overview Page to Mapsettings

salisdemos:
### Ticket
DPLAN-4020

The menu on the right hand side has to point to an overview page instead of the map settings itself, because we want to extend that section to allow creating and editing layer-settings.

### How to review/test

Clicking on "Karteneinstellungen" should lead to a minimap with some links

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
backend looks good.

PR #3295, opened 2024-06-19
feat(refs DPLAN-4020): Use Api 2.0 instead of 1.0

salisdemos:
### Ticket
DPLAN-4020

Use API 2.0 instead of Api 1.0 to fetch/put GisLayers and GisLayerCategories

This is a Part of https://github.com/demos-europe/demosplan-core/pull/3312


MoritzMandler (approved):
bachend looks good

client/js/store/map/Layers.js
```diff
@@ -253,9 +272,11 @@ const LayersStore = {
     },
 
     save ({ state, commit, dispatch }) {
+      console.log('save', state.apiData.data)
```
sakutademos, line 275:
```suggestion
```

client/js/store/map/Layers.js
```diff
@@ -200,15 +200,34 @@ const LayersStore = {
...
-      return dpApi({
-        method: 'GET',
-        url: Routing.generate('dplan_api_procedure_layer_list',
-          {
-            procedureId: procedureId,
-            include: ['categories', 'gisLayers'].join()
-          }
-        )
-      })
+      return dpApi.get(Routing.generate('api_resource_list', {
```
sakutademos, line 203: this call returns a 400. I don't know if it is intended to be like this here and if it is fixed in other PRs. I am a bit lost... The request looks good to me

sakutademos (approved):

PR #3296, opened 2024-06-19
feat(refs DPLAN-11888): Pass Number instead of null

salisdemos:
DPLAN-11888

- Add fallback to pager. null is not allowed anymore.
- Add missing Api-1.0 Routes (Just to be sure)


hwiem (commented):
Should the api route changes be in this PR?

salisdemos:
> Should the api route changes be in this PR?

got me. I was lazy. I thought its small enough to fit in the same PR

hwiem:
> > Should the api route changes be in this PR?
> 
> got me. I was lazy. I thought its small enough to fit in the same PR

I don't even mind that :sweat_smile: but I think it should be included in the PR description (in the future)

hwiem (approved):
lgtm, code review only

salisdemos:
> > > Should the api route changes be in this PR?
> > 
> > 
> > got me. I was lazy. I thought its small enough to fit in the same PR
> 
> I don't even mind that 😅 but I think it should be included in the PR description (in the future)

Actually it is in the description, but not in the title

hwiem:
> > > > Should the api route changes be in this PR?
> > > 
> > > 
> > > got me. I was lazy. I thought its small enough to fit in the same PR
> > 
> > 
> > I don't even mind that 😅 but I think it should be included in the PR description (in the future)
> 
> Actually it is in the description, but not in the title

:D in that case, note to self: read the description

PR #3297, opened 2024-06-19
chore: Bump Yarn@4.2.2

salisdemos:
with the new container 1.8.45 we have yarn@4.2.2 to use that properly, we have to declaire it in the package.json


hwiem (approved):
lgtm

PR #3299, opened 2024-06-19
fix: Remove Preinstall hook in Package json

salisdemos:
with yarn 4 this doesn't work anymore :-/
We have to use a different solution


spiess-demos (approved):
lgtm!

ahmad-demos (approved):

PR #3301, opened 2024-06-19
fix: Remove adding none-existing folder in updateCommand

salisdemos:
This looks like some leftovers from decades ago


mussbach (approved):
when they are not needed any more the commands can well be deleted, thx

salisdemos:
> when they are not needed any more the commands can well be deleted, thx

Usage Error: The file:client/ui string didn't match the required format (package-name@range). Did you perhaps forget to explicitly reference the package name?

$ yarn add [--json] [-F,--fixed] [-E,--exact] [-T,--tilde] [-C,--caret] [-D,--dev] [-P,--peer] [-O,--optional] [--prefer-dev] [-i,--interactive] [--cached] [--mode #0] ...
Usage Error: The file:client/js-utils string didn't match the required format (package-name@range). Did you perhaps forget to explicitly reference the package name?

--> Folders don't exist anymore

PR #3310, opened 2024-06-20
allow multiple ids in permission:enable command

mussbach:
### Ticket
DPLAN-11556

To avoid calling the permission:enable command 34 times a possibility is added to provide an array of customers and roles. Along the way a bug was fixed that citizen orga should not be provided with any permissions. Moreover code was deduplicated

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


MoritzMandler (approved):
lgtm - tested

PR #3313, opened 2024-06-21
fix: Adjust Type Casing for APi Types

salisdemos:
we have to carefully check if the Types for API 1.0 Routes have to be Pascal or camel. So do, some don't.
And with that and since to type doesn't get transformed anymore, we had to adjust the casing for `report` in the backend. Along there where couple of occurrences, where changing to Pascal simply got overseen.

### Ticket
DPLAN-11889
-- DPLAN-11828
-- DPLAN-11886
-- DPLAN-11896
-- DPLAN-11897
-- DPLAN-11900
-- DPLAN-11902


sakutademos (approved):
lgtm!

PR #3315, opened 2024-06-21
Feat(DPLAN-32689): create modal window for exporting files

sakutademos:
### Ticket
[DPLAN-3268](https://demoseurope.youtrack.cloud/issue/DPLAN-3268/3-3-Als-Fachplanung-mochte-ich-die-Spaltennamen-in-den-Synopsen-anpassen-konnen-um-sie-nach-meinen-Anforderungen)

**Description:** This PR introduces a statement export modal that allows renaming of DOCX columns during export.

- docx headers are sent via query parameters (col1, col2, and col3)
- the option to change titles is available only for docx files (zip with docx files)
- the modal is accessible on both the statements list and the STN detail page

zugehörige UX Discovery: [DPLAN-11773](https://demoseurope.youtrack.cloud/issue/DPLAN-11773/UX-Discovery-Modal-fur-konfigurierbare-Synopsen)

### How to review/test

- [ ] Tests updated/created
- [ ] Update documentation


client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,241 @@
...
+
+    getDefaultPlaceholderByKey (key) {
+      const defaultPlaceholders = {
+        left: Translator.trans('segments.export.segment.id'),
+        middle: Translator.trans('segments.export.statement.label'),
+        right: Translator.trans('segment.recommendation')
+      }
+
+      return defaultPlaceholders[key]
+    },
```
salisdemos, line 172: I looks like you do something weird with the placeholders.
Why do You have to reset them?

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,241 @@
...
+  methods: {
+    applyDefaultPlaceholdersForEmptyNames () {
+      Object.keys(this.docxColumns).forEach(key => {
+        if (!this.docxColumns[key].title) {
+          this.docxColumns[key].title = null
+          this.docxColumns[key].placeholder = this.getDefaultPlaceholderByKey(key)
+        }
+      })
+    },
+
```
salisdemos, line 157: I suggest to remove this method here, the @input="handleInput" in the Template and the handleInput Method.
You should not have to change the Placeholder.

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,241 @@
...
+        if (columnTitle) {
+          this.updateSessionStorage(storageKey, columnTitle)
+          this.docxColumns[key].placeholder = columnTitle
+          columnTitles[key] = columnTitle
+        } else {
+          this.removeFromSessionStorage(storageKey)
+          this.docxColumns[key].placeholder = this.getDefaultPlaceholderByKey(key)
+          columnTitles[key] = null /** Setting the value to null will trigger the display of the default column titles */
+        }
```
salisdemos, line 194:
```suggestion
        if (columnTitle) {
          columnTitles[key] = columnTitle === '' ? null : columnTitle
          this.updateSessionStorage(storageKey, columnTitles[key])
        } else {
          this.removeFromSessionStorage(storageKey)
          columnTitles[key] = null /** Setting the value to null will trigger the display of the default column titles */
        }
```

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,241 @@
...
+    openModal () {
+      this.setInitialValues()
+      this.$refs.exportModalInner.toggle()
+    },
+
+    resetColumnsUnsavedValues () {
+      Object.keys(this.docxColumns).forEach(key => {
+        this.setTitleAndPlaceholderByKey(key)
+      })
+    },
```
salisdemos, line 221: this is redundant with setInitialValues. I think we can delete this one here and only set the values on open. Doing it twice is a bit unnecessary ;-)

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,241 @@
...
+
+    setTitleAndPlaceholderByKey (key) {
+      const storageKey = `exportModal:docxCol:${key}`
+      const storedColumnTitle = this.getItemFromSessionStorage(storageKey)
+      this.docxColumns[key].title = storedColumnTitle || null /** Setting the value to null will trigger the display of the default column titles */
+      this.docxColumns[key].placeholder = storedColumnTitle || this.getDefaultPlaceholderByKey(key)
+    }
+  },
```
salisdemos, line 235:
```suggestion
    setInitialValues () {
      Object.keys(this.docxColumns).forEach(key => {
        const storageKey = `exportModal:docxCol:${key}`
        const storedColumnTitle = this.getItemFromSessionStorage(storageKey)
        this.docxColumns[key].title = storedColumnTitle || null /** Setting the value to null will trigger the display of the default column titles */
        this.docxColumns[key].placeholder = storedColumnTitle || this.getDefaultPlaceholderByKey(key)
      })
    }
  },
```

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,241 @@
...
+
+  mounted () {
+    this.setInitialValues()
+  }
```
salisdemos, line 239: I think we don't need that. The method gets triggered on open anyway

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,241 @@
...
+      this.$emit('export', {
+        route: this.exportTypes[this.active].uploadPath,
+        docxHeaders: ['docx', 'zip'].includes(this.active) ? columnTitles : null
+      })
+      this.closeModalAndResetValues()
+    },
+
+    handleInput () {
+      this.applyDefaultPlaceholdersForEmptyNames()
+    },
```
salisdemos, line 206: As I wrote above. This can be removed. But in general: If a method simply calls one other method, You can use that one straight away.

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -915,11 +902,10 @@ export default {
       }
     },
 
-    showHintAndDoExport (route) {
+    showHintAndDoExport ({ route, docxHeaders }) {
       if (window.dpconfirm(Translator.trans('export.statements.hint'))) {
-        window.location.href = this.exportRoute(route)
+        window.location.href = this.exportRoute(route, docxHeaders)
```
salisdemos, line 907: could that be archived by an APi-request?

salisdemos (changes requested):

salisdemos:
Please merge main into this Branch

client/js/components/statement/listStatements/ExportModal.vue
sakutademos: because the placeholders could be different depending on what is saved in sessionStorage. I mean, if there is nothing saved in sessionStorage, the default placeholders are used. If there are saved values in sessionStorage, they will be used as the placeholders...the user needs to know/see the titles of the table headers are at the moment when the window is opened

salisdemos: Yes. but if there is something stored, that values should not be displayed as placeholders, but as value from that input.

salisdemos: The placeholder it self should not change

client/js/components/statement/listStatements/ExportModal.vue
sakutademos: done

client/js/components/statement/listStatements/ExportModal.vue
sakutademos: done here https://github.com/demos-europe/demosplan-core/pull/3315/commits/fce45b714549e364a05dce66e055ef6de1c8bf85

client/js/components/statement/listStatements/ListStatements.vue
sakutademos: it could be, but it should be adjusted in the BE,  it was implemented in a different way. Should it be done via API?

salisdemos: not necessarily right away, but it would be better I guess. with that type of requests, we always got a page reload wich is not so user friendly

sakutademos: hm, there is no page reload after exporting the files.

I do not know how complex it would be for the BE to switch to an Api, but it should have been done a bit earlier.. it could be discussed with the be, but I am not sure if it is necessary because there is no difference for the user. I do not see a page reload

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,200 @@
...
+    closeModal () {
+      this.$refs.exportModalInner.toggle()
+    },
+
+    handleAbort () {
+      this.$emit('abort')
+      this.closeModal()
+    },
+
+    handleExport () {
```
salisdemos, line 157: if you find the time to write test, I think it would be good to test if the emitted value looks the way its suppose to look

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -451,22 +426,34 @@ export default {
...
+            ...this.searchFieldsSelected !== null ? { fieldsToSearch: this.searchFieldsSelected } : {}
+          },
+          sort: this.selectedSort
+        }
+
+        if (docxHeaders) {
+          parameters.tableHeaders = {
+            c_left: docxHeaders.left,
+            c_middle: docxHeaders.middle,
+            c_right: docxHeaders.right
```
salisdemos, line 451: if its not too complicated to refactor, can we just use numbers for the cols. Than we can simpler extend if there will come more or we want to use the logic in other situations.
```suggestion
            col1: docxHeaders[0],
            col2: docxHeaders[1],
            col3: docxHeaders[2]
```

salisdemos (changes requested):

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,200 @@
...
+  },
+
+  mixins: [sessionStorageMixin],
+
+  data () {
+    return {
+      active: 'docx',
+      docxColumns: {
+        left: {
+          width: 'col-span-1',
```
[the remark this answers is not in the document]
sakutademos: ask Hans :) Here is a design template --> https://www.figma.com/proto/qKBgvF8gUAM1xIBA8Ubqgh/DPLAN-11773-Export-Modal?node-id=7-26&t=QGu9ukLPDXMD2ugA-1&scaling=min-zoom&content-scaling=fixed&page-id=4%3A5

It is probably because the column/title for Id needs less space than the other columns. As I see in the export document this column is smaller as well.

client/js/components/statement/listStatements/ExportModal.vue
sakutademos: right, i will write tests in the separate PR

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,200 @@
...
+            :id="key"
+            type="text"
+            :data-cy="column.defaultValue"
+            :placeholder="Translator.trans(column.placeholder)"
+            :width="column.width"
+            v-model="column.title" />
+        </div>
+      </section>
+
+      <div class="flex justify-end">
```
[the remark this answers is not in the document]
sakutademos: I have to place the button on the right side. Is a flex a bad choice for this?

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,200 @@
...
+    }
+  },
+
+  methods: {
+    closeModal () {
+      this.$refs.exportModalInner.toggle()
+    },
+
+    handleAbort () {
+      this.$emit('abort')
```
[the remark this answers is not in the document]
sakutademos: maybe we have to do smth on the abort event in the parent component, but you are right... It is not needed here right now

client/js/components/statement/listStatements/ListStatements.vue
sakutademos: We renamed it to `col1`, `col2`, and `col3`. it could be easier to handle in the future, you are right. https://github.com/demos-europe/demosplan-core/pull/3315/commits/540aa687fc1655e7584f92e760a451857def3161

However, why do you prefer an array in this case and not an object? I think it is clearer to use 'docxHeaders' as a  `key: value`

translations/messages+intl-icu.de.yml
```diff
@@ -1577,10 +1579,11 @@ export.portrait: "Hochformat"
 export.portrait.fragments: "Hochformat (mit Datensätzen)"
 export.potentialAreas: "Potenzialflächen"
 export.settings.hint: "Wählen Sie aus, welche Daten im Export der Stellungnahmen in der Abwägungstabelle enthalten sein sollen. Die Einstellungen betreffen nur den DOCX-Export \"Querformat\"."
-export.statements.zip: "Stellungnahmen als einzelne DOCX-Dateien"
-export.statements.docx: "Stellungnahmen in einer DOCX-Datei"
+export.statements: "Stellungnahmen exportieren"
+export.statements.zip: "einzelne DOCX-Dateien"
+export.statements.docx: "DOCX-Datei"
```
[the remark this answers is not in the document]
sakutademos: done here https://github.com/demos-europe/demosplan-core/pull/3315/commits/d581ed4fb58e7fe2bfa9e76ed3421f1f0c0c86af. Overwrite it in ewm project: https://github.com/demos-europe/demosplan-project-ewm/pull/122

client/js/components/statement/listStatements/ExportModal.vue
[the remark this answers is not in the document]
salisdemos: Do You know, that the DPButtonRow-Component has a Prop "align" wich should do the trick?

salisdemos (changes requested):
Just one tiny thing left

translations/messages+intl-icu.de.yml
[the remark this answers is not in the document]
salisdemos: please no Links to projects :pray:

client/js/components/statement/listStatements/ExportModal.vue
```diff
@@ -0,0 +1,200 @@
...
+    <dp-modal
+      ref="exportModalInner"
+      content-classes="w-11/12 sm:w-10/12 md:w-8/12 lg:w-6/12 xl:w-5/12">
+      <h2 class="mb-5">
+        {{ Translator.trans('export.statements') }}
+      </h2>
+
+      <section>
+        <h3 class="text-lg">
+          {{ Translator.trans('file.format') }}
```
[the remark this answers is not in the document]
salisdemos: please no Links to project-PRs. Only the other way round!

translations/messages+intl-icu.de.yml
```diff
@@ -1577,10 +1579,11 @@ export.portrait: "Hochformat"
...
 export.statements.hint: "Achtung: Falls weitere Einreichende zu einer Stellungnahme hinzugefügt wurden, werden diese auch im DOCX-Export aufgeführt."
-export.statements.xlsx: "Stellungnahmen in einer XLSX-Datei"
+export.statements.xlsx: "XLSX-Datei"
```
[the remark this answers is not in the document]
sakutademos: done

client/js/components/statement/listStatements/ExportModal.vue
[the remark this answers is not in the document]
salisdemos: Thanks for investigating. Thats weired. But yeah. we should then rename the prop. (And find out, why the default doesnt work)

salisdemos (approved):

MoritzMandler (approved):
lgtm

salisdemos (approved):

PR #3332, opened 2024-06-26
dplan 4020 refactor controller

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-4020/3-5-Als-Bearbeiterin-mochte-ich-eigene-Karten-zur-Planung-einbinden-konnen-um-Lagebeziehungen-oder-nicht-unmittelbar-erkennbare

Description
1. I created a new action in the controller: DemosPlanDocumentController::saveImportedElementsAdminAction
- This action is called in the template elements_admin_import.html.twig
- More exactly, it is called when the user clicks "save" button
- What this action does is to save the imported documents
2. That action then redirects to the existing one: DemosPlanDocumentController::elementAdminListAction
- I adjusted the existing one inspired on the work that you did on DemosPlanMapController

### How to review/test
- Case 1: Go to diplanfest  and bobsh --> verfahren/PROCEDURE_ID/verwalten/planunterlagen
- import document
- cancel import and play around with it.
- It should work as expected

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Controller/Document/DemosPlanDocumentController.php
```diff
@@ -634,103 +634,101 @@ public function elementAdminListAction(
...
-        // properly transformed into FlashBag messages, since the method for that is called in the
-        // DemosPlanResponseListener, see bug T17790.
-        if (0 !== (is_countable($requestPost) ? count($requestPost) : 0)) {
-            return $this->redirectToRoute('DemosPlan_element_administration', ['procedure' => $procedure]);
-        }
-
         return $this->renderTemplate(
             '@DemosPlanCore/DemosPlanDocument/elements_admin_list.html.twig',
-            compact('templateVars', 'procedure', 'title', 'procedureSettings')
+            compact('templateVars', 'title', 'procedureSettings')
```
salisdemos, line 679: Am I right, that the `'/verfahren/{procedure}/verwalten/planunterlagen` elementAdminListAction-Controller now would look more or less the same as the `/verfahren/{procedure}/verwalten/karte` elementAdminListAction-Controller and one of them can be deleted in the End?

salisdemos (commented):
First of all. For me it looks really good.
But I like to have a second pair of eyes on that from a backend person.
And: I think this is more an improvement than part of the US. So we should not merge this before Featurefreeze and the point it to main. Or is there anything within here, that is required for the Feature?

demosplan/DemosPlanCoreBundle/Controller/Document/DemosPlanDocumentController.php
sierrazamudiodemos: hey @salisdemos exactly, one of them can be deleted, I would recommend to delete the /verwalten/karten

config/menus.yml
```diff
@@ -319,14 +319,14 @@ submenu_procedures:
...
-        path: 'DemosPlan_map_administration_overview'
+        path: 'DemosPlan_element_administration'
         child_paths: [
             'DemosPlan_map_administration_gislayer',
             'DemosPlan_map_administration_gislayer_new',
             'DemosPlan_map_administration_map',
         ]
         permission: ['area_admin_initial_map_view'] # Only controls the menu item. 'area_admin_map' is required for the map settings
-        path_params: ['procedureId']
+        path_params: ['procedure']
```
martendemos, line 329: I guess the parameter contains a string? (procedureId)
Therefore I would prefer the name "procedureId" if possible.

sierrazamudiodemos: hey, this url is used in another templates that are not touched in this PR, if I adjust it, I will have to adjust the other templates (which is more a FE thing) I can do it but then it becomes a bit of a bigger scope

salisdemos (approved):

salisdemos (approved):

martendemos (dismissed):
got an permission denied when I try to access the "/verwalten/planunterlagen" route.

salisdemos (approved):

salisdemos:
we doublechecked the concerns from @martendemos so it should be fine to overrule him.

PR #3343, opened 2024-06-28
fix (DPLAN-11957) Create permission to display options

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-11957/Verschiedene-Kartenmoglichkeiten-gelten-nicht-fur-EWM

Create permissions instead of overriding the template to display checkboxes

### How to review/test
- test ewm: there should not be the checkboxes
- test other project e.g. blp there should be the checkboxes


jalalifardemos (approved):
lgtm

PR #3344, opened 2024-06-28
fix(refs DPLAN-11801): Prevent negative reports with content

salisdemos:
- if negative report is set, it should not be possible to send any 'meaningful' data.
- on the other hand if there is already something selected / typed in, it should not be possible to select 'negative_report'

Can be tested on public detail if the procedure is in participation. Check especially the documents relation. for that You need some documents/elements with the posibility to reference them....

### DIFF
--- a/client/js/components/statement/publicStatementModal/StatementModal.vue
+++ b/client/js/components/statement/publicStatementModal/StatementModal.vue
@@ -110,6 +110,7 @@
               }"
               value="0" />
             <dp-radio
+              :disabled="canNotBeNegativeReport"
               name="r_isNegativeReport"
               id="negative_report_true"
               data-cy="statementModal:indicationerror"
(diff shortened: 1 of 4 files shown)


sakutademos (approved):
lgtm, code review

client/js/components/statement/publicStatementModal/StatementModal.vue
```diff
@@ -906,6 +910,14 @@ export default {
       userId: 'userId'
     }),
 
+    canNotBeNegativeReport () {
```
ahmad-demos, line 913: I see this name is long!
maybe `isNotNegativeReport` better?

salisdemos: Its long but still readable imho

PR #3347, opened 2024-06-28
feature (ref 11961): allow to delete Gislayer

jalalifardemos:
### Ticket: 
DPLAN-11961

Description: get access to delete GisLayer

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/ResourceTypes/GisLayerResourceType.php
+++ b/demosplan/DemosPlanCoreBundle/ResourceTypes/GisLayerResourceType.php
@@ -86,7 +86,7 @@ final class GisLayerResourceType extends DplanResourceType
 
     public function isDeleteAllowed(): bool
     {
-        return true;
+        return $this->currentUser->hasPermission('area_map_participation_area');
     }
 
     public function isGetAllowed(): bool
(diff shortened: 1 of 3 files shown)


demosplan/DemosPlanCoreBundle/ResourceTypes/GisLayerResourceType.php
```diff
@@ -84,6 +84,11 @@ public function isAvailable(): bool
         return true;
     }
 
+    public function isDeleteAllowed(): bool
```
sierrazamudiodemos, line 87: Lets add the permission area_map_participation_area as it is done with the isGetAllowed, and isListAllowed, in this way, the deletion is also protected

sierrazamudiodemos (changes requested):
Add area_map_participation_area permission to isDeleteAllowed (as it is done with isGetAllowed, and isListAllowed)

demosplan/DemosPlanCoreBundle/ResourceTypes/GisLayerResourceType.php
jalalifardemos: done in [5250f95](https://github.com/demos-europe/demosplan-core/pull/3347/commits/5250f95a3e681ed932017b8b3b13dcb706a4156f)

sierrazamudiodemos (approved):
BE works. FE does not refresh directly

sierrazamudiodemos:
@salisdemos the BE part is fixed, could you the FE part? it seems that the after deleting, the FE does not detect the changes directly, let us know if there is something to do in the BE side

sierrazamudiodemos (approved):
thanks florian and saeed for expanding the resource type :) it works

sakutademos (approved):
lgtm, code review

PR #3352, opened 2024-07-01
bug(refs DPLAN-11849): Set boundingbox only if it exists

gruenbergerdemos:
### Ticket
DPLAN-11849

- Boundingbox is an array, so we need to check for length
- Remove unused features ()

### How to review/test
Choose a procedure where the initial extent is not set yet (Karteneinstellungen -> Startkartenausschnitt)
Verfahren -> Abschnitte -> Erwiderung verfassen -> Ortsbezug -> There should be a map

### PR Checklist


salisdemos (approved):
:see_no_evil:

PR #3358, opened 2024-07-02
chore: fix tests that fail the build

mussbach:
There are a few issues with our current phpunit tests that make the jenkins build fail

### How to review/test
run all tests with `phpunit`. no fatal errors should occur

### PR Checklist

- [x] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/tests/backend/core/Core/Unit/Logic/PermissionsTest.php
+++ b/tests/backend/core/Core/Unit/Logic/PermissionsTest.php
@@ -20,6 +20,7 @@ use demosplan\DemosPlanCoreBundle\Entity\User\OrgaType;
 use demosplan\DemosPlanCoreBundle\Entity\User\Role;
 use demosplan\DemosPlanCoreBundle\Entity\User\User;
 use demosplan\DemosPlanCoreBundle\Entity\User\UserRoleInCustomer;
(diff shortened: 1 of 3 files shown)


tests/backend/core/Core/Unit/Logic/PermissionsTest.php
```diff
@@ -129,6 +130,7 @@ protected function getPermissionsInstance(bool $ownsProcedure, bool $inviteOrgaF
         $validator = self::$container->get(ValidatorInterface::class);
         $procedureRepository = $this->getProcedureRepositoryMock();
         $permissionsClass = $this->getPermissionsClass();
+        $accessControlService = self::$container->get(AccessControlService::class);
```
sierrazamudiodemos, line 133: question: this is added to potentially be used in the future for permissions in other tests?

sierrazamudiodemos (approved):
Thanks for doing this, I run the tests and we have 1698, some failed and this is already a good step

tests/backend/core/Core/Unit/Logic/PermissionsTest.php
mussbach: No, this is added because the it was added to the constructor of `Permissions` and therefore needed to be passed here as otherwise all permission tests broke

PR #3370, opened 2024-07-03
chore (refs DPLAN-12006): register default users for new customer

mussbach:
### Ticket
DPLAN-12006

When the exist the citizen user and the ai api user should be automatically registered to avoid errors

### How to review/test
create a new customer via `bin/<project> dplan:data:generate-customer` it should contain two users

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


sierrazamudiodemos (approved):
It works: 2 users are stored in DB under this new customer. Question: why do we need an anonymous user?

mussbach:
In principle, we could assume that "no user" (null) is the anonymous user, but the design decision was made decades ago that a user is always needed. There are still places where this assumption holds true, and while getting rid of the user might have some advantages, we would have to refactor a lot of code first. We can definitely discuss this topic and update where we stand.

PR #3379, opened 2024-07-05
fix: Interpret String as String

salisdemos:
If we got a real string from TWIG, we can't try to compute it.


mussbach (approved):
darn, I did this in my initial PR but forgot it here. Thank you

PR #3380, opened 2024-07-05
fix(refs dplan-11923): Use Title instead of Name to display Tags

salisdemos:
### Ticket
DPLAN-11923

To display the Tags of a Group we need a title, not a name

I wonder if/how this ever worked. Looks like the code hasn't changed for years

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_convert_annotated_pdf.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_convert_annotated_pdf.html.twig
@@ -1,7 +1,14 @@
 {% extends '@DemosPlanCore/DemosPlanCore/procedure.html.twig' %}
 
 {% block component_part %}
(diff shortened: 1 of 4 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_convert_annotated_pdf.html.twig
```diff
@@ -1,7 +1,14 @@
 {% extends '@DemosPlanCore/DemosPlanCore/procedure.html.twig' %}
 
 {% block component_part %}
-    {% set allTags = templateVars.availableTopics|default([])|map(el => { id: el.id, title: el.title, tags: el.tags|map(tag => { id: tag.id, name: tag.title }) }) %}
+    {% set allTags = templateVars.availableTopics|default([])|map(el => {
+        id: el.id,
+        title: el.title,
+        tags: el.tags|map(tag => {
+            id: tag.id,
+            title: tag.title
```
salisdemos, line 9: the only real change I did in this block is to rename `name: tag.title` to `title: tag.title` the rest ist just reformating

PR #3399, opened 2024-07-09
fix(refs dplan-12020): Prevent Layer Url from doubling the Service param

salisdemos:
### Ticket
DPLAN-12020

Somewhere in OpenLayers the Service-Param gets added. If its a part of the stored Url, we have to remove it, before creating the Layer.

### How to review/test

Try to use this Url in the segments area 
https://maps.deges.de/server/services/AWF_EWMS_WMS_WFS/VKE3251_2_S84_Achse_dxf_25833/MapServer/WMSServer?SERVICE=WMS
It should display something around Coswig near Dresden


gruenbergerdemos (approved):
lgtm

PR #3400, opened 2024-07-09
fix(refs dplan-12021): Pass Opacity to Segments Map

salisdemos:
### Ticket
DPLAN-12021

If the Opacity is not 100% but less it should be visible in the segments-map.

### DIFF
--- a/client/js/components/map/map/DpOlMap.vue
+++ b/client/js/components/map/map/DpOlMap.vue
@@ -107,6 +107,7 @@
           :key="layer.name"
           :attributions="layer.attribution || ''"
           :order="layer.mapOrder + 1"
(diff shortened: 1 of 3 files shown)


salisdemos:
> This seems to be not working, my layers still have 100% opacity. Is `this.opacity` missing in client/js/components/map/map/DpOlMapLayer.vue:110 `const layer = createTileLayer(this.title, this.name, this.source, this.opacity)`?

thanks for spotting. I don't know where this got lost. But i did.
024ad35854a234bf0e7ae7b78ed8cde3350f8ffd

gruenbergerdemos (approved):
lgtm!

PR #3405, opened 2024-07-11
fix(refs DPLAN-11803): properly calculate inParticipation phase

salisdemos:
all phases with a "write" permission are participation phases. So we don't have to rely of the string spelling.

### Ticket
DPLAN-11803


ahmad-demos (approved):
lgtm, codeReview!

PR #3407, opened 2024-07-11
tests: fix unit tests

gruenbergerdemos:
Fix unit tests

- Vuex module names are case sensitive -> fix names
- Update snapshot for changed component


salisdemos (approved):
Thanks a lot!

PR #3408, opened 2024-07-12
fix(refs DPLAN-11488): Prevent error on unset departments

salisdemos:
### Ticket
DPLAN-11488

sometimes a department is not set (yet). So we have to prevent the script from throwing errors in that case. (same with allowed roles)


gruenbergerdemos (approved):
lgtm

PR #3420, opened 2024-07-12
fix(refs DPLAN-11964): Set empty array as empty Object for Territory

salisdemos:
The Components expects an Object and can't handle (empty) Arrays. Maybe the fix should be done where ever the Array is set. But I couldn't find out where that happens (in acceptable time)

### Ticket
DPLAN-11964

### DIFF
--- a/client/js/components/map/admin/MapAdmin.vue
+++ b/client/js/components/map/admin/MapAdmin.vue
@@ -37,7 +37,7 @@
       :max-extent="procedureMapSettings.attributes.defaultMapExtent"
       :procedure-id="procedureId"
       :procedure-coordinates="procedureMapSettings.attributes.coordinate"
-      :procedure-init-territory="procedureMapSettings.attributes.territory.length > 0 ? procedureMapSettings.attributes.territory : {}"
+      :procedure-init-territory="Array.isArray(procedureMapSettings.attributes.territory) ? {} : procedureMapSettings.attributes.territory"
       :scales="procedureMapSettings.attributes.availableScales"
       @field:update="setField" />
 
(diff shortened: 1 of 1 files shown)


client/js/components/map/admin/MapAdmin.vue
```diff
@@ -37,7 +37,7 @@
       :max-extent="procedureMapSettings.attributes.defaultMapExtent"
       :procedure-id="procedureId"
       :procedure-coordinates="procedureMapSettings.attributes.coordinate"
-      :procedure-init-territory="procedureMapSettings.attributes.territory"
+      :procedure-init-territory="procedureMapSettings.attributes.territory.length > 0 ? procedureMapSettings.attributes.territory : {}"
```
gruenbergerdemos, line 40: I don't see how this can work. 
If territory is an array: It breaks when territory is not empty, because mapView expects an object.
If territory is an object: It does not work when territory is not empty, because checking for length in an object returns undefined and mapView gets {} instead of the correct object

Locally it does not work like this, but I get always an object (empty or not). No idea, where the empty array on prod comes from.

gruenbergerdemos (changes requested):

client/js/components/map/admin/MapAdmin.vue
salisdemos, line 40:
```suggestion
      :procedure-init-territory="Array.isArray(procedureMapSettings.attributes.territory) ? {} : procedureMapSettings.attributes.territory"
```

salisdemos: > I don't see how this can work.

You are right. I only tested it with the corrupt data, not with something that should work anyhow.
changed in [6ccdcaa](https://github.com/demos-europe/demosplan-core/pull/3420/commits/6ccdcaaee72f5eabb622029a7faaba1f4a3b8897)


> No idea, where the empty array on prod comes from.

Me too. but it does. I couldn't find out how it happens, but I checked the DB and found quite a hand full

gruenbergerdemos (approved):
Works so far. 

Optional suggestion: Maybe `client/js/store/map/ProcedureMapSettings.js:83` would be a better place for the check?

PR #3431, opened 2024-07-16
fix(refs DPLAN-12009): Fix file copies on STN copy to different procedure

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-12009/ROBOB-Prod-Bei-verschobenen-STN-wurden-in-der-Ubersicht-die-Links-von-Anhangen-nicht-angepasst

Description:
All original statement creations where handling their file copy tasks correct as far as setting individual file-entries for all of them. The only problem was a procedureId that was set to late on the new statement in order to be used for those files.

on copy there was a method StatementAttachmentService::copyToStatement that made use of a createAttachment
method that did not create a new file-entry for the attachment but set the same file-reference. Changed this method to use the correct one from the statement-repository.

on move it was more tricky.
When creating new statements - after their originals are done we create the (normal) statements by... ```StatementHandler::createNonOriginalStatement::copyStatementObjectWithinProcedureWithRelatedFiles::copyStatementObjectWithinProcedure::copyAttachmentEntries::copyToStatement``` ... in the end calling the same problematic method as described above. Therefore The statement and its original use the same file-reference inside their containers.
This makes it necessary to create new file-entries for all moved files - as otherwise the files would be shared with the original-statement which remains inside the source procedure.
But not that quick - by changing the ```copyToStatement``` method - the from now on newly-created statements wont share their statement-attachment files with its original anymore. So only the weitere-attachments need to get new files created on move. The statement-attachment is fine with just an update of its procedure.

The picked changes from the here linked PR are necessary to access the correct now procedure related file by a given hash.

### Linked PRs (optional)
picked part of the fix from:
https://github.com/demos-europe/demosplan-core/pull/2850

### Tasks (optional)
Fix wrong db-entries of failed copies via migration?

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Repository/StatementRepository.php
+++ b/demosplan/DemosPlanCoreBundle/Repository/StatementRepository.php
@@ -138,7 +138,7 @@ class StatementRepository extends CoreRepository implements ArrayInterface, Obje
     /**
      * Add new ClusterStatement, what basically is a Statement with a cluster of Statements.
      *
-     * @return statement - headStatement of the created statement-cluster
+     * @return Statement - headStatement of the created statement-cluster
      *
      * @throws Exception
      */
@@ -479,7 +479,7 @@ class StatementRepository extends CoreRepository implements ArrayInterface, Obje
     /**
      * @return array<int, Statement|Segment>
      */
-    public function getEntities(array $conditions, array $sortMethods, int $offset = 0, int $limit = null): array
+    public function getEntities(array $conditions, array $sortMethods, int $offset = 0, ?int $limit = null): array
     {
         return parent::getEntities($conditions, $sortMethods, $offset, $limit);
     }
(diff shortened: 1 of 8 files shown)


mussbach (approved):
lgtm, code review only

demosplan/DemosPlanCoreBundle/Repository/StatementRepository.php
```diff
@@ -2032,9 +2031,11 @@ public function copyFileContainers(Statement $originalToCopy, Statement $newOrig
             ->getStatementFileContainers($originalToCopy->getId());
 
         $fileStrings = [];
+        $copiedFileContainers = [];
```
mussbach, line 2034: this change does not do anything

mussbach (approved):
lgtm, one change is not necessary, but no blocker

demosplan/DemosPlanCoreBundle/Repository/StatementRepository.php
bakridemos: u re right, thx, it's removed here https://github.com/demos-europe/demosplan-core/pull/3431/commits/9642bf72c7b8bd1b666480db653a1bdef637f762

bakridemos (approved):

PR #3432, opened 2024-07-16
fix(refs DPLAN-11488): Make departments updateable again

salisdemos:
### Ticket
DPLAN-11488

I think the main issue was the wrong data while updating the select. But during the way to find that one I tried to make thinks more resilient. I hope its still reviewable :-/


ahmad-demos (approved):
lgtm, codeReview!

PR #3434, opened 2024-07-17
feature (ref DPLAN11903): use timezone to export right time

jalalifardemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-11903/falsche-Zeitangabe-PDF-Export-Verfahrensprotokoll

Description: set timezone to export right time

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Report/DemosPlanReportController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Report/DemosPlanReportController.php
@@ -75,7 +75,7 @@ class DemosPlanReportController extends BaseController
         $slugify = new Slugify();
         $procedure = $procedureHandler->getProcedureWithCertainty($procedureId);
 
-        $currentTime = Carbon::now();
+        $currentTime = Carbon::now()->timezone('Europe/Berlin');
         $reportMeta = [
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Controller/Report/DemosPlanReportController.php
```diff
@@ -70,12 +70,12 @@ public function exportProcedureReportAction(
...
         PermissionsInterface $permissions,
         ProcedureHandler $procedureHandler,
-                              $procedureId
+        $procedureId
     ): Response {
         $slugify = new Slugify();
         $procedure = $procedureHandler->getProcedureWithCertainty($procedureId);
 
-        $currentTime = Carbon::now();
+        $currentTime = Carbon::now()->timezone('Europe/Berlin');
```
sierrazamudiodemos, line 78: I would not hard code 'Europe/Berlin', rather try to figure out how to get it from a system configuration., I wonder if this happens in prod (it can be only the test env that does not have Europe Berlin configured

sierrazamudiodemos (changes requested):

PR #3448, opened 2024-07-18
feat(refs DPLAN-3796): Add input for file name in exports

gruenbergerdemos:
### Ticket
DPLAN-3796

Add input for file name for docx exports. You can use {ID}, {NAME} and {EINGANGSNR} as placeholders. 

### How to review/test
Verfahren -> Stellungnahmen -> Exportieren -> einzelne DOCX-Dateien
Verfahren -> Stellungnahmen -> Details und Erwiderungen -> Exportieren
Try to use the placeholders and export. Also see if example file name beneath the input shows the right file name

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-core/pull/3395


client/js/components/statement/StatementExportModal.vue
```diff
@@ -154,6 +177,29 @@ export default {
   computed: {
     exportModalTitle () {
       return this.isSingleStatementExport ? Translator.trans('statement.export.do') : Translator.trans('export.statements')
+    },
+
+    exampleFileName () {
+      let exampleFileName = 'm101-jacob-meier-e5089.docx'
+      const exampleId = 'm101'
+      const exampleName = 'jacob-meier'
+      const exampleInternId = 'e5089'
```
spiess-demos, line 186: Nitpick: i'd suggest to chose an uppercase "M" within the value of `exampleId`, to better match the "real system".

gruenbergerdemos: Yeah, I checked and in the exported name it is lower case. That's why I chose lowercase here. I think it needs to be adjusted in BE if we want to keep upper/lower case. At the moment it seems like everything is lowercase.

spiess-demos: Thanks for the clarification, i did not actually look into the exported files.

client/js/components/statement/StatementExportModal.vue
```diff
@@ -54,24 +55,45 @@
...
+            aria-labelledby="docxFileName"
+            :text="Translator.trans('docx.export.file_name.hint')" />
+          <dp-input
+            id="fileName"
+            v-model="fileName"
+            :placeholder="Translator.trans('docx.export.file_name.placeholder')"
+            type="text"/>
+          <small>
+            <span
+              class="font-semibold"
```
spiess-demos, line 82: `font-semibold` sadly does not play well with the demosplan-core font stack as some configurations use custom fonts where not all weights are supported. For the time being, use `weight--bold` here.

client/js/components/statement/StatementExportModal.vue
```diff
@@ -54,24 +55,45 @@
...
+          </h3>
+          <dp-contextual-help
+            aria-labelledby="docxFileName"
+            :text="Translator.trans('docx.export.file_name.hint')" />
+          <dp-input
+            id="fileName"
+            v-model="fileName"
+            :placeholder="Translator.trans('docx.export.file_name.placeholder')"
+            type="text"/>
+          <small>
```
spiess-demos, line 80: When looking at the example text, i feel it is very small. Also, hints in DpLabel are styled with `font-size-small`. I'd use that here, too.

client/js/components/statement/StatementExportModal.vue
```diff
@@ -54,24 +55,45 @@
...
             :placeholder="Translator.trans(column.placeholder)"
-            :width="column.width"
-            v-model="column.title" />
+            type="text"
+            :width="column.width" />
         </div>
+        <template v-if="this.active === 'zip' || isSingleStatementExport">
+          <h3
+            id="docxFileName"
+            class="inline-block text-lg mr-1">
```
spiess-demos, line 69: `text-lg` is not yet configured to use demosplan sizing. This will be soon rolled out, changing the value of `text-lg` to 19.2px. However in the designs i meant those to be "label-like" in terms of font-size and font-weight (16px / segoeUi semibold). When seeing this on a Debian System i see that this is where system fonts make it hard to match the design, as Ubuntu uses `DejaVu Sans` which has no weight of 600 which is semibold, leading to the bold weight to be used. I feel this needs to be tackled globally, i guess by ditching system fonts entirely. For the time being i think `text-lg` can be used.

client/js/components/statement/StatementExportModal.vue
```diff
@@ -54,24 +55,45 @@
...
+            v-model="column.title"
             :data-cy="column.dataCy"
             :placeholder="Translator.trans(column.placeholder)"
-            :width="column.width"
-            v-model="column.title" />
+            type="text"
+            :width="column.width" />
         </div>
+        <template v-if="this.active === 'zip' || isSingleStatementExport">
+          <h3
```
spiess-demos, line 67: First i thought, why isn't this a `label` prop in DpInput. But then, the headings before would also need to be labels and for the `Spaltennamen in DOCX-Exporten` this would not work out?

I think there is a semantic element which could be used for that:

```
<fieldset>
  <legend class="styled exactly like label">Spaltennamen in DOCX-Exporten</legend>
  <dp-input /> <dp-input /> <dp-input />
</fieldset>
```

However as this needs a little more refining, i just wanted to drop it here but no need to do it now.

client/js/components/statement/StatementExportModal.vue
gruenbergerdemos: Done in [47a88bc](https://github.com/demos-europe/demosplan-core/pull/3448/commits/47a88bc1302b10013a094a604ce9dd2bd60fed7a)

client/js/components/statement/StatementExportModal.vue
gruenbergerdemos: Done in [47a88bc](https://github.com/demos-europe/demosplan-core/pull/3448/commits/47a88bc1302b10013a094a604ce9dd2bd60fed7a)

client/js/components/statement/StatementExportModal.vue
gruenbergerdemos: Thanks for pointing out. I'd like to look into that later, maybe next week, so the feature freeze can happen tomorrow (and I'm not working tomorrow).

spiess-demos (approved):
lgtm!

PR #3468, opened 2024-07-22
bug (refs DPLAN-12079): Remove comments relationship

gruenbergerdemos:
### Ticket
DPLAN-12079

Somehow, after editing the recommendation text and saving, the comment relationship is added to the payload in the save request for updating the map. Comments need to be removed as updating them is technically not supported. After completing the request, they are added again to the store to be able to display them.

### How to review/test
Verfahren -> Abschnitte -> Erwiderung verfassen -> Bearbeiten and save recommendation text. Then open 'Ortsbezug' and draw something in the map and save it.

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
+++ b/client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
@@ -241,6 +241,22 @@ export default {
       })
     },
 
+    /**
+     * Restore non-updatable comments from segments relationships after update request
+     */
+    restoreComments (comments) {
+      if (comments) {
+        const segmentWithComments = {
+          ...this.segment,
+          relationships: {
+            ...this.segment.relationships,
+            comments: comments
+          }
+        }
+        this.setItem({ ...segmentWithComments, id: this.segment.id })
+      }
+    },
+
     save () {
       this.setItem({
         ...this.segment,
@@ -249,13 +265,24 @@ export default {
           polygon: JSON.stringify(this.featuresObject)
         }
       })
+      const comments = { ...this.segment.relationships.comments } || null
+
+      /**
+       *  Comments need to be removed as updating them is technically not supported
+       *  After completing the request, they are added again to the store to be able to display them
+       */
+      if (this.segment.relationships.comments) {
+        delete this.segment.relationships.comments
+      }
 
       return this.saveSegmentAction(this.segmentId)
         .then(checkResponse)
         .then(() => {
+          this.restoreComments(comments)
           dplan.notify.confirm(Translator.trans('confirm.saved'))
         })
         .catch(() => {
+          this.restoreComments(comments)
           dplan.notify.error(Translator.trans('error.changes.not.saved'))
         })
     },


client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
```diff
@@ -249,13 +265,24 @@ export default {
           polygon: JSON.stringify(this.featuresObject)
         }
       })
+      const comments = { ...this.segment.relationships.comments } || null
```
salisdemos, line 268: I think (and sonarcloud thinks that too) that {} won't resolve to false and with that it will never be null.

client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
```diff
@@ -241,6 +241,22 @@ export default {
...
+    restoreComments (comments) {
+      if (comments) {
+        const segmentWithComments = {
+          ...this.segment,
+          relationships: {
+            ...this.segment.relationships,
+            comments: comments
+          }
+        }
+        this.setItem({ ...segmentWithComments, id: this.segment.id })
```
salisdemos, line 256: why do You need `this.segment.id` ? Shouldn't it be already set with `this.segment` ?

client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
```diff
@@ -249,13 +265,24 @@ export default {
...
+      }
 
       return this.saveSegmentAction(this.segmentId)
         .then(checkResponse)
         .then(() => {
+          this.restoreComments(comments)
           dplan.notify.confirm(Translator.trans('confirm.saved'))
         })
         .catch(() => {
+          this.restoreComments(comments)
```
salisdemos, line 285: You can also use `.finally()` to avoid the repetition.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

salisdemos (changes requested):
Is this a general problem with the json:api update-behaviour? If so, we should try to find a solution within the vuex-json-api lib (not for this fix though, but in general)

gruenbergerdemos:
> Is this a general problem with the json:api update-behaviour? If so, we should try to find a solution within the vuex-json-api lib (not for this fix though, but in general)

Maybe, I'm not sure. I'm trying to understand the problem. I linked you in slack.

client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
gruenbergerdemos: Seems logical and it works without the additional id, so I removed it in [c88767f](https://github.com/demos-europe/demosplan-core/pull/3468/commits/c88767f0073d4e9650963acedb7281eefc944899)

client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
gruenbergerdemos: Thanks! Done in [8f9c738](https://github.com/demos-europe/demosplan-core/pull/3468/commits/8f9c73861bb42af6f4aab9169d7a3155a5c52316)

client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
gruenbergerdemos: You're right, this makes no sense. Changed in [48cc5c7](https://github.com/demos-europe/demosplan-core/pull/3468/commits/48cc5c75c358cf4e45eb582697207087f25e3f5e)

salisdemos (approved):

PR #3470, opened 2024-07-22
fix(refs DPLAN-11449): Print warning for short segments

salisdemos:
### Ticket
DPLAN-11449

IF One selects a segment wich is less than 10 chars long it is not possible to save it. But to let the user know what happens, there has to be a warning about it.
And it should be prevented before the cardPane closes..

### How to review/test

Goto segementation-Page and try to produce a segment shorter than 10 chars. Then try to modify it to a allowed length and save again...

### PR Checklist


gruenbergerdemos (changes requested):
Just one translation nitpick

gruenbergerdemos:
And does base branch need to be changed to `release`?

gruenbergerdemos (approved):
lgtm

PR #3474, opened 2024-07-22
chore: use MessageBagInterface in DI to show all messages

mussbach:
MessageBagInterface should be used in DI everywhere, as otherwise another instance of the MessageBag is instantiated and the collected messages are never shown in the interface

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Document/DocumentHandler.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Document/DocumentHandler.php
@@ -288,21 +288,21 @@ class DocumentHandler extends CoreHandler
 
             if ($fileInfo->isDir()) {
                 $result[] = [
-                  'isDir'   => true,
-                  'title'   => $fileInfo->getFilename(),
-                  'path'    => $fileInfo->getPathname(),
-                  'entries' => $this->elementImportDirToArray(
-                      $fileInfo->getPathname()
-                  ),
+                    'isDir'   => true,
+                    'title'   => $fileInfo->getFilename(),
+                    'path'    => $fileInfo->getPathname(),
+                    'entries' => $this->elementImportDirToArray(
+                        $fileInfo->getPathname()
+                    ),
                 ];
             } else {
                 // utf8_decode filename, weil Zip Umlaute kaputt macht
                 $filename = utf8_decode($fileInfo->getFilename());
 
                 $result[] = [
-                  'isDir'  => false,
-                  'title'  => $filename,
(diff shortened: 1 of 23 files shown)


sierrazamudiodemos (changes requested):
Double message when creating an stn

- I create a stn im ewm (Direkt stn)
- Double success message is displayed.
- It does not happen in dev

mussbach:
> Double message when creating an stn
> 
>     * I create a stn im ewm (Direkt stn)
> 
>     * Double success message is displayed.
> 
>     * It does not happen in dev

yes, most likely we will have duplicated notifications now. The new messages where introduced to fix this bug, but did not fix the root cause. Not sure how to handle this as wit this PR are quite likely a bunch of messages fixed that where not even found as bugs. Do you think we should find any duplicated messages beforehand @sierrazamudiodemos ?

sierrazamudiodemos:
> > Double message when creating an stn
> > ```
> > * I create a stn im ewm (Direkt stn)
> > 
> > * Double success message is displayed.
> > 
> > * It does not happen in dev
> > ```
> 
> yes, most likely we will have duplicated notifications now. The new messages where introduced to fix this bug, but did not fix the root cause. Not sure how to handle this as wit this PR are quite likely a bunch of messages fixed that where not even found as bugs. Do you think we should find any duplicated messages beforehand @sierrazamudiodemos ?

How can we find newly created messages? or is there a way that maybe you have on top of your mind to find duplicated messages?

sierrazamudiodemos:
> Do you think we should find any duplicated messages beforehand @sierrazamudiodemos 

I tend towards a yer because I want to avoid qa labeling as a blocker or red bug, or introducing a differnent behaviour. On other hand it is good to let people know that we should use MessageInterface

when commenting from line 2920 to line 2927 in demosplan/DemosPlanCoreBundle/Logic/Statement/StatementHandler.php then only one notification is displayed.

I wonder if looking for $this->getMessageBag()->addObject and testing the functionality there will remove the duplicated notifications.

mussbach:
These are more or less exactly my thoughts about this :-) We could also as a supporting measure briefly check the recent git history of all the files that are within this PR and scan the commit messages for occurrences of "message" "notification" or the like

sierrazamudiodemos:
Oki I understand. I will approve it and I think we could merge in main, so we have the following sprints to fix it, what do you think? the idea of the commits history is good

sierrazamudiodemos (approved):

sierrazamudiodemos:
Thanks for the discussion :-)

PR #3477, opened 2024-07-23
fix(refs DPLAN-941): Adjust Filter to find names with Blanks

salisdemos:
### Ticket
DPLAN-941

### How to review/test

As Support goto the Users List.

search for a Name (Firstname AND Lastname) eg. "Lisa Bauer". The search should find all users with "Lisa" and all users with "bauer"


hwiem (approved):
lgtm

PR #3478, opened 2024-07-23
bug (refs DPLAN-1916): show warning

gruenbergerdemos:
### Ticket
DPLAN-1916

- Show warning only if segment has draft list
- Fixed another bug with console errors if segment has no items


salisdemos (approved):
looks legit

PR #3479, opened 2024-07-23
fix(refs DPLAN-12128): Add Place to Segment relationship

salisdemos:
### Ticket
DPLAN-12128

If we don't have the place in the relationship when we want to create a comment, it breaks.

### How to review/test

as FPA goto a single statement (segments-list) and click the Tab "erwiedern". (If You go to the same place trough another way you wont find the error)
try to add a comment to a segment.

### PR Checklist


hwiem (approved):
code lgtm

PR #3480, opened 2024-07-24
bug (refs DPLAN-12137): Change header field translation

gruenbergerdemos:
### Ticket
DPLAN-12137

Change translation for header field from Status to STN-Status to avoid confusion. 

### How to review/test
Verfahren -> Abschnitte -> Spalten auswählen


hwiem (approved):
lgtm, code review only

PR #3483, opened 2024-07-24
feat (DPLAN-11982): images in segments export

breierDemos:
### Ticket
[DPLAN-11982](https://demoseurope.youtrack.cloud/issue/DPLAN-11982/2-2-Als-Bearbeiterin-mochte-ich-Bilder-in-Stellungnahmetexten-und-dem-Export-anzeigen-konnen-um-Einwendungen-korrekt)

Referenced images in segment text are added to docx export. (segment export)
DATA generates links to images in statement texts in the format "Darstellung_Stell_XXX". DATA adds a class to the link tags so we can filter for the right links in a text. This pr adds logic to adjust these new links to match the format "statementExternId_Darstellung_Stell_XXX", and adds the linked images to the docx exports. The adjusted format is only relvant in the exported file. So in the interface the format that comes from DATA will be used.

### How to review/test
code review, For local testing you need "Stellungnahme aufteilen", Images in recommendations should work as before,
execute tests.

### PR Checklist

- [x] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/ImageLinkConverter.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/ImageLinkConverter.php
@@ -12,42 +12,73 @@ declare(strict_types=1);
 
 namespace demosplan\DemosPlanCoreBundle\Logic;
 
+use demosplan\DemosPlanCoreBundle\Entity\Statement\Segment;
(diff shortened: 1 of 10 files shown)


jalalifardemos (approved):
Thanks, lgtm

PR #3488, opened 2024-07-25
feat(refs DPLAN-12088): Delay before trying to set the marker bounds

salisdemos:
### Ticket
DPLAN-12088
ADO [17346](https://www.dev.diplanung.de/DefaultCollection/DiPlan%20Incidents/_workitems/edit/17346)

We have to wait a little moment before we can fit to the bounds. Somehow the map needs its time.

With that fix, we should "zoom" to the procedure pins on load.
https://diplanbau-release.demos-europe.eu/app_dev.php/

### DIFF
--- a/client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
+++ b/client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
@@ -331,10 +331,6 @@ export default {
     fitToBounds () {
       setTimeout(() => {
         if (window.markersLayer.getLayers().length > 0) {
-          /*
-           * {
-           * }
-           */
           const bounds = window.markersLayer.getBounds().pad(0.2)
           window.map.fitBounds(bounds)
 
--- a/client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
+++ b/client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
@@ -328,6 +328,27 @@ export default {
   },
 
   methods: {
+    fitToBounds () {
+      setTimeout(() => {
+        if (window.markersLayer.getLayers().length > 0) {
+          /*
+           * {
+           *   # Here, the
+           *   fitBounds
+           *   #
+           * }
+           */
+          const bounds = window.markersLayer.getBounds().pad(0.2)
+          window.map.fitBounds(bounds)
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
```diff
@@ -328,6 +328,26 @@ export default {
...
 
   methods: {
+    fitToBounds () {
+      setTimeout(() => {
+        if (window.markersLayer.getLayers().length > 0) {
+          /*
+           * {
+           *   # Here, the
+           *   fitBounds
+           *   #
```
mussbach, line 338: what is this? Do you still need it?

client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
```diff
@@ -328,6 +328,26 @@ export default {
   },
 
   methods: {
+    fitToBounds () {
+      setTimeout(() => {
+        if (window.markersLayer.getLayers().length > 0) {
```
mussbach, line 333: When no procedures/markers are found at all this will poll forever, right?

ahmad-demos (approved):
lgtm, only codeReview!

mussbach (changes requested):
basically fine, just two small questions

client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
salisdemos: no. setTimeOut is only called once. https://developer.mozilla.org/en-US/docs/Web/API/setTimeout.
setIntervall is the one wich calls itself gain and again.

client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
salisdemos: its just moved code without touching. Can be deleted. I don't really care.
```suggestion
```

client/js/components/procedure/publicindex/DpSearchProcedureMap.vue
mussbach: indeed, mixed it up with setInterval

PR #3495, opened 2024-07-26
fix(refs dplan-12026): Prevent undefined

salisdemos:
### Ticket
DPLAN-12026

We cant check if the element is not available due to some permissions..

as FP-A -> 
/verfahren/<procedure-id>verwalten/gislayer/neu

--> check if the console throws errors.


gruenbergerdemos (approved):
lgtm

PR #3496, opened 2024-07-29
bug(refs DPLAN-12154): Fix segmentation buttons

gruenbergerdemos:
### Ticket
DPLAN-12154

- Fix segmentation buttons: We need `inline-block` here for the `div` in the addonWrapper and some margin for the icon in the addon
- Fix another bug where we get a console error when recommendationContainer does not exist

~~Move condition for rendering addon component from inside the addon to here, so we don't have an empty div lingering around.
I'm not sure about this approach. We don't have the empty div and therefor no styling issues, but does it make sense to have this condition in core? Or should this logic remain in the addon?~~

### DIFF
--- a/client/js/components/statement/splitStatement/CardPaneCard.vue
+++ b/client/js/components/statement/splitStatement/CardPaneCard.vue
@@ -58,13 +58,12 @@
       :text="Translator.trans('segment.edit')"
       @click="$emit('edit-segment', segment.id)" />
     <addon-wrapper
-      :addon-props="{
-        segmentStatus: segment.status
-      }"
+      v-if="segment.status !== 'confirmed'"
+      class="inline-block mr-1"
       hook-name="split.statement.buttons"
       @segment:confirm="$emit('segment:confirm', segment.id)" />
     <dp-button-icon
-      class="u-ml-0_25"
+      class="ml-1"
       icon="fa-trash"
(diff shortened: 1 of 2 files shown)


salisdemos:
> does it make sense to have this condition in core? Or should this logic remain in the addon?

I think I should remain in the addon. The concept about the addon-wrapper is, that any addon could hook into it. So it will happen that other addons like to use the same wrapper (in different configurations) wich quite likely will follow other conditions.

salisdemos (changes requested):
see comment

client/js/components/statement/splitStatement/CardPaneCard.vue
```diff
@@ -58,13 +58,12 @@
       :text="Translator.trans('segment.edit')"
       @click="$emit('edit-segment', segment.id)" />
     <addon-wrapper
-      :addon-props="{
-        segmentStatus: segment.status
-      }"
+      v-if="segment.status !== 'confirmed'"
```
graupnerdemos, line 61: This would mean addon mounting here is only ever allowed for non-confirmed segments. Is this what we want?

graupnerdemos (changes requested):

graupnerdemos (dismissed):

salisdemos (approved):

PR #3500, opened 2024-07-29
fix(refs DPLAN-12130): Don't make unneccessary drawing updates

salisdemos:
### Ticket
DPLAN-12130

Don't make unneccessary drawing updates. This leads to adding a new poligons while dragging.

To understand what happens, I reorganized the code a bit.

As FP-A
- Goto /verfahren/<procedure-id>/verwalten/globaleGisEinstellungen
- draw a terroitory (Geltungsbereich)
- edit the drawing.


client/js/components/map/map/DpOlMapEditFeature.vue
```diff
@@ -202,13 +183,41 @@ export default {
...
+      if (this.selectedFeatureId.indexOf(id) === -1) {
+        this.selectedFeatureId.push(id)
+        if (hasOwnProp(event, 'element')) {
+          event.element.set('id', id)
+          this.disabled = false
+        }
+      }
+    },
+
+    clearAll () {
```
salisdemos, line 198: this is only moved to order it alphabetically.
I didn't want to reorder everything to not make it impossible to see the changes.

client/js/components/map/admin/MapAdmin.vue
```diff
@@ -141,12 +141,6 @@ export default {
     }
   },
 
-  provide () {
```
hwiem, line 144: Is it safe to remove this?

client/js/components/map/map/DpOlMapDrawFeature.vue
```diff
@@ -176,12 +177,6 @@ export default {
     }
   },
 
-  watch: {
```
hwiem, line 179: Did you make sure that removing this does not break anything in the context it was added (segment Ortsbezug)?

gruenbergerdemos: This did break segment location map. But the watch created more problems than solving, so now the init of the map is done via event ([8db883e](https://github.com/demos-europe/demosplan-core/pull/3500/commits/8db883e1f553873bb6a973d83510df751de865dc)).

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -740,6 +740,7 @@ export default {
       this.$parent.$parent.resetSlidebar()
       this.toggleSlidebarContent({ prop: 'slidebar', val: { isOpen: true, segmentId: this.segment.id, showTab: 'map' } })
       this.$root.$emit('show-slidebar')
+      this.$root.$emit('segmentMap:show')
```
hwiem, line 743: I think we should try to avoid adding `$root.$emit`s to the code base since they are not supported anymore in vue 3. But since we haven't decided on a pattern to replace it yet, I don't consider this a blocker.

hwiem (approved):
code lgtm

PR #3561, opened 2024-08-14
bug (refs AB#19037): use permission to access role list

jalalifardemos:
### Ticket: https://www.dev.diplanung.de/DefaultCollection/BOP-HH/_workitems/edit/19037

Description: use permission to access user list
- Use `AdministratableUserResourceType` instead of `UserResourceType`, because the second one also includes deleted users
- Use store for departments, DpApi Call is not needed
- Set default empty string for department name
- Use default page 1 if there is no page

- [ ] Tests updated/created


breierDemos (approved):
be lgtm

salisdemos (approved):
looks good. Thanks

PR #3563, opened 2024-08-15
feature (ref AB#18943): access to endet participation

jalalifardemos:
### Ticket:https://www.dev.diplanung.de/DefaultCollection/BOP-HH/_workitems/edit/18943

Description: give access to procedure if internal phase is endet participation

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/procedure/admin/AdministrationPlaces.vue
+++ b/client/js/components/procedure/admin/AdministrationPlaces.vue
@@ -56,6 +56,7 @@
           }" />
         <dp-button-row
           :busy="isLoading"
+          data-cy="addNewPlace"
           primary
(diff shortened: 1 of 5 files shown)


breierDemos (changes requested):
I think this is not correct. In the specific project 'evaluating' not longer exists as internal phase key. The correct key is 'internalphase5'. But in the database 'evaluating' is still present. This needs a project migration that changes the key from 'evaluating' to 'internalphase5', only for internal phases.
We would also have to check whether we have hard-coded 'evaluating' somewhere in the code for this project, so that it can end up back in the database. This would make the migration pointless.

PR #3582, opened 2024-08-22
refactor: Use tailwind classes

gruenbergerdemos:
Some refactorings

- Use tailwind classes
- Order attributes


client/js/components/statement/assessmentTable/DpExportModal.vue
```diff
@@ -45,25 +45,25 @@
             role="tabpanel">
             <fieldset
               v-if="options.pdf.anonymize || options.pdf.obscure"
-              class="u-mb-0_5 u-pb-0_5">
+              class="mb-2 pb-2">
```
hwiem, line 48: Here and below: The `mb-2` is not applied, since the fieldset has `margin-bottom: 0 !important`. `u-mb-0_5` worked because it also has the `!important` declaration.

(Before:
spacing_before
After:
spacing_after)

client/js/components/statement/assessmentTable/DpExportModal.vue
```diff
@@ -103,24 +103,25 @@
...
               v-if="options.pdf.exportTypes && exportChoice.pdf.template == 'condensed' && view == 'assessment_table'"
-              class="u-mb-0_5 u-pb-0_5">
+              class="mb-2 pb-2">
               <legend
                 class="sr-only"
                 v-text="Translator.trans('export.data')" />
               <dp-radio
-                class="mb-1"
                 id="pdfExportTypeStatementsOnly"
+                :checked="exportChoice.pdf.exportType === 'statementsOnly'"
```
hwiem, line 112: I notice that I'm not always happy with the attribute order we defined, e.g. for me, `checked` and `@change` belong together. I don't mean this as a change request, I'd just like to know your opinion on it.

hwiem (changes requested):

client/js/components/statement/assessmentTable/DpExportModal.vue
gruenbergerdemos: Thanks, I changed it here [e8844ac](https://github.com/demos-europe/demosplan-core/pull/3582/commits/e8844ac2723cb99ae252471c8e55d791c4fe4c89)

client/js/components/statement/assessmentTable/DpExportModal.vue
gruenbergerdemos: I see your point, but I personally am fine with the alphabetic order. Other props may also belong together with events or other things and I think it could get confusing how to group and order them.

hwiem (approved):
lgtm, thanks for the cleanup!

PR #3592, opened 2024-08-27
F dplan 12254 add custom drupal condition factory

bakridemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12254/Filter-Bearbeiter-Nicht-zugewiesen-funktioniert-nicht

Description:  add DemosPlanDrupalConditionFactory: in some cases however conditions do have values, we do need the OPERATORS 'IS NULL' and 'IS NOT NULL' and since they are only available in the condition without values, some filters are not working anymore. The DemosPlanDrupalConditionFactory extends the PredefinedDrupalConditionFactory and adds two operators (IsNull and IsNotNull) to the operators with value

Not sure with the name. I am open to any suggestion.

### DIFF
--- a/config/services.yml
+++ b/config/services.yml
@@ -1292,13 +1292,13 @@ services:
 
     EDT\JsonApi\Validation\FieldsValidator: ~
 
-    EDT\Querying\ConditionParsers\Drupal\DrupalConditionParser:
-        arguments:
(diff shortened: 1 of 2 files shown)


bakridemos:
this Issue will be handled otherwise in a new pr that's why this pr will be closed

PR #3597, opened 2024-08-27
fix: Ignore cachesum changes on yarn install

salisdemos:
for some dependencies (vuex-json-api) the cachesum somehow is not stable. It be be related to the fact, that we get the package straight from GH.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,3 +1,4 @@
 enableTelemetry: 0
 
 nodeLinker: node-modules
+checksumBehavior: 'ignore'


hwiem (commented):
Why do we link to a commit of vuex-json-api in the `package.json` instead of releasing a new version?

salisdemos:
> Why do we link to a commit of vuex-json-api in the `package.json` instead of releasing a new version?

Good question. Ask the maintainer ;-) 
But even if we release now vuex-json-api. We will run into similar issues sooner or later i think

graupnerdemos (changes requested):
As we discovered, we're using the hash because the dependency branch in question got lost in PR-ization. Maintainer and branch author are wondering how that happened. For now, there should be no need to add this change, if and when we encounter such an issue again, we can re-evaluate making the change. It should however in my opinion always be a temporary solution.

salisdemos:
not the right approach

PR #3598, opened 2024-08-27
Fix(dplan-2307): map toolbar layout issue

sakutademos:
### Ticket
[DPLAN-2307
](https://demoseurope.youtrack.cloud/issue/DPLAN-2307/Karteneben-ein-ausblenden-verdeckt-Einzeichnen-Werkzeuge)

**Description:** As mentioned in the bug ticket, the toolbar overflows the upper box when it is open. This issue appears to be caused by the toolbars absolute positioning. It is placed at the bottom and opens upwards, which makes it difficult to determine if this behavior is a bug or a feature. 
However, the such behavior can be confusing for users. Adjusted the design as suggested by Ines in the ticket:

_Die aufklapbare Box nach unten akflappen lassen, mit Scrollmöglichkeit.
Die rote (Aktion) Box soll mit im Scrollbereich sein und nicht mehr abgedenkt._

image
image
image

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_public_participation_detail.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_public_participation_detail.html.twig
@@ -1,7 +1,7 @@
 {#  map toolbar #}
 {% set mapAndCountyGroupEnabled = templateVars.statementFieldDefinitions|filter(el => el.name == 'mapAndCountyReference')[0].enabled == true|default(true) %}
 {% set isLocationEnabled = mapAndCountyGroupEnabled or procedureStatementPriorityArea %}
-<div class="{{ 'c-map__toolbar u-mb-0_5-palm'|prefixClass }}">
+<div class="{{ 'c-map__toolbar flex flex-col u-mb-0_5-palm'|prefixClass }}">
 
     {% if hasPermission('feature_new_statement') %}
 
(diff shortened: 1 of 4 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_public_participation_detail.html.twig
```diff
@@ -1,369 +1,371 @@
...
-                        </a>
-                    </li>
-                </ul>
-
-                <div class="{{ 'c-actionbox__panelwrapper u-pb-0 u-mb-0_5'|prefixClass }}">
-                    <div v-show="activeActionBoxTab === 'talk'" id="statementAction" class="{{ 'c-actionbox__panel'|prefixClass }}"
-                         :class="activeActionBoxTab === 'talk' ? prefixClass('is-visible-actionbox') : ''"
-                         data-statement-state-started>
+<div class="c-map__toolbar">
+    <div class="{{ 'c-map__toolbar--container u-mb-0_5-palm'|prefixClass }}">
```
spiess-demos, line 5: The `--` separator denotes a [modifier in BEM](https://getbem.com/naming/#modifier). As the newly introduced element does not implement a variation of the toolbar, but introduces a new "block", the appropriate naming would be `c-map__toolbar-container`. Also see our [conventions on selector naming](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/css.html#selector-naming).

On the other hand, this whole thing could possibly be expressed using Tailwind with `sm:overflow-y-auto sm:height-[531px] sm:pr-1`. I know, the `sm:` breakpoint kicks in 1px earlier than the `media-query('lap-up')` include but maybe we can ignore that.

spiess-demos (changes requested):
Generally speaking i find the solution valid. Only implementation concerns :)

templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_public_participation_detail.html.twig
sakutademos: right, thx for the information. I refactored it here

Regarding the Tailwind, I am not really sure why, but it did not work with prefixes like `sm:` there

spiess-demos (approved):
lgtm!

PR #3600, opened 2024-08-28
feat(refs AB#18561): Change trans key for export

gruenbergerdemos:
### Ticket
https://www.dev.diplanung.de/DefaultCollection/BOP-HH/_workitems/edit/18561

Change translation key to be more specific about what happens.


hwiem (approved):
lgtm

PR #3601, opened 2024-08-28
fix(DPLAN-12326): image overlaps text

sakutademos:
### Ticket
[DPLAN-12326](https://demoseurope.youtrack.cloud/issue/DPLAN-12326/Bild-uberdeckt-Mitteilungstext-Portal-Mitteilungen)

Description: This PR fixes an issue with overlapping images. The `c-image--detail` class adds a `max-width: 530px`, which conflicts with the `max-w-full `class from Tailwind. To resolve this, the `c-image--detail `class has been removed from the Twig template.

- [ ] Tests updated/created
- [ ] Update documentation


muellerdemos (approved):
Not sure if we want this but my gut feeling would be to replace the` c-image` with tailwind classes all together too, but since I did not sprint a dplan project for quite a while now I don't know if this is useful or not. however this PR fixes what it wants to fix so -  lgtm

PR #3605, opened 2024-08-29
test(refs #AB18560): add unit test for ExportModal.vue

hwiem:
### Ticket
https://www.dev.diplanung.de/DefaultCollection/EfA%20DiPlanung/_workitems/edit/18560

- the tests mostly cover if the template is rendered correctly depending on given conditions
- additional test cases (e.g. for interactions) may be added later

### How to review/test

- run `yarn test ExportModal.spec.js` in the container
(to see the results of each test case, uncomment `verbose: true` in `jest.config.js`

### PR Checklist

- [x] Tests updated/created

### DIFF
--- a/tests/frontend/unit/ExportModal.spec.js
+++ b/tests/frontend/unit/ExportModal.spec.js
@@ -1,0 +1,434 @@
+import { afterEach, beforeEach, describe, expect, it } from '@jest/globals'
+import ExportModal from '@DpJs/components/statement/assessmentTable/ExportModal.vue'
+import shallowMountWithGlobalMocks from '@DpJs/VueConfigLocal'
+
+describe('ExportModal', () => {
+
+  const props = {
+    options: {
+      docx: {
+        _defaults: {
+          anonymous: false,
+          exportType: 'statementsOnly',
+          sortType: 'default',
+          template: 'condensed'
+        },
+        anonymize: true,
+        buttonLabel: 'export.docx',
+        buttonLabelSingle: 'export.trigger.docx',
+        exportTypes: true,
+        obscure: true,
+        tabLabel: 'export.docx',
+        templates: {
+          condensed: {
+            name: 'export.compact'
+          },
+          landscape: {
+            explanation: 'explanation.export.docx',
+            name: 'export.landscape'
+          },
+          portrait: false
+        }
+      },
+      pdf: {
+        _defaults: {
+          anonymous: false,
+          exportType: 'statementsOnly',
+          sortType: 'false',
+          template: 'condensed'
+        },
+        anonymize: true,
+        buttonLabel: 'export.pdf',
+        buttonLabelSingle: 'export.trigger.pdf',
+        exportTypes: true,
+        newPagePerStn: true,
+        obscure: true,
+        tabLabel: 'export.pdf',
+        templates: {
+          condensed: {
+            name: 'export.compact'
+          },
+          landscape: {
+            name: 'export.landscape'
+          },
+          portrait: {
+            name: 'export.portrait'
+          }
+        }
+
+      },
+      xlsx: {
+        _defaults: {
+          anonymous: false,
+          exportType: 'topicsAndTags',
+          sortType: 'false',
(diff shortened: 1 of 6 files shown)


tests/frontend/unit/ExportModal.spec.js
```diff
@@ -0,0 +1,432 @@
...
+          }
+        }
+      })
+
+      const checkboxStub = wrapper.find('[datacy="exportModal:newPagePerStn"]')
+      expect(checkboxStub.exists()).toBe(false)
+    })
+
+    it('displays radio buttons for selecting a format in the pdf tab if pdf.templates option is true', () => {
+      const radioButtons = wrapper.findAll('[datacy^="exportModal:pdfTemplate"]')
```
muellerdemos, line 209:
```suggestion
      const radioButtons = wrapper.findAll('[datacy="exportModal:pdfTemplate"]')
```
I think this is a typo but since this can be found on other occurrences too, maybe this is a syntax I do not know?

tests/frontend/unit/ExportModal.spec.js
```diff
@@ -0,0 +1,432 @@
...
+      wrapper.vm.handleSubmit()
+
+      const event = wrapper.emitted('submit')
+
+      expect(wrapper.emitted()).toHaveProperty('submit')
+      expect(event).toHaveLength(1)
+    })
+
+  describe('ExportModal: pdf export', () => {
+    it('displays "anonymized" checkbox in the pdf tab if pdf.anonymize or pdf.obscure option is set to true', async () => {
```
muellerdemos, line 148: Reading this kind of makes me expect to see the setting of one of those props within the test rather than having it set as default but I this is rather my personal preference I guess.

tests/frontend/unit/ExportModal.spec.js
```diff
@@ -0,0 +1,432 @@
...
+      expect(checkboxStub.exists()).toBe(true)
+    })
+
+    it('does not display "anonymized" checkbox in the xlsx tab if xlsx.anonymize and xlsx.obscure options are set to false',  () => {
+      const checkboxStub = wrapper.find('[datacy="exportModal:xlsxAnonymous"]')
+
+      expect(checkboxStub.exists()).toBe(false)
+    })
+
+    // fails because 'feature_admin_assessmenttable_export_statement_generic_xlsx' permission is always evaluated as true
```
muellerdemos, line 377: Shouldn't we rather remove this and craft another test which doesn't require the evaluation of `feature_admin_assessmenttable_export_statement_generic_xlsx`?

muellerdemos (commented):
Nice set of tests, just some questions in general otherwise this lgtm

tests/frontend/unit/ExportModal.spec.js
hwiem: It's not a typo, it looks for all elements that have a `datacy` attribute that has a value that starts with the given string

tests/frontend/unit/ExportModal.spec.js
hwiem: Yeah I was wondering the same, but then what should be the default values?

muellerdemos: I assume setting both to false messes up things for other tests so I have no other idea too.

tests/frontend/unit/ExportModal.spec.js
hwiem: I'm not sure how? Right now, all permission checks evaluate to true in our unit tests because of
`const hasPermission = jest.fn(() => true)`
in VueConfigLocal.js. This means there will always be the third radio button (which is rendered if `feature_admin_assessmenttable_export_statement_generic_xlsx` is true), so this test cannot succeed.
Of course we should fix permissions in our unit tests asap, but I suggest doing it in a separate PR.

muellerdemos: Got it, thanks for the explanations :+1:

muellerdemos (approved):
nice work

PR #3620, opened 2024-09-03
fix(refs DPLAN-12267): add consideration as BoilerplateCategory permission based by default

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12267/Beim-CSV-Import-von-Schlagworten-und-Textbausteinen-werden-Textbausteine-nicht-automatisch-im-Bereich-Erwiderung-angezeigt

Description:
add consideration as BoilerplateCategory permission based by default

```
Adjust BoilerplateCategoryRepository ->
refactor method getByTitle to actually work
- it was not used nor working before;
Adjust BoilerplateRepository ->
refactor the add method to handle categories
in a working way. The key categories was
never set before so no old behavior got
compromised.
added private helper method to make adding
categories work;
ProcedureService -> alter logic to add
the category consideration to all boilerplate
if permission is active;
StatementHandler -> fix buggy code to
return a bool all the time and not a
string in some cases before used inside an
if condition!
```

- [x] Tests updated/created
- [ ] Update documentation


bakridemos (approved):
lgtm, tested too

PR #3622, opened 2024-09-04
bug(refs DPLAN-1879): Change aria live to assertive

gruenbergerdemos:
### Ticket
DPLAN-1879

- Screen reader does not always read the result with "polite".
- Order some attributes and add cypress selectors


ahmad-demos (approved):
lgtm, only codeReview!

PR #3623, opened 2024-09-04
bug(refs DPLAN-1953): Set full width to true

gruenbergerdemos:
### Ticket
DPLAN-1953

We don't have an aside block here, so we can show the content over the whole width.

### How to review/test
Login as Privatperson -> Profil icon -> Benachrichtigungen: The table should use the whole width of the page


muellerdemos (approved):
lgtm

PR #3634, opened 2024-09-05
bug(refs DPLAN-12379): Add z index to draw feature

gruenbergerdemos:
### Ticket
DPLAN-12379

Set z index for drawing layer to 1000 to make sure it is always on top. If other layers are on top of it, we can't edit our drawings.

### How to review/test
Verfahren -> Planungsdokumente und Planzeichnung -> Startkartenausschnitt -> Edit Startkartenausschnitt and/or Kartenbegrenzung and then try to edit a drawing.


spiess-demos (approved):
As this is the desired behavior for all instances of the component (drawing should be above all other layers) i think this is an appropriate solution. Lgtm (code review).

PR #3649, opened 2024-09-10
bug(refs DPLAN-12404): Check for location contacts in institutions

gruenbergerdemos:
### Ticket
DPLAN-12404

Add check for location Contacts so we don't run into an error.  

### How to review/test
Verfahren -> Institutionen verwalten -> Institution hinzufügen -> institutions list should load

### DIFF
--- a/client/js/components/procedure/admin/DpAddOrganisationList.vue
+++ b/client/js/components/procedure/admin/DpAddOrganisationList.vue
@@ -181,7 +181,7 @@ export default {
 
     rowItems () {
       return Object.values(this.invitableToebItems).reduce((acc, item) => {
-        const locationContactId = item.relationships.locationContacts.data.length > 0 ? item.relationships.locationContacts.data[0].id : null
+        const locationContactId = item.relationships.locationContacts?.data.length > 0 ? item.relationships.locationContacts.data[0].id : null
         const locationContact = locationContactId ? this.getLocationContactById(locationContactId) : null
         const hasNoEmail = !item.attributes.participationFeedbackEmailAddress
 
--- a/client/js/components/procedure/admin/DpAddOrganisationList.vue
+++ b/client/js/components/procedure/admin/DpAddOrganisationList.vue
@@ -31,22 +31,22 @@
               {{ Translator.trans('address') }}
             </dt>
             <dd
-              v-if="locationContacts.street"
+              v-if="locationContacts?.street"
               class="ml-0">
               {{ locationContacts.street }}
             </dd>
             <dd
-              v-if="locationContacts.postalcode"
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/admin/DpAddOrganisationList.vue
```diff
@@ -31,22 +31,22 @@
               {{ Translator.trans('address') }}
             </dt>
             <dd
-              v-if="locationContacts.street"
+              v-if="locationContacts?.street"
```
hwiem, line 34: If `locationContacts` is not defined, wouldn't that mean that only the text "address" is displayed, but no address data? So maybe we could add a `v-if="locationContacts"` to the `dl` above, which should also make adding all the `?` below unnecessary?

client/js/components/procedure/admin/DpAddOrganisationList.vue
```diff
@@ -179,8 +179,8 @@ export default {
 
     rowItems () {
       return Object.values(this.invitableToebItems).reduce((acc, item) => {
-        const locationContactId = item.relationships.locationContacts.data[0].id
-        const locationContact = this.getLocationContactById(locationContactId)
+        const locationContactId = item.relationships.locationContacts.data.length > 0 ? item.relationships.locationContacts.data[0].id : null
```
hwiem, line 182: Is `locationContacts` always defined?

client/js/components/procedure/admin/DpAddOrganisationList.vue
gruenbergerdemos: No, if there is no adress data, the translation `Translator.trans('notspecified')` is shown. But changed it a bit here to make it more clear: [23f1f9b](https://github.com/demos-europe/demosplan-core/pull/3649/commits/23f1f9bbf351eaf5ada2a7c69d9a9ce530170ab7)

client/js/components/procedure/admin/DpAddOrganisationList.vue
gruenbergerdemos: I added an additional check to be safe: [2f9be6a](https://github.com/demos-europe/demosplan-core/pull/3649/commits/2f9be6a36bae17e98ba62533279f2cea6315f8e3)

client/js/components/procedure/admin/DpAddOrganisationList.vue
gruenbergerdemos: And here: [beba1ae](https://github.com/demos-europe/demosplan-core/pull/3649/commits/beba1ae0399c5261458267769c03810d3646a2a2) forgot the check in the method earlier

PR #3665, opened 2024-09-11
chore (refs DPLAN-12415): adjust Hooks/Selectors for gisLayer links.

ahmad-demos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12415

- adjust Hooks/Selectors for gisLayer links.

### DIFF
--- a/client/js/components/document/DpMapSettingsPreview.vue
+++ b/client/js/components/document/DpMapSettingsPreview.vue
@@ -379,13 +379,6 @@ export default {
   },
 
   methods: {
-    dataCyFormating (str) {
-      return str
-        .replace(/\s(.)/g, (match, group1) => group1.toUpperCase())
-        .replace(/\s+/g, '')
-        .replace(/^(.)/, (match, group1) => group1.toLowerCase())
-    },
-
     href (link) {
       return Routing.generate(link.routeName, {
         procedureId: this.procedureId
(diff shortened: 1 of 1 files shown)


client/js/components/document/DpMapSettingsPreview.vue
```diff
@@ -72,7 +72,7 @@
             v-tooltip="Translator.trans(link.tooltipContent)"
             class="o-link"
             :class="{'color-status-complete-text': link.done()}"
-            :data-cy="`gisLayerLink:${index}`"
+            :data-cy="`gisLayerLink:${dataCyFormating(Translator.trans(link.label))}`"
```
salisdemos, line 75: why not like so?
```suggestion
            :data-cy="`gisLayerLink:${link.label}`"
```
then we don'T have to transform the text.
And using a transated Text as hook is unstable anyhow. that can change and vary over different projects quite often.

salisdemos (changes requested):

client/js/components/document/DpMapSettingsPreview.vue
ahmad-demos: Yes you are right, thank you!

ahmad-demos: https://github.com/demos-europe/demosplan-core/pull/3665/commits/eb968af9414b7abc99de14804ae5d7ecbe723373

salisdemos (approved):

PR #3670, opened 2024-09-11
fix: DPLAN-12009 copy statement with file container

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12009/ROBOB-Prod-Bei-verschobenen-STN-wurden-in-der-Ubersicht-die-Links-von-Anhangen-nicht-angepasst

Description:
- use 'copyStatementObjectWithinProcedureWithRelatedFiles' instead of 'copyStatementObjectWithinProcedure' to copy statements with container files.
- refactor 'persistAndFlushReportEntries': this method use transaction and try fo flush report while copying statements which cause an exception and close the EntityManager. To avoid that we have to seprate the persisting part from the transaction, and this is done by refactoring 'persistAndFlushReportEntries'. The method 'persistReportEntry' can be used now in the case when we don't want to flush report entities directly after persisting (which is always the case with statements copying).

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Report/ReportService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Report/ReportService.php
@@ -49,8 +49,7 @@ class ReportService extends CoreService
         private readonly StatementReportEntryFactory $statementReportEntryFactory,
         private readonly TranslatorInterface $translator,
         private readonly ValidatorInterface $validator,
-        private readonly EntityManagerInterface $entityManager
-
+        private readonly EntityManagerInterface $entityManager,
     ) {
     }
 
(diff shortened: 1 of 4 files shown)


demosplan/DemosPlanCoreBundle/Logic/Report/ReportService.php
```diff
@@ -347,4 +341,15 @@ protected function addAnonymizationReport(string $category, StatementAnonymizeRp
 
         return $this->reportRepository->addObject($report);
     }
+
+    public function persistReportEntry(array $reportEntries, EntityManagerInterface $em): void
```
mussbach, line 345: should be named `persistReportEntries` when an array is passed. Why do you inject the entity manager? Isn't it possible to use DI in the ReportService?

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/Logic/Report/ReportService.php
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/3670/commits/b73a623b3a91c6f4af9793abb24e3245455f28c0

mussbach (approved):
lgtm, thx code reviewn only

PR #3671, opened 2024-09-12
bug(refs DPLAN-12405): Fix merging layers

gruenbergerdemos:
### Ticket
DPLAN-12405

Fix merging map layers. 
- I think this `'undefined' !== typeof activeLayer.id || '' !== hoverLayerId` check is not necessary, because it doesn't hurt to have the a tag always there and only show/not show the icon on hover and click. With the check however it does not work correctly and it is quite hard to link the layers, because the icon "blinks".
- Use correct permission for setting layer active. We only need this if the feature for merging layers is enabled. 
- Add missing field to Api request
- Some refactorings

### How to review/test
Verfahren -> Planungsdokumente und Planzeichnung -> Kartenebenen -> Add some overlay layers (https://sgx.geodatenzentrum.de/wms_basemapde?SERVICE=WMS eg) -> Click on layer -> the visibilty group icons should appear and you should be able to group them


hwiem (commented):
It doesn't work for me, maybe we could take a look together?

hwiem (approved):
After checking again today, it works :tada: 
Code lgtm

PR #3674, opened 2024-09-12
F dplan 11379 add customer to news

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-11379/ADO-Issue-15401-Portal-Mitteilungen-nur-im-Mandanten-sichtbar-in-dem-sie-erstellt-wurden

- Global news are not linked to any specific customer.
- In this PR, each global news item from _platform_content will be duplicated for every customer. This allows each customer to choose whether to keep or delete the news.
- Three tables are involved:
  - _platform_content: Stores the news.
  - _platform_content_categories: Links global news to a category.
  - _manually_list_sort: Stores the order of the news.
- The news order will not be included in the migration. News will be copied to customers without retaining the original order.

### How to review/test
- First go to  diplanbau project and then /news/verwalten
- Check that there are those news
- Go to your DB and check the table _platform_content (there is not customer id)
- Then Run the migration bin <project> dp:migrate
- Then check the table _platform_content, you will see the customer id column and another entries (as the existing entries where duplicated per customer)
- Refresh the page /news/verwalten
- You will see the news have a different order (that is ok) they should be sorted by created_at
- Then Edit - delete the news (all should work)

- [ ] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Entity/GlobalContent.php
```diff
@@ -186,6 +188,13 @@ class GlobalContent extends CoreEntity implements UuidEntityInterface, GlobalCon
      */
     protected $categories;
 
+    /**
+     * @ORM\ManyToOne(targetEntity="demosplan\DemosPlanCoreBundle\Entity\User\Customer")
+     *
+     * @ORM\JoinColumn(name="customer_id", referencedColumnName="_c_id", onDelete="CASCADE", nullable=false)
```
mussbach, line 194: Just to be sure: It is the current wish of the team to use database cascades?

mussbach (approved):
lgtm, code review only. Maybe you could try to add some tests

sierrazamudiodemos:
@mussbach I adjusted the files: demosplan/DemosPlanCoreBundle/Logic/ContentService.php
-  public function getContentList(User $user, $limit = null): array --> to get news only for current customer (when user is not logged. this is the Startseit news)
- public function getContentAdminList($categoryName = null): array  --> to get news only for current customer (when user is loggedin)
-     public function getSingleContent($ident) --> to get one single news. This is to make sure that with the url having the id of the news and switching to another customer, then the news will not be rechable
-     public function addContent($data) --> create new news


I also adjusted the migration to insert _platform_content_roles so that the news are visible per role

breierDemos (approved):
lgtm code review only

PR #3679, opened 2024-09-13
fix(refs DPLAN-12428): cast MaxPerPage Param from string to int

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-12428/Fehlermeldung-beim-Original-STn-List

Description:
The params retrieved from the form are parsed as strings. We need an integer to call the method $paginator->setMaxPerPage().

This change within the PagerFanta library was introduced here: https://github.com/BabDev/Pagerfanta/commit/36ab48a0a25e55186b1f002acf5c0ed655b81dae and we have updated to version >3 since the 06.06.24

Added a simple type cast for uses of PagerFanta::setMaxPerPage

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm code review only

PR #3680, opened 2024-09-16
fix(refs DPLAN-1917): Add map-params to newly created customers via GenerateCustomerCommand

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-1917/Basemap-wird-nicht-per-Default-auf-Startseitenkarte-bei-neuem-Mandanten-eingetragen

Description:

Adds the map-parameters to the GenerateCustomerCommand process.
You cann add the option ```-m default``` to be able to skip the question regarding map values and automatically use the default values. Otherwise you will now be asked for a BaseLayerUrl, BaseLayerLayers and a MapAttribution.

I would like to move these 3 constants into the CustomerInterface after this PR gets approved: 
```
    private const DEFAULT_BASE_LAYER_URL = 'https://sgx.geodatenzentrum.de/wms_basemapde';
    private const DEFAULT_BASE_LAYER_LAYERS = 'de_basemapde_web_raster_farbe';
    private const DEFAULT_MAP_ATTRIBUTION = '© basemap.de / BKG ({currentYear})';
```
In order to test this it will be easier to not have those constants within a demosplan-addon release candidate....

You can run the tests in the terminal via: 
```
SYMFONY_DEPRECATIONS_HELPER=disabled  vendor/bin/phpunit tests/backend/core/Core/Functional/GenerateCustomerCommandTest.php
```

Demosplan-addon update with new constants:
https://github.com/demos-europe/demosplan-addon/pull/124
Core update to use new demosplan-addon version v0.40:
https://github.com/demos-europe/demosplan-core/pull/3705

- [x] Tests updated/created
- [ ] Update documentation


demosplan/DemosPlanCoreBundle/Command/Data/GenerateCustomerCommand.php
```diff
@@ -86,6 +100,13 @@ protected function configure(): void
             InputOption::VALUE_REQUIRED,
             'The subdomain of the customer to be created. If omitted it will be asked interactively.'
         );
+        $this->addOption(
```
mussbach, line 103: i would set the default value as default, so that when nothing is given the default values are set

mussbach (changes requested):
Please include these Changes into GenerateCustomerCommandTest

demosplan/DemosPlanCoreBundle/Command/Data/GenerateCustomerCommand.php
MoritzMandler: I wanted to allow the option to insert empty strings as well.
But you can call the command vie ```-m default``` to use those defaults without any further interaction

mussbach: I see, fair point, but is this possible when the values are validated like this? (had no opportunity to check it myself)

mussbach (approved):
lgtm, code review only, thx a lot

PR #3682, opened 2024-09-16
chore (refs DPLAN-12460): Fix hidden flyout.

ahmad-demos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12460/Bei-Abschnittsliste-mit-wenigen-Eintragen-sind-nach-Klick-auf-drei-Punkte-die-Optionen-abgeschnitten

- Fix hidden flyout.
- If a section list has only a few entries, then don't full display for flyout menu.

### DIFF
--- a/client/js/components/procedure/SegmentsList/SegmentsList.vue
+++ b/client/js/components/procedure/SegmentsList/SegmentsList.vue
@@ -107,7 +107,11 @@
         <dp-data-table
           ref="dataTable"
           class="overflow-x-auto pb-3"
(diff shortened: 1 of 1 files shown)


spiess-demos (changes requested):
while this fixes the symptom when there are very few items, the bug will persist with the following scenario: there are more items (say, 10 items) and the user hovers the action menu on the last item. 

i'd prefer a solution that tackles the root cause, which would involve taking the element outside the normal flow and positioning it with @popperjs/core (which is also used by tippy.js).

salisdemos:
already fixed in another PR

PR #3683, opened 2024-09-17
bug(refs DPLAN-12445): Improve accessibility

gruenbergerdemos:
### Ticket
DPLAN-12445

Make focus with keyboard more clear. 
- Add keyboard focus on editor toolbar items
- Add 50% opacity for box shadow on focused buttons (recommended by ui/uix team)

### How to review/test
Öffentlichkeitsansicht Verfahren -> Reden Sie mit -> focus with keyboard and see if everything has visible outline


spiess-demos (commented):
Why do you use outline here, and change the box-shadow opacity of the scss mixin "keyboard-focus"?

spiess-demos (approved):
lgtm!

hwiem (changes requested):
The flyout button is entirely skipped with the keyboard, isn't it?  <- Edit: I realized there is a separate bug ticket for that and I now saw your comment in the ticket, so nevermind this.

And the Diplanung/Home link focus style looks different from the navigation links' one - if it's not too complicated, I think we should make it consistent.

Apart from that I think it's very nice. :tada: 

I also noticed that the "Markieren" button in the tiptap editor is not aligned properly, should we fix it here also or should we create a new ticket?

gruenbergerdemos:
> And the Diplanung/Home link focus style looks different from the navigation links' one - if it's not too complicated, I think we should make it consistent.

I changed it in the linked project PRs. 

> I also noticed that the "Markieren" button in the tiptap editor is not aligned properly, should we fix it here also or should we create a new ticket?

I think we need to fix that in demosplan-ui, so I think we should make a new ticket.

hwiem (approved):
lgtm!

PR #3688, opened 2024-09-18
bug(refs DPLAN-11964): Get Code from Main to make mapFeature Interaction work again

salisdemos:
Don't update Id all the time to not redraw feature on map

cherry picked 651885eb37d9afadca59472f5e4cced12b606c9c (https://github.com/demos-europe/demosplan-core/pull/3500)

AND

Set z index for drawing layer to 1000 to make sure it is always on top. If other layers are on top of it, we can't edit our drawings.

(cherry picked from commit c611491af866c6a9d580f767d87f704df58177b5) https://github.com/demos-europe/demosplan-core/pull/3634

### Ticket
DPLAN-11964

When drawing in the admin-map, the init extent and max extent should not be draggable, but the created polygons should be editable and deleteable as expected.

### DIFF
--- a/client/js/components/map/admin/MapAdmin.vue
+++ b/client/js/components/map/admin/MapAdmin.vue
@@ -141,6 +141,12 @@ export default {
     }
   },
 
+  provide () {
+    return {
+      olMapState: this.olMapState
+    }
+  },
+
   data () {
     return {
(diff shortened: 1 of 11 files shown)


sakutademos (approved):
code review, lgtm

PR #3690, opened 2024-09-18
f (refs DPLAN-12009) picked fix Copy statement with file container 

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12009/ROBOB-Prod-Bei-verschobenen-STN-wurden-in-der-Ubersicht-die-Links-von-Anhangen-nicht-angepasst

Description:
- use 'copyStatementObjectWithinProcedureWithRelatedFiles' instead of 'copyStatementObjectWithinProcedure' to copy statements with container files.
- refactor 'persistAndFlushReportEntries': this method use transaction and try fo flush report while copying statements which cause an exception and close the EntityManager. To avoid that we have to seprate the persisting part from the transaction, and this is done by refactoring 'persistAndFlushReportEntries'. The method 'persistReportEntry' can be used now in the case when we don't want to flush report entities directly after persisting (which is always the case with statements copying).


MoritzMandler (approved):
lgtm - just picked from main

PR #3700, opened 2024-09-20
Fix(DPLAN-12449): dp-obscure tag not being applied

sakutademos:
**Ticket:** [DPLAN-12449](https://demoseurope.youtrack.cloud/issue/DPLAN-12449/Geschwarzte-Stellungnahme-Text-ist-sichtbar-in-Offentlichkeit-Ansicht)

**Description:** This PR fixes an issue where the `<dp-obscure>` tag was not being applied because the `transformObscureTag` function was not triggered. This function should be executed with every update action, otherwise, the `<dp-obscure>` tag will not be transformed as expected.

However, executing it with every update action causes another bug, [DPLAN-12305 Prevent Cursor from Jumping to End of File](https://github.com/demos-europe/demosplan-ui/pull/995). To resolve this, the transformed text should be processed outside the input event to avoid updating the full text directly.

I think it's because `fullText` is a `v-model` of the editor, and if its value is different from the HTML, it updates the field and causes the cursor to jump. So, we shouldn't update `fullText` directly but store it in a different variable. I haven't found a better solution, I tried different approaches... If you have a better idea, feel free to share :)

### Linked PRs (optional)
- https://github.com/demos-europe/demosplan-ui/pull/1027

- [ ] Tests updated/created
- [x] Update documentation


salisdemos (approved):
Thank You to make me understand what went wrong. Until now, I didn't get why the cursor jumped all the time.

PR #3697, opened 2024-09-19
chore(refs dplan-12315): Replace draggable with sortable in newsAdminList

salisdemos:
### Ticket
DPLAN-12315

### How to review/test

Login as FPA
goto /verfahren/<procedure-id>verwalten/aktuelles

Drag drop articles and see if everything work as expected

Requires: f_wip_vue3_draggables in demosplan-ui

### PR Checklist

Delete the checkbox if it doesn't apply/isn't necessary.

- [ ] Tests updated/created
- [ ] Update documentation
- [ ] Link all relevant tickets
- [ ] Move the tickets on the board accordingly
- [ ] Data-Cy attributes added/updated ([conventions](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/twig_html.html#guideline-for-naming-cypress-hooks))

PR #3701, opened 2024-09-20
chore(refs dplan-12315): Replace draggable with sortable for places

salisdemos:
### Ticket
DPLAN-12315

uses the new way of the dp-draggable, where the return-Value changed.

### How to test:

goto places ( `/verfahren/<procedure-id>/schritte` )
drag-drop and reload page. It should work like before

Builds on top of
https://github.com/demos-europe/demosplan-core/pull/3697


client/js/components/procedure/admin/AdministrationPlaces.vue
```diff
@@ -400,12 +403,12 @@ export default {
         })
     },
 
-    updateSortOrder (placeData) {
+    updateSortOrder ({ id, newIndex }) {
```
muellerdemos, line 406: I really like that we use a parameter object here!

muellerdemos (approved):
nice refactoring, lgtm

PR #3707, opened 2024-09-23
bug(refs DPLAN-12484): Fix target name for ol map draw point

gruenbergerdemos:
### Ticket
DPLAN-12484

The layer name was changed [here](https://github.com/demos-europe/demosplan-core/pull/3688), so we need to also change the target name for DpOlMapDrawPoint. 

### How to review/test
Verfahren -> Grundeinstellungen -> Verortung des Verfahrens -> You should be able to draw a point.


hwiem (approved):
Lgtm, thanks!

PR #3710, opened 2024-09-24
bug(refs DPLAN-12484): Features should be an array, not an object

gruenbergerdemos:
### Ticket
DPLAN-12484

OL expects features as an array instead of an object. 

### How to review/test
This bug only appears with ne procedures. 

New procedure -> Grundeinstellungen -> There should be no errors in console and you should be able to draw a point for Ortsbezug


hwiem (approved):
Lgtm + works!

PR #3714, opened 2024-09-25
fix (refs DPLAN-12489): update Map to accept  as an array instead of …

ahmad-demos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12489

- update Map to accept  as an array instead of an object.

### How to review/test

- Go to blp release Server.
- Login as `Fachplanung-Admin`
- Verfahren -> Grundeinstellungen -> Ensure Map without console errors and allows point drawing for Ortsbezug.


mussbach (approved):
lgtm, code review only, thx

PR #3724, opened 2024-09-26
bug(refs DPLAN-12530): Set initial segments to empty array

gruenbergerdemos:
### Ticket
DPLAN-12530

- Initial segments is expected to be an array, so set the default to empty array instead of null.
- Add error message if no places are defined. Tell user what to do with link to places. 

### How to review/test
Choose a procedure with no defined places (Bearbeitungsschritte)
Verfahren -> Stellungnahmen -> Aufteilen -> Create a segment and check error message and link to places


hwiem (approved):
Nice! Lgtm + works

PR #3738, opened 2024-09-30
bug(refs DPLAN-12552): Distinguish between master and procedure

gruenbergerdemos:
### Ticket
DPLAN-12552

- Use `useGlobalInformationUrl` instead of `featureInfoUrl.global`. It seems like this was renamed. `featureInfoUrl` is not available anymore. 
- Add prop `isMaster` to check if we are in procedure or in master
- Make resource type dynamically depending on procedure or master

### How to review/test
Blaupausen -> Planungsdokumente und Planzeichnung -> Startkartenausschnitte -> Everything should be working as expected

### DIFF
--- a/client/js/components/map/admin/MapAdmin.vue
+++ b/client/js/components/map/admin/MapAdmin.vue
@@ -300,7 +300,7 @@ export default {
   },
 
   async mounted () {
-    const settings = await this.fetchProcedureMapSettings(this.procedureId, this.isMaster)
+    const settings = await this.fetchProcedureMapSettings({ procedureId: this.procedureId, isMaster: this.isMaster })
     this.procedureMapSettings = JSON.parse(JSON.stringify(settings))
   }
 }
--- a/client/js/components/map/admin/MapAdmin.vue
+++ b/client/js/components/map/admin/MapAdmin.vue
@@ -50,7 +50,7 @@
         @suitableScalesChange="value => areScalesSuitable = value" />
 
       <dp-input
-        v-if="!procedureMapSettings.attributes.featureInfoUrl.global && hasPermission('feature_map_feature_info')"
+        v-if="!procedureMapSettings.attributes.useGlobalInformationUrl && hasPermission('feature_map_feature_info')"
         v-model="procedureMapSettings.attributes.informationUrl"
         id="informationURL"
         class="u-mb"
@@ -135,6 +135,12 @@ export default {
   },
 
   props: {
+    isMaster: {
+      type: Boolean,
+      required: false,
+      default: false
+    },
+
     procedureId: {
       type: String,
       required: true
@@ -155,13 +161,13 @@ export default {
       },
       procedureMapSettings: {
         id: '',
-        type: 'ProcecdureMapSetting',
+        type: 'ProcedureMapSetting',
         attributes: {
           coordinate: [],
           copyright: '',
           defaultMapExtent: [],
           defaultBoundingBox: [],
-          featureInfoUrl: { global: false },
+          useGlobalInformationUrl: false,
           informationUrl: '',
           showOnlyOverlayCategory: false,
           mapExtent: [],
@@ -294,7 +300,7 @@ export default {
   },
 
   async mounted () {
-    const settings = await this.fetchProcedureMapSettings(this.procedureId)
+    const settings = await this.fetchProcedureMapSettings(this.procedureId, this.isMaster)
     this.procedureMapSettings = JSON.parse(JSON.stringify(settings))
   }
 }
--- a/client/js/store/map/ProcedureMapSettings.js
+++ b/client/js/store/map/ProcedureMapSettings.js
@@ -39,10 +39,10 @@ export default {
         .then(response => checkResponse(response))
     },
 
-    fetchProcedureMapSettings ({ commit }, args = { procedureId, isMaster: false }) {
+    fetchProcedureMapSettings ({ commit }, { procedureId, isMaster = false }) {
       try {
-        const resourceType = args.isMaster ? 'ProcedureTemplate' : 'Procedure'
-        const url = Routing.generate('api_resource_get', { resourceId: args.procedureId, resourceType: resourceType })
+        const resourceType = isMaster ? 'ProcedureTemplate' : 'Procedure'
+        const url = Routing.generate('api_resource_get', { resourceId: procedureId, resourceType: resourceType })
         const procedureMapSettingFields = ['availableScales',
           'boundingBox',
           'defaultBoundingBox',
--- a/client/js/store/map/ProcedureMapSettings.js
+++ b/client/js/store/map/ProcedureMapSettings.js
@@ -41,7 +41,7 @@ export default {
 
     fetchProcedureMapSettings ({ commit }, procedureId) {
       try {
-        const url = Routing.generate('api_resource_get', { resourceId: procedureId, resourceType: 'Procedure' })
+        const url = Routing.generate('api_resource_get', { resourceId: procedureId, resourceType: 'ProcedureTemplateResourceType' })
         const procedureMapSettingFields = ['availableScales',
           'boundingBox',
           'defaultBoundingBox',
@@ -75,7 +75,7 @@ export default {
 
         const params = {
           fields: {
-            Procedure: [
+            ProcedureTemplate: [
               'mapSetting'
             ].join(),
             ProcedureMapSetting: procedureMapSettingFields.join()
--- a/client/js/store/map/ProcedureMapSettings.js
+++ b/client/js/store/map/ProcedureMapSettings.js
@@ -39,9 +39,10 @@ export default {
         .then(response => checkResponse(response))
     },
 
-    fetchProcedureMapSettings ({ commit }, procedureId) {
+    fetchProcedureMapSettings ({ commit }, procedureId, isMaster) {
       try {
-        const url = Routing.generate('api_resource_get', { resourceId: procedureId, resourceType: 'ProcedureTemplate' })
+        const resourceType = isMaster ? 'ProcedureTemplate' : 'Procedure'
+        const url = Routing.generate('api_resource_get', { resourceId: procedureId, resourceType: resourceType })
         const procedureMapSettingFields = ['availableScales',
           'boundingBox',
           'defaultBoundingBox',
@@ -58,7 +59,7 @@ export default {
 
         if (hasPermission('feature_map_feature_info')) {
           procedureMapSettingFields.push('informationUrl')
-          procedureMapSettingFields.push('featureInfoUrl')
+          procedureMapSettingFields.push('useGlobalInformationUrl')
         }
 
         if (hasPermission('feature_map_attribution')) {
@@ -75,7 +76,7 @@ export default {
 
         const params = {
           fields: {
-            ProcedureTemplate: [
+            Procedure: [
               'mapSetting'
             ].join(),
             ProcedureMapSetting: procedureMapSettingFields.join()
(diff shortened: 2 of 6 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_admin.html.twig
```diff
@@ -6,7 +6,9 @@
         heading: 'drawing.mapsection.define'|trans
     } %}
 
-    <map-admin procedure-id="{{ procedure }}">
+    <map-admin
+        :is-master="{{ master | json_encode() }}"
```
salisdemos, line 10: nitpic
```suggestion
        :is-master="{{ master|json_encode }}"
```

client/js/store/map/ProcedureMapSettings.js
```diff
@@ -39,9 +39,10 @@ export default {
         .then(response => checkResponse(response))
     },
 
-    fetchProcedureMapSettings ({ commit }, procedureId) {
+    fetchProcedureMapSettings ({ commit }, args= { procedureId, isMaster }) {
```
salisdemos, line 42: this looks un-javascript
```suggestion
    fetchProcedureMapSettings ({ commit }, { procedureId, isMaster }) {
```

client/js/store/map/ProcedureMapSettings.js
```diff
@@ -39,9 +39,10 @@ export default {
...
       try {
-        const url = Routing.generate('api_resource_get', { resourceId: procedureId, resourceType: 'Procedure' })
+        const resourceType = args.isMaster ? 'ProcedureTemplate' : 'Procedure'
+        const url = Routing.generate('api_resource_get', { resourceId: args.procedureId, resourceType: resourceType })
```
salisdemos, line 45:
```suggestion
        const resourceType = isMaster ? 'ProcedureTemplate' : 'Procedure'
        const url = Routing.generate('api_resource_get', { resourceId: procedureId, resourceType: resourceType })
```

client/js/store/map/ProcedureMapSettings.js
sakutademos, line 42: s there be an `args` argument or can we skip it and just use `{ procedureId, isMaster }`? It would be shorter

client/js/store/map/ProcedureMapSettings.js
```diff
@@ -75,7 +76,7 @@ export default {
 
         const params = {
           fields: {
-            Procedure: [
+            [resourceType]: [
```
sakutademos, line 79: should it be an array?

client/js/store/map/ProcedureMapSettings.js
hwiem, line 42: Should we give `isMaster` a default value of `false`? I know that it will also work if it is `undefined`, but it seems cleaner to define a default value.

demosplan/DemosPlanCoreBundle/Controller/Map/DemosPlanMapController.php
```diff
@@ -74,9 +75,11 @@ public function mapAdminAction(
             ]
         );
 
+        $procedure = $procedureService->getProcedure($procedureId);
+
         return $this->renderTemplate(
             '@DemosPlanCore/DemosPlanMap/map_admin.html.twig',
-            ['procedure' => $procedureId, 'title' => 'drawing.admin.adjustments.gis']
+            ['procedure' => $procedureId, 'master' => $procedure?->getMaster(), 'title' => 'drawing.admin.adjustments.gis']
```
salisdemos, line 82: Am I right, that 'master' is a bool? If so, can we rename this to 'isMaster' ?
And would this be a good opportunity the rename 'procedure' to 'procedureId' . That would be great.

client/js/store/map/ProcedureMapSettings.js
gruenbergerdemos: It is not an array but a computed property name in an object. This syntax allows us to use the value of the resourceType variable as a key in the fields object.

demosplan/DemosPlanCoreBundle/Controller/Map/DemosPlanMapController.php
breierDemos: adjusted in [ee2df92](https://github.com/demos-europe/demosplan-core/pull/3738/commits/ee2df92119435d2c1c577d758d4f1a13e4b93cd8)

templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_admin.html.twig
hwiem: Just a question because I don't know it: when do we use `|json_encode()` and when do we use `|json_encode|e('js')` (or is it the same?)? Is it needed here? I'd have used `Boolean({{ master }})` :thinking:

demosplan/DemosPlanCoreBundle/ResourceTypes/ProcedureTemplateResourceType.php
```diff
@@ -25,21 +27,31 @@
...
     {
-        return [
-            $this->createIdentifier()->readable()->sortable()->filterable(),
-            $this->createAttribute($this->agencyMainEmailAddress)->readable(true)->sortable()->filterable(),
-            $this->createAttribute($this->description)->readable()->aliasedPath($this->desc),
-            $this->createToManyRelationship($this->agencyExtraEmailAddresses)->readable()->filterable(),
-            $this->createToOneRelationship($this->owningOrganisation)->readable()->aliasedPath($this->orga)->sortable()->filterable(),
-        ];
+        if ($this->currentUser->hasAnyPermissions('area_public_participation', 'area_admin_map')) {
+            $properties[] = $this->createAttribute($this->coordinate)->readable()->aliasedPath(Paths::procedure()->settings->coordinate);
```
sierrazamudiodemos, line 44: I think rather than modifying the resource type file to include the coordinate, we should consume it from the existing resourcetype  where it's already referenced. What I mean is:

- The coordinate is an attribute belongs to the ProcedureMapSetting. That ResourceType is available when calling the line below: 
 ```
$properties[] = $this->createToOneRelationship($this->mapSetting)->aliasedPath(Paths::procedure()->settings)->readable();
        }
```
- I think the FE could potentially call the relationship, and get the coordinate from there. 
- then we can add the permission area_procedure_adjustments_general_location the permissions file in the needed project as `\demosplan\DemosPlanCoreBundle\ResourceConfigBuilder\ProcedureMapSettingResourceConfigBuilder::@$coordinate` requires that permission.

What do you think about this approach?

**Review by sierrazamudiodemos — CHANGES_REQUESTED** [review]

File: demosplan/DemosPlanCoreBundle/Controller/Map/DemosPlanMapController.php (hunk 5 above)
Comment (gruenbergerdemos) [reply demosplan/DemosPlanCoreBundle/Controller/Map/DemosPlanMapController.php]: Done here [25c7ba5](https://github.com/demos-europe/demosplan-core/pull/3738/commits/25c7ba5058b51c9b31a9896923cdd97754e6a7c8)

File: client/js/store/map/ProcedureMapSettings.js (hunk 2 above)
Comment (gruenbergerdemos) [reply client/js/store/map/ProcedureMapSettings.js:42]: I could not make it work otherwise.

Comment (gruenbergerdemos) [reply client/js/store/map/ProcedureMapSettings.js:42]: This does not work. `isMaster` is always undefined this way.

File: client/js/store/map/ProcedureMapSettings.js (hunk 3 above)
Comment (gruenbergerdemos) [reply client/js/store/map/ProcedureMapSettings.js:45]: See above

File: client/js/store/map/ProcedureMapSettings.js (hunk 2 above)
Comment (salisdemos) [reply client/js/store/map/ProcedureMapSettings.js:42]: Thats weired. But procedureId is set proppery?

Comment (salisdemos) [reply client/js/store/map/ProcedureMapSettings.js:42]: maybe then: 
```suggestion
    fetchProcedureMapSettings ({ commit }, args= { procedureId, isMaster = false }) {
```
?

I really don't understand, why it doesn't work as expected :-/

sakutademos (approved):
it works,lgtm, but some nitpicks

client/js/components/map/admin/MapAdmin.vue
```diff
@@ -294,7 +300,7 @@ export default {
   },
 
   async mounted () {
-    const settings = await this.fetchProcedureMapSettings(this.procedureId)
+    const settings = await this.fetchProcedureMapSettings({ procedureId: this.procedureId, isMaster: this.isMaster })
```
salisdemos, line 303: the method is also used here:
client/js/components/procedure/StatementSegmentsList/StatementSegmentsList.vue
If we change the structure to an Object, we have to change that there too.

salisdemos (changes requested):

templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_admin.html.twig
gruenbergerdemos: I'm not sure either. Twig syntax is always confusing for me. Actually I did not know `Boolean({{ master }})`, but I think it is better, thanks! Done here [d35b132](https://github.com/demos-europe/demosplan-core/pull/3738/commits/d35b132a4d076af6c2e082dc499cd07de68c3c39)

demosplan/DemosPlanCoreBundle/ResourceTypes/ProcedureTemplateResourceType.php
breierDemos: The frontend actually already calls it via ProcedureMapSetting. There was a misunderstanding. I've removed the line in [6565401](https://github.com/demos-europe/demosplan-core/pull/3738/commits/6565401073294817e9ab301f16745213b3d851ec)

sierrazamudiodemos: ah, is the permission active for the coordinate in the needed project?

client/js/store/map/ProcedureMapSettings.js
gruenbergerdemos: Probably better, I added it here: [58ef831](https://github.com/demos-europe/demosplan-core/pull/3738/commits/58ef83174e231523c97a87673294f5961a2d34c8)

demosplan/DemosPlanCoreBundle/ResourceTypes/ProcedureTemplateResourceType.php
breierDemos: No, I think it should be turned off for this project.

sierrazamudiodemos:
BE wise reviewed and approved

client/js/components/map/admin/MapAdmin.vue
gruenbergerdemos: Thanks for catching, done here [6e61bac](https://github.com/demos-europe/demosplan-core/pull/3738/commits/6e61bac7c3e0defe817f0d323c71bc4843506680)

client/js/store/map/ProcedureMapSettings.js
hwiem: Does it work like that? I think it has to be `isMaster = false`

gruenbergerdemos: > Thats weired. But procedureId is set proppery?

Yes.

> I really don't understand, why it doesn't work as expected :-/

I'm not sure. I think it has something to do with how we call it? I tried several ways, this was the only way I could make it work.

gruenbergerdemos: Added space here [621a77c](https://github.com/demos-europe/demosplan-core/pull/3738/commits/621a77c837fdb7c5ad735e6ab1e590658b1c1b39)

gruenbergerdemos: `isMaster = false` throws an error with `args`. But somehow now it works as expected without the args. Done here [ecf51f2](https://github.com/demos-europe/demosplan-core/pull/3738/commits/ecf51f259e7f05b7e00021b5d6accb312866f014)

gruenbergerdemos: I have no idea what I did wrong before, but now it seems to work like expected without the `args`. Done here [ecf51f2](https://github.com/demos-europe/demosplan-core/pull/3738/commits/ecf51f259e7f05b7e00021b5d6accb312866f014) Sorry for the confusion.

salisdemos (approved):

salisdemos:
Thanks a lot

PR #3742, opened 2024-09-30
fix(refs DPLAN-12496): reduce margin

hwiem:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12496/Filterung-bricht-um-wenn-alle-Filter-gesetzt-sind

- with reduced margin, all filter flyout triggers still fit in one line if a filter is selected in each of them
- also, the filter label was vertically aligned to align properly with the filter flyout triggers

### How to review/test

- go to segments list
- select a filter in each filter flyout
- make sure all filter flyout triggers are still displayed next to each other:
filter_flyout_triggers

### PR Checklist


gruenbergerdemos (approved):
lgtm and works!

PR #3751, opened 2024-10-02
fix(refs DPLAN-2153): remove file filter from assessment table filters in planfestsh

muellerdemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-2153/Abwagungstabellen-Filter-In-Planfest-ist-Datei-Bezug-nicht-moglich-entweder-Bezugs-Option-rein-oder-Filter-raus

This PR removes the file filter from the assessment table filters for the planfestsh project since it is not possible to create a file relation to a statement hence this filter is simply redundant.

### How to review/test
- checkout
- login as fpa user
- navigate to assessment table and select filters
- the file filter for statements should not be rendered anymore (only for the planfestsh project)

grafik

### PR Checklist


client/js/components/statement/assessmentTable/FilterModalSelectItem.vue
```diff
@@ -9,7 +9,7 @@
 
 <template>
   <div
-    v-if="false === hidden"
+    v-if="false === hidden && (dplan.projectName === 'Planfeststellungsverfahren' && filterItem.attributes.label !== 'Datei')"
```
muellerdemos, line 12: I am honestly not sure if this ain't too much of buisness logic within this component. However I have no better idea how this could be solved instead, since this should only be valid for the planfestsh project. Having an template override instead appears to be kind of an overkill to me but maybe this is rather the way to go?

hwiem, line 12: Afaik we want to avoid using the `projectName` for condition checks like this (`if projectName === 'someString'`) because we don't want to have the project names in the code. I think we should create a permission for this and handle it in the backend

hwiem (changes requested):

client/js/components/statement/assessmentTable/FilterModalSelectItem.vue
muellerdemos: This is what I was afraid of. Thanks, I will get in touch with the php team for that and adjust this afterwards with this PR.

muellerdemos: Now using the permission check and a computed to determine if the file filter should be rendered here: https://github.com/demos-europe/demosplan-core/pull/3751/commits/6e4a4891d083e58cdb37865940f474a0175d75c7
(the permission is only disabled for the planfestsh project)

And removed a duplicate of the computed here https://github.com/demos-europe/demosplan-core/pull/3751/commits/f3119bf98e268f190600e906b6774e699a9f4284 (which i forgot to delete after fixing the sort order :/ )

salisdemos (commented):
Thanks for muddeling trough this. You one genral thought: please don't use project spefic namings here in the core project. Please keep wordings generalized.

client/js/components/statement/assessmentTable/FilterModalSelectItem.vue
hwiem: I think that we don't really need to do anything in the FE; if I understand correctly, the permission can be checked in `getAvailableFiltersAssessmentTable` in the BE. Or am I missing something?

config/permissions.yml
```diff
@@ -2020,6 +2020,11 @@ feature_statement_delete:
     expose: true
     label: 'Delete (non-draft, non-original) statements.'
     loginRequired: true
+feature_statement_file_filter_set:
+    description: 'Set file to filter in statement list'
+    expose: true
+    label: 'Allow User to can set File as a filter'
```
sakutademos, line 2026:
```suggestion
    label: 'Allow the user to set a file as a filter'
```

client/js/components/statement/assessmentTable/FilterModalSelectItem.vue
bakridemos: i agree with @hwiem , i think this can be only in BE handled

config/permissions.yml
muellerdemos: Thx, done here: https://github.com/demos-europe/demosplan-core/pull/3751/commits/21c94a4306395bc061ed1d72f2e7e7235a2fd1c2

client/js/components/statement/assessmentTable/FilterModalSelectItem.vue
muellerdemos: Thanks, I did not know of the existence of the `getAvailableFiltersAssessmentTable` function but it is the cleaner solution indeed, removed the frontend checks here: https://github.com/demos-europe/demosplan-core/pull/3751/commits/0ec36c5685baa61b21f9b17c7b652679f74227bf and @jalalifardemos will handled the BE part of this PR.

hwiem (approved):
Assuming that the permission is disabled for the project in a separate PR, this lgtm (code review only)

PR #3762, opened 2024-10-04
chore: sync version of ol used in the masterportal

muellerdemos:
### Description:
Like mentioned here: https://github.com/demos-europe/demosplan-core/pull/3172#issuecomment-2233597383 the version of ol should be in sync with the one used by the masterportal. The version of masterportal we are currently using is 2.40.0 which uses ol 9.2.4 hence to prevent updates of this dependency which could be potentially incompatible, this version of ol should be set fixed via the dependabot configuration file.

The downside is we need to track masterportal updates and adjust the dependabot.yml accordingly and I am not sure where and if we should document this (further than in a comment)

Masterportal dependencies can be investigated here: https://bitbucket.org/geowerkstatt-hamburg/masterportalapi/src/master/package.json

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,6 +39,12 @@ updates:
       - dependency-name: chalk
         versions:
           - ">= 5"
+
+      # We need to sync the ol version with the one used in the masterportal dependency.
+      # Currently, we are useing masterportal v2.40.0 which depends on ol v9.2.4
+      - dependency-name: ol
+        versions:
+          - "> 9.2.4"
     groups:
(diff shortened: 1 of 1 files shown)


spiess-demos:
After reading through #3776, maybe we should lock openLayers to 7.5.1 instead of 9.2.4?

muellerdemos:
My proposal would be rather to work with `do-not-merge` labels for this but I thought initially the same with managing this via the dependabot config, but this would also have its downsides like it would be tougher to keep track of updates for masterportalapi

PR #3788, opened 2024-10-07
bug(refs DPLAN-12555): Add styling for administration email

gruenbergerdemos:
### Ticket
DPLAN-12555

We get the email text from BE and need a parent class to style the lists correctly. 

### How to review/test
Verfahren -> Email an alle Einreichenden -> Add an email text with ordered and unordered lists -> Go to confirmation -> Lists should be styled properly.


templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_send_email_confirm.html.twig
```diff
@@ -31,7 +31,7 @@
             {{ 'email.body'|trans }}
         </label>
 
-        <div>{{ templateVars.preparationMail.mailBody|wysiwyg }}</div>
+        <div class="email-body">{{ templateVars.preparationMail.mailBody|wysiwyg }}</div>
```
hwiem, line 34: I think this could also be solved by applying `c-styled-html`, which I believe we use for cases like this (when we get html from the BE)

hwiem (changes requested):

hwiem (approved):
lgtm, thanks!

PR #3789, opened 2024-10-07
F dplan 12579 adjust text width in pdf export

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12579/Pdf-Formatierung-Querformat-Stellungnahme-Text-nimmt-nicht-den-gesamten-Whitespace

Description: 
use list width as reference to adjust text width :  the list width is already adjusted for every export to take up all the whitespaces, the text should actually have the same width and take up all the whitespaces too.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/AssessmentTableExporter/Enum/ListLineWidth.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/AssessmentTableExporter/Enum/ListLineWidth.php
@@ -15,7 +15,7 @@ namespace demosplan\DemosPlanCoreBundle\Logic\Statement\AssessmentTableExporter\
 enum ListLineWidth: int
 {
     case VERTICAL_SPLIT_VIEW = 7;
-    case VERTICAL_NOT_SPLIT_VIEW = 17;
+    case VERTICAL_NOT_SPLIT_VIEW = 18;
     case HORIZONTAL_SPLIT_VIEW = 12;
-    case HORIZONTAL_NOT_SPLIT_VIEW = 24;
+    case HORIZONTAL_NOT_SPLIT_VIEW = 27;
 }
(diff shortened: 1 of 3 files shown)


demosplan/DemosPlanCoreBundle/Logic/Statement/AssessmentTableExporter/Enum/ListLineWidth.php
```diff
@@ -15,7 +15,7 @@
 enum ListLineWidth: int
 {
     case VERTICAL_SPLIT_VIEW = 7;
-    case VERTICAL_NOT_SPLIT_VIEW = 17;
+    case VERTICAL_NOT_SPLIT_VIEW = 18;
     case HORIZONTAL_SPLIT_VIEW = 12;
-    case HORIZONTAL_NOT_SPLIT_VIEW = 24;
+    case HORIZONTAL_NOT_SPLIT_VIEW = 27;
```
MoritzMandler, line 20: These enum entries need to keep their values.
This is because STN texts could contain lists - and lists need extra space for their intendation
[Export-81.pdf](https://github.com/user-attachments/files/17279905/Export-81.pdf)

MoritzMandler (changes requested):
Works fine - just the enum values need to stay put.

demosplan/DemosPlanCoreBundle/Logic/Statement/AssessmentTableExporter/Enum/ListLineWidth.php
bakridemos: weird, i already tested it with a list and i didn,t get that, i tried it again with all export and it seems  in order, here are the exports :
[Export-251.pdf](https://github.com/user-attachments/files/17282601/Export-251.pdf)
[Export-252.pdf](https://github.com/user-attachments/files/17282625/Export-252.pdf)
[Export-253.pdf](https://github.com/user-attachments/files/17282627/Export-253.pdf)

bakridemos: I see now what u mean, i had to export all statements to see that. I thought may be i can use the list width to adjust the text width but i have to do it separately i think

bakridemos: i fixed it here https://github.com/demos-europe/demosplan-core/pull/3789/files

MoritzMandler (approved):
This works fine - thx :)

PR #3790, opened 2024-10-07
fix(DPLAN-2825): layout issue: orga street number is positioned too far to the right

sakutademos:
### Ticket
[DPLAN-2825](https://demoseurope.youtrack.cloud/issue/DPLAN-2825/Hausnummer-ist-weit-ins-Abseits-geraten)

**Description:** This PR fixes an issue with the layout of the address in the orga data entry: remove `flex-grow `from the street name input to resolve the issue, but some adjustments need to be done: to ensure the input displays its value without being cut off, the size of the input should be set to match the street name length. Otherwise, if flex-grow is not set, the input may be cropped.

Since the other project where inputs have borders, it makes sense to keep the design for this project unchanged.

### How to review/test
Login -> Usermenu -> auf Organamen klicken

affects: other projects as well


sakutademos:
also, a question: does it make sense to create a Vue component for it instead of Twig?

salisdemos:
> also, a question: does it make sense to create a Vue component for it instead of Twig?

absolute

salisdemos:
please be aware of not using project names within core PRs :pray: . Please use generalized wordings instead

salisdemos (changes requested):
Looks like something is a bit messed up. I see quite a bunch of changes, that look like they don't belong here. (like package.json Dockerfile, ...)

PR #3791, opened 2024-10-07
bug(refs DPLAN-12600): Remove link to old details page

gruenbergerdemos:
### Ticket
DPLAN-12600

The link was only added to compare old and new details page and should be removed now.

### How to review/test
Verfahren -> Stellungnahmen -> Click on three dots -> There should be only a link to "Details und Erwiderungen" and no link to "Detailansicht"


hwiem (approved):
lgtm

PR #3793, opened 2024-10-08
bug(refs DPLAN-1967): Add min height

gruenbergerdemos:
### Ticket
DPLAN-1967

The overflow-x: auto on the container sets overflow-y to hidden. To solve this, min height is added to the container so flyout has always enough space.

### How to review/test
You need a procedure with only one small segment. 
Verfahren -> Abschnitte -> Check fly out


sakutademos (approved):
code review, lgtm

PR #3794, opened 2024-10-08
chore: Bump ol@9 and masterportal@2.40.0

salisdemos:
To Bump Masterportalapi to the latest version, we had to bump OL and olcs too.
And for some reason resolving internal imports from olcs are not working out of the box. To fix thatwe hd to add some pathResolving on our side. If somebody has any Idea how to resolve this in a better/ more general way: please tell me.


muellerdemos (approved):
thanks a lot for taming that beast - build runs perfectly fine with this :heart:

PR #3798, opened 2024-10-08
bug(refs DPLAN-12608): Remove permission

gruenbergerdemos:
### Ticket
DPLAN-12608

Access to the twig is already restricted with this permission: feature_statement_data_input_orga. Everyone who has this permission is also allowed to view the source statement.


breierDemos (approved):
lgtm code review only

hwiem:
Just curious: Why did we not enable the `feature_read_source_statement_via_api` permission for the role? Doesn't it make the application somewhat unpredictable if in some places, the original statement is wrapped in this permission, but in other places it's not? (But I also do not understand entirely if the permission is used for the original statement in general or just for api usage (as the name suggests))

gruenbergerdemos:
`feature_read_source_statement_via_api` is used in various places and I was not sure if the role `PROCEDURE_DATA_INPUT` should have this permission. I thought it to be a little safer this way, but overall not too sure about it.

PR #3801, opened 2024-10-08
fix(refs DPLAN-12614): isNotEmptyFeatureCollection function returns false always

muellerdemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12614/Nach-dem-Geltungsbereich-angezeichnet-ist-bleibt-der-Titel-mit-Lage-einzeichnen

This function appears to be outdated since `this.territory.features` is not an array but an object so the `JSON.parse(string).features.length > 0` comparison could never work properly because it always returned false. To solve this we need to iterate over the keys of the features and check if the length is bigger than 0.

Also if the `isNotEmptyFeatureCollection` encounters errors, I think the correct way to handle this is to return false instead so the user gets the information that no territory has been set up so far. Also I think we should return the error in a `console.error()`

### How to review/test
- checkout branch
- login as FPA (e.g.)
- select any procedure
- navigate to "Antragsunterlagen" via the side menu
- add a territory
- save
- make sure if there is a territory set that the link appears in green color otherwise it should indicate there is no territory by rendering a default text color (blueish)

### PR Checklist


gruenbergerdemos (changes requested):
As far as I know features should be an array. OL can't process it correctly otherwise. Maybe BE could have a look why this is given to twig as an object?
For comparison you can check the territory at "Lage einzeichnen" (MapAdmin.vue). Features should be an array there. (Don't know if there is already a bug for that, but territory also is not shown on the map at "Antragsunterlagen und Planzeichnung")

salisdemos (commented):
There have been some changes around territory lately to fix several bugs. As far as I remember where a lot of them related to obj/array issues. 
For some reason territory get stored inconsistantly sometimes as array and sometimes as object. I wasn't able to find out when and I thought by fixing all the little things the problem is solved. But now it seems, that its not.
So I would please you to try to understand where its set as Obj and where as Array and what it should be. If its now a bit tight on time, it would be great to try to solve this after codefreeze. Maybe we can get a separate investigation-Ticket for that :thinking:

muellerdemos:
@gruenbergerdemos @salisdemos thanks for the infos, I was unaware the territory drawing did not render. Nevertheless there are some major underlying issues. It is true, that the territory is in an invalid format which will be passed from the BE to FE via procedureSettings but adding a valid format did not resolve the issue.

I have tried to investigate this and for some reason the territory will not be rendered via the functional DpOlMapLayerVector component. Even after creating a valid geojson via https://geojson.io/#map=10.49/54.255/9.9523 the component just didn't want to render and debugging was also troublesome. 

However I have an idea how we could tackle this issue. My plan would be to create some helper functions to create vector layers instead of relying on a functional component (for vue3 this doesn't make much sense anyways) and render those layers directly in the `DpOlMap` component and pass the features which shall be rendered as optional props. This would require some time since this would mean a bigger refactoring but I think we should rather do this sooner than later and within this sprint appears to be a good moment. 

I already tested this approach lightly and it looks really promising and would remove potentially a lot of lines of code, which would also be a great benefit.

PR #3827, opened 2024-10-14
f DPLAN-12620 adjust InvitablePublicAgencyResourceType :

bakridemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12620/Institutionen-hinzufugen-nicht-moglich

Description: 
 since the last EDT update, we have to check and disallow empty lists for propertyHasNotAnyOfValues. In this case the 'invitedOrgaIdsCondition' has to be returned even with a empty value ( when there are no invited institutions yet ) otherwise the condition will return null.


jalalifardemos (approved):
Thanks, tested too

PR #3828, opened 2024-10-14
bug(refs DPLAN-12601): Fix layer legend

gruenbergerdemos:
### Ticket
DPLAN-12601

- Add height: auto to layer legend item, because the legend should take so much space as it needs
- Make toggle button disabled if there is no legend
- Hide horizontal scrollbar, because we have a slider button for that

### How to review/test
You can add a legend for a layer at Planungsdokumente und Planzeichnung -> Kartenebene definieren -> Neue Kartenebene (eg https://sgx.geodatenzentrum.de/wms_basemapde)
Verfahren -> Öffentlichkeitsansicht -> Legende should be disabled if there is no legend
Verfahren -> Öffentlichkeitsansicht -> legend should take the place it needs vertically, but there should be no horizontal scrollbar


hwiem (approved):
Lgtm & works!
Beyond the fix, I'm wondering if we want to display something like "Keine Legende verfügbar" as a title when hovering over the disabled button? But it's a nitpick :)

PR #3838, opened 2024-10-15
bug(refs DPLAN-12623): Fix adding and editing places

gruenbergerdemos:
### Ticket
DPLAN-12623

- Add Checks if the given place name is unique within the list of places. To update an existing place the currentPlaceId is given and the place with this id is excluded from the check.
- Add error handling for update request
- Add id from request to new places to fix sorting

### How to review/test
Verfahren -> Übersicht -> Bearbeitungsschritte
- Add a new place -> Sort it -> Reloade page -> Sorting should be saved 
- Edit a place -> Take same name as another place -> Saving should throw an error explaining that you cannot have the same name


client/js/components/procedure/admin/AdministrationPlaces.vue
```diff
@@ -392,9 +392,13 @@ export default {
       }
 
       dpApi.patch(Routing.generate('api_resource_update', { resourceType: 'Place', resourceId: rowData.id }), {}, payload)
-        .then(dplan.notify.confirm(Translator.trans('confirm.saved')))
+        .then(checkResponse)
         .catch((err) => console.error(err))
-        .finally(() => {
+        .finally((response) => {
```
hwiem, line 397: nitpick

```suggestion
        .finally(response => {
```

client/js/components/procedure/admin/AdministrationPlaces.vue
```diff
@@ -302,19 +303,17 @@ export default {
...
      *
      * @param placeName { string }
-     * @param isNewPlace { boolean }
+     * @param currentPlaceId { string }
      * @returns { boolean }
      */
-    isUniquePlaceName (placeName, isNewPlace = false) {
-      const foundSimilarName = this.places.filter(el => el.name === placeName)
-      return isNewPlace ? foundSimilarName.length === 0 : foundSimilarName.length === 1
+    isUniquePlaceName (placeName, currentPlaceId = '') {
```
hwiem, line 314: nitpick: since `name` and `id` both refer to the same place, I think we should use consistent naming (e.g. `currentPlaceName` and `currentPlaceId`)

PR #3839, opened 2024-10-15
fix(refs DPLAN-12651): Prevent undefined header fields

salisdemos:
when permission field_organisation_competence is false, it produced an empty object. in some logic it ws tried to find an attribute field within that object, which lead to an error.

### Ticket
DPLAN-12651


gruenbergerdemos (approved):
lgtm, thanks

PR #3840, opened 2024-10-15
fix (DPLAN-12617) Adjust permission to use same PDF import to display original pdf

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12617/Stellungnahmen-Original-PDF-Link-weiterleitet-zur-Seite-nicht-gefunden-Seite-und-weitere-Anhange-werden-in-der-Detailansicht

Add additional permission to display "Original PDF" button only when the "PDF Import" feature is active in the project

### How to review/test
- Go to diplanrog and go to stn liste
- Original PDF btn is not there
- Go to ewm (same branch no need to add any addon)
- Go to stn liste
- Original PDF btn is there

- [ ] Tests updated/created
- [ ] Update documentation


hwiem (approved):
This is only a temporary fix, right? Assuming that, this lgtm :+1: 
But in order to understand why this change was made, I think it could be helpful to add the reasoning in the PR description/commit message

hwiem (approved):

sierrazamudiodemos:
Hey @hwiem , I believe this is a long-term solution, though there might be a better one. Currently, the "Import PDF" feature is only available for EWM. The "Original PDF" button is part of this feature, but I think the permissions for the button weren’t set correctly when it was created. Since we're now using the same template for Diplanrog in EWM, the "Original PDF" button has appeared there unintentionally.

hwiem:
> Hey @hwiem , I believe this is a long-term solution, though there might be a better one. Currently, the "Import PDF" feature is only available for EWM. The "Original PDF" button is part of this feature, but I think the permissions for the button weren’t set correctly when it was created. Since we're now using the same template for both Diplanrog and EWM, the "Original PDF" button has appeared there unintentionally.

Ah, then I misunderstood. I think we will make the 'original pdf' download available independently of the import center soon, so this check might not work anymore. But more importantly, I tested this with both permissions enabled and there are still instances where the pdf can't be downloaded. I noticed that in those instances `originalPdf` is `undefined` (leading to the error "Unter dem angefragten Pfad /file/c4756d1d-a6e8-4c82-ada7-1eaf0fc3c07e/undefined wurde kein Inhalt gefunden." , while in other cases it is `null`. In the FE, we disable the button if `originalPdf` is `null`, but not if it is `undefined`. I think this might be a FE issue or an issue in the vuex-json-api library - I checked quickly now and we are trying to access the file `hash` in client/js/components/statement/listStatements/ListStatements.vue:765, but the file `attributes` appear to be an empty array. I could look into this later or maybe @gruenbergerdemos can check it

sierrazamudiodemos:
@hwiem would you prefer to keep the PR open for the FE changes ? or a new PR for FE can be created?

hwiem:
> @hwiem would you prefer to keep the PR open for the FE changes ? or a new PR for FE can be created?

Let's keep this one :+1:

breierDemos (approved):
lgtm thanks!

hwiem (approved):
The code as such looks good to me, but the issue in the statement detail view mentioned in the ticket is not fixed (afaics) and while the original pdf link is not displayed in projects without the pdf import anymore, the underlying problem is not fixed (so in projects with pdf import, the error still happens).

gruenbergerdemos:
Attachments in detail view is fixed here https://github.com/demos-europe/demosplan-project-diplanrog/pull/146. 
@hwiem In which projects does the original PDF link not work? It does work in EWM for me.

hwiem:
> Attachments in detail view is fixed here [demos-europe/demosplan-project-diplanrog#146](https://github.com/demos-europe/demosplan-project-diplanrog/pull/146). 

Ah I didn't know that, thanks :+1: 

> @hwiem In which projects does the original PDF link not work? It does work in EWM for me.

Interesting, for me it doesn't work in ewm. I didn't find a statement with an original pdf link that worked - either they didn't have the link enabled or it didn't work.

gruenbergerdemos:
> Interesting, for me it doesn't work in ewm. I didn't find a statement with an original pdf link that worked - either they didn't have the link enabled or it didn't work.

I tested it again and it seems to only work with new statements. With old ones the I also get the "Seite nicht gefunden" error :disappointed:

PR #3848, opened 2024-10-17
fix (DPLAN-12622) Adjust permission to display button

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12622/Als-Fachplanung-kann-ich-aus-der-Offentlichkeitsansicht-nicht-zuruck-in-die-Bearbeitungsansicht-navigieren

The button "Verfahren bearbeiten" was initially linked to the `"area_admin_assessmenttable"` permission but is no longer dependent on it. I have decided to use a different permission that better aligns with the button's purpose.

- The `"area_admin_assessmenttable"`permission was originally set using the setProcedurePermissions method for ownsProcedure (see: demosplan/DemosPlanCoreBundle/Permissions/Permissions.php:528).
- With the removal of the assessment table, this permission has been eliminated from the core and shifted to projects that still use the assesment table.
- Since `"area_admin_assessmenttable"` permission  will eventually be deprecated along with the assessment table, I explored other permissions.
- I identified `"area_public_participation`" as a suitable alternative. This permission "controls access to a procedure's public details."
- I thought about creating a new permission called `"statement_list"` to replicate the old function. However, since EWM works fine without it, I didnt see the need.
- Thus, the `"area_public_participation"` permission was chosen as the most fitting option.

### How to review/test
- Login as Walter West
- Go to diplanrog, check any Verfahren
- Click on Öffentlichkeit/Institutions-Ansicht
- You should see the Public details page of the Verfahren
- Then you should see a button on the right "Verfahren bearbeiten" click on it and it will take you to the Verfahren
- Do the same with blp for example (as I removed the old permission, then this will impact the other projects) 

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/pageheader.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/pageheader.html.twig
@@ -5,7 +5,7 @@
         procedure is defined and
         procedure is not null and
         (
-                hasPermission('area_statements') or hasPermission('area_admin_assessmenttable')
+                hasPermission('area_statements') or hasPermission('area_public_participation')
         )
     )
 %}
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/pageheader.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/pageheader.html.twig
@@ -5,7 +5,7 @@
         procedure is defined and
         procedure is not null and
         (
-                hasPermission('area_statements') or hasPermission('area_public_participation')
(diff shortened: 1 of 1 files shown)


salisdemos (changes requested):
I think this is not fitting. What the permission should say is "i have access to the admin area". The Permission You chose is used to control the public section. What do You think about the permission "area_admin" ?

sierrazamudiodemos:
Good idea! I updated it now ;) @salisdemos 946899299b7d71825a305aca8aa4cc546adc996a

salisdemos (approved):
I hope we find the propper permission :crossed_fingers:

PR #3852, opened 2024-10-17
chore: Bump Babel-Jest@29 and dependents

salisdemos:
- since the output of snapshot changed, I hat to update them
- with babel-jest@29 we hd to bump @vue/vue2-jest and jest too.
- with jest@29 jest-environment-jsdom has to be installed explicitly


sakutademos (approved):
code review, lgtm

PR #3772, opened 2024-10-04
Adjust File name in pdf export

jalalifardemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-2197/langer-PDF-Name-fuhrt-zu-Formatierungfehler

Description: Adjust Long file name in PDF export

Delete the checkbox if it doesn't apply/isn't necessary.

- [ ] Tests updated/created
- [ ] Update documentation
- [x] Link all relevant tickets
- [x] Move the tickets on the board accordingly
- [ ] Data-Cy attributes added/updated ([conventions](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/twig_html.html#guideline-for-naming-cypress-hooks))

PR #3853, opened 2024-10-17
Revert "Adjust File name in pdf export"

martendemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-2197

Reverts demos-europe/demosplan-core#3772
There are missing latex packages at suse. This fix isn't needed for the current release, therefore it can be postponed.


MoritzMandler (approved):
lgtm

PR #3862, opened 2024-10-18
f DPLAN-12677 flush statement without container file while copying

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12677

Description: 
We do have to flush the new copied statement here if the original statement has no FileContainers otherwise the new copied statement is already flushed while copying FileContainers in the previous method 'addFilesToCopiedStatement'.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementCopier.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementCopier.php
@@ -521,6 +521,8 @@ class StatementCopier extends CoreService
         $this->getDoctrine()->getManager()->persist($newStatement);
         if ($statement->getFiles() !== []) {
             $this->statementService->addFilesToCopiedStatement($newStatement, $statement->getId());
+
+            return $newStatement;
         }
 
         // We do have to flush the new copied statement here if the original statement has no FileContainers otherwise
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementCopier.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementCopier.php
@@ -519,7 +519,13 @@ class StatementCopier extends CoreService
         }
         // persist to get an ID for the FileContainer copying below
         $this->getDoctrine()->getManager()->persist($newStatement);
-        $this->statementService->addFilesToCopiedStatement($newStatement, $statement->getId());
+        if ($statement->getFiles() !== []) {
+            $this->statementService->addFilesToCopiedStatement($newStatement, $statement->getId());
+        }
+
+        // We do have to flush the new copied statement here if the original statement has no FileContainers otherwise
+        // the new copied statement is already flushed while copying FileContainers in the previous method 'addFilesToCopiedStatement'.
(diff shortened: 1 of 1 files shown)


jalalifardemos (approved):
Thanks, Lgtm

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementCopier.php
```diff
@@ -519,7 +519,13 @@ public function copyStatementObjectWithinProcedureWithRelatedFiles(
...
         // persist to get an ID for the FileContainer copying below
         $this->getDoctrine()->getManager()->persist($newStatement);
-        $this->statementService->addFilesToCopiedStatement($newStatement, $statement->getId());
+        if ([] !== $statement->getFiles()) {
+            $this->statementService->addFilesToCopiedStatement($newStatement, $statement->getId());
+        }
+
+        // We do have to flush the new copied statement here if the original statement has no FileContainers otherwise
+        // the new copied statement is already flushed while copying FileContainers in the previous method 'addFilesToCopiedStatement'.
+        $this->getDoctrine()->getManager()->flush();
```
MoritzMandler, line 528: This would avoid a double flush...

// We do have to flush the new copied statement here if the original statement has no FileContainers otherwise
// the new copied statement is already flushed while copying FileContainers in the previous method 'addFilesToCopiedStatement'
if (0 === count($statement->getFiles())) {
    $this->getDoctrine()->getManager()->flush();
}

MoritzMandler (changes requested):
target branch

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementCopier.php
bakridemos: thx done here https://github.com/demos-europe/demosplan-core/pull/3862/commits/c03677c7dc3c84833d9dba95fade1095f3282ae9, differently but double flush is avoided

bakridemos:
> target branch

i think main is alright, release is already back merged in main

MoritzMandler:
replaced by https://github.com/demos-europe/demosplan-core/pull/3864

bakridemos:
this issue is already fixed in https://github.com/demos-europe/demosplan-core/pull/3864

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementCopier.php
martendemos: Guys I don't get it. Double flush isn't a thing at all, afaik.
The flush of doctrine just writes everything in the database what was persisted before. After all of this is written, doctrine cleared this "persisted storage". An additional flush will not find anything to persist.
This additional logic, which is used here is misleading, I would prefer to avoid it.

bakridemos: we want to avoid double flush here but this was not the problem. We do flush copied statement once while copying file container which is part of the copy process. The problem was that we don't flush anywhere copied statement if they don't have any filecontainer

PR #3871, opened 2024-10-21
F dplan 12665 enable available phases using api

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12665/Verfahrensschritte-in-Detailansicht-funktionieren-nicht-richtig

- Phase field needed to be adjusted so that it can be displayed on the Statement API
- Additionally, availableInternalPhases and availableExternalPhases also needed to be exposed in the API

FE: 
- Get availableInternalPhases and availableExternalPhases via Api
- Remove deleting isManual when saving, because we need it as check for several inputs and only updated fields are sent via request
- Use phase as value for both availableInternalPhases and availableExternalPhases, because there can only be one of them

### How to review/test
1. Go to projects where phases are available
2. Go to a Verfahren and create a STN with a particular phase
3. Then go to the created STN, check that the selected phase is there
4. Also update the phase and check that the selected phase is there

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
BE Pair review: It's looking good so far. For now, it's working and should fix the bug. However, we want to improve it further on a new branch. Thanks!

hwiem (approved):
FE code lgtm + it works

PR #3875, opened 2024-10-22
feat(refs DPLAN-12697): add custom html exchange format for prosemirror data

muellerdemos:
### Ticket

**WIP**
#### what is currently working:
- tests :)
- basic transformation of prosemirror formatted data into a html string and vice versa
#### what is missing:
- make this clean and remove redundancies
- not all parsed attributes are available currently (e.g. segment_text)

Basically a PoC state but certain aspects already functional even though not cleaned up yet. But still quite a bit to do:
- make sure the prosemirror data is valid and useable in editors
- make sure the html string contains all relevant data

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


client/js/lib/prosemirror/converter.js
```diff
@@ -0,0 +1,88 @@
...
+ * @param htmlString - string
+ */
+export class ProseMirrorConverter {
+  constructor() {
+    this.prosemirrorData = null
+    this.htmlString = null
+  }
+
+  /**
+   * Returns a ProseMirrorConverter instance filled with ProseMirror data.
```
spiess-demos, line 20: Nitpick: At this point it is not perfectly clear what "ProseMirror data" may be.

client/js/lib/prosemirror/converter.js
```diff
@@ -0,0 +1,88 @@
+/**
+ * This ProseMirrorConverter class is responsible for converting
+ * ProseMirror data to an HTML string and to convert proseMirror data
+ * to an HTML string.
```
spiess-demos, line 4: Nitpick: something seems to be duplicated here.

client/js/lib/prosemirror/converter.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * Returns a ProseMirrorConverter instance filled with an HTML string.
+   * @param htmlString
+   * @returns { ProseMirrorConverter }
+   */
+  fromHtml(htmlString) {
+    try {
+      const parser = new DOMParser()
```
spiess-demos, line 55: I'd suggest to instanciate the parser with the class constructor so it is not instanciated everytime again, when `fromHtml` runs.

client/js/lib/prosemirror/converter.js
muellerdemos: I see, I tried to be more explicit here: https://github.com/demos-europe/demosplan-core/pull/3875/commits/4ba7fe8b305eb22aa506d0e9ae95102779fa1faa and hope by adding a documentation link too, things will be more clear.

client/js/lib/prosemirror/converter.js
muellerdemos: Ups, thanks a lot, this was adjusted here, too https://github.com/demos-europe/demosplan-core/pull/3875/commits/4ba7fe8b305eb22aa506d0e9ae95102779fa1faa

client/js/lib/prosemirror/converter.js
muellerdemos: This is a really great suggestion, thanks a lot!
Done here: https://github.com/demos-europe/demosplan-core/pull/3875/commits/269959511ba697df98843ec0a812649697c6cd31

muellerdemos:
@spiess-demos thank you for having an early look at this piece of work already. I was also not so sure if we want this class syntax since I rarely see this done in dplan projects (I personally like it from time to time since it enables method chaining quite easily) but I am not sure if this is a shared view.

client/js/lib/prosemirror/converter.js
spiess-demos: thanks! i think this makes it clear for "first time readers" of that code.

muellerdemos:
Unfortunately this piece of work was redundant hence this PR can be deleted. We never need to transform the exchange format from html to prosemirror or vice versa. The format will be provided by data and we need to use it then.

PR #3879, opened 2024-10-22
f DPLAN-12711 adjust statement intern id length

bakridemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12711/Eingangsnummer-Lange-muss-angepasst-werden

Description: The input/entry number length can have a maximum of 35 characters but it can be sometimes too long, that' why it has to be adjusted


mussbach (approved):
lgtm, thx

PR #3893, opened 2024-10-25
fix(refs DPLAN-12672): rename datacy hook

hwiem:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12672

The hook in the statement list should be changed back to `createStatement` as that is how it is used in cypress tests already (and the original statement list should have the same hook).

### PR Checklist


spiess-demos (approved):
i thought we did agree on the hooks to have some sort of namespacing but maybe that is not needed.

hwiem:
> i thought we did agree on the hooks to have some sort of namespacing but maybe that is not needed.

Yes, but maybe only for new hooks? The change seemed urgent, so I did not want to start a discussion. But I'll ask about it. :)

PR #3905, opened 2024-10-30
F dplan 12618 check if one layer al least is available

bakridemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12618/Es-muss-immer-ein-Gis-Dienst-hinterlegt-sein-dieser-muss-aber-nicht-zwingend-die-Basemap-sein

Description: in some projects, a procedure had to have at least one layer, that's means that users supposed not be able to all existing layers. The resource type is adjusted in this pr to check and make sure that users can not anymore delete all layers and introduce a new permission that will be only used in projects where procedure had to have at least one layer.

Note:  we don't have in this case a bulk delete that's why this approach cover not that.

### DIFF
--- a/client/js/components/map/admin/AdminLayerList.vue
+++ b/client/js/components/map/admin/AdminLayerList.vue
@@ -137,6 +137,7 @@
         <h3 class="mt-4">
           {{ Translator.trans('map.bases') }}
         </h3>
+        <h5>{{ Translator.trans('info.gislayer.delete') }}</h5>
         <!-- List-Head -->
         <div class="color--grey mb-1 mt-2 mr-2">
           <div class="c-at-item__row-icon pl-0">
--- a/client/js/components/map/admin/AdminLayerList.vue
+++ b/client/js/components/map/admin/AdminLayerList.vue
@@ -137,7 +137,9 @@
         <h3 class="mt-4">
           {{ Translator.trans('map.bases') }}
         </h3>
-        <h5>{{ Translator.trans('info.gislayer.delete') }}</h5>
+        <h5 v-if="hasPermission('feature_map_required')">
+          {{ Translator.trans('info.gislayer.delete') }}
+        </h5>
         <!-- List-Head -->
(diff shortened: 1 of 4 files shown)


client/js/components/map/admin/AdminLayerList.vue
```diff
@@ -137,6 +137,7 @@
         <h3 class="mt-4">
           {{ Translator.trans('map.bases') }}
         </h3>
+        <h5>{{ Translator.trans('info.gislayer.delete') }}</h5>
```
salisdemos, line 140: Here is a permission-Check missing.
```suggestion
        <h5 v-if="hasPermission('feature_map_required')">
          {{ Translator.trans('info.gislayer.delete') }}
        </h5>
```

demosplan/DemosPlanCoreBundle/ResourceTypes/GisLayerResourceType.php
```diff
@@ -88,6 +94,15 @@ public function isAvailable(): bool
 
     public function isDeleteAllowed(): bool
     {
+        if ($this->currentUser->hasPermission('area_map_participation_area')) {
```
salisdemos, line 97: how do we ensure, that there is one base layer in the beginning? A blueprint doesn't have to have a layer at all.

salisdemos (changes requested):

client/js/components/map/admin/AdminLayerList.vue
bakridemos: oh yee thx

demosplan/DemosPlanCoreBundle/ResourceTypes/GisLayerResourceType.php
bakridemos: While creating a procedure the layers will be copied from the master blueprint which had already a basemap layer. So one layer will be direct available

salisdemos (approved):
lets give it a try

PR #3909, opened 2024-10-30
f revert unneeded adjustments 

bakridemos:
Description: these changes are not needed anymore since the issue is already treated differently in https://github.com/demos-europe/demosplan-core/pull/3908


salisdemos (approved):
legit

PR #3915, opened 2024-11-01
chore: Remove dp-form-row

salisdemos:
we can archive the same result with just some tw classes

### DIFF
--- a/client/js/components/procedure/admin/DpNewProcedure/DpNewProcedure.vue
+++ b/client/js/components/procedure/admin/DpNewProcedure/DpNewProcedure.vue
@@ -53,15 +53,14 @@
         v-text="Translator.trans('procedure.data')" />
 
       <addon-wrapper hook-name="procedure.fields" />
-
       <template v-if="hasPermission('feature_use_plis')">
-        <dp-form-row class="u-mb-0_75">
-          <dp-select
-            id="r_plisId"
-            :label="{ text: Translator.trans('name'), hint: Translator.trans('explanation.plis.procedurename') }"
-            name="r_plisId"
-            :options="plisNameOptions" />
-        </dp-form-row>
+        <dp-select
+          id="r_plisId"
+          class="mb-4"
+          :label="{ text: Translator.trans('name'), hint: Translator.trans('explanation.plis.procedurename') }"
+          name="r_plisId"
+          :options="plisNameOptions" />
+
         <dl>
           <dt
             v-text="Translator.trans('public.participation.desc')"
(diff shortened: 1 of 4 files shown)


client/js/components/procedure/admin/NewBlueprintForm.vue
```diff
@@ -35,49 +35,47 @@
...
-          data-cy="newMasterName"
-          :label="{
-            text: Translator.trans('name')
-          }"
-          maxlength="200"
-          name="r_name"
-          required />
-      </dp-form-row>
+      <dp-input
+        class="mb-0.5"
```
spiess-demos, line 39: I'd suggest that we settle on a method for vertical spacing: either apply a top- or bottom margin, never both. What do you think?

client/js/components/procedure/admin/NewBlueprintForm.vue
```diff
@@ -120,7 +120,7 @@
 
         <p
           v-if="isCustomerMasterBlueprintExisting && hasPermission('feature_admin_customer_master_procedure_template')"
-          class="lbl__hint u-ml-0_75 u-mb">
+          class="flash-warning lbl__hint p-1 ml-4 mb-4 mt-1">
```
spiess-demos, line 123: We could also use DpInlineNotification here:

```
      <dp-inline-notification
          v-if="isCustomerMasterBlueprintExisting && hasPermission('feature_admin_customer_master_procedure_template')"
          :message="Translator.trans('explanation.customer.masterblueprint.uncheck.existing')"
          type="warning" />
```

client/js/components/user/citizenRegisterForm/CitizenRegisterForm.vue
```diff
@@ -25,9 +25,10 @@
             :class="prefixClass('font-size-large u-mb')"
             v-text="Translator.trans('login.email')" />
           <slot />
-          <dp-form-row :class="prefixClass('u-mb-0_75 space-stack-s')">
+          <div :class="prefixClass('mb-3 grid gap-4')">
```
spiess-demos, line 28: this is freaking simple. l like it. much.

client/js/components/procedure/admin/NewBlueprintForm.vue
salisdemos: Yeah. I thought something similar several times.
I would say, lets always append top-spacing, with that we don't have to handle the last Element in a loop. 
On the other hand, most defaults add spacing to the bottom :thinking: 

What Do You prefer?

spiess-demos: as @hwiem also mentioned in the related slack discussion, i'd prefer placing the spacing class to the container, and in cases where this is not possible, add top-spacing to elements (as in `<element class="mt-0.5">`).

client/js/components/procedure/admin/NewBlueprintForm.vue
salisdemos: good suggestion
8bbf938d21ae91c40efe848c87f3e6285e658043

client/js/components/procedure/admin/NewBlueprintForm.vue
salisdemos: since the spacing was a bit off anyway, I tried to clean it a bit instead of reworking "everything".
52f41f0bdb705a4fb3f635d9ea368d47c0762e55

client/js/components/procedure/admin/DpNewProcedure/DpNewProcedure.vue
```diff
@@ -53,15 +53,14 @@
...
-        <dp-form-row class="u-mb-0_75">
-          <dp-select
-            id="r_plisId"
-            :label="{ text: Translator.trans('name'), hint: Translator.trans('explanation.plis.procedurename') }"
-            name="r_plisId"
-            :options="plisNameOptions" />
-        </dp-form-row>
+        <dp-select
+          id="r_plisId"
+          class="mb-4"
```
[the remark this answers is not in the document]
salisdemos: true 3c347b098febbc79b6bb776c6b41c2386a682348

client/js/components/user/orgaRegisterForm/OrgaRegisterForm.vue
```diff
@@ -87,43 +88,40 @@
...
+                :class="prefixClass('md:col-span-2')"
+                data-cy="useremail"
+                :label="{
+                  bold: false,
+                  text: Translator.trans('email.address')
+                }"
+                name="r_useremail"
+                required
+                type="email" />
+              <dp-input
```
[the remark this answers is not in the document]
salisdemos: yes. Tested again.

sakutademos (approved):
code review, lgtm! Thx!

PR #3919, opened 2024-11-04
chore: use runtime component

mussbach:
### Ticket
https://www.dev.diplanung.de/DefaultCollection/EfA%20DiPlanung/_workitems/edit/20738

The runtime component https://symfony.com/doc/5.x/components/runtime.html is the default way of symfony to boot the application. We should adhere with this.

### How to review/test
Everything should work as expected or break immediately

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


sierrazamudiodemos:
- When I run  `bin/diplanbau dp:c:c` nothing is displayed in the console
- When I run `y dev:diplanbau`  I get an error

sierrazamudiodemos (changes requested):
- when I run the command bin/diplanbau dp:c:c nothing is displayed in console
- when I run the command y dev:diplanbau   there is an error

PR #3923, opened 2024-11-04
fix(refs dplan-12776): Make TS-Components mount

salisdemos:
It looks like the name of the components for written in TS is not accessible like before. So we have to put the name straight away which is even cleaner imho

### Ticket
DPLAN-12776

Part of the vue3-Migration DPLAN-12315 
https://github.com/demos-europe/demosplan-core/pull/1151

### DIFF
--- a/client/js/InitVue.js
+++ b/client/js/InitVue.js
@@ -94,7 +94,7 @@ function initialize (components = {}, storeModules = {}, apiStoreModules = [], p
 
     Object.keys(components).forEach(comp => {
       if (components[comp]) {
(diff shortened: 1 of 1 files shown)


elmasdemos (commented):
lgtm, code review

elmasdemos (commented):
btw didnt know that this would be a problem for ts

elmasdemos (approved):

salisdemos:
> btw didnt know that this would be a problem for ts

I was surprised too. And I am still not sure if its a problem with our way to build demosplan-ui.

PR #3928, opened 2024-11-04
fix(refs dplan-12776): Remove Email Boilerplates in Customer Settings Update mail

salisdemos:
since we are not in a procedure, we can't have a procedureId which is required by the boilerplate component.
An Error Occured only in dev-mode. Bit also in prod-Mode no Boilerplates where shown, So we can remove the code.

On my way I moved something closer to vue

### Ticket
DPLAN-12776


hwiem (approved):
lgtm, code review only

PR #3933, opened 2024-11-04
fix(refs dplan-12770): Don't transform store-state (publicStatement)

salisdemos:
### Ticket
DPLAN-12770

this.formData is bind to the store and is not allowed to be transformed without mutation. in vue2 this was just a warning. now, in vue3 it leads to a hard error.

Part of the vue3-Migration [DPLAN-12315](https://demoseurope.youtrack.cloud/issue/DPLAN-12315)
https://github.com/demos-europe/demosplan-core/pull/1151


elmasdemos (approved):
lgtm, code review. its a known error in vue3

PR #3935, opened 2024-11-04
fix(refs dplan-12775): Move customerBranding logic to Store

salisdemos:
it always was supposed to work like so
and it makes more sense.
Now we don't have to pass the data to the component anymore.

Before it worked, but it was'nt possible to save styling-changes if no img where present.

As Support goto customer-settings page and try to change the styling without an uploaded image

### Ticket
DPLAN-12775


elmasdemos (approved):
lgtm, code review

PR #3962, opened 2024-11-07
fix(refs DPLAN-12699): Don't request Municipalities if not enabled

salisdemos:
### Ticket
DPLAN-12699

This issue is related to the statement-List/a-table refactoring [i_T35963_integrate_project_features](https://github.com/demos-europe/demosplan-core/tree/i_T35963_integrate_project_features)

As FPA goto A-Table -> Table should load.

Adjust permission-Check that municipalities only get requested if they are present. 

May be untestable because an Pagerfanta update crashes the a-table atm.

### DIFF
--- a/client/js/store/statement/Statement.js
+++ b/client/js/store/statement/Statement.js
@@ -510,7 +510,7 @@ export default {
         fields.County = 'name'
       }
 
-      if (hasAnyPermissions(['field_statement_municipality', 'area_admin_assessmenttable'])) {
+      if (hasPermission('field_statement_municipality')) {
         includes.push('municipalities')
         statementFields.push('municipalities')
         fields.Municipality = 'name'
@@ -865,7 +865,6 @@ export default {
             'elements',
             'files',
             'fragmentsElements',
-            'municipalities',
             'paragraph',
             'priorityAreas',
             'tags'


hwiem (commented):
While the code lgtm, I do not understand:
- how this related to [i_T35963_integrate_project_features](https://github.com/demos-europe/demosplan-core/tree/i_T35963_integrate_project_features)
- why `'area_admin_assessmenttable'` was added here before. Should it be an 'and' instead of an 'or' connection? Is it safe to remove it?

sierrazamudiodemos (commented):
Thanks for this

sierrazamudiodemos:
I think it is safe to remove it, because the projects that have area_admin_assessmenttable do not have the permission field_statement_municipality. Only one project has that permission active, so to me these 2 permissions are not related to each other

salisdemos:
> * why `'area_admin_assessmenttable'` was added here before. Should it be an 'and' instead of an 'or' connection? Is it safe to remove it?

As far as I was able to reconstruct it, it looks like it was added "accidentially" during a refactoring. Not knowing, that the area permission would be enabled anyway to display the page.
I further fear, that other permissions set nearby may be not 100% fitting too.

hwiem (approved):

PR #3963, opened 2024-11-07
feat (refs DPLAN-12824): Don't delete customer master template

breierDemos:
### Ticket
[DPLAN-12824](https://demoseurope.youtrack.cloud/issue/DPLAN-12824/Es-muss-sichergestellt-werden-dass-immer-eine-Plattform-Blaupause-aktiviert-ist)

If the user tries to delete the "Plattform-Blaupause", the deletion won't be executed. Instead it skips the template and prints a warning.

### How to review/test
Login as Mandanten Admin and try to delete the Plattform Blaupause, Code Review

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


sierrazamudiodemos (approved):
Thank you for this! I would add to make a difference between Procedure and Blaupause on the logs, but I think it is ok, thanks for this!!

PR #3964, opened 2024-11-07
chore: fix pagerfanta pagination after major update

mussbach:
Interface and property visibility changed in major update

### How to review/test
call assessmenttable

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (changes requested):
I think with $this->array, we are not getting the ArrayAdapter::array. The value is copied during the constructor call, but then the values of the two arrays could theoretically be changed separately. Could this be a problem?

mussbach:
As the ArrayInterface does not contain any setters (nor direct getters) the contents of the array cannot be altered, so the copy of the array should be no problem @breierDemos

breierDemos (approved):
I see, thank you
code review only

PR #3971, opened 2024-11-11
feat (DPLAN-12730): show original attachment link

breierDemos:
### Ticket
[DPLAN-12730](https://demoseurope.youtrack.cloud/issue/DPLAN-12730/Original-STN-Anhang-wird-in-Detailansicht-nicht-angezeigt
)

I removed this permission check to show original attachment link. Do you think we need a permission check here? If yes, maybe we need a new permission, because the previous one is not suitable anymore.

### How to review/test
code review, check link in interface

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


salisdemos (changes requested):
Lets wait how the discussion ends. And if we want to solve it without Permission, I would add the `v-if="originalAttachment.hash"` for the label too

salisdemos (approved):
legit. Thanks

PR #3984, opened 2024-11-12
fix(refs dplan-12823): Reset orga form only for editable fields

salisdemos:
### Ticket
DPLAN-12823

On Orga-Edit page (e.g. as FPA)
`/organisation/edit/<orga-id>`

When resetting the page, non-editable should not change (not be cleared)
Other Fields should change to the initial value (not be cleared if there was initial content)

In addition I moved some code. Sorry. That should have been a separate PR. If wanted I can guide through the changes.
The Commits are quite clear to follow.

### DIFF
--- a/client/js/components/user/orgaDataEntry/OrganisationDataForm.vue
+++ b/client/js/components/user/orgaDataEntry/OrganisationDataForm.vue
@@ -305,12 +305,6 @@ export default {
       default: false
     },
 
-    hasPaperCopySection: {
-      type: Boolean,
-      required: false,
-      default: false
-    },
-
     projectName: {
       type: String,
       required: true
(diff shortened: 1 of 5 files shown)


client/js/components/user/orgaDataEntry/OrganisationDataForm.vue
```diff
@@ -1,15 +1,34 @@
...
         {{ Translator.trans('organisation.data') }}
       </legend>
 
         <!-- Name -->
         <dp-input
           id="orga_name"
           v-model="organisation.name"
-          class="mb-2"
+          :class="prefixClass('mb-2')"
           data-cy="organisationData:name"
```
gruenbergerdemos, line 32: Do we still need these prefixClasses?

client/js/components/user/orgaDataEntry/OrganisationDataForm.vue
```diff
@@ -322,16 +353,47 @@ export default {
...
+        || this.hasNotificationSection
+        || this.showDetailedInfo
+        || hasPermission('feature_change_submission_type')
     },
-
-    displayCustomer () {
-      return hasPermission('feature_display_customer_names') && this.customers?.length
+  },
+  methods: {
+    precheckSubmit (e) {
```
gruenbergerdemos, line 378: This seems to not get called on submit (tested locally)

gruenbergerdemos (changes requested):
Nice refactoring, thanks :)
Just a general question and the pre submit check seems not be working

client/js/components/user/orgaDataEntry/OrganisationDataForm.vue
salisdemos: Yes. we should keep that functionality. ;-)

client/js/components/user/orgaDataEntry/OrganisationDataForm.vue
salisdemos: Not at all? Or did you just don't run into the if?

gruenbergerdemos: I set the breakpoint before the if and it didn't stop

hwiem: Fixed in c18f3b966f51b0ad650dc9b53d3c387f8d0bc088

gruenbergerdemos (approved):
lgtm, thanks for the fix

PR #3985, opened 2024-11-12
feat (DPLAN-11405): delete_command adjustment and test adjustment

martendemos:
### Ticket
DPLAN-11405

This branch was used to check and test manually the deletion of procedures.
In this context I noticed that the existing test only deletes a single and small procedure, which seems to be insufficient to me.
So I add an additional test which is deleting all existing procedure-fixtures.

- [x] Tests updated/created


breierDemos (approved):
lgtm, test works.

PR #3992, opened 2024-11-13
feat(refs DPLAN-12526): add statement quicksave button after ocr text recognition

muellerdemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12526/3-2-Als-Importerin-mochte-ich-meine-Anpassungen-nach-der-OCR-Texterkennung-zwischenspeichern-konnen-um-sicherzugehen-dass-meine

This PR adds the possibility to quick save the annotated statement text so adjustments can be persisted to be used at a later time without needing to finish this step completely, which could result in a loss of the work.

### Linked PRs (optional)

PR Backend: https://github.com/demos-europe/demosplan-addon-demospipes/pull/247

### How to test:
- install this demospipes version: https://github.com/demos-europe/demosplan-addon-demospipes/tree/f-DPLAN-12526-add-quicksave-property
- checkout branch
- run migration `bin/ewm doc:mig:migrate --configuration=addons/cache/demosplan-addon-demospipes-f-DPLAN-12526-add-quicksave-property/config/addon_migrations.yml`
- after ocr text recognition a new button should be visible under the editor component and the text should be savable now (resource type gets an update request)

### PR Checklist


client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -323,6 +323,8 @@
         hidden-input="r_text"
         v-model="values.text" />
 
+      <slot />
+
```
salisdemos, line 327: why do You split the business-logic from the button? 
If there is a case, where we don't want to offer the quicksave, shouldnt we then use a flag or move the logic to the triggering component?

salisdemos (commented):
Sorry, can't check out for testing atm.
Since I can't test, just general thoughts in a comment

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
muellerdemos: That is a good question. Basically I thought we only need the quick save button after the ocr text recognition step right before confirming the pdf import. But it makes me wonder now if this is a useful feature for direct imports in the import center. 

I mean I have added the slot so it does not interfere with the import center but if we want to quick save the statement text there too, I think you are right by moving the button directly into DpSimplifiedNewStatementForm.vue but here I am not sure about the user workflow. I always thought they use the direct import only with copy+paste but if this is not the case I agree and we should move the button to the form directly.

salisdemos: No. I think You are right, a  we don't need it in the import center. But then I would suggest to move the methods to the dpConvertAnotatedPdf file, to gather button-template and method in the same file.

muellerdemos: Thanks for the suggestion. Moved the function to `DpConvertAnnotatedPdf.vue` here: https://github.com/demos-europe/demosplan-core/pull/3992/commits/411080c03c604beb58c87e75ed090ecd12c0b5ae but I think we still need the ref usage since the v-model of `values.text` contains the current text of the statement text editor (which exists in `DpSimplifiedNewStatementForm.vue`).

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -578,6 +582,11 @@ export default {
   methods: {
     setInitialValues () {
       this.values = { ...this.initValues }
+
+      if (this.values.quickSave) {
+        this.values.text = this.values.quickSave
+      }
+
```
salisdemos, line 589: isn't this be redundant with DpConvertAnnotatedPdf.vue:213 ?
if not exactly, I think this logic too should only live in the parent component

salisdemos (approved):

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
muellerdemos: Ah yes indeed, this could be move too, thanks! https://github.com/demos-europe/demosplan-core/pull/3992/commits/e10aa989fe41133723dd986e16ff8d93218b3703

PR #3999, opened 2024-11-14
Feat beaa2 organisation id field

sakutademos:
### Ticket
DPLAN-XXXXX
or
https://www.dev.diplanung.de/DefaultCollection/EfA%20DiPlanung/_workitems/edit/XXXXX

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
```diff
@@ -177,6 +177,18 @@
         </p>
       </div>
 
+      <addon-wrapper
+        hook-name="mein.berlin.additional.field"
+        :addon-props="{
+          class: 'ml-4',
+          relationshipId: this.organisationId,
+          resourceType: 'MeinBerlinAddonOrganisation',
+          label: Translator.trans('organisation.mein.berlin.id'),
```
mussbach, line 186: Addon names and specifics should not be exposed in core

sakutademos: yes, I got it and moved it to the addon. is it ok now? https://github.com/demos-europe/demosplan-core/pull/3999/files#diff-972e4459afe8be8598e4910ee7db76f0eee580c7c0ee7d8ce78224f0593230edR180-R190

mussbach: Thank you, this is way better. I think in the long run we need some kind of hook here that multiple addons could add their fields here but for now this is probably enough

client/js/components/procedure/basicSettings/DpBasicSettings.vue
```diff
@@ -120,11 +124,24 @@ export default {
...
+
+    procedureId: {
+      required: true,
+      type: String
     }
   },
 
   data () {
     return {
+      addonPayload: {
```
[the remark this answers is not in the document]
sakutademos: is it ok?  https://github.com/demos-europe/demosplan-core/pull/3999/commits/7411968ea7ec3529e9b43ae749b62887cc454b99

client/js/components/procedure/basicSettings/DpBasicSettings.vue
```diff
@@ -120,11 +124,24 @@ export default {
...
+
+    procedureId: {
+      required: true,
+      type: String
     }
   },
 
   data () {
     return {
+      addonPayload: { /** The payload required for addon requests. When a value is entered in the addon field, it emits data that must include the following fields */
```
hwiem, line 137: Thanks for adding the comment :heart:

client/js/components/procedure/basicSettings/DpBasicSettings.vue
```diff
@@ -156,6 +189,34 @@ export default {
         })
     },
 
+    handleAddonRequest () {
+      const payload = this.createAddonPayload()
+
+      const addonRequest = dpApi({
+        headers: {
+          ...(dplan.csrfToken && { 'x-csrf-token': dplan.csrfToken }) // TODO: should be adjusted in UI: api2defaultHeaders
```
hwiem, line 197: Why isn't it adjusted in demosplan-ui? It's a very small change, isn't it?

client/js/components/procedure/basicSettings/DpBasicSettings.vue
```diff
@@ -168,8 +229,31 @@ export default {
       this.selectedPublicPhase = phase
     },
 
+    submit () {
+      const addonExists = Boolean(window.dplan.loadedAddons['addon.additional.field'])
+      const addonHasValue = this.addonPayload.value || this.addonPayload.initValue
```
hwiem, line 234: Don't we want this to be of type Boolean, too?

client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
```diff
@@ -972,7 +986,12 @@ export default {
       })
       this.emitOrganisationUpdate()
       this.resetRegistrationStatus()
-    }
+    },
+
+    updateAddonPayload (payload) {
+      console.log(payload)
```
hwiem, line 992: Should the console.log() be removed?

client/js/components/procedure/basicSettings/DpBasicSettings.vue
```diff
@@ -146,6 +163,22 @@ export default {
...
+        attributes: this.addonPayload.attributes,
+        relationships: this.addonPayload.url === 'api_resource_update' ? undefined : {
+          procedure: {
+            data: {
+              type: 'Procedure',
+              id: this.procedureId
+            }
+          }
+        },
+        ...(this.addonPayload.url === 'api_resource_update' ? { id: this.addonPayload.id } : {}),
```
hwiem, line 178: What do you think about using destructuring here?

```suggestion
      const { attributes, id, resourceType, url } = this.addonPayload
      
      return {
        type: resourceType,
        attributes,
        relationships: url === 'api_resource_update' ? undefined : {
          procedure: {
            data: {
              type: 'Procedure',
              id: this.procedureId
            }
          }
        },
        ...(url === 'api_resource_update' ? { id } : {}),
```

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -108,8 +108,10 @@
         :init-procedure-categories="JSON.parse('{{ selectedProcedureCategories|default([])|json_encode|e('js', 'utf-8') }}')"
         init-procedure-phase-internal="{{ proceduresettings.phase }}"
         init-procedure-phase-public="{{ proceduresettings.publicParticipationPhase }}"
-        :init-similar-recommendation-procedures="JSON.parse('{{ selectedAllowedSegmentAccessProcedures|json_encode|e('js', 'utf-8') }}')">
+        :init-similar-recommendation-procedures="JSON.parse('{{ selectedAllowedSegmentAccessProcedures|json_encode|e('js', 'utf-8') }}')"
+        procedure-id="{{ proceduresettings.id|default() }}">
```
hwiem, line 112:
```suggestion
        procedure-id="{{ proceduresettings.id|default }}">
```

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -723,6 +725,17 @@
                                 </div>
                             {% endif %}
 
+                            <addon-wrapper
+                                class="mb-4"
+                                hook-name="addon.additional.field"
+                                :addon-props="{
+                                    'relationshipId': '{{ proceduresettings.id }}',
+                                    'relationshipKey': 'procedure',
+                                    'isInput': true
```
hwiem, line 734:
```suggestion
                                :addon-props="{
                                    relationshipId: '{{ proceduresettings.id }}',
                                    relationshipKey: 'procedure',
                                    isInput: true
```

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -1277,13 +1315,20 @@
 
             {# form controls when not in wizard mode #}
             <div class="text-right space-inline-s">
+                {# A hidden input is used here to prevent the form from being submitted in wizard mode when the next button is clicked #}
                 <input
+                    type="submit"
+                    class="hidden"
+                    value="{{ "save"|trans }}">
```
hwiem, line 1322: Why is the input needed?

translations/messages+intl-icu.de.yml
```diff
@@ -2707,8 +2707,11 @@ procedure.phases.external.evaluating: "Auswertung Öffentlichkeit"
 procedure.phases.external.earlyparticipation: "Frühzeitige Beteiligung Öffentlichkeit"
 procedure.phases.external.participation: "Öffentliche Auslegung"
 procedure.phases.external.closed: "Beschlussfassung Öffentlichkeit"
+procedure.pictogram: "Übersichtskarte/Bild"
+procedure.pictogram.altText: "Alternativtext des Bildes"
+procedure.pictogram.altText.toolTipp: "Geben Sie eine kurze Beschreibung des Bildinhalts ein, um die Barrierefreiheit zu gewährleisten."
+procedure.pictogram.copyright: "Copyright des Bildes"
```
hwiem, line 2713: Do we say 'Copyright' or 'Urheberrecht'?

hwiem (changes requested):

client/js/components/procedure/basicSettings/DpBasicSettings.vue
sakutademos: yes, i was sick for two weeks, so was not able to do it :) i will do it

hwiem: Ah, of course. :) (Hope you're better!)

client/js/components/procedure/basicSettings/DpBasicSettings.vue
muellerdemos: The honor belongs to @sakutademos =)

client/js/components/procedure/basicSettings/DpBasicSettings.vue
muellerdemos: I need to look that up, as far as I know the header should already be a part of `dpApi` and the backend should check for either the `x-csrf-token ` header or use the hidden input in case no such header is provided

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
muellerdemos: Thanks, done here https://github.com/demos-europe/demosplan-core/pull/3999/commits/a084c693d6dc0c0de9c497dc083f969a5476b743

translations/messages+intl-icu.de.yml
muellerdemos: I haven't thought about this before but mixing german and english words is kind of odd, however I couldn't find in slack or in the ticket anything in that regard hence I trust my gut here (and that you find it strange too) and change it to `Urheberrecht`. https://github.com/demos-europe/demosplan-core/pull/3999/commits/dc3a96c8068ac567a51ebc216b65b19c497ba8fa

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
muellerdemos: Oh, actually I can confirm that this input prevents the form to be submitted when in assistant or wizard mode, but I have absolutely no idea why. @sakutademos could you explain this to me? I really don't get it but it works =)

client/js/components/procedure/basicSettings/DpBasicSettings.vue
muellerdemos: I find it more precise and probably less error prone. Done here https://github.com/demos-europe/demosplan-core/pull/3999/commits/05b4d0a61fe1af1d235ee5290657a05239eec642

client/js/components/procedure/basicSettings/DpBasicSettings.vue
muellerdemos: Long story short, we do not need it here and for occurrences we need it, it is already included in dpApi.

sakutademos: hmm, we have `apiDefaultHeaders` and `api2DefaultHeaders`. We set it explicitly in the headers for API 1.0 but not for API 2.0. I wasn't sure why, so I thought it might be a fix to add it there

muellerdemos: We actually only need it for form posts and I have added the csrf token to dpApi a while ago so we do not need to include the hidden csrf input field in all forms (just for convenience reasons)  https://github.com/demos-europe/demosplan-core/pull/3999/commits/2eca50b44d393164e8b8d6d9aba0ef62477b9dcb

sakutademos: hm, but if you remove it, the lilla messages will reappear on the POST request. Or you want to add in demosplan ui?

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
sakutademos: to add some actions when the user clicks the save button, I had to change it from input `type="submit"` to button and add a `submit()` method in `DpBasicSettings`, where I included additional actions before submitting the form. However, as I understand it, the wizard mode requires `type="submit"` because, without it, every button is treated as a submit button. That is my assumption

muellerdemos: Thank you, this makes sense to me and I suggest we leave it like that since this does the trick.

client/js/components/user/DpOrganisationList/DpOrganisationListItem.vue
```diff
@@ -159,6 +168,42 @@ export default {
...
+        ...(this.addonPayload.url === 'api_resource_update' ? { id: this.addonPayload.id } : {}),
+      }
+    },
+
+    handleAddonRequest () {
+      const payload = this.createAddonPayload()
+
+      const addonRequest = dpApi({
+        headers: {
+          ...(dplan.csrfToken && { 'x-csrf-token': dplan.csrfToken }) // TODO: should be adjusted in UI: api2defaultHeaders
```
hwiem, line 192: Shouldn't we remove the csrf token here, too?
Also (but not in this PR), maybe we could extract these functions into another file so we don't have to duplicate them.

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -723,6 +725,17 @@
                                 </div>
                             {% endif %}
 
+                            <addon-wrapper
```
spiess-demos, line 728: Nitpick (not to be implemented in this PR): `addon.additional.field` is a _very_ generic key. Also: does the actual hook name need to contain "addon" in it? as the whole thing is in `<addon-wrapper />`, i think it is clear what we're looking at. I'd suggest to name it more specific to this view and section, like `basicSettings.nameAddress.additionalFields`.

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -898,6 +911,31 @@
                                             )
                                             }}
                                         {% endif %}
+
+                                        <dp-input
```
spiess-demos, line 915: Curious question: Both fields are not wrapped within a permission... this means that they're gonna be visible for all projects, along with the rather opinionated microcopy ("Übersichtskarte/Bild", "Laden Sie ein Bild des Bebauungsplanausschnitts auf einer Karte hoch."). Did you take that decision on purpose?

PR #4003, opened 2024-11-15
getEmailAuthCode can to return null

jalalifardemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12868

Description: remove logic exception and set return empty string, if Email Auth code is not set. Because in `portal_profile_2fa.html.twig` we need null value, that user can to set 2fa .

- [x] Tests updated/created
- [ ] Update documentation


mussbach (changes requested):
please add a test

mussbach (approved):
tests would be nice but no direct blocker in this case

jalalifardemos:
done in https://github.com/demos-europe/demosplan-core/pull/4003/commits/e110bf4297615a028ab8f1554ffee3d7cf6561b9

PR #4010, opened 2024-11-15
fix(refs dplan-12874): Request isPrint for layers to allow cheicking against the attr

salisdemos:
### Ticket
DPLAN-12874

we need isPrint for the layers in the public detail to determine if we wasnt to show them or not. When we dont get the info, we will show them by default.


mussbach (approved):
lgtm, thx, code review only

PR #4022, opened 2024-11-19
feat(refs DPLAN-3217 AB#15115): Fix update votes

gruenbergerdemos:
### Ticket
DPLAN-3217

Only send update request to BE if attributes have changed.

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementPublicationAndVoting.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementPublicationAndVoting.vue
@@ -450,9 +450,10 @@ export default {
 
     sendUpdateVote () {
       const promises = Object.values(this.initialVotes).map(vote => {
-        if (this.votes[vote.id]) {
-          const currentVote = this.votes[vote.id]
-          const hasChanged = Object.keys(vote.attributes).some(key => vote.attributes[key] !== currentVote.attributes[key])
+        const  { id, attributes } = vote
+        const currentVote = this.votes[id]
+        if (currentVote) {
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementPublicationAndVoting.vue
```diff
@@ -451,14 +451,18 @@ export default {
...
+          if (hasChanged) {
+            this.saveStatementVoteAction(vote.id)
+            .then(() => {
+              return true
+            })
+            .catch(() => {
+              dplan.notify.error(Translator.trans('error.api.generic'))
+              return false
+            })
+          }
```
hwiem, line 465: suggestion:

```
const  { id, attributes } = vote
const currentVote = this.votes[id]

if (currentVote) {
  const hasChanged = Object.keys(attributes).some(key => attributes[key] !== currentVote.attributes[key])
  
  if (hasChanged) {
    this.saveStatementVoteAction(id)
      .then(() => {
        return true
      })
      .catch(() => {
        dplan.notify.error(Translator.trans('error.api.generic'))
          return false
        })
      }
```

hwiem (approved):
Nice solution!
I have only a small suggestion that makes the code a little better to read in my opinion, but it's a nitpick :)

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementPublicationAndVoting.vue
gruenbergerdemos: Nice suggestion, thanks

PR #4026, opened 2024-11-19
fix(refs DPLAN-3217): re-add div for consistent spacing

hwiem:
### Ticket
[DPLAN-3217](https://demoseurope.youtrack.cloud/issue/DPLAN-3217/ADO-Issue-15115-Integration-EWM-Part-2-Stellungnahmedetailansicht-aus-EWM-und-dplan-vereinen)

The div was erroneously removed, but is actually needed in case there is no location reference defined. In that case, it ensures there is enough spacing between this and the next field below.

### PR Checklist


hwiem:
Replaced by https://github.com/demos-europe/demosplan-core/pull/4027

PR #4027, opened 2024-11-19
fix(refs DPLAN-3217): re-add div for consistent spacing

hwiem:
the div is needed when there is no location, because without it there is not enough space between this field and the one below


spiess-demos (approved):
lgtm! code review but "what should possibly go wrong?"

PR #4034, opened 2024-11-20
fix: make eslint 9 work with eslintrc.js

hwiem:
In eslint 9, the default config format is `eslint.config.js`. To be able to continue using `eslintrc.js`, we need to set `ESLINT_USE_FLAT_CONFIG` to `false`.
We should move to using `eslint.config.js`, since `eslintrc.js` will not be supported anymore in eslint 10.

To be able to run `yarn lint` and only lint the files we actually want to lint, I added some ignorePatterns to the `eslintrc.js`.

Also, `eslint-plugin-n` needed to be added to the dev dependencies.

And finally, I ran `yarn lint --fix` to get from
```
✖ 992 problems (291 errors, 701 warnings)
```
to
```
✖ 396 problems (100 errors, 296 warnings)
```


muellerdemos (approved):
nice, makes linting work again :heart: 

for the rest i trust eslint here :)

PR #4040, opened 2024-11-21
feat(refs DPLAN-12526): add success and error notifications for quicksave

muellerdemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12526/3-2-Als-Importerin-mochte-ich-meine-Anpassungen-nach-der-OCR-Texterkennung-zwischenspeichern-konnen-um-sicherzugehen-dass-meine

The success and error notifications for quick saving a statement text were missing hence are added with this PR. Also a translation was added to achieve that.

### PR Checklist


client/js/components/procedure/imageAnnotator/DpConvertAnnotatedPdf.vue
```diff
@@ -229,6 +230,13 @@ export default {
       }
 
       dpApi.patch(Routing.generate('api_resource_update', { resourceType: 'AnnotatedStatementPdf', resourceId: this.documentId }), {}, payload)
+        .then(response => {
+          if (response.ok) {
+            dplan.notify.confirm(Translator.trans('statement.save.quickSave.success'))
+          } else {
```
hwiem, line 236: Does it not work to use a `.catch()`?

muellerdemos: unfortunately not. I was actually not so sure so i tested it like that: 
```
dpApi.patch(Routing.generate('api_resource_update', { resourceType: 'AnnotatedStatementPdf', resourceId: '123' }), {}, payload)
        .then(response => { dplan.notify.confirm(Translator.trans('statement.save.quickSave.success')) })
        .catch(error => { dplan.notify.error(Translator.trans('error.api.generic')) })
```

but even though the api returns a response with status code `400` the confirm notification will still be displayed.

hwiem: Do you think it would make sense to still add the `.catch()` block in case the BE throws an exception? I'm not sure that we will handle all errors if we only have the `else` block.

muellerdemos: Indeed, this would be a lot more resilient with your suggestion, thanks, done here: https://github.com/demos-europe/demosplan-core/pull/4040/commits/4822513a1348ae7aad2224e1fd9b6f5096e9de29

hwiem (approved):
lgtm, thanks!

PR #4045, opened 2024-11-22
feat(refs DPLAN-12916): add function to redirect back to import center

muellerdemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12916/Abbrechen-Button-funktioniert-nicht-im-Schritt-Import-bestatigen-und-auch-im-Schritt-Uberprufen

This PR adds a secondary action explicitly for the `DpSimplifiedStatementForm.vue`'s dp-button-row component to redirect the user back to the import center instead of the overview.

- for some reason I still do not understand the `<a>`-tag does not redirect when clicking the secondary button of dp-button-row
- hence a secondary action was added for redirecting back to the import center

### PR Checklist

### DIFF
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -376,7 +376,8 @@
         :href="Routing.generate('DemosPlan_procedure_dashboard', { procedure: procedureId })"
         primary
         secondary
-        @primary-action="submit" />
+        @primary-action="submit"
+        @secondary-action="abort" />
     </form>
   </div>
 </template>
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -580,6 +581,11 @@ export default {
   },
 
   methods: {
+    abort () {
+      const href = `${Routing.generate('DemosPlan_procedure_import', { procedureId: this.procedureId })}/#import#StatementPdfImport`
```
hwiem, line 585: Is the '#' needed before the 'import#StatementPdfImport'?

hwiem (commented):
> for some reason I still do not understand the <a>-tag does not redirect when clicking the secondary button of dp-button-row

I think it's probably because of the `@click.prevent` on the secondary button in `DpButtonRow`. While I think we should fix this in `DpButtonRow` at some point, this workaround seems fine to me, too. 
Just have 1 question

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
muellerdemos: Yes this is needed since we want to redirect directly to the import section of the import center after aborting the confirmation of a pdf import.

hwiem (approved):
Lgtm, code review only

PR #4054, opened 2024-11-25
chore: only load aws s3 sdk api

mussbach:
The package aws/aws-sdk-php contains a lot of aws apis that are not needed. According to https://github.com/aws/aws-sdk-php/tree/master/src/Script/Composer they can be reduced, which shrinks the size of the dependency (and thus the container) by 45 MB.

### How to review/test
run composer install and check that vendor/aws/aws-sdk-php/src/data contains only a few subfolders (including s3), not 50+

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -184,6 +184,9 @@
         }
     },
     "extra": {
+      "aws/aws-sdk-php": [
+        "S3"
+      ],
       "bamarni-bin": {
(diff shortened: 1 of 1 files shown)


breierDemos (approved):
lgtm

PR #4061, opened 2024-11-27
chore: Ratelimit api requests

mussbach:
We should avoid that an attacker could use a captured JWT Token without limits. 

The deletion of unused Injections into AbstractStatementResourceType and StatementResourceType where only needed to make the test run.

### How to review/test
Copy any api request in the browser network panel and fire it as curl (or similar tools) eleven times. The last time should receive a 429 response

### PR Checklist

- [ ] Tests updated/created
- [ ] Update documentation


breierDemos (approved):
lgtm code review, test works :)

PR #4063, opened 2024-11-27
f-DPLAN-12816-retain-layers-preselection-after-saving-or-updating-layers

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12816/ROBOBSH-Stage-Basemap-bleibt-nicht-vorausgewahlt-nach-Speichern-neuer-Kartenlayer

Description: 
adjust condition :
- 'VisibilityGroupId' can be null (default value) or can be an empty string. Layers in the same group can be updated only when 'VisibilityGroupId' is not null and has to be not empty too otherwise the layer does not have any group. That's why the condition has to be adjusted to have the expected behavior.
- the 'getVisibilityGroupId' method is adjusted too : we must retrieve only layers that's belong to the current procedure. Using only the 'visibilityGroupId' to get procedure relevant layers leads to have wrong result and updating layers that's does not belong to the current procedure


salisdemos (approved):
Thank You! I feel confident enough to approve that  ;-)

PR #4065, opened 2024-11-28
chore: add reminders to PR template

hwiem:
Two "reminders for handling PRs" are added:
- Run `yarn lint`
- Run `yarn test`
Also, the reminders are cleaned up a bit so the wording is more consistent.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -28,12 +28,14 @@ A list of all related tasks that need to be done before this can be merged.
 -->
 
 ### PR Checklist
-<!-- Reminders for handling PRs -->
+<!-- Reminders for handling PRs 
 
-Delete the checkbox if it doesn't apply/isn't necessary.
+Delete the checkbox if it doesn't apply/isn't necessary. -->
 
-- [ ] Tests updated/created
+- [ ] Create/Update tests
 - [ ] Update documentation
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -28,12 +28,14 @@ A list of all related tasks that need to be done before this can be merged.
...
 
-Delete the checkbox if it doesn't apply/isn't necessary.
+Delete the checkbox if it doesn't apply/isn't necessary. -->
 
-- [ ] Tests updated/created
+- [ ] Create/Update tests
 - [ ] Update documentation
+- [ ] Add/Update data-cy attributes ([conventions](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/twig_html.html#guideline-for-naming-cypress-hooks))
+- [ ] Run `yarn lint`
+- [ ] Run `yarn test`
```
muellerdemos, line 39: epically good :heart: !

muellerdemos (approved):
thanks, i really like it

salisdemos (commented):
Shouldn't that run as PR-Checks? It may be not required for the moment, but it should be.

hwiem:
> Shouldn't that run as PR-Checks? It may be not required for the moment, but it should be.

You mean both? I think the tests don't run because jenkins is not working, and I'm not sure we ever ran eslint as PR check? At least not for a long time. Right now we have >100 errors and they need to be fixed before we can add the check.
In any case I find it useful to have these reminders in addition to the checks, but especially while the checks are not working.

spiess-demos (approved):
Lgtm, thanks for the additions!

PR #4076, opened 2024-11-28
fix(refs dplan-12315): Send correct data on statement modal edit

salisdemos:
### Ticket
DPLAN-12315
DPLAN-12921

It seems, that the reactivity broke in the statmentModal while upgrading to vue3. I should't have worked before, but it did. Now its not clean, but consistent.

Since I moved a block to a separate method, the changes are not so easy to spot. Sorry for that. 
The main thing I did was using the local `dataToSend` instead of `this.setStatementData` which interacts with the store to prepare the statement data before updating

### DIFF
--- a/client/js/components/statement/publicStatementModal/StatementModal.vue
+++ b/client/js/components/statement/publicStatementModal/StatementModal.vue
@@ -1416,7 +1416,7 @@ export default {
         r_document_id: data.draftStatement.document?.id ?? '',
         r_document_title: data.draftStatement.document?.title ?? '',
         r_represents: data.draftStatement.represents ?? '',
-        r_location: Object.keys(data.draftStatement.statementAttributes)[0] ?? 'mapLocation',
+        r_location: Object.keys(data.draftStatement.statementAttributes).pop() ?? 'mapLocation',
         r_location_geometry: data.draftStatement.polygon,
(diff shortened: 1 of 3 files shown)


client/js/components/statement/publicStatementModal/StatementModal.vue
```diff
@@ -1414,6 +1403,35 @@ export default {
...
+        r_ident: this.draftStatementId,
+        r_isNegativeReport: data.draftStatement.negativ ? '1' : '0',
+        r_element_id: data.draftStatement.elementId || '',
+        r_element_title: data.draftStatement.element?.title ?? '',
+        r_paragraph_id: data.draftStatement.paragraphId ?? '',
+        r_paragraph_title: data.draftStatement.paragraph?.title ?? '',
+        r_document_id: data.draftStatement.document?.id ?? '',
+        r_document_title: data.draftStatement.document?.title ?? '',
+        r_represents: data.draftStatement.represents ?? '',
+        r_location: Object.keys(data.draftStatement.statementAttributes)[0] ?? 'mapLocation',
```
sakutademos, line 1419: hm, I cannot test it now, but I am wondering if it is okay as it is. The previous logic was a bit different: the reduce() method was used before, and now you are returning the first element

PR #4092, opened 2024-12-02
chore(refs dplan-12315): Set deep for watcher explicitly

salisdemos:
### Ticket
DPLAN-12315
DPLAN-12921

even though `deep: false` is the default, we should set it during the migration process so we can easily check if we already checked that occurrence. Otherwise we will get a warning again and again and never know if we can ignore it.


spiess-demos (approved):
code review lgtm!

PR #4095, opened 2024-12-02
chore: Bump Sass-loader to 16.0.3

salisdemos:
Since I had to adjust some options, I had to create the PR manually.
For details, why I had to change things see https://webpack.js.org/loaders/sass-loader/#api

### DIFF
--- a/client/fe/webpack/moduleRules.js
+++ b/client/fe/webpack/moduleRules.js
@@ -160,13 +160,16 @@ const moduleRules =
         {
           loader: 'sass-loader',
           options: {
-            implementation: require('sass-embedded'),
-            additionalData: `$url-path-prefix: '${config.urlPathPrefix}';`,
+            api: 'modern',
             sassOptions: {
-              includePaths: [
-                config.projectRoot + 'web/',
-                config.publicPath
-              ]
+              implementation: require('sass-embedded'),
+              additionalData: `$url-path-prefix: '${config.urlPathPrefix}';`,
+              sassOptions: {
+                includePaths: [
+                  config.projectRoot + 'web/',
+                  config.publicPath
+                ]
+              }
             }
           }
         }
--- a/client/fe/webpack/moduleRules.js
+++ b/client/fe/webpack/moduleRules.js
@@ -160,16 +160,14 @@ const moduleRules =
         {
           loader: 'sass-loader',
           options: {
-            api: 'modern',
+            api: 'modern-compiler',
             sassOptions: {
               implementation: require('sass-embedded'),
               additionalData: `$url-path-prefix: '${config.urlPathPrefix}';`,
-              sassOptions: {
-                includePaths: [
-                  config.projectRoot + 'web/',
-                  config.publicPath
-                ]
-              }
+              loadPaths: [
+                config.projectRoot + 'web/',
+                config.publicPath
+              ]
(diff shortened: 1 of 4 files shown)


client/fe/webpack/moduleRules.js
```diff
@@ -160,13 +160,16 @@ const moduleRules =
         {
           loader: 'sass-loader',
           options: {
-            implementation: require('sass-embedded'),
-            additionalData: `$url-path-prefix: '${config.urlPathPrefix}';`,
+            api: 'modern',
```
hwiem, line 163: Just wondering: Why not 'modern-compiler', as it is strongly recommended?

> Using modern-compiler and sass-embedded together significantly improve performance and decrease built time. We strongly recommend their use. We will enable them by default in a future major release.

salisdemos: oh no. Something went wrong. These changes where not supposed to be here in this PR.

salisdemos: When I use "modern-compiler" the build doesn't finish. No errors, no output, but it doesn't stop.

client/fe/webpack/moduleRules.js
```diff
@@ -160,13 +160,16 @@ const moduleRules =
...
             sassOptions: {
-              includePaths: [
-                config.projectRoot + 'web/',
-                config.publicPath
-              ]
+              implementation: require('sass-embedded'),
+              additionalData: `$url-path-prefix: '${config.urlPathPrefix}';`,
+              sassOptions: {
```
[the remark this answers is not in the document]
salisdemos: Thanks for digging this out. I'll have a deeper look

client/fe/webpack/moduleRules.js
salisdemos: added the workaround @spiess-demos mentioned here: 2179f0d7194c62e1c74e18fe509a829a52ce53f7

client/fe/webpack/moduleRules.js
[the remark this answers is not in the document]
salisdemos: lets try it: 2179f0d7194c62e1c74e18fe509a829a52ce53f7

client/fe/webpack/moduleRules.js
```diff
@@ -160,10 +160,11 @@ const moduleRules =
         {
           loader: 'sass-loader',
           options: {
-            implementation: require('sass-embedded'),
-            additionalData: `$url-path-prefix: '${config.urlPathPrefix}';`,
+            api: 'modern-compiler',
             sassOptions: {
-              includePaths: [
+              implementation: require('sass-embedded'),
+              additionalData: `$url-path-prefix: '${config.urlPathPrefix}';`,
```
spiess-demos, line 166: When i tried it like this, the build failed with some error about scss variables inside css files. My suggestion would be to drop this line entirely and hardcode the path in the project that uses the var.

salisdemos: seriousely? My build builds.

spiess-demos: then lets go with it!

spiess-demos (approved):
lgtm (apart from the non-reproducable-we'll-see-what-happens error).

PR #4109, opened 2024-12-04
chore: Use flat option for esLint

salisdemos:
since esLint 9 they restructured the config file to a "flat" way.
And eslintrc.js is replaced by eslint.config.js

The linter itself runs, but when building the fe-assets, some error occurs.

Since the webpack-hook didn't help us to reduce linting issues we decided to move the linter checks to GH and remove it from webpack. there the new format breaks the build for some reason


hwiem (changes requested):
Thanks for setting this up :heart: 
It works and looks good, but I'd like to keep the rules defined in https://github.com/standard/standard/blob/master/RULES.md. If I remember correctly, stylisticJs can be used for this (see https://github.com/demos-europe/planningcloud-hub/blob/main/eslint.config.js). 

In general I think we should try to have a base eslint config for all projects to keep things consistent (see https://github.com/spiess-demos/eslint-config-demos/blob/main/eslint.config.js),  which of course isn't something to solve in this PR, but maybe to keep in mind.

mussbach:
@salisdemos what should happen with this PR?

salisdemos:
> @salisdemos what should happen with this PR? I tried to fix the reviewdog eslint task and stumbled over this.

Its on my list of PRs that will be focused after the vue3-PR is merged

hwiem:
Replaced by https://github.com/demos-europe/demosplan-core/pull/5102 to have a common remote branch

PR #4110, opened 2024-12-04
Use if condition to access voting and publication area

jalalifardemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12986

Description: Use If condition to access Voting and public area in Statement

- [x] Tests updated/created
- [ ] Update documentation

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMeta.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMeta.vue
@@ -58,7 +58,7 @@
           @save="(data) => save(data)" />
 
         <statement-publication-and-voting
-          v-if="hasPermission('feature_statements_vote')"
+          v-if="hasPermission('feature_statements_vote') || hasPermission('feature_statements_publication')"
           :editable="editable"
           :statement="statement"
           @save="(data) => save(data)"
--- a/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMeta.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMeta.vue
@@ -58,6 +58,7 @@
           @save="(data) => save(data)" />
 
         <statement-publication-and-voting
+          v-if="hasPermission('feature_statements_vote')"
           :editable="editable"
           :statement="statement"
           @save="(data) => save(data)"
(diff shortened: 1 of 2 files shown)


client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMeta.vue
```diff
@@ -58,6 +58,7 @@
           @save="(data) => save(data)" />
 
         <statement-publication-and-voting
+          v-if="hasPermission('feature_statements_vote')"
```
salisdemos, line 61: this one is a bit tricky. We already use the permission inside the component for a part of what happens there. 
The publication should be displayed even if the permission is off.
client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementPublicationAndVoting.vue:31

salisdemos (changes requested):

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMeta.vue
salisdemos: Maybe, we need another permission like "feature_statement_publication"

jalalifardemos: Maybe we can remove the Permission inside the Component and use that in StatementMeta Component, because it shows user this part anyway
`<fieldset data-dp-validate="statementPublicationAndVotingData">
    <legend
      id="publicationAndVoting"
      class="mb-3 color-text-muted font-normal">
      {{ Translator.trans('publication.and.voting') }}
    </legend>
    <div class="font-semibold">
      {{ Translator.trans('publish.on.platform') }}
    </div>`
    before to check permission. Or we can use Permission at the top of the StatementPublicationAndVoting Component in main template

jalalifardemos: Because I think create a new Permission for this thing is not useful when we have already one for it

salisdemos: but the votes and the publication are two different sections as far as I can see. And the permission is already in use. So we can't just move it the break the logic in some projects.

PR #4131, opened 2024-12-10
f DPLAN 13016 use element id to get paragraphs titles

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-13016/HHBOP-Eingeschrankte-Anzahl-an-Originalstellungnahmen-in-BOP-Prod-Verfahren

Description: use element id to get paragraphs titles

### DIFF
--- a/client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
+++ b/client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
@@ -255,8 +255,8 @@ export default {
       if (hasOwnProp(this.statement, 'paragraph')) {
         const elementParagraphs = this.selectedElementParagraph()
 
-        if (elementParagraphs && this.statement.paragraphParentId) {
-          const paragraph = elementParagraphs.find((el) => el.id === this.statement.paragraphParentId)
+        if (elementParagraphs && this.statement.paragraphParentId && this.statement.elementId) {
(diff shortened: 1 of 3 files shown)


client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
```diff
@@ -255,8 +255,8 @@ export default {
       if (hasOwnProp(this.statement, 'paragraph')) {
         const elementParagraphs = this.selectedElementParagraph()
 
-        if (elementParagraphs && this.statement.paragraphParentId) {
-          const paragraph = elementParagraphs.find((el) => el.id === this.statement.paragraphParentId)
+        if (elementParagraphs && this.statement.paragraphParentId && this.statement.elementId) {
```
salisdemos, line 258:
```suggestion
        if (elementParagraphs && this.statement.elementId) {
```

salisdemos (changes requested):

client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
bakridemos: it is tricky here, i thought we don't need it too, but actualy we do. the statement must have a paragraph otherwise we don't have to do this. 'paragraphParentId' is not used to compare here but to make sure that the statement has paragraphs

salisdemos (approved):

PR #4135, opened 2024-12-11
chore: To be aligned with the changes in demosplan-ui.

salisdemos:
In demosplan-ui@0.4.3 we change mention name to mention label. to make it work, we have to change it here too


spiess-demos (approved):
lgtm (code review)

PR #4146, opened 2024-12-13
fix(refs dplan-13039) Add missing attributes for layer

salisdemos:
### Ticket
DPLAN-13039

this is a cherrypick from 01aae4563c06f635ef46215a933520e3159ceb1a

### DIFF
--- a/client/js/components/map/publicdetail/controls/layerlist/DpPublicLayerList.vue
+++ b/client/js/components/map/publicdetail/controls/layerlist/DpPublicLayerList.vue
@@ -23,7 +23,7 @@
         :layer-groups-alternate-visibility="layerGroupsAlternateVisibility" />
       <dp-public-layer-list-category
         v-else
-        :key="`category:${layer.id}?`"
+        :key="`category:${layer.id}`"
         :group="layer"
         :layer-type="layerType"
         :visible="true"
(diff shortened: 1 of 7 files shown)


client/js/components/map/publicdetail/controls/layerlist/DpPublicLayerList.vue
```diff
@@ -23,7 +23,7 @@
         :layer-groups-alternate-visibility="layerGroupsAlternateVisibility" />
       <dp-public-layer-list-category
         v-else
-        :key="layer.id"
+        :key="`category:${layer.id}?`"
```
muellerdemos, line 26: nitpick: do we really need the question mark as a string here?

muellerdemos (approved):
just found one nitpick but not sure if the question mark is really needed or not.

However the diff is the same as in https://github.com/demos-europe/demosplan-core/commit/01aae4563c06f635ef46215a933520e3159ceb1a so this lgtm

client/js/components/map/publicdetail/controls/layerlist/DpPublicLayerList.vue
salisdemos: no. That makes no sense
```suggestion
        :key="`category:${layer.id}`"
```

PR #4149, opened 2024-12-16
Filter query to not show guest only procedures

jalalifardemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12843/BEDatenerfasser-Navigieren-zu-einem-Verfahren-Ein-Fehler-ist-aufgetreten-Seite

Description: User as a data input role has access to some procedures and not some else. It happens because procedure can be set to show guest only or not. If procedure be set to show just for guest, data input user ca not access to that and see an error.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


mussbach (changes requested):
I do not think that this is the right direction to fix this. I think Datenerfasser should be able to access participationGuestOnly procedures

jalalifardemos:
https://github.com/demos-europe/demosplan-core/pull/4811

PR #4151, opened 2024-12-16
chore: Remove unused Code

salisdemos:
- success_message, fail_message are not filled anywhere
- organisationList is not called in this file


bakridemos (approved):
lgtm, there are some projects overrides where it must be removed too

PR #4159, opened 2024-12-17
fix(refs DPLAN-12994): Add Roles to create-user-Dropdown after they are loaded

salisdemos:
### Ticket
DPLAN-12994

Due to an unfair race condition, the roles options in the create user form was always empty.
To resolve this, We have to check the store if the roles "arrives" there to then update the options.

### DIFF
--- a/client/js/components/user/DpUserList/DpUserFormFields.vue
+++ b/client/js/components/user/DpUserList/DpUserFormFields.vue
@@ -266,6 +266,17 @@ export default {
     }
   },
 
+  watch: {
+    '$store.state.Role': {
+      handler () {
+        if (this.allowedRolesForOrga.length === 0) {
+          this.setInitialOrgaData()
+        }
+      },
+      deep: false
+    }
+  },
+
   methods: {
     ...mapActions('Orga', {
       organisationList: 'list'
@@ -331,7 +342,7 @@ export default {
      * - user is not set: all roles
      * - user is set: roles for current organisation
      */
-    getAllowedRolesForOrga () {
+    async getAllowedRolesForOrga () {
       let allowedRoles
 
       if (this.currentUserOrga.id === '') {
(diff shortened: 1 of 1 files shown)


client/js/components/user/DpUserList/DpUserFormFields.vue
```diff
@@ -348,14 +349,17 @@ export default {
...
+          this.setOrga(orga.data.data)
 
-      return allowedRoles
+          if (this.currentUserOrga.id && hasOwnProp(this.organisations[this.currentUserOrga.id].relationships, 'allowedRoles')) {
+            allowedRoles = this.organisations[this.currentUserOrga.id].relationships.allowedRoles.list()
+          }
+        })
+        .finally(() => {
+          return allowedRoles
+        })
```
gruenbergerdemos, line 362: Do we always want to return `allowedRoles` even if the then block does not execute?

client/js/components/user/DpUserList/DpUserFormFields.vue
```diff
@@ -313,7 +314,7 @@ export default {
         return dpApi.get(url)
       }
 
-      return new Promise((resolve, reject) => resolve(true))
+      return new Promise.resolve(resolve => resolve(true))
```
gruenbergerdemos, line 317: Is this correct syntax? I think Promise.resolve expects only a value, not a method

gruenbergerdemos (changes requested):
The roles are now correctly shown and I can select them, but when saving a new user I get this error: 
vue.esm.js:3826 TypeError: Cannot read properties of undefined (reading 'then')
    at VueComponent.save (DpCreateItem.vue?./node_modules/vue-loader/lib/index.js??vue-loader-options!./node_modules/source-map-loader/dist/cjs.js:221:13)

client/js/components/user/DpUserList/DpUserFormFields.vue
gruenbergerdemos: I'm not sure. Does it make sense to return roles in line 350 if the request fails? I think it could be maybe confusing if the user does not see all the roles he is expecting. On the other hand, is it better if we return an empty array in case of error and the user cannot select any roles? I think you can leave it like this.

salisdemos: returning an empty array would lead to an not savable user since role is an required field

gruenbergerdemos:
I still get the same error as mentioned above. Or is this another bug and you want to handle it with a new PR?

salisdemos:
> I still get the same error as mentioned above. Or is this another bug and you want to handle it with a new PR?

I thought it was fixed with my latest changes and I only forget to reassign you. Sorry for bothering.

salisdemos:
this was already fixed elsewhere

PR #4163, opened 2024-12-18
B dplan 13034 use api2 to create user

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-13034

- Move to using AdministratableUserResourceType instead of old 1.0 api
- Important things to check: 
- Updating Roles,Department and Orga relationships
- When a user is created, login field is set same as email field

### How to review/test
1. Create a user and check that emails has been sent (not possible to test in local)
2. Update user all possible combinations: roles, orga, email.
3. when updating email, check that login is also updated

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/client/js/components/user/DpCreateItem.vue
+++ b/client/js/components/user/DpCreateItem.vue
@@ -175,13 +175,6 @@ export default {
   },
 
   computed: {
-    /**
-     * Needed for entity === 'user'
-     */
-    ...mapState('Role', {
-      roles: 'items'
-    }),
-
     dynamicComponent () {
       return this.customComponent[this.entity].componentName
     },
@@ -195,8 +188,9 @@ export default {
     },
 
     itemResource () {
+      const type = this.entity === 'user' ? 'AdministratableUser' : this.entity
       return {
-        type: this.entity,
+        type,
         ...this.item
       }
     }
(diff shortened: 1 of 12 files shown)


client/js/components/user/DpCreateItem.vue
```diff
@@ -195,8 +188,9 @@ export default {
     },
 
     itemResource () {
+      const type = this.entity === 'user' ? 'AdministratableUser' : this.entity
```
hwiem, line 191: I don't know how urgent this PR is, but can't we use `'AdministratableUser'` as the value for `this.entity` instead of `'user'`? Also maybe it makes sense to rename `entity` to `resourceType`? And shouldn't the resourceType for organisation be `'Organisation'`, too? :thinking:

hwiem (changes requested):

client/js/components/user/DpCreateItem.vue
salisdemos: The PR already is way larger then supposed to be. And in my experience its not simple to rename a ResourceType or anything along. I thought about it too and quickly became the feeling of "one not simply.... "

hwiem (approved):
FE lgtm, code review only

bakridemos (approved):
updatable method will be deprecated from next update, this as spoken has to be readjusted later. Otherwise seem good to me.

PR #4186, opened 2025-01-03
f remove unused controller

bakridemos:
Description: Unused controllers should be deleted to improve system performance by reducing resource consumption and eliminating unnecessary clutter. This streamlining makes the system easier to maintain and reduces the risk of software bugs or security vulnerabilities.


bakridemos:
a wrong controller is by mistake deleted here that why this will be closed

PR #4195, opened 2025-01-08
fix(BEAA2-14): add required attribute to the fileupload component

sakutademos:
### Ticket
[BEAA2-14](https://demoseurope.youtrack.cloud/issue/BEAA2-14/Piktogramm-wird-nicht-als-Pflichtfeld-angezeigt)

**Description:** This PR adds a required attribute to the pictogram.

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
@@ -877,7 +877,7 @@
                                 {% if hasPermission('field_procedure_pictogram') and hasPermission('area_public_participation') %}
                                     <div class="u-mb">
                                         <label class="inline-block u-mb-0" for="r_pictogram">
-                                            {{ 'procedure.pictogram'|trans }}
+                                            {{ 'procedure.pictogram'|trans }}*
                                         </label>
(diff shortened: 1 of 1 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -907,7 +907,8 @@
                                                 "img",
                                                 "form.button.upload.file",
                                                 1,
-                                                true
+                                                true,
+                                                required: 'true'
```
hwiem, line 911: Shouldn't it also work to just pass `'true'` (instead of `required: 'true'`)? We don't normally pass the parameter key

sakutademos: yes, I thought it should work that way, but I do not really understand how it knows that I mean true specifically for the required attribute. Is it based on the order of the attributes?

sakutademos: ok, it works this way --> https://github.com/demos-europe/demosplan-core/pull/4195/commits/2683cbe272b9c2415dcb7d11b32a4f67e5b67888

hwiem (approved):
lgtm, code review only

PR #4197, opened 2025-01-08
fix(BEAA2-19): remove label from the pictogram

sakutademos:
### Ticket
https://demoseurope.youtrack.cloud/issue/BEAA2-19/Text-zu-Piktogramm-widerspricht-sich-in-Tooltip-und-in-Hinweistext

**Description:** This PR removes the label from the pictogram.

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
@@ -879,11 +879,6 @@
                                         <label class="inline-block u-mb-0" for="r_pictogram">
                                             {{ 'procedure.pictogram'|trans }}
                                         </label>
-                                        {% include '@DemosPlanCore/Extension/contextual_help.html.twig' with {
-                                            helpText: 'text.procedure.edit.external.pictogram'|trans,
-                                            cssClasses:'mb-1',
-                                            showPlainHint: true
-                                        } %}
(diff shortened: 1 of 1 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -900,10 +900,7 @@
                                         {% else %}
                                             {{  fileupload(
                                                 "r_pictogram",
-                                                '<p class="lbl__hint">'
-                                                ~ "explanation.procedure.pictogram.missing"|trans
-                                                ~ "explanation.procedure.pictogram.dimensions"|trans
-                                                ~ '</p>',
+                                                "hide",
```
hwiem, line 903: Shouldn't the label text be replaced with the tooltip text, and the tooltip be removed? (That's how I understand the ticket)

hwiem (changes requested):

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
sakutademos: you understand it correctly :D https://github.com/demos-europe/demosplan-core/pull/4197/commits/c077ad4cc84c34b6d6e6e0c105984f1619a34427

hwiem (approved):
lgtm :)

PR #4200, opened 2025-01-09
feat(refs DPLAN-12968, #AB21840): set margin outside of component

hwiem:
### Ticket
[DPLAN-12968](https://demoseurope.youtrack.cloud/issue/DPLAN-12968/ADO-Issue-21840-Admin-Institutionsliste-Institutionen-konnen-durchsucht-werden)

Previously, margin was set via the `flash` class on the root `div` in `DpInlineNotification`. When setting dismissible to true and then dismissing the notification in the interface, the hint icon has no margin. Instead of defining it directly in the component, the margin can now simply be set from outside the component and will be applied to the root `div` (whether it's the notification or the icon):

```
  <dp-inline-notification
      classs="mt-3 mb-2"
      dismissible
      :dismissible-key="helpTextDismissibleKey"
      :message="helpText"
      type="info" />
```

I tested only some usages and found that sometimes, the extra margin was not needed. In other cases I added the same margin as was previously set in the component, so things should look as before.

### PR Checklist

- [x] Run `yarn test`
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## UNRELEASED
 
+### Changed
+
+DpInlineNotification: Set margin from outside the component (instead of inside)
+
 ## 2.20.0
 
 - Added: Admin institution list: Institutions can now be tagged filtered by categories.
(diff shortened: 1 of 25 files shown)


salisdemos (approved):
Code looks good. And if the should be ocurences, where the spacing is - against the expectation - wrong, its easy to fix in a follow up. ;-)

PR #4220, opened 2025-01-16
fix: Move Sass-loader settings to the right place

salisdemos:
two issues in one: The implementation option was set in the wrong scope and for whatever reason the modern-compiler option prevented finishing the process. Although we want to have that option enabled and it worked with that quirk for dev. compiling the prod-assets did not work.


mussbach (approved):
it works, thanks a lot

PR #4227, opened 2025-01-17
fix(DPLAN-12125): display layer-type label only if the permission is on

sakutademos:
### Ticket
[DPLAN-12125](https://demoseurope.youtrack.cloud/issue/DPLAN-12125/Kartentyp-kann-beim-Anlegen-eines-GIS-Layers-nicht-ausgewahlt-werden)

**Description:** This PR fixes an issue with the display of the label for the part that should not be shown in projects with disabled permissions:

- move label to the `if` block


hwiem (approved):
lgtm, code review only

PR #4237, opened 2025-01-20
feat(refs DPLAN-12780,#AB21086): Make own component for orga competence

gruenbergerdemos:
### Ticket
DPLAN-12780

Move organisation competence to own component so it does not depend on paper copy permissions.

### How to review/test
Login as Inst-Koordination -> Daten der Organisation -> there should be a new fieldset "organisation competence" and no paper copy fields


client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <fieldset
+    v-if="hasPermission('field_organisation_competence')"
```
hwiem, line 3: I think we should move the permission check outside of the component:

```
<organisation-competence
  v-if="hasPermission('field_organisation_competence')"
  :organisation="organisation" />
```

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
```diff
@@ -0,0 +1,39 @@
...
+    id="organisationCompetence"
+    class="w-3/4">
+    <legend class="font-size-large weight--normal u-mb-0_75">
```
hwiem, line 6: nitpick: Can we use tailwind classes here?

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
```diff
@@ -0,0 +1,39 @@
...
+      {{ Translator.trans('competence') }}
+    </legend>
+    <div
+      class="w-full mb-3">
+      <dp-text-area
+        id="orga_competence"
```
hwiem, line 12: nitpick: sort attributes alphabetically

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <fieldset
+    v-if="hasPermission('field_organisation_competence')"
+    id="organisationCompetence"
+    class="w-3/4">
+    <legend class="font-size-large weight--normal u-mb-0_75">
+      {{ Translator.trans('competence') }}
+    </legend>
+    <div
+      class="w-full mb-3">
```
hwiem, line 10: Did you run `yarn lint`? I don't know if it's the github interface, but it seems there are a couple of formatting issues in this file

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
```diff
@@ -0,0 +1,39 @@
...
+      <dp-text-area
+        id="orga_competence"
+        data-cy="organisationData:paperCopy:competence"
+        :name="`${organisation.id}:competence`"
+        :value="organisation.competence"
```
hwiem, line 15: I know it was like this before, but shouldn't we use v-model here? :thinking:

hwiem (changes requested):

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
gruenbergerdemos: Why would you prefer it this way?

hwiem: If we check the permission outside, the component is not rendered at all if the permission is `false`. We could even import it dynamically based on the permission. In that case, the component will not be imported or initialized if the permission is `false`.

gruenbergerdemos: True. But I also like to have it in the component, because then I don't have to worry about permission checks when importing the component (and can't forget it). Done in [2aa9cdc](https://github.com/demos-europe/demosplan-core/pull/4237/commits/2aa9cdc387a0acf1c5da018a1b382877577924f6)

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
gruenbergerdemos: I think it doesn't matter in this case, but I changed it here [785572d](https://github.com/demos-europe/demosplan-core/pull/4237/commits/785572dcbbcbd10c664361ed3411ab48bbd8de84)

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
gruenbergerdemos: Thanks, done in [785572d](https://github.com/demos-europe/demosplan-core/pull/4237/commits/785572dcbbcbd10c664361ed3411ab48bbd8de84)

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
gruenbergerdemos: I switched them where it was possible [785572d](https://github.com/demos-europe/demosplan-core/pull/4237/commits/785572dcbbcbd10c664361ed3411ab48bbd8de84)

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
gruenbergerdemos: Can't figure out how to run it on just one file atm, but it looks fine to me (I changed only the div) here [785572d](https://github.com/demos-europe/demosplan-core/pull/4237/commits/785572dcbbcbd10c664361ed3411ab48bbd8de84)

client/js/components/user/orgaDataEntry/OrganisationCompetence.vue
hwiem: That's also a good point. Maybe we should discuss it in the FE team?

hwiem (approved):
lgtm!

PR #4246, opened 2025-01-22
Fix(DPLAN-14776): formatting issue after the spit statement view - release_ewm

sakutademos:
### Ticket
[DPLAN-14776](https://demoseurope.youtrack.cloud/issue/DPLAN-14776/Nach-Aufteilen-von-eine-Stellungnahme-ist-die-Formatierung-von-Text-komplett-weg)

Description: This PR fixes an issue with text formatting after splitting the statement view. The problem was that after extending `schema.spec.marks`, the standard marks from ProseMirror were overwritten. So found a way to extend it without overwriting the default marks and to added an underline mark as well.

### How to review/test
EIne STN mit Formatierung erstellen -> Aufteilen -> Abschnitte durch checken

### DIFF
--- a/client/js/components/statement/splitStatement/SegmentationEditor.vue
+++ b/client/js/components/statement/splitStatement/SegmentationEditor.vue
@@ -52,37 +52,53 @@ export default {
 
   data () {
     return {
+      customMarks: {
+        underline: {
+          parseDOM: [{ tag: 'u' }],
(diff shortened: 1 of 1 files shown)


salisdemos (approved):
Thanks!

salisdemos (approved):
nice. Thanks!

PR #4250, opened 2025-01-23
f DPLAN-14775 use the right translation key

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-14775

Description: use the right translation key

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
@@ -400,7 +400,7 @@
                                         }]) %}
                                     {% endfor %}
                                     <participation-phases
-                                        autoswitch-hint="{{ 'period.autoswitch.hint'|trans({ phase: 'procedure.phases.internal.evaluating'|trans }) }}"
+                                        autoswitch-hint="{{ 'period.autoswitch.hint'|trans({ phase: 'procedure.phases.analysis'|trans }) }}"
                                         data-cy="internalPhase"
                                         field-name="r_phase"
                                         help-text="{{ 'text.procedure.edit.internal.change_phase'|trans }}"


templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -400,7 +400,7 @@
                                         }]) %}
                                     {% endfor %}
                                     <participation-phases
-                                        autoswitch-hint="{{ 'period.autoswitch.hint'|trans({ phase: 'procedure.phases.internal.evaluating'|trans }) }}"
+                                        autoswitch-hint="{{ 'period.autoswitch.hint'|trans({ phase: 'procedure.phases.analysis'|trans }) }}"
```
mussbach, line 403: This phase key only exists in one distinct project. You need to find a phase that always exists or override this for projects where the key differs

mussbach (changes requested):

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
bakridemos: we don't have and use the same keys in every project. The value is different depending on the project too. Should i use one key from other project which has the same value ???

mussbach: This is hard to fix properly then. I fear that we have to find out which phase is participationstate: Procedure::PARTICIPATIONSTATE_FINISHED and pass this to the twig, so that the phase key is used as a variable in the transkey variable

bakridemos:
this pr will be closed and the issue will be handled in other pr https://github.com/demos-europe/demosplan-core/pull/4480

PR #4256, opened 2025-01-23
refacor(dplan-12013): topical tags - add new form section

sakutademos:


client/js/components/tags/TagsListForm.vue
salisdemos, line 1: I suggest to name the file TagsCreateForm or something like that

client/js/components/tags/TagsListForm.vue
```diff
@@ -0,0 +1,262 @@
...
+        topic: ''
+      },
+      newTopic: {
+        title: ''
+      },
+      // This is necessary to allow extending the Tags-Resource
+      tagAttributes: {
+        boilerplate: '',
+        title: ''
+      }
```
salisdemos, line 140: actually. This is not necessary, because wee don't extend the Tag-resource with addons. Addons hold their own reourceTypes which relate to the Tag-resourcetype in core. and boilerplate is an relationship anyway. I know, this is just copy-past from my stuff, but I think we should take the chance and clean it up.

client/js/components/tags/TagsListForm.vue
```diff
@@ -0,0 +1,262 @@
...
+                  id: newTagId
+                })
+              }
+            }
+          })
+
+          this.$root.$emit('tag:created', newTagId)
+          // reset form data
+          this.closeForm()
+          this.newTag = this.tagAttributes
```
salisdemos, line 232: this is done in `closeForm` inherently

client/js/components/tags/TagsList.vue
```diff
@@ -201,16 +121,6 @@ export default {
       return Object.keys(this.tagAttributes)
```
salisdemos, line 121: I wonder if we want to keep it like this, or just type the two keys in the Request "hard coded"

translations/messages+intl-icu.de.yml
```diff
@@ -3742,7 +3742,6 @@ tag.name.empty.error: "Das Schlagwort muss einen Namen haben."
 tag.topic.name.empty.error: "Das Thema muss einen Namen haben."
 tag.topic.name.duplicate.error: "Das Thema existiert bereits. Bitte verwenden Sie einen anderen Namen."
 tag.new: "Neues Schlagwort"
```
salisdemos, line 3744: shouldn't we refactor this and several other keys to `entity.create` too? Maybe that would be a good onboarding-task :thinking:

salisdemos (changes requested):
Thanks for cleaning my stuff. Looks good! Just some small things.

client/js/components/tags/TagsList.vue
sakutademos: hm, no i think its unnecessarily and could be refactored: https://github.com/demos-europe/demosplan-core/pull/4256/commits/a66f487c978895bde9aaa4336418b9f8781d59ec. would this be ok?

client/js/components/tags/TagsListForm.vue
sakutademos: https://github.com/demos-europe/demosplan-core/pull/4256/commits/a66f487c978895bde9aaa4336418b9f8781d59ec

client/js/components/tags/TagsListForm.vue
sakutademos: yes, i did not refactor this part, because i was not sure how you use it. done here https://github.com/demos-europe/demosplan-core/pull/4256/commits/a66f487c978895bde9aaa4336418b9f8781d59ec

client/js/components/tags/TagsListForm.vue
sakutademos: right, done here https://github.com/demos-europe/demosplan-core/pull/4256/commits/3c2b39eadff2eb6a9bfcae647d2e98e7c611bb2b. Use TagsCreateForm for better clarity instead of the simple TagsForm

salisdemos (approved):
looks good. Thank You

PR #4221, opened 2025-01-16
fix: display messagebag from vuex-json-api

salisdemos:
A long time, we wondered, why the messages from the Backend where not displayed when we used the vuex-json-api. 
We thought something in the lib swallowed the messages.
Now I found a way by co incident to display all the fine informations provided by the ressourceTypes.

PR #4275, opened 2025-01-28
fix(refs DPLAN-14848): check content type

hwiem:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-14848/Speichern-von-Institution-Tags-geht-nicht

Without checking the content type, this error occurred:
```
SyntaxError: Failed to execute 'json' on 'Response': Unexpected end of JSON input
```
It seems that this happens when the response body is empty and no content type is set.

### Linked PRs 

- https://github.com/demos-europe/demosplan-core/pull/4221

### PR Checklist

### DIFF
--- a/client/js/store/core/initStore.js
+++ b/client/js/store/core/initStore.js
@@ -109,6 +109,8 @@ function initStore (storeModules, apiStoreModules, presetStoreModules) {
 
                   return Promise.reject(response)
                 }
+
+                return Promise.reject(error)
               }
             ]
           }),
--- a/client/js/store/core/initStore.js
+++ b/client/js/store/core/initStore.js
@@ -72,16 +72,22 @@ function initStore (storeModules, apiStoreModules, presetStoreModules) {
             },
             successCallbacks: [
               async (success) => {
-                const response = await success.json()
+                // If the response body is empty, contentType will be null
+                const contentType = success.headers.get('Content-Type')
 
-                const meta = response.data?.meta
-                  ? response.data.meta
-                  : response.meta || null
-                if (meta?.messages) {
-                  handleResponseMessages(meta)
+                if (contentType && contentType.includes('application/json')) {
+                  const response = await success.json()
+
+                  const meta = response.data?.meta
+                    ? response.data.meta
+                    : response.meta || null
+                  if (meta?.messages) {
+                    handleResponseMessages(meta)
+                  }
+
+                  return Promise.resolve(response)
                 }
-
-                return Promise.resolve(response)
+                return Promise.resolve(success)
               }
             ],
             errorCallbacks: [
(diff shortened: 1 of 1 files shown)


client/js/store/core/initStore.js
```diff
@@ -72,16 +72,22 @@ function initStore (storeModules, apiStoreModules, presetStoreModules) {
...
+                  const meta = response.data?.meta
+                    ? response.data.meta
+                    : response.meta || null
+                  if (meta?.messages) {
+                    handleResponseMessages(meta)
+                  }
+
+                  return Promise.resolve(response)
+                }
+                return Promise.resolve(success)
```
salisdemos, line 90:
```suggestion
                }

                return Promise.resolve(success)
```

client/js/store/core/initStore.js
```diff
@@ -72,16 +72,22 @@ function initStore (storeModules, apiStoreModules, presetStoreModules) {
             },
             successCallbacks: [
               async (success) => {
-                const response = await success.json()
+                // If the response body is empty, contentType will be null
+                const contentType = success.headers.get('Content-Type')
```
salisdemos, line 76: i can assume, that "success" can also be an already resolved json object..
Could we implement that case easily too? Or should we do that in a separat PR

client/js/store/core/initStore.js
```diff
@@ -72,16 +72,22 @@ function initStore (storeModules, apiStoreModules, presetStoreModules) {
             },
             successCallbacks: [
               async (success) => {
-                const response = await success.json()
+                // If the response body is empty, contentType will be null
```
salisdemos, line 75: please adapt these changes for the error-case too.

salisdemos (changes requested):

client/js/store/core/initStore.js
hwiem: Done in d5c798cda19c1efa26c0d832acb88cf257fbdd8e

client/js/store/core/initStore.js
hwiem: My main worry here is that I wouldn't be able to test the code, it might affect a lot of places in the application and we have code freeze today - so I'd prefer to do that in a separate PR.

client/js/store/core/initStore.js
hwiem: Done in d1470c717c28094fabe8fcc0b6e7fa0cc47f3e82

client/js/store/core/initStore.js
```diff
@@ -72,31 +72,43 @@ function initStore (storeModules, apiStoreModules, presetStoreModules) {
...
+                    ? response.data.meta
+                    : response.meta || null
 
-                return Promise.reject(response)
+                  if (meta?.messages) {
+                    handleResponseMessages(meta)
+                  }
+
+                  return Promise.reject(response)
+                }
```
salisdemos, line 111: is the `return Promise.reject(error)` missing on purpose?

salisdemos (changes requested):

salisdemos (approved):
Thanks for catching ;-)

PR #4280, opened 2025-01-29
fix(refs DPLAN-12498): check for length of object keys

hwiem:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-12498/Nach-zeichnen-von-Geltungsbereich-der-Text-andert-sich-nicht-zu-Geltungsbereich-ist-gezeichnet

Since `features` is an object, checking for its length did not work

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/document/DpMapSettingsPreview.vue
+++ b/client/js/components/document/DpMapSettingsPreview.vue
@@ -387,7 +387,8 @@ export default {
 
     isNotEmptyFeatureCollection (string) {
       try {
-        return Object.keys(JSON.parse(string).features).length > 0
+        const parsedFeatures = JSON.parse(string).features
+        return Array.isArray(parsedFeatures) ? parsedFeatures.length > 0 : Object.keys(JSON.parse(string).features).length > 0
       } catch (e) {
         return true
       }
--- a/client/js/components/document/DpMapSettingsPreview.vue
+++ b/client/js/components/document/DpMapSettingsPreview.vue
@@ -388,7 +388,8 @@ export default {
     isNotEmptyFeatureCollection (string) {
       try {
         const parsedFeatures = JSON.parse(string).features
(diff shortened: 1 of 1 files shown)


client/js/components/document/DpMapSettingsPreview.vue
```diff
@@ -387,7 +387,7 @@ export default {
 
     isNotEmptyFeatureCollection (string) {
       try {
-        return JSON.parse(string).features.length > 0
+        return Object.keys(JSON.parse(string).features).length > 0
```
salisdemos, line 390: is features always an object? I have in mind that sometimes it is also an array and we have not yet been able to find out where this comes from.
```suggestion
        const parsedFeatures = JSON.parse(string).features
        return Array.isArray(parsedFeatures) ? parsedFeatures.length > 0 : Object.keys(JSON.parse(string).features).length > 0
```

client/js/components/document/DpMapSettingsPreview.vue
```diff
@@ -387,7 +387,8 @@ export default {
 
     isNotEmptyFeatureCollection (string) {
       try {
-        return JSON.parse(string).features.length > 0
+        const parsedFeatures = JSON.parse(string).features
+        return Array.isArray(parsedFeatures) ? parsedFeatures.length > 0 : Object.keys(JSON.parse(string).features).length > 0
```
salisdemos, line 391: sorry, was not clean
```suggestion
        const parsedFeatures = JSON.parse(string).features

        return Array.isArray(parsedFeatures) ? parsedFeatures.length > 0 : Object.keys(parsedFeatures).length > 0
```

salisdemos (changes requested):

salisdemos (approved):
Thanks! looks good

PR #4281, opened 2025-01-29
refac(DPLAN-12013): some frontend adjustements

sakutademos:
**Description:** 
- to avoid warnings when updating the resource type adjust `successCallbacks` in the `initStore`
- reload the tags and topics when deleting a tag
- adjust the layout for the sortable list: reduce padding-bottom to 18px to prevent breaking the layout. Additional PR in UI
- remove the unused computed value topicsAsOptions, as it is already present in another component

### DIFF
--- a/client/js/components/tags/TagsList.vue
+++ b/client/js/components/tags/TagsList.vue
@@ -7,6 +7,7 @@
     <dp-tree-list
       v-if="transformedCategories"
       class="mb-4"
+      align-toggle="center"
       :tree-data="transformedCategories"
       :options="{
         branchesSelectable: false,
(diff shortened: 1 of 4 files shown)


client/js/components/tags/TagsList.vue
```diff
@@ -235,8 +225,9 @@ export default {
     deleteItem (item) {
       dpRpc('bulk.delete.tags.and.topics', { ids: [item] })
         .then(checkResponse)
-        .then(response => {
-          console.log(response, 'response delete')
+        .then(() => {
+          this.dataIsRequested = false
```
salisdemos, line 229: i would remove  this here again, but add it in line 257 in a new `then()`. Otherwise we undermine the mechanism to prevent the race condition.

client/js/store/core/initStore.js
```diff
@@ -73,12 +73,19 @@ function initStore (storeModules, apiStoreModules, presetStoreModules) {
...
+
+                if (contentType && contentType.includes('application/json')) {
+                  try {
+                    response = await success.json()
+                  } catch (e) {
+                    console.warn('Could not parse response. It might have already been parsed.', e);
+                    response = success
+                  }
+                } else {
+                  response = success
```
salisdemos, line 88: this should already be there after getting latest main.

salisdemos (changes requested):

salisdemos (approved):
Thanks

PR #4286, opened 2025-01-30
refac(DPLAN-12013): issue with displaying messages from the msgBag

sakutademos:
Description: There was a problem with displaying messages/errors after merging with the main branch.

- error messages are not being displayed: there was an issue with handling callbacks in initStore: the contentType can be not only `application/json` but also other strings like `application/vnd.api+json`. Therefore, it must be checked if the type contains `'json'` and not just if it equals `'application/json'`
- some code style adjustment


salisdemos (approved):
true

PR #4293, opened 2025-01-31
style: fix linting errors

hwiem:
This PR contains both auto- and manual fixes for eslint errors.

before: ✖ 426 problems (130 errors, 296 warnings)
after: ✖ 324 problems (29 errors, 295 warnings)

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`


client/js/components/user/orgaDataEntry/EmailNotificationSettings.vue
```diff
@@ -118,7 +118,7 @@ export default {
 
   computed: {
     hasNotificationSectionPermission () {
-      return this.willReceiveNewStatementNotification && hasPermission('feature_notification_statement_new') ||
+      return (this.willReceiveNewStatementNotification && hasPermission('feature_notification_statement_new')) ||
```
gruenbergerdemos, line 121: Do we use this for better readability?

gruenbergerdemos (approved):
lgtm, thanks for fixing :)

client/js/components/user/orgaDataEntry/EmailNotificationSettings.vue
hwiem: Yes, eslint was complaining that we're mixing `&&` and `||` and that we should add parentheses so it's easer to understand

PR #4297, opened 2025-02-03
F dplan 14866 export synopsis without personal data

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-14866/EWM-RVMO-Personenbezogene-Daten-aus-Synopse-entfernen

Description: 

FE : 
- add new export type docxCensored.
- add censorParameter : it will take the value of the new parameter censor depends on the export type and false if it not existing.
BE :
- add the possibility to export synopsis without personal data. The new parameter censorParameter will allow to check if the export can include orga and user infos or not.

- [X] Update changelog 
- [X] Update tests


client/js/components/statement/StatementExportModal.vue
```diff
@@ -28,23 +28,25 @@
...
-            :label="{
-              hint: active === key ? exportType.hint : '',
-              text: Translator.trans(exportType.label)
-            }"
-            :value="key"
-            :checked="active === key"
-            @change="active = key" />
+          <div v-for="(typeDetail, typeKey) in exportTypes" :key="typeKey">
+              <dp-radio
+                v-for="(subTypeDetail, subTypeKey) in typeDetail"
```
salisdemos, line 33: Why do You nest the loop?
I think it would be better readable, if we flatten the Object

client/js/components/statement/StatementExportModal.vue
```diff
@@ -210,6 +232,11 @@ export default {
   },
 
   methods: {
+    getActiveTypeDetail() {
+      const [typeKey, subTypeKey] = this.active.split('-');
+
+      return this.exportTypes[typeKey][subTypeKey];
+    },
```
salisdemos, line 239: if we flaten the object, we can remove this ;-)

salisdemos (changes requested):

client/js/components/statement/StatementExportModal.vue
bakridemos: i changed the exportTypes structure, every export type can have 1 or many subtypes ( in our case two subtype normal and censored ) and to do that i think i have to use a nested loop. I need it to access the subtypes too to combine type and subtype while setting the active state too.

client/js/components/statement/StatementExportModal.vue
bakridemos: Since the exportTypes stuctrure changed the active state will reflect the currently selected radio button, uniquely identifying them by combining type and sub-type where necessary and not anymore only type. To access the subType parameters i needed to do it like this.

client/js/components/statement/StatementExportModal.vue
salisdemos: I thought about something like this:
```
exportTypes: {
  'docx-normal': {...},
  'docx-censored': { ...},
  'zip-nomal': {...},
  'zip-censored': {...},
  'xlsx-normal': {}
}

```

bakridemos: we can do it this way too, if we don't want adjust the div but we ll get something like 
image
 the way i did it is a little bit complex but it will be like this 
image
which its better in my opinion

bakridemos: i adjut it anyway and used the 
exportTypes: {
  'docx-normal': {...},
  'docx-censored': { ...},
  'zip-nomal': {...},
  'zip-censored': {...},
  'xlsx-normal': {}
},

sakutademos: i add some styling adjustments here: https://github.com/demos-europe/demosplan-core/pull/4297/commits/888174c67b43be21378f1c6e91166f505f2d9cc9 to place the anonymize inputs below

client/js/components/statement/StatementExportModal.vue
```diff
@@ -156,23 +155,37 @@ export default {
         }
       },
       exportTypes: {
-        docx: {
+        docx_normal: {
           label: 'export.docx',
           hint: '',
           exportPath: 'dplan_statement_segments_export',
-          dataCy: 'exportModal:export:docx'
+          dataCy: 'exportModal:export:docx',
```
sakutademos, line 162:
```suggestion
          dataCy: 'exportModal:export:docx'
```

sakutademos (approved):
FE lgtm, just tiny nitpick

MoritzMandler (approved):
lgtm - tested and works fine :)

PR #4302, opened 2025-02-04
feat(refs DPLAN-12969,#AB21841): Fix tags in filters 

gruenbergerdemos:
### Ticket
DPLAN-12969

We need to call the institution tag categories again after applying a filter to get all available tags, because the filter request also updates the institution tag category store.

### How to review/test
Login as Mandatenadmin -> Institutionen verschlagworten -> Apply filter -> Check other categories if there are still filter options (tags) available -> Click Zurücksetzen and check if filter options are still valid

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
+++ b/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
@@ -83,7 +83,7 @@
             </ul>
           </template>
           <template
-            v-for="(category, idx) in institutionTagCategories"
+            v-for="(category, idx) in institutionTagCategoriesCopy"
             v-slot:[category.attributes.name]="institution">
             <dp-multiselect
               v-if="institution.edit"
@@ -192,6 +192,12 @@ export default {
     initialFilter: {
       type: [Object, Array],
       default: () => ({})
+    },
+
+    isActive: {
+      type: Boolean,
+      required: false,
+      default: false
     }
   },
 
@@ -203,6 +209,7 @@ export default {
       editingInstitution: null,
       editingInstitutionTags: {},
       initialSelection: [],
+      institutionTagCategoriesCopy: {},
       isLoading: true,
       searchTerm: ''
     }
@@ -308,7 +315,7 @@ export default {
     },
 
     institutionTagCategoriesValues () {
-      return Object.values(this.institutionTagCategories)
+      return Object.values(this.institutionTagCategoriesCopy)
     },
 
     selectableColumns () {
@@ -325,6 +332,14 @@ export default {
           category: relationships?.category?.data
         }
       })
+    }
+  },
+
+  watch: {
+    isActive (newValue) {
+      if (newValue) {
+        this.getInstitutionTagCategories()
+      }
     }
   },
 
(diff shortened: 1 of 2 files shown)


client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
```diff
@@ -502,13 +503,18 @@ export default {
           if (categoryId) {
             this.setIsFilterFlyoutLoading({ categoryId, isLoading: false })
           }
+          // We need to call getInstitutionTagCategories again to get all available tags for the filter,
+          // because fetchInvitableInstitution will update the institutionTagCategory store with the filtered tags
+          if (!isInitial) {
+            this.getInstitutionTagCategories()
```
hwiem, line 509: I wonder if there isn't a way of avoiding requesting the categories again here, because categories and tags don't really change, so if we could store them in a local variable after fetching them initially, we wouldn't have to request them again? Maybe we don't really need the `InstitutionTagCategory` store module here?
Only when creating tags and categories in the "Schlagworte verwalten" tab and switching to the "Institutionen gruppieren" tab without reloading the page we would have to request tags and categories again.
What do you think?

hwiem (commented):
Nice that you figured out what the issue is :tada: :)

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
hwiem: We could do that, but on `@change="setActiveTabId"` we could also check the tab id and only update the tag categories if it's 'institutionList'. And maybe we could even add a custom store module 'InstitutionTagManagement' (e.g., maybe there's a better name) that holds a boolean 'needToUpdate' (or something more specific) which we can set to `true` in TagList.vue, then check on `@change="setActiveTabId"` in InstitutionTagManagement.vue?

gruenbergerdemos: I decided to leave the store and use a copy of the institutionTagCategories instead, because it is way easier to use the store object (with all the required relationships and so on) instead of having to create it. 
To detect the switch I added an isActive prop for the institution list and watch it. I think it is ok to get the tag categories after every switch to institution list, because normally the user will only switch to Schlagworte verwalten if he wants to edit something there, I think. 
What do you think about these changes? 
Done in [32c73fb](https://github.com/demos-europe/demosplan-core/pull/4302/commits/32c73fb1c6817e1f286d8410122f4047d93c1d5e)

hwiem: I think the `isActive` prop is a nice idea. :+1: I agree that it's ok to request the tag categories on every tab switch to the institution list as it shouldn't happen so often.
I tested the changes and they work. :tada: However when applying filter options from different dropdowns, I do not get the expected results, but I don't know if that's related to your changes (I suspect the request still needs adjustments)

hwiem: I checked, it should be an issue with the request parameters (so unrelated to your changes)

hwiem (approved):
lgtm + works!

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
gruenbergerdemos: > However when applying filter options from different dropdowns, I do not get the expected results

Yeah, I also saw this, but thought we could probably solve this with a new PR

PR #4303, opened 2025-02-04
fix(refs DPLAN-14901): Catch mapSettings-Request

salisdemos:
In some circumstances - not understood by now - the request fails to load the map settings in the admin area. If so, no map is displayed. To prevent error on the side.
This doesn't fix the issue, that the map is not displayed when the request fails. Without informations, we can't display the map.
And hence we get an 404 from the Backend

ado: 22818

### Ticket
DPLAN-14901
https://www.dev.diplanung.de/DefaultCollection/EfA%20DiPlanung/_workitems/edit/22818

### DIFF
--- a/client/js/store/map/ProcedureMapSettings.js
+++ b/client/js/store/map/ProcedureMapSettings.js
@@ -94,7 +94,7 @@ export default {
             const procedureMapSettings = {
               attributes: {
                 availableScales: data.availableScales.map(scale => ({ label: `1:${scale.toLocaleString('de-DE')}`, value: scale })) ?? [],
-                coordinate: convertExtentToFlatArray(data.coordinate) ?? '',
+                coordinate: convertExtentToFlatArray(data.coordinate) ?? [],
                 copyright: data.copyright ?? '',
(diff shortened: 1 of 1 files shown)


client/js/store/map/ProcedureMapSettings.js
```diff
@@ -111,8 +111,12 @@ export default {
             }
 
             commit('setItem', { key: 'procedureMapSettings', value: procedureMapSettings })
+
             return procedureMapSettings
           })
+          .catch(() => {
+            dplan.notify.error(Translator.trans('error.api.generic'))
+          })
```
gruenbergerdemos, line 119: I think it does not catch. I could reproduce the map not showing, but it didn't show any error messages and the breakpoint did not stop (instead it went into the response part).

salisdemos: Then You got a different error. That might be. But since I can't reproduce any, I try to minimize the effects of everything that could happen. And I have been reported, that the request fails "sometimes"

gruenbergerdemos (approved):

PR #4305, opened 2025-02-05
chore(refs DPLAN-12697): Slicing Tagging new exchange format

salisdemos:
This is a Draft to exchange the index based exchange format.

More infos to be defined

**Ticket**
DPLAN-12697


mussbach:
Was solved differently

PR #4310, opened 2025-02-06
fix(refs DPLAN-14736): reduce number of requests

hwiem:
### Ticket
[DPLAN-14736](https://demoseurope.youtrack.cloud/issue/DPLAN-14736/Orga-Filter-funktioniert-nicht-richtig)

- in the mein berlin addon additional field, after requesting the dropdown options once, there is not really any need to request them again without reloading the page
- when filtering the organisation list, the list items are destroyed and mounted again, so the request was fired 10 times (for each list item once) with each click on a filter checkbox, leading to a 'too many requests' error when changing the filters a lot
- now, after requesting the dropdown options in the addon component once, they are emitted to the organisation list component which is not re-mounted when filtering. the organisation list component then passes them down via prop back to the addon component. if available, the request is not fired again. that way, we still have 10 requests initially (which is not ideal, and we might want to find a solution for that), but no more requests when filtering

### How to review/test

It may take a while to get the addon running and test the changes locally, so I recommend only doing a code review. 
My main question is: Do you think this approach is fine from an architectural point of view or could there be a better solution?

### Linked PRs

- [x] https://github.com/demos-europe/demosplan-addon-mein-berlin/pull/28

### Tasks
- [x] Release new version of demosplan-addon-mein-berlin
- [x] Bump version of demosplan-addon-mein-berlin in demosplan-core

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/user/DpOrganisationList/DpOrganisationList.vue
+++ b/client/js/components/user/DpOrganisationList/DpOrganisationList.vue
@@ -31,14 +31,15 @@
           class="o-list o-list--card u-mb"
           data-cy="pendingOrganisationList">
           <dp-organisation-list-item
-            class="o-list__item"
             v-for="(item, idx) in pendingOrgs"
             :key="`pendingOrganisation:${idx}`"
             :available-orga-types="availableOrgaTypes"
+            class="o-list__item"
+            data-cy="pendingOrganisationListBlk"
+            module-name="Pending"
             :organisation="item"
             :selectable="false"
-            module-name="Pending"
-            data-cy="pendingOrganisationListBlk" />
+            @addonOptions:loaded="setAdditionalFieldOptions" />
         </ul>
         <dp-sliding-pagination
           v-if="pendingOrganisationsTotalPages > 1"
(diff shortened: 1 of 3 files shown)


client/js/components/user/DpOrganisationList/DpOrganisationList.vue
```diff
@@ -31,14 +31,15 @@
           class="o-list o-list--card u-mb"
           data-cy="pendingOrganisationList">
           <dp-organisation-list-item
```
gruenbergerdemos, line 33: Not sure why we have to differente list items. Do we need the prop `:additional-field-options="additionalFieldOptions"` here, too?

gruenbergerdemos (commented):
Just one question, otherwise code lgtm. I also don't have another idea atm.

client/js/components/user/DpOrganisationList/DpOrganisationList.vue
hwiem: I'm not sure if it's needed either, but it seemed safer to add it, and I don't think it does any harm.

gruenbergerdemos: Did you push the changes? I don't see it added

hwiem: Ah, I think I misunderstood. Your concern was that it _should_ be added? I remember adding it, but you're right, it's definitely not there. :smile: I'll add it

gruenbergerdemos (approved):

PR #4313, opened 2025-02-07
bug(refs DPLAN-12481): news edit leads to line break

riechedemos:
### Ticket
DPLAN-12481

Description:  fixed line break issue by refactoring twig to DpCheckbox.vue

### How to review/test
Edit an existing news

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanNews/news_admin_edit.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanNews/news_admin_edit.html.twig
@@ -150,27 +150,23 @@
                 }) }}
 
             {% else %}
-                {{ uiComponent('form.row', {
-                    elements: [
-                        {
-                            label: { text: 'citizen.for'|trans },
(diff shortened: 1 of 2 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanNews/news_admin_edit.html.twig
```diff
@@ -150,27 +150,23 @@
...
-                            elementSize: 'medium'
-                        }
-                    ],
-                    heading: 'visible',
-                    required: true,
-                    isFieldset: true
-                }) }}
+                <fieldset>
+                    <legend class="font-semibold mb-1">{{ 'visible'|trans }}&#42</legend>
+                    <dp-checkbox
```
hwiem, line 155: This is my bad, because I suggested to use DpCheckbox, but now I realize that we should use DpCheckboxGroup, which contains exactly the markup you used here (fieldset + legend) :see_no_evil:

hwiem (changes requested):

hwiem (approved):
Since we found out that `DpCheckboxGroup` can only be used in vue context: code lgtm! Thanks for your patience. :)

PR #4319, opened 2025-02-10
bug(refs DPLAN-12402): Remove redundant error message

gruenbergerdemos:
### Ticket
DPLAN-12402

We don't need the error notification in the else block, because data-dp-validate in the form already throws an error if it is not validated. With the else block there are two error notifications shown. 

### How to review/test
Verfahren -> Neues Verfahren erstellen -> Leave the Stellungnahmetext empty -> There should be only one error notification

### PR Checklist

- [x] Run `yarn test`


hwiem (approved):
lgtm, code review only

PR #4321, opened 2025-02-10
feat (refs DPLAN-14844): fix: Bug in case of unset parameter

martendemos:
### Ticket
DPLAN-14844

fix for case of empty array / no external_links defined

https://github.com/demos-europe/demosplan-project-diplanbau/pull/278
https://github.com/demos-europe/demosplan-project-diplanfest/pull/122
https://github.com/demos-europe/demosplan-project-diplanrog/pull/167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - **Patch Version**: Incremented for bug fixes.
 
 ## UNRELEASED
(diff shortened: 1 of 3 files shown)


martendemos:
Needs to be done on release, instead of main.
https://github.com/demos-europe/demosplan-core/pull/4322

PR #4322, opened 2025-02-10
B dplan 14844 fix

martendemos:
### Ticket
DPLAN-14844

fix for case of empty array / no external_links defined

https://github.com/demos-europe/demosplan-project-diplanbau/pull/278
https://github.com/demos-europe/demosplan-project-diplanfest/pull/122
https://github.com/demos-europe/demosplan-project-diplanrog/pull/167

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/DefaultTwigVariablesService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/DefaultTwigVariablesService.php
@@ -188,7 +188,7 @@ class DefaultTwigVariablesService
         $externalLinks = $this->globalConfig->getExternalLinks();
 
         // simple list of urls are given, without intention to filter
-        if (!is_array(array_values($externalLinks)[0])) {
+        if (empty($externalLinks) || !is_array(array_values($externalLinks)[0])) {
             return $externalLinks;
         }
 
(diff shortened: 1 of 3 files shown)


demosplan/DemosPlanCoreBundle/Logic/DefaultTwigVariablesService.php
```diff
@@ -188,7 +188,7 @@ private function getFilteredExternalLinks(): array
         $externalLinks = $this->globalConfig->getExternalLinks();
 
         // simple list of urls are given, without intention to filter
-        if (!is_array(array_values($externalLinks)[0])) {
+        if (empty($externalLinks) || !is_array(array_values($externalLinks)[0])) {
```
mussbach, line 191: as getExternalLinks() always returns an array it would be more clear when 0 === count($externalLinks) would be used, but empty is ok here

mussbach (approved):
lgtm, code review only

demosplan/DemosPlanCoreBundle/Logic/DefaultTwigVariablesService.php
martendemos: For me `empty()` is more readable than `0 === count()`, thats why I prefer it.
Do we have commited us to use `0 === count()` prefered to `empty()`?

PR #4333, opened 2025-02-13
fix(refs DPLAN-14939): set initial column width

hwiem:
### Ticket
[DPLAN-14939](https://demoseurope.youtrack.cloud/issue/DPLAN-14939/Institutionsliste-Layout-Spalten-bleiben-auf-rechts)

- when initially loading the page, column widths should have reasonable values and for some reason, this was not the case without setting an initial column width - the first column was very wide, while the others were way too narrow
- now, all category columns have the same width initially and can then be adjusted as needed by the user

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


hwiem:
Was solved in https://github.com/demos-europe/demosplan-core/pull/6227

PR #4334, opened 2025-02-13
feat(refs DPLAN-14634, #AB22463): sort categories + tags by creationDate

hwiem:
### Ticket
[DPLAN-14634](https://demoseurope.youtrack.cloud/issue/DPLAN-14634/ADO-Issue-22463-Admin-Institutionsliste-Filteroptionen-konnen-nach-Schlagwortkategorien-angepasst-werden)

- instead of displaying the categories and tags in a random order, it seems more intuitive to sort them by creationDate
- while sorting by the user is not possible in the interface, this way we give them the possibility to create their desired order by adjusting the order in which they create categories and tags
- also, in the tag list, after creating a category, it is displayed at the top of the list now, so it's easier to confirm that it has indeed been created

### PR Checklist

- [x] Run `yarn lint`

### DIFF
--- a/client/js/components/procedure/admin/InstitutionTagManagement/TagList.vue
+++ b/client/js/components/procedure/admin/InstitutionTagManagement/TagList.vue
@@ -140,10 +140,12 @@ export default {
       this.listInstitutionTagCategories({
         fields: {
           InstitutionTagCategory: [
+            'creationDate',
             'name',
             'tags'
           ].join(),
           InstitutionTag: [
+            'creationDate',
             'isUsed',
             'name',
             'category'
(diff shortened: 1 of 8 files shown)


client/js/components/procedure/admin/InstitutionTagManagement/TagList.vue
```diff
@@ -222,27 +224,31 @@ export default {
...
+        .sort((a, b) => new Date(b.attributes.creationDate) - new Date(a.attributes.creationDate))
+        .map(category => {
+          const { attributes, id, type } = category
+          const tags = category.relationships?.tags?.data.length > 0 ? category.relationships.tags.list() : []
+
+          return {
+            id,
+            name: attributes.name,
+            children: Object.values(tags)
+              .sort((a, b) => new Date(b.attributes.creationDate) - new Date(a.attributes.creationDate))
```
spiess-demos, line 237: Do we also want to sort the tags by creationDate? I think alphabetic (or better, manual sorting) would be better here.

hwiem: I think tags should follow the same logic as categories, isn't it confusing otherwise? Why do you think alphabetic order might be better?
Since this is just an improvement on the side I didn't want to get into manual sorting, and as long as we don't have that sorting by `creationDate` gives at least some kind of control to the user.

spiess-demos: ok i think we can go with creationDate for now, but ultimately this forces users to understand the sorting mechanism by trial & error, and then creating the needed tags in reverse order of the desired "end order". 

to do better, we have to either add a setting per category (which is kind of overkill i guess), or implement a manual sorting mechanism for tags, which may add overhead for users who then are forced to sort large lists alphabetically, but maybe some tags are better sorted manually (like "Verfahrensschritte", which need to appear in a custom order). so maybe a 3rd improvement would be to offer a "sort alphabetically" action on each category within the admin interface.

spiess-demos (approved):
lgtm (code review)

client/js/components/procedure/admin/InstitutionTagManagement/TagList.vue
hwiem: This
> this forces users to understand the sorting mechanism by trial & error, and then creating the needed tags in reverse order of the desired "end order"

is a really good point, and it is also true for categories. Should the sort order be descending by default, so at least you don't have to reverse the order as a user?

The manual sorting (either by drag+drop or via button/flyout) I think we should discuss with the product owner and create a user story.

spiess-demos: >Should the sort order be descending by default, so at least you don't have to reverse the order as a user?

Maybe this is a good idea, yes.

hwiem: Done in 0b7f97a09a3229982613ffbc29f9ee92effe5111

bakridemos (approved):
lgtm, code review only

PR #4336, opened 2025-02-13
Refac(dplan 12188) no reset button

riechedemos:
DPLAN-12188

Description: Refactored search to DbSearchField for new design and reset function.

### How to review/test
Login as support/ use search

### DIFF
--- a/client/js/components/user/DpUserList/DpUserList.vue
+++ b/client/js/components/user/DpUserList/DpUserList.vue
@@ -17,17 +17,17 @@
   <div class="u-mt">
     <!-- search -->
     <div class="layout u-mt">
-      <div class="layout__item u-1-of-1">
+      <div class="layout__item u-1-of-1 flex">
         <dp-search-field
           data-cy="search:currentSearchTerm"
           :placeholder="Translator.trans('searchterm')"
           @search="handleSearch"
           @reset="handleReset" />
+        <dp-contextual-help
+          class="float-right"
+          :text="tooltipContent" />
       </div>
     </div>
-    <dp-contextual-help
-      class="float-right"
-      :text="tooltipContent" />
     <dp-loading
       v-if="isLoading"
       class="u-ml u-mt" />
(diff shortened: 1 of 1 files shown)


sakutademos (commented):
It works well, thank you! The question is: we want to have the same search bar across all places in the future. Is this search field's design suitable for other areas

sakutademos:
could it be changed here as well? It should not be too complicated, I hoope
image

sakutademos (approved):
lgtm, thx :)

spiess-demos:
> The question is: we want to have the same search bar across all places in the future? Is this search field suitable for other areas. Question for UI team @spiess-demos. Can it be implemented in the STN List and across other places

We have agreed on DpSearchField to be used in all of the cases where a search field controls a DpDataTable or list view.

spiess-demos:
> could it be changed here as well? I mean DpSearchFiled instaed of DpInput. It should not be too complicated, I hoope

It should be changed there, too, but i'd suggest in a separate PR.

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -18,25 +18,16 @@
...
-          :text="tooltipContent" />
+        <dp-search-field
+          data-cy="search:currentSearchTerm"
+          :placeholder="Translator.trans('searchterm')"
+          @search="handleSearch"
+          @reset="handleReset" />
       </div>
     </div>
-
+    <dp-contextual-help
```
spiess-demos, line 28: Is there a reason you put the tooltip onto a new line? placing it within the same div as the search field would (imho) look a little more sorted.

riechedemos: c24432c

riechedemos: c24432c

spiess-demos (approved):
lgtm (code review)

PR #4338, opened 2025-02-13
f DPLAN-14978 : adjust condition to enable and show automatically procedure phase change area

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-14978

Description: adjust condition to enable and show automatically procedure phase change area :
this area is displayed even if the permission is not enabled which is confusing. It has to be showed only when the relevant permission is enabled and when a participation phase is selected. When the permission is not enabled this area should be desabled .


bakridemos:
the logic seems to be right like it's , thats why this pr will be closed

PR #4345, opened 2025-02-14
f_DPLAN-14976 limit user administration scope for organisation-administration to own orga via new permission

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-14976

Description:
The user Fachplanung-Masteruser should only be able to administrate the users of their own organisation.

There was no condition available to limit the amount of organisation available via the route used to fill the dropdown for the user-create menu. 
image

Therefore the user could create and update users of different organisations. This was prevented by adding a new permission and checking it within the the used routes.

The result now looks like this where the organisation is still not preset - but the only choice within the dropdown 
image


demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
```diff
@@ -101,14 +101,23 @@ public function listAction(
...
         PermissionsInterface $permissions,
         Request $request,
+        CurrentUserService $currentUser,
         JsonApiPaginationParser $paginationParser,
     ) {
         try {
             if ($permissions->hasPermission('area_organisations_view_of_customer')
                 || $permissions->hasPermission('area_manage_orgas_all')
             ) {
                 $currentCustomer = $customerHandler->getCurrentCustomer();
```
sierrazamudiodemos, line 111: Hey Thanks @MoritzMandler for this.
What do you think about writing something like:
 if (false === $permissions->hasPermission('area_organisations_view_of_customer')
               && false === $permissions->hasPermission('area_manage_orgas_all')) {
 throw new AccessDeniedException('User has no access rights to get $orgalist.');
}

and then rest of the code?

I was thinking to avoid the if else and have it right at the beginning of the code to make it more readable.
I know it is not part of the changes. and I thought it could be a nice touch for the PR

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
```diff
@@ -101,14 +101,23 @@ public function listAction(
...
-                $orgaList = $orgaService->getOrgasInCustomer($currentCustomer);
+                if ($permissions->hasPermission('feature_organisation_own_users_list')) {
+                    $orga = $currentUser->getUser()->getOrga();
+                    if (null === $orga) {
+                        throw new AccessDeniedException('User has no orga and no access rights to get $orgalist.');
+                    }
+                    $orgaList = [$orga];
+                } else {
+                    $orgaList = $orgaService->getOrgasInCustomer($currentCustomer);
+                }
```
sierrazamudiodemos, line 120: What do you think about moving it to another method in the same class?

demosplan/DemosPlanCoreBundle/ResourceTypes/AdministratableUserResourceType.php
```diff
@@ -291,6 +292,13 @@ protected function getProperties(): ResourceConfigBuilderInterface
             );
 
         $configBuilder->addCreationBehavior(new FixedSetBehavior(function (User $user, EntityDataInterface $entityData): array {
+            if ($this->currentUser->hasPermission('feature_organisation_own_users_list')) {
+                $orgaToSet = $entityData->getToOneRelationships()[$this->orga->getAsNamesInDotNotation()];
+                if ($this->currentUser->getUser()->getOrga()->getId() !== $orgaToSet['id']) {
+                    $this->messageBag->add('error', 'error.user.administration.limited.to.own.organisation');
+                    throw new InvalidArgumentException('User is only allowed to administrate users of their own organisation.');
+                }
+            }
```
sierrazamudiodemos, line 301: Thank you for this. I was thinking: what about formalizing in a wiki that using addCreationBehavior is the accurate place to add permissions check like this type? I can add it to a wiki, and before adding I wanted to hear your opinion @MoritzMandler

MoritzMandler: CreationBehavior?

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
sierrazamudiodemos:
```
if (!$permissions->hasPermission('area_organisations_view_of_customer') && $permissions->hasPermission('area_manage_orgas_all')
            ) { 
 throw new AccessDeniedException('User has no access rights to get $orgalist.');
}
```

and then you can remove the else of the line 139

MoritzMandler: ok, got it - I thought you mean the orgaList if-else blcok

demosplan/DemosPlanCoreBundle/ResourceTypes/AdministratableUserResourceType.php
MoritzMandler: I completely mixed two prs same resourceType in my mind - sry....
That would be very nice of you :)

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
MoritzMandler: done in 7b97cc9f8977e3bf1327704f6b44fada0776b544

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
sierrazamudiodemos: Good idea! thank you!

sierrazamudiodemos (approved):
Thank you Moritz

PR #4351, opened 2025-02-17
chore: check that image exists in export to avoid error

mussbach:
Export should not fail in case some image is not available. At this point the images need to be present locally, not via flysystem

### How to review/test
you could delete a file that should exist locally and docx export assessmenttable

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


breierDemos (changes requested):
In case the image is not there, should we perhaps still add a note (something like "The image could not be found / does not exist.") instead of the image? So that the reference in the text shows this note.

breierDemos:
> @breierDemos where would you display the text? In the export itself?

yes

mussbach:
As the bug occurred again I will merge this and implement you request as future improvement

PR #4360, opened 2025-02-19
fix(DPLAN-15035): issue with saving the unchanged Tag/Topic title

sakutademos:
### Ticket
[DPLAN-15035](https://demoseurope.youtrack.cloud/issue/DPLAN-15035/Fehlermeldung-wenn-man-ein-Schlagwort-nicht-andert-und-auf-Checkmark-klickt)

**Description:** This PR fixes an issue with saving the unchanged Tag/Topic title.

- check on the FE if the title attribute has been changed and make an update request only if there are any changes.

### How to review/test
Auf edit Icon von einem Schlagwort klicken -> nichts ändern -> auf Checkmark klicken -> Fehlermeldung

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [x] Update changelog

### DIFF
--- a/client/js/components/tags/TagsList.vue
+++ b/client/js/components/tags/TagsList.vue
@@ -44,7 +44,7 @@
           :is-in-edit-state="isInEditState"
           :procedure-id="procedureId"
           type="TagTopic"
-          @abort="abort"
+          @abort="closeEditForm"
           @delete="deleteItem"
           @edit="setEditState"
           @save="save" />
@@ -55,7 +55,7 @@
           :is-in-edit-state="isInEditState"
           :procedure-id="procedureId"
           type="Tag"
-          @abort="abort"
+          @abort="closeEditForm"
           @delete="deleteItem"
           @edit="setEditState"
           @save="save" />
@@ -177,7 +177,7 @@ export default {
       saveTagTopic: 'save'
     }),
 
-    abort () {
+    closeEditForm () {
       this.isInEditState = ''
     },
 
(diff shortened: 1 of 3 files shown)


client/js/components/tags/TagsList.vue
```diff
@@ -277,23 +277,30 @@ export default {
...
-      this[`update${type}`]({
-        attributes,
-        id,
-        relationships: this[type][id].relationships,
-        type
-      })
-      this[`save${type}`](id)
-        .then(() => {
-          this.isInEditState = ''
+      if (typeof this[updateMethod] === 'function' && typeof this[saveMethod] === 'function') {
```
salisdemos, line 290: I am not sure if this is the best approach. Why can it be not a function? 
Because, if its no function, we still can't save. That don't fixes the issue imho.

salisdemos (changes requested):

client/js/components/tags/TagsList.vue
sakutademos: I understand you , but I thought it would be a nice check, especially when the type we receive from the outside could change and may no longer be a function in the FE. However, I understand your perspective and c[an remove it if you want](https://github.com/demos-europe/demosplan-core/pull/4360/commits/e3f97f0204573beba2a61ad11d50172012c9e017). but if we use typesript, we should always include such checks

salisdemos: Maybe. For me it feels wrong, but maybe I just don't see the benefits.

salisdemos (approved):
Thanks. Looks good.

PR #4368, opened 2025-02-20
tests: Fix test for StatementExportModal

gruenbergerdemos:
`StatementExportModal` was changed recently, this fixes the test. 

Related PR 
https://github.com/demos-europe/demosplan-core/pull/4350


hwiem (approved):
Thanks for fixing!

PR #4371, opened 2025-02-20
Fix or skip BE tests

sierrazamudiodemos:
This PR is about fixing some tests and skipping others, so that the BE TeamCity check is required for future PR

### How to review/test
- tests/backend/core/Core/Functional/Command/AddonAutoinstallCommandTest.php It is fixed: by removing the check that clear the cache. That is because this command does not clear the cache anymore at the moment the test expects it
- tests/backend/core/Core/Unit/Logic/PermissionsTest.php It is fixed: It was failing due to a permission that was in core was moved to projects
- tests/backend/core/Core/Unit/Utilities/Twig/RouterTest.php the test was skipped
- tests/backend/core/Document/Unit/DocumentBundleImporterTest.php It is fixed: The test was asserting the wrong string
- tests/backend/core/Import/ZipImportServiceTest.php It is fixed: the test was asserting a temp file that changes between local and the TeamCity server
- tests/backend/core/News/Functional/GlobalNewsHandlerTest.php It is fixed
- tests/backend/core/Procedure/Functional/ProcedureServiceTest.php it is fixed: Now it is using another test produre that is not the default customer procedure
- tests/backend/core/Report/Functional/ReportMessageConverterTest.php it is fixed: now it is using mocks
- tests/backend/core/Statement/Functional/ImageLinkConverterTest.php it is skipped


mussbach (approved):
Great thanks a lot!

PR #4374, opened 2025-02-21
feat DPLAN-15068 Add template for tag topic import

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-15068/Add-Beispieldatei-fur-Schlagwort

The user should have a template file that describes how to import the tags - topics

### How to review/test
- Go to Verfahren > Schlalwort
- There should be a link called BeispielDatei and when clicking , a csv should be downloaded as a template


salisdemos (approved):
Thanks for adding the examples

PR #4376, opened 2025-02-21
add missed onDelete cascade to Boilerplate and BoilerplateCategory

jalalifardemos:
### Ticket: https://www.dev.diplanung.de/DefaultCollection/BOP-HH/_workitems/edit/24405

Description: We have an error when run `app/console dplan:maintenance -e prod --no-debug`.
Error:
´´´
Error thrown while running command "'demos:maintenance' -e prod --no-debug"
"Doctrine\ORM\Mapping\DefaultQuoteStrategy::getJoinTableName(): Argument #1 ($association) must be of type array, null given, called in /data/bobhh/vendor/doctrine/orm/src/Persisters/Entity/BasicEntityPersister.php on line 1754
´´´
The error could indeed be related to the lack of onDelete='cascade' in Doctrine ORM entity mappings. If an entity is deleted and its associations are not properly handled with cascading operations, it can lead to null values being passed where arrays are expected.  To resolve this, ensure that our entity associations are correctly configured with onDelete='cascade' where necessary. This will ensure that related entities are properly deleted when the parent entity is deleted, preventing null values in associations. In our case Boilerplate and BoilerplateCategory should fix.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/Entity/Procedure/Boilerplate.php
```diff
@@ -64,8 +64,8 @@ class Boilerplate extends CoreEntity implements UuidEntityInterface, Boilerplate
      *
      * @ORM\JoinTable(
      *     name="predefined_texts_categories",
-     *     joinColumns={@ORM\JoinColumn(name="_pt_id", referencedColumnName="_pt_id")},
-     *     inverseJoinColumns={@ORM\JoinColumn(name="_ptc_id", referencedColumnName="ptc_id")}
+     *     joinColumns={@ORM\JoinColumn(name="_pt_id", referencedColumnName="_pt_id", onDelete="CASCADE")},
```
mussbach, line 67: we should use doctrine cascading here. When really a database cascade is needed a migration is missing. Did you test it?

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/Entity/Procedure/Boilerplate.php
jalalifardemos: I add doctrine Cascade in https://github.com/demos-europe/demosplan-core/pull/4376/commits/b772f4e5900e1bccadebe1a71c651e6883e7928e. I didn't test it (I don't know how can I dot that) but pretty sure we need this cascade here, because If user delete a boilerplate manually, it's relation should be removed too

martendemos: 1. Whenever possible we want to use doctrine cascade in favor of database cascade.
2. Afaik the doctrine cascading works exactly in the other direction than the database cascading. Thats why tests (automatically and or manual) are very important.
3. While database cascade can't be covered by automatically test (because the temp-test-DB are not perform these kind of DB-sided-cascading), doctrine cascade will be executed by automatically tests. So this is one of the reasons we want to use doctrine cascade instead of database cascade.
4. Using doctrine cascade as well as database cascade like in this case, is a duplication of the cascade logic and don't seems to be correct at all.

My suggestion: Remove the database cascades and cover the doctrine cascades with automatically tests.

martendemos: In Addition important to know is, that a cascade deletion on manyToMany (cross) tables, will only delete the entries in the cross-table (and therefore the reletion between the two entities), not the related entity.

jalalifardemos: done in https://github.com/demos-europe/demosplan-core/pull/4376/commits/4090ae6ddef1ded6254b486eb9e05df381fb75f2

mussbach (approved):
lgtm, code review only

PR #4377, opened 2025-02-21
chore: Adjust Dependencies to match the dependency Version Tree

salisdemos:
Unfortunately the required dependecies by the latest `eslint-config-standard@17.1.0` are a bit outdated. With that we can't require the latest versions of `eslint-plugin-promise`, `eslint-plugin-n` and `eslint`.

cesium and @types/leaflet have to be added to satisfy masterportalapi


hwiem (approved):
lgtm, thanks

PR #4378, opened 2025-02-21
chore: Bump Stylelint@16.14.1 and run lint --fix

salisdemos:
we that we could run `yarn lint:scss` in the ci as required check

`yarn lint:scss` should now run without errors.
Maybe its necessary to clear the yarn cache before.

I had to adjust the stylelint.config to match the corrents version options.

All content-changes are generated with `yarn lint:scss --fix` (except of the no-lint-entry)


spiess-demos (approved):
lgtm (code review)

hwiem (approved):
Lgtm, thank you :heart:

PR #4398, opened 2025-02-26
feature (ref: 14941): Add abbitMq Parameter to parameters default

jalalifardemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-14941/ADO-5545-Konfiguration-DiPlanBeteiligung-zum-RabbitMQ-Nachrichtenaustausch-mit-DiPlanCockpit

Description: Add rabbitMQ Parameter from XBeteiligung Addon to parameter default. Parameters.yml should be updated with true value in diplanbau.

### DIFF
--- a/config/parameters_default.yml
+++ b/config/parameters_default.yml
@@ -427,6 +427,8 @@ parameters:
     # try to create procedures from the messages
     enable_xta_communication: false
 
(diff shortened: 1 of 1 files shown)


mussbach (changes requested):
I do not think this should live in the core

PR #4402, opened 2025-02-26
chore (refs DPLAN-14978): throw event with unmodified values

mussbach:
### Ticket
DPLAN-14978

unfortunately php/doctrine is not able to fetch the unmodified procedure from the database **before** the update of the modified object in updateProcedureObject, so we need to dispatch the event again. Moreover during clone the objects that are attached to the entity are not cloned automatically. I tried __clone which did not work, I exprerimented with https://github.com/myclabs/DeepCopy which is quite a nice package, but triggered new problems. Finally I found no other way than to trigger the event again with the original values as well

### How to review/test
code review should be best

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


breierDemos (approved):
lgtm
Maybe we can create a ticket to discuss this further. What do you think?

PR #4404, opened 2025-02-27
fix(DPLAN-12263): the item was still displaying in the list after deleting it

sakutademos:
### Ticket
[DPLAN-12263](https://demoseurope.youtrack.cloud/agiles/141-63/current?settings&tab=allgemeines&issue=DPLAN-12263)

**Description:** This PR fixes an issue where the item was still displaying in the list after deleting it. 

The problem is related to reactivity: the `statementsObject` varible updates properly, but the computed value `items()` that transforms the `statementsObject` does not detect the change. I tried different ways to trigger the reactivity, but there were always some problems. The only solution without issues was to fetch the items by page after the delete action.

- [ ] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/client/js/components/statement/listStatements/ListStatements.vue
+++ b/client/js/components/statement/listStatements/ListStatements.vue
@@ -945,8 +945,8 @@ export default {
       if (window.confirm(Translator.trans('check.statement.delete'))) {
         this.deleteStatement(id)
           .then(response => checkResponse(response, {
-              200: { type: 'confirm', text: 'confirm.statement.deleted' },
-              204: { type: 'confirm', text: 'confirm.statement.deleted' }
+            200: { type: 'confirm', text: 'confirm.statement.deleted' },
+            204: { type: 'confirm', text: 'confirm.statement.deleted' }
           }))
           .then(() => {
             this.getItemsByPage(this.pagination.currentPage)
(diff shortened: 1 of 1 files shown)


client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -663,7 +663,6 @@ export default {
       if (assigneeId !== this.currentUserId) {
         this.claimStatement(statementId)
       } else {
-        console.log('unclaim')
```
salisdemos, line 666: :heart:

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -945,10 +944,13 @@ export default {
...
-            200: { type: 'confirm', text: 'confirm.statement.deleted' },
-            204: { type: 'confirm', text: 'confirm.statement.deleted' }
-          }))
+          .then(response => {
+            this.getItemsByPage(this.pagination.currentPage)
+            checkResponse(response, {
+              200: { type: 'confirm', text: 'confirm.statement.deleted' },
+              204: { type: 'confirm', text: 'confirm.statement.deleted' }
+            })
+          })
```
salisdemos, line 953: shouldn't the getItemsByPage be fired after the checkResponse?
what if the request fails. Is it on purpose, that the page gets reloaded then?

salisdemos (approved):
Just a qustion out of curiosity. No blocker

client/js/components/statement/listStatements/ListStatements.vue
sakutademos: of course, you are totally right (: fixed it

salisdemos (approved):

PR #4405, opened 2025-02-27
DPLAN-DS31 filter deleted single and paradocs when retrieved via resourceTypes

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/tickets/DS-31/Evtl.-Bug-in-den-Planunterlagendarstellungen?backToIssues

Description:
filter deleted single and paradocs when retrieved via resourceTypes

### How to review/test
delete a single-doc from an element after referencing it in a statement - this will lead to a soft-deletion.
Those ones were not filtered out correctly before but are now.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/demosplan/DemosPlanCoreBundle/ResourceTypes/ParagraphResourceType.php
+++ b/demosplan/DemosPlanCoreBundle/ResourceTypes/ParagraphResourceType.php
@@ -21,6 +21,7 @@ use EDT\PathBuilding\End;
  *
  * @property-read End $title
  * @property-read PlanningDocumentCategoryResourceType $element
+ * @property-read End $deleted
  */
 final class ParagraphResourceType extends DplanResourceType
 {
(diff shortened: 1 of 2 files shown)


bakridemos:
> lgtm, code review only, could you perhaps provide a test?

i tried to provide a test but resourcetypes are not really easy to test. The get and list action are not allowed in SingleDocumentResourceType and is only available when the right permission is enabled which we can't while we can't login with a test user with the required permissions set

mussbach:
@bakridemos I assume you tried it via `AbstractApiTest::enablePermissions()`? No merge blocker though

bakridemos:
> @bakridemos I assume you tried it via `AbstractApiTest::enablePermissions()`? No merge blocker though

exactly

PR #4407, opened 2025-02-28
feat DPLAN-14864 EWM integration into diplanfest

sierrazamudiodemos:
### Ticket
DPLAN-14864
https://demoseurope.youtrack.cloud/issue/EFA-61/EWM-Integration-DiPlanBeteiligung-Planfeststellung

- Make EWM integration possible in diplanfest. Use as an inspiration the existing PR of diplanrog.

### How to review/test
FE wise:
- The "Schlussmitteilung versendet"  has some logic that was in BE and was moved to FE. for that, the StatementResourceType was adjusted to expose new attributes to the FE for that logic

BE wise:
1. Big change was to enable the "send" action as RpCRequest in  "Schlussmitteilung versendet" 
2. That said, there are 3 important things:
a) create a new RpcRequest route called StatementEmailSender that the FE is calling when the user presses "send" button
b) create a new class called StatementEmailSender which is in charge of sending the email.
c) adjust the existing class demosplan/DemosPlanCoreBundle/Logic/AssessmentTable/AssessmentTableServiceStorage.php. which is used in the old controller. This class now calls the newly created class StatementEmailSender to send the email. In this way, we centralized on one class the logic of sending the email from the old controller and the newly created rpc request

 
3. for documenting reasons, you can also query the RPC request from http request in Phpstorm:
```
POST http://diplanfest.dplan.local/app_dev.php/rpc/2.0
X-JWT-Authorization: Bearer REPLACE_WITH_YOUR_TOKEN
X-Demosplan-Procedure-Id: REPLACE_WITH_YOUR_PROCEDURE_ID
Content-Type: application/json

{
  "jsonrpc":"2.0",
  "method":"statement.email.sender",
  "id":"REPLACE_WITH_A_VALID_ID",
  "params":
  {
    "statementId": ""REPLACE_WITH_STATEMENT_ID",
    "subject": "Email Subject",
    "body": "This is the body of the email.",
    "sendEmailCC" : "cc@example.com,cc2@example.com",
    "emailAttachments": [
      "sunflower_is_amazing_and*_more.jpg:ff1abbb3-dc5f-4a59-90ba-e15983fa0b47:260050:image/jpeg",
      "2_Fehlanzeige-Bsp.PNG:f7af6bc8-a26d-48a7-9b3e-7deea998f83d:260050:image/jpeg"
    ]
  }
}
```

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMetaFinalEmail.vue
```diff
@@ -0,0 +1,320 @@
+<license>
```
hwiem, line 1: I think the non-edit-mode (i.e. when the statement is not assigned to me) still needs some adjustments:

final_mail

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMetaFinalEmail.vue
```diff
@@ -0,0 +1,320 @@
...
+(c) 2010-present DEMOS plan GmbH.
```
hwiem, line 2: When i send the final email, i get a notification that i should change the statement status to "Schlussmitteilung versendet". Is that possible? Or should we adjust the text?

hwiem (changes requested):
FE code looks good in general (just some nitpicks), but when testing I noticed some issues

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMetaFinalEmail.vue
gruenbergerdemos: Mmh, no, good point, we can't change the status manually. I change the text.

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMetaFinalEmail.vue
gruenbergerdemos: Thanks for noticing. Done with [3596706](https://github.com/demos-europe/demosplan-core/pull/4407/commits/3596706305d34344349c34bec2d01471f9f285a8)

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMetaFinalEmail.vue
```diff
@@ -0,0 +1,320 @@
...
+
+    explanationNoSendingEmail () {
+      const { authorFeedback, feedback, publicStatement } = this.statement.attributes
+      let sendFinalEmail = false
+      if ((feedback === 'snailmail' && this.statement.relationships.votes.data) ||
+          feedback === 'email') {
+        sendFinalEmail = true
+      }
+
+      if (!sendFinalEmail) {
```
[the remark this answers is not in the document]
gruenbergerdemos: Done with [d71ef30](https://github.com/demos-europe/demosplan-core/pull/4407/commits/d71ef30a12668d959c0769de69a27b8c91b26cd9)

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMetaFinalEmail.vue
hwiem: I think that we should also display something below the "Kopie" label in non-edit-mode. I found the text very confusing because also it was not clear to me what "Kopie" meant. Maybe we could rename it to "Kopie an"? And the label hint is not needed when you can't input anything. After looking at what is displayed for other fields on the page when they are disabled and empty, I'd display a message like "Keine E-Mail-Adressen angegeben" or something like that?

gruenbergerdemos: I removed the hint when disabled and changed the label to "Kopie an" with [ea9eed7](https://github.com/demos-europe/demosplan-core/pull/4407/commits/ea9eed7ffa7235d3725a0f80e3ad493bc7f99a77).

> After looking at what is displayed for other fields on the page when they are disabled and empty, I'd display a message like "Keine E-Mail-Adressen angegeben" or something like that?

Most input fields also have nothing displayed in disabled mode. I suggest to to fix this later overall in a bug ticket.

breierDemos (changes requested):
**Summary of Pair Review**
1. the $emailcc is not defined
2. maybe remove try catch or find a better way
3. check $emailcc --> $ccEmailAddresses
4. 730 line the emailCC is emtpy and it is not reset
5. in the sendFinalStatementemail the $emailcc does not have to be reset all the time (because of case above)
6. in the sendFinalStatementEmail, there is a foreach per receipient email, maybe it does not make sense because the email address is not logged... ask Flora
7. replace getEmail2 by getParticipationEmail()
8. rename method detectRecipientEmailAddress ? or maybe add comment -> detectRecipientParticipationEmailAddresses
9. blocker PR: dont user reference &


related to case 7
image

- E-Mail Verfahrensträger* -> email2
- Weitere Empfänger*innen -> ccEmail2

sierrazamudiodemos:
Point 7: 89ed6e49ace25e0c485c34ae14e19b75dbcce9e2

breierDemos (approved):
Thank you :) BE lgtm

Point 6 can be handled later.

PR #4419, opened 2025-03-04
fix(refs  dplan-14909): Always send assignee and place in pair

salisdemos:
### Ticket
DPLAN-14909

For some reason, sometimes, when both assignee and place changed, only place where send. Because it happens soemtimes, I think it is some sort of race condition between the request and an update of the initial segment. Because I monitored that the initial segment gets updated aswell even though, It should not at that very moment.

To prevent all that, I decided to bypass the vuex-json-api and update the resourcetype "manually".

### How to review/test

in the segmentslist, when changing plce and assignee, it always should send both.
Hence the bug appears just randomly, its really hard to test.

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
@@ -623,7 +623,7 @@ export default {
         }
       }
 
-      return dpApi.patch(Routing.generate('api_resource_update', { resourceType: 'StatementSegment', resourceId: this.segment.id }), {}, payload)
+      dpApi.patch(Routing.generate('api_resource_update', { resourceType: 'StatementSegment', resourceId: this.segment.id }), {}, payload)
         .then(checkResponse)
         .then(() => {
           this.claimLoading = false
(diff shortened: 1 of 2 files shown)


client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -863,9 +872,12 @@ export default {
       }
 
       this.setSegment({ ...updated, id: this.segment.id })
+
+      return relations
     },
 
     updateSegment (key, val) {
+      console.log('updateSegment', key, val)
```
hwiem, line 880: This can be removed I guess

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -678,9 +678,18 @@ export default {
...
       const comments = this.segment.relationships.comments ? { ...this.segment.relationships.comments } : null
+      const rel = this.updateRelationships()
+
+      const payload = {
+        data: {
+          id: this.segment.id,
+          type: 'StatementSegment',
+          relationships: { assignee: rel.assignee, place: rel.place }
+        }
+      }
```
hwiem, line 689: nitpick; indentation may need to be fixed in my suggestion

```suggestion
      const { assignee, place } = this.updateRelationships()

      const payload = {
        data: {
          id: this.segment.id,
          type: 'StatementSegment',
          relationships: {
            assignee,
            place
            }
        }
      }
```

hwiem (approved):
Code lgtm, just nitpicks

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
salisdemos: I did it like that in the beginning, but for some reason sonarCloud complained about it. Now its fine . Don't get it.
0c1aec2fd9c492da1f7e43eee516588b545fafca

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
salisdemos: 0c1aec2fd9c492da1f7e43eee516588b545fafca

PR #4420, opened 2025-03-04
ADO-24228 attempt to fix email attachments - at least enhance logging

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-15048/ADO-Issue-24228-Produktion-Integration-Diplan-Beteiligung-Bauleitplanung-Bestatigungsmails-STN-Einreichung-enthalten-keinen-PDF

Description:
pass content type and name for the email attachments explicitly to the message->attach method to be flysystem compatible.
Delete Attachments only after the mail has been marked as sent successfully.
enhance logging for future investigations.

In order to be compatible with the FlySystem storage we had to implement the ```Email::attach``` method instead of the previously used ```Email::attachFromPatch``` method. They handle their defaults for 'name' and 'contentType' slightly different, as there is no way to guess a name or mime-type from the filepath when just a stream is provided.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/MailService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/MailService.php
@@ -26,6 +26,8 @@ use Doctrine\ORM\EntityNotFoundException;
 use Exception;
 use League\Flysystem\FilesystemException;
 use League\Flysystem\FilesystemOperator;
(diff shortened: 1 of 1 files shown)


mussbach (approved):
lgtm, code review only

PR #4425, opened 2025-03-06
fix(refs DPLAN-15179): reset filters in localStorage

hwiem:
### Ticket
[DPLAN-15179](https://demoseurope.youtrack.cloud/issue/DPLAN-15179/Filter-konnen-nicht-zuruckgesetzt-werden-Institutionen-verschlagworten)

- reset button is no longer disabled when filters are applied; this was fixed by ensuring that `appliedFilterQuery` is an array only when empty, and an object when not empty
- on reset, applied filters are also removed from storage, so when reloading the page, filters are not re-applied

### How to review/test

- login as customer admin
- go to "Institutionen verschlagworten"
- filter the list and try to reset the applied filters - it should now work without problems and list should remain unfiltered after reloading the page

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
+++ b/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
@@ -668,6 +668,10 @@ export default {
         })
     },
 
+    resetFilterQueryInLocalStorage () {
+      localStorage.setItem('filterQuery', JSON.stringify({}))
+    },
+
     resetQuery () {
       this.searchTerm = ''
       Object.keys(this.allFilterCategories).forEach((filterCategoryId, idx) => {
@@ -690,6 +694,8 @@ export default {
           }
         }
       })
+
+      this.resetFilterQueryInLocalStorage()
       this.appliedFilterQuery = []
       this.getInstitutionsByPage(1)
     },
--- a/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
+++ b/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
@@ -232,6 +232,7 @@ export default {
   mixins: [tableScrollbarMixin],
 
   props: {
+    // Array when empty, object when not empty
     initialFilter: {
       type: [Object, Array],
       default: () => ({})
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
```diff
@@ -232,6 +232,7 @@ export default {
   mixins: [tableScrollbarMixin],
 
   props: {
+    // Array when empty, object when not empty
```
salisdemos, line 235: :-/

salisdemos: can't we send always an object?

salisdemos (approved):
I can imagine, that it would bloat the PR, but i left the question anyway.
For a quick fix  the given solution is legit.

hwiem:
> For me locally it did not work. The isQueryApplied stays false and does not change to true and the reset button stays disabled after applying a filter.

I checked now and I think it happens when you visit the page for the first time (since the `localStorage` was introduced). Once you reload the page it works. I'll take another look at it. :disappointed:

gruenbergerdemos:
> I checked now and I think it happens when you visit the page for the first time (since the localStorage was introduced). Once you reload the page it works.

I reloaded it a bunch of times, but after every reload the first time applying filter the button stayed disabled :/

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
hwiem: As a matter of fact, we can. I thought that since it is initially set to `this.initialFilter`, we couldn't, because it has type `[Object, Array]`, and normally that's because of how we get it from the BE via twig. But as I realized now, `initialFilter` is never passed to the component, so it can be removed, and then `appliedFilterQuery` can just be an object. Thanks for asking!

hwiem: Done in df3e25905f6f5aa131743aa791ba8a52a6e61cca

hwiem:
> I reloaded it a bunch of times, but after every reload the first time applying filter the button stayed disabled :/
Could you check again if it works now? And if it doesn't, maybe we can take a look together?

PR #4426, opened 2025-03-06
Bug(refs DPLAN-15155): New stn select user

riechedemos:
### Ticket
DPLAN-15155

Extended the if statement to ensure length does not cause an error.

### How to review/test
Create new Stn and choose existing user from drowdown


salisdemos (approved):
Thanks for fixing

hwiem (approved):
Nice, lgtm!

PR #4427, opened 2025-03-06
fix(refs DPLAN-15179): reset filters in localStorage

hwiem:
cherry-pick of https://github.com/demos-europe/demosplan-core/pull/4425 from main


salisdemos (approved):
legit

PR #4432, opened 2025-03-07
fix(refs DPLAN-15170, #AB22463): persist flyout and filter option selection

hwiem:
### Ticket
[DPLAN-15170](https://demoseurope.youtrack.cloud/issue/DPLAN-15170)

- after reloading the page, the selection of filter flyouts made by the user via the 'more filters' flyout is still visible
- filter options from those flyouts that are applied will remain applied after reloading the page
- if filter options from a filter flyout are selected and applied, that filter flyout cannot be hidden via the 'more filters' flyout, since it would not be clear which filters are applied. instead, the checkbox is disabled. also, the count of selected options is now displayed in the 'more filters' flyout behind each category label.

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/procedure/SegmentsList/FilterFlyout.vue
+++ b/client/js/components/procedure/SegmentsList/FilterFlyout.vue
@@ -538,6 +538,12 @@ export default {
       const isInitialWithQuery = true
 
       this.requestFilterOptions(isInitialWithQuery)
+
+      if (this.itemsSelected) {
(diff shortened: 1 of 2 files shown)


riechedemos:
If you uncheck all categories, they disappear, but they reappear after reloading the page.

sakutademos (approved):
code review, lgtm, thx

PR #4439, opened 2025-03-10
fix DPLAN-15211 Set default email body

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-15211

- I remove emailDefaultText varible and set default email body to emailBodyText directly. 
- In this way, behaviour is consistent to how default emailSubject is set


salisdemos (approved):
for whatever reason it was implemented the way it was before. Now it looks correct

PR #4441, opened 2025-03-11
fix DPLAN-15223: Check if permission is enable before making request

sierrazamudiodemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-15223/Fehlermeldung-bei-Klick-auf-Erwiderung-verfassen-uber-Abschnittsansicht

Check if permission is enable before requesting ProcedureMapSettingResourceType

### DIFF
--- a/client/js/store/map/ProcedureMapSettings.js
+++ b/client/js/store/map/ProcedureMapSettings.js
@@ -39,8 +39,6 @@ export default {
         .then(response => checkResponse(response))
     },
 
-
-
     fetchProcedureMapSettings ({ commit }, { procedureId, isMaster = false }) {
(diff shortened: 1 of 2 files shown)


client/js/store/map/ProcedureMapSettings.js
```diff
@@ -40,6 +40,7 @@ export default {
     },
 
     fetchProcedureMapSettings ({ commit }, { procedureId, isMaster = false }) {
+      if (!hasPermission('area_admin_map')) return
```
salisdemos, line 43:
```suggestion
      if (!hasPermission('area_admin_map')) return

```

salisdemos (approved):
just a nitpick missing blank line ;-)

PR #4446, opened 2025-03-12
fix(refs DPLAN-15146): Pass propper Values to orgafilter and orgaRequest

salisdemos:
- needs demosplan-ui update

### Ticket
DPLAN-15146

Only send fields to fetch the orga which are permitted

To fix the Bug https://github.com/demos-europe/demosplan-core/pull/4446 is required


gruenbergerdemos (approved):
lgtm, code review only

PR #4457, opened 2025-03-14
F create deletion report entry when procedure is deleted

bakridemos:
Description: 

- use the method 'updateProcedure' instead 'updateProcedureWithoutReport' : a deletion report has to be created to archive and provide details about deletion records.
- add the parameter '$createReports' : this parameter will be true by default,  it allows with false value to update procedure without creating report entries
- add method 'createProcedureDeletionEntry'

### How to review/test
- delete a procedure and a new deletion report has to be created in the reports table

- [x] Update changelog


MoritzMandler (approved):
lgtm - tested and works fine - even though these entries are not displayed anywhere they are nice to have

PR #4458, opened 2025-03-14
chore: do not send a not existing hash to the frontend

mussbach:
In case of an error or found virus during upload only the id is reset and an invalid hash is sent. We should also remove the temporary hash

### How to review/test
code review might be easiest. At the frontend side you may assess, what happens, when no hash is sent back from the upload file. I could not find any misbehaviour

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


salisdemos (approved):
fe wise legit

MoritzMandler (approved):
lgtm

PR #4459, opened 2025-03-14
chore(refs DPLAN-15021): Enhance upload tags form

salisdemos:
### Ticket
DPLAN-15021

to display the uploaded file, so the user get a visual effect that the upload was successfull

To test this: It should be possible to upload a csv file with tags (at /verfahren/<procedure-id>/schlagworte)
and the uploaded file should be displayed before it gehts imported.


gruenbergerdemos (approved):
lgtm and works

PR #4461, opened 2025-03-17
bug: Remove demosplan-ui link from package.json

gruenbergerdemos:
The demosplan-ui link should only be added locally and not be committed.


salisdemos (approved):
Thank You!

PR #4466, opened 2025-03-18
fix(refs DPLAN-15247): Always fetch places and assignableUsers for segments once and just once

salisdemos:
### Ticket
DPLAN-15247

Before, the places and assignableUsers where fetch for every segment.
To prevent these mass of unneccesary request it got checkt if there 
are already places/users in the store. If so, don't fire the request.
But if the place or assignee was already set, it was delivered as 
relationship with the segment. With that the list wasnt empty anymore.
Because of that and because the if logic wasn't the most elegant way, I moved the logic to the wrapper. there we don't have to check if it was already laoded. we justs fetch on mounted.

** How to Test **
Check if the List of places and assignableUsers is as long as expected

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/SegmentsRecommendations.vue
+++ b/client/js/components/procedure/StatementSegmentsList/SegmentsRecommendations.vue
@@ -296,7 +296,6 @@ export default {
 
   mounted () {
     if (Object.keys(this.segments).length === 0) {
-
       this.fetchSegments()
     }
   }


client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -916,42 +916,45 @@ export default {
...
-        include: 'department',
-        sort: 'lastname'
-      })
-        .then(() => {
-          this.setSelectedAssignee()
-        })
-    } else {
-      this.setSelectedAssignee()
-    }
+    console.log('mounted segment')
```
hwiem, line 919: Should the changes in this file be committed?

hwiem (changes requested):

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
salisdemos: yes. This is the most important line of the PR ;-)

sakutademos (approved):

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
salisdemos: this and some more leftovers where removed here 17f49e6c75b43d1271a3a511ba4a8ef72a5d754f

client/js/components/procedure/StatementSegmentsList/SegmentsRecommendations.vue
```diff
@@ -188,9 +196,33 @@ export default {
         })
     },
 
-    fetchSegments () {
+    async fetchSegments () {
```
hwiem, line 199: nitpick: this function should be renamed because it does a lot more than fetching segments

hwiem (approved):
nice!

PR #4475, opened 2025-03-19
fix(refs DPLAN-15169): enable vue 3 render function api

hwiem:
### Ticket
[DPLAN-15169](https://demoseurope.youtrack.cloud/issue/DPLAN-15169/Nicht-moglich-um-ein-Bild-zu-einem-Kapitel-zu-hinzufugen)

This is needed to make tiptap/vue3 NodeViewWrapper work (used in demosplan-ui/DpResizableImage), which uses the new api.

### How to review/test

- Login as FPA
- Go to Verfahren > Planungsdokumente und Planzeichnung
- Edit an existing category or create one first
- In the edit view, scroll down and click the "create new chapter" button
- In the new view, try to upload a picture to the editor - this should work without errors

### Linked PRs (optional)
- [x] https://github.com/demos-europe/demosplan-ui/pull/1217

### Tasks 
- [x] Merge https://github.com/demos-europe/demosplan-ui/pull/1217
- [x] Release new demosplan-ui version
- [x] Use new demosplan-ui version in demosplan-core https://github.com/demos-europe/demosplan-core/pull/4483
- [ ] Merge this PR

### PR Checklist

### DIFF
--- a/client/js/InitVue.js
+++ b/client/js/InitVue.js
@@ -31,11 +31,16 @@ import loadSentry from './loadSentry'
 import NotificationStoreAdapter from '@DpJs/store/core/NotificationStoreAdapter'
 import NotifyContainer from '@DpJs/components/shared/NotifyContainer'
 import RegisterFlyout from '@DpJs/components/user/RegisterFlyout'
+import { configureCompat } from '@vue/compat'
 
 function initialize (components = {}, storeModules = {}, apiStoreModules = [], presetStoreModules = {}) {
   bootstrap()
 
   return initStore(storeModules, apiStoreModules, presetStoreModules).then(store => {
+    configureCompat({
+      RENDER_FUNCTION: false
+    })
+
     const app = createApp({
       mounted () {
         window.dplan.notify = new NotificationStoreAdapter(this.$store)


salisdemos (approved):
Even though, the solution is legit, I am not sure if we want to use the option like so, or if we want to refactor the renderfunctions instead.

hwiem:
> Even though, the solution is legit, I am not sure if we want to use the option like so, or if we want to refactor the renderfunctions instead.

Why do you think we should refactor the render functions?
Apart from that general question, this fix is needed because a dependency (tiptap/vue-3) is using a render function with the vue 3 api, so I don't really see a way around adding this to the config. Once we use vue 3 instead of the compat mode, we can remove it again.

salisdemos:
> > Even though, the solution is legit, I am not sure if we want to use the option like so, or if we want to refactor the renderfunctions instead.
> 
> Why do you think we should refactor the render functions? Apart from that general question, this fix is needed because a dependency (tiptap/vue-3) is using a render function with the vue 3 api, so I don't really see a way around adding this to the config. Once we use vue 3 instead of the compat mode, we can remove it again.

Sorry. Then I understood the setting the wrong way round. Just forget about it.

PR #4480, opened 2025-03-20
F dplan 14775 pass the right translation key

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-14775

Description: 
- pass the right translation key when it exists.
- set empty string as default value : the reason is that there are twigs that are not using this translation key but they include and overwrite the core twig where the translation key is needed.


demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
```diff
@@ -1399,7 +1399,17 @@ public function administrationEditAction(
             } else {
                 $template = '@DemosPlanCore/DemosPlanProcedure/administration_edit.html.twig';
                 $title = 'procedure.adjustments';
+
+                $evaluatingPhase = null;
+                foreach ($templateVars['internalPhases'] as $internalPhase) {
```
mussbach, line 1404: I think we can check for         participationstate: !php/const demosplan\DemosPlanCoreBundle\Entity\Procedure\Procedure::PARTICIPATIONSTATE_FINISHED
        here to keep things dynamically and double check, that this key is set in any procedurephases.yml

mussbach (changes requested):

PR #4482, opened 2025-03-20
bug(refs DPLAN-15256): Remove emits from root

gruenbergerdemos:
### Ticket
DPLAN-15256

In vue 3 the method $root.$on does not exist anymore. This PR fixes the failing tests because of this.

### How to review/test
Run `yarn test`. There is still one test failing because of $root.$on (`DpPublicLayerListCategory`). This will be fixed in another PR.

PR #4472, opened 2025-03-19
test(refs DPLAN-15256): fix several jest tests

gruenbergerdemos:
### Ticket
DPLAN-15256

- Use vue3 in jest config
- createLocalVue no longer exists
- Use createApp for localVue
- Prototype no longer exists, use config.globalProperties instead

**Please note:** originally, the target branch was the vue 3 branch, but it has been merged into `main`. github did not manage to update the commit list accordingly, but the file changes have been updated.

### How to review/test

- `yarn test` should run the tests
- several errors remain (some are fixed on https://github.com/demos-europe/demosplan-core/pull/4482)
  - DpLayerLegend.spec.js
  - DpAdminLayerList.spec.js
  - DpEditFieldMultiSelect.spec.js
  - FilterModalSelectItem.spec.js
  - DpPublicLayerListCategory.spec.js
  - DpBulkEditStatement.spec.js
  - StatementExportModal.spec.js
- to not make this PR even bigger, they will be fixed separately

### Linked PRs
- [ ] https://github.com/demos-europe/demosplan-core/pull/4482

### PR Checklist

- [x] Create/Update tests
- [x] Link all relevant tickets

PR #4491, opened 2025-03-24
test(refs DPLAN-15256): fix export modal test

hwiem:
### Ticket
[DPLAN-15256](https://demoseurope.youtrack.cloud/issue/DPLAN-15256)

Fixes all errors in ExportModal.spec.js.

### How to review/test

Run ExportModal.spec.js, there should be no errors.

### PR Checklist

- [x] Create/Update tests
- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/statement/assessmentTable/AssessmentTableFilter.vue
+++ b/client/js/components/statement/assessmentTable/AssessmentTableFilter.vue
@@ -61,7 +61,8 @@
       class="c-at__controls o-sticky o-sticky--border space-stack-xs u-pt-0_5 u-pb-0_25"
       ref="header">
       <search-and-sorting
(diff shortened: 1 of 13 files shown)


salisdemos (approved):
As far as I can see it looks fine. I think in this state we should just merge quickly and check then in main if it has sideeffects :see_no_evil:

hwiem:
> I think in this state we should just merge quickly and check then in main if it has sideeffects 🙈

This makes me wonder how we should continue with the test branches (https://github.com/demos-europe/demosplan-core/pull/4472, https://github.com/demos-europe/demosplan-core/pull/4482). Maybe we should merge them in the current state and fix the remaining tests on one or several new branches?

PR #4492, opened 2025-03-24
f(refs:DPLAN-12867)  delete vote explanation

riechedemos:
### Ticket
DPLAN-12867

Delete vote explanation because it's already part of statement-publish

### How to review/test
login as walter west -> new stn -> check if there is just one info box "
Mit der Freischaltung wird die Mitzeichnungsfunktion automatisch aktiviert."


hwiem (approved):
Nice, thank you!

PR #4493, opened 2025-03-24
b(refs:DPLAN-15164):  radio btn spacing

riechedemos:
### Ticket
DPLAN-15164

Refactored to DpRadio to fix spacing issue.

### How to review/test
walter west -> verfahrensliste -> einreichende suchen

### DIFF
--- a/client/js/components/procedure/DpSearchProcedures.vue
+++ b/client/js/components/procedure/DpSearchProcedures.vue
@@ -28,14 +28,14 @@
           <dp-radio
             id="searchall"
             name="searchselection"
-            :value="all"
+            value="all"
             v-model="searchIn"
             :label="{ text: Translator.trans('search.all.procedures') }" />
         </div>
         <dp-radio
           id="searchselected"
           name="searchselection"
-          :value="selected"
+          value="selected"
           v-model="searchIn"
           :label="{ text: Translator.trans('select.procedures.search') }" />
       </fieldset><!--
--- a/client/js/components/procedure/DpSearchProcedures.vue
+++ b/client/js/components/procedure/DpSearchProcedures.vue
@@ -28,18 +28,16 @@
           <dp-radio
             id="searchall"
             name="searchselection"
-            :value="'all'"
+            :value="all"
             v-model="searchIn"
-            :label="{ text: Translator.trans('search.all.procedures') }"
-          />
+            :label="{ text: Translator.trans('search.all.procedures') }" />
         </div>
         <dp-radio
           id="searchselected"
           name="searchselection"
-          :value="'selected'"
+          :value="selected"
           v-model="searchIn"
-          :label="{ text: Translator.trans('select.procedures.search') }"
-        />
+          :label="{ text: Translator.trans('select.procedures.search') }" />
       </fieldset><!--
    --><div
         class="layout__item u-2-of-3 u-pl-0"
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/DpSearchProcedures.vue
```diff
@@ -25,30 +25,21 @@
...
             id="searchall"
             name="searchselection"
-            value="all"
-            v-model="searchIn">
-          <label
-            for="searchall"
-            class="inline">
-            {{ Translator.trans('search.all.procedures') }}
-          </label>
+            :value="'all'"
```
hwiem, line 31: Here and below: Why not
```
value="all"
```
?

client/js/components/procedure/DpSearchProcedures.vue
```diff
@@ -25,30 +25,21 @@
...
+            v-model="searchIn"
+            :label="{ text: Translator.trans('search.all.procedures') }"
+          />
```
hwiem, line 34:
```suggestion
            :label="{ text: Translator.trans('search.all.procedures') }" />
```

client/js/components/procedure/DpSearchProcedures.vue
```diff
@@ -25,30 +25,21 @@
...
-          v-model="searchIn">
-        <label
-          for="searchselected"
-          class="inline align-text-bottom">
-          {{ Translator.trans('select.procedures.search') }}
-        </label>
+          :value="'selected'"
+          v-model="searchIn"
+          :label="{ text: Translator.trans('select.procedures.search') }"
+        />
```
hwiem, line 42:
```suggestion
          :label="{ text: Translator.trans('select.procedures.search') }" />
```

hwiem (commented):
Nice! Just a few nitpicks

client/js/components/procedure/DpSearchProcedures.vue
riechedemos: Here’s why each dp-radio needs a unique value:
- **Purpose of Radio Buttons**: Radio buttons are designed for exclusive selection. When a group of radio buttons share the same name attribute, only one can be selected at a time. The value attribute of the selected radio button is stored in the v-model data property. If both radio buttons had the same value, selecting either would yield the same result in the v-model.
- **Differentiate Between Options**: In this case, one radio button is for searching "all" procedures, and the other is for searching "selected" procedures. These two options should correspond to different values ("all" and "selected") to function correctly.

riechedemos: fixed 95ac40bc1ec5ccafe21c4669e3703a80162057da

client/js/components/procedure/DpSearchProcedures.vue
riechedemos: fixed 95ac40bc1ec5ccafe21c4669e3703a80162057da

client/js/components/procedure/DpSearchProcedures.vue
riechedemos: fixed 95ac40bc1ec5ccafe21c4669e3703a80162057da

client/js/components/procedure/DpSearchProcedures.vue
hwiem: It has to be without the colon:
```
value="all"
```

riechedemos: it is now :)

salisdemos (approved):
looks legit. Thanks

client/js/components/procedure/DpSearchProcedures.vue
hwiem: Did you maybe not push the changes? It still shows me `:value="all"` :thinking:

riechedemos: Ups sorry. But now 6993696875bd350d82d24c62a499df252a609985

hwiem (approved):
Nice, thank you!

PR #4494, opened 2025-03-25
feat (refs DPLAN-15242): null is a valid return type

martendemos:
### Ticket
DPLAN-15242

Null as return is also allowed.
Additionally set string as parameter type and remove now unnecessary docblock


mussbach (approved):
lgtm, code review only

PR #4421, opened 2025-03-05
F dplan 23404 obscure text segments

riechedemos:
### Ticket
DPLAN-XXXXX
or
https://www.dev.diplanung.de/DefaultCollection/EfA%20DiPlanung/_workitems/edit/XXXXX

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Add/Update data-cy attributes ([conventions](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/twig_html.html#guideline-for-naming-cypress-hooks))
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Link all relevant tickets
- [ ] Move the tickets on the board accordingly
- [ ] Update changelog

PR #4499, opened 2025-03-26
b(refs-DPLAN-13110) delete obscure option in new statement 

riechedemos:
### Ticket
DPLAN-15298

Deleted obscure option in DpEditor in NewStatement component since it should just be available in segements. 

### How to review/test
Create new stn and take a look at the icons

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-core/pull/4421


salisdemos (approved):
legit

PR #4501, opened 2025-03-26
DPLAN-4237 set NextValidExternalIdForProcedure on stn sync instead ofstrictly setting its source STN value

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-4237/STN-Nr-wird-doppelt-vergeben

Description:
set NextValidExternalIdForProcedure on stn sync instead ofstrictly setting its source STN value
image
will become in targetProcedure that already holds a M2 id now
image

If you couple a procedure with a token obtained by an hearing-agency-procedure and you have your own statements created  within that same procedure before the hearing-agency starts to sync theirs - multiple STNs with the same externalId will be created since we strictly set the externalId of the source Statement.
image

I am not certain this issue is the root cause of this ticket - but this hearing-agency sync-feature got introduced before the ticket got created - and that makes it possible at least. 

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


breierDemos (approved):
lgtm code review only

PR #4506, opened 2025-03-27
b(refs: DPLAN-15161) conditional rendering DpTreeListToggle

riechedemos:
### Ticket
DPLAN-15161

Add conditional rendering for the DpTreeListToggle component with a new hasContent prop and a computed property for evaluation.

### How to review/test
Verfahren -> Schlagworte -> Compare behavior of the little toggle icon : no tags (no toggle icon) /  tags (toggle icon)

### DIFF
--- a/client/js/components/tags/TagsList.vue
+++ b/client/js/components/tags/TagsList.vue
@@ -8,9 +8,10 @@
 
     <dp-tree-list
       v-if="transformedCategories"
-      class="mb-4"
       align-toggle="center"
-      :tree-data="transformedCategories"
+      :branch-identifier="isBranch"
+      class="mb-4"
+      :has-content="hasContent"
       :options="{
         branchesSelectable: false,
         dragAcrossBranches: true,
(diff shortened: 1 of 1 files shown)


client/js/components/tags/TagsList.vue
```diff
@@ -126,6 +127,10 @@ export default {
       TagTopic: 'items'
     }),
 
+    hasContent () {
+      return this.transformedCategories && this.transformedCategories.length > 0
```
salisdemos, line 131: You can beautify this by:
```suggestion
      return this.transformedCategories?.length > 0
```

And: Are You sure, You have to test that at all? as far as i can see this.transformedCategories should always return an array, even if there are no TagTopics

salisdemos (approved):
Thanks. Just a noneblocking nitpic

client/js/components/tags/TagsList.vue
riechedemos: true, I changed it 138c965233086da83fe4d013c2e95bedccf8beb4

PR #4510, opened 2025-03-28
use meaningful data-cy names instead of loop values

sakutademos:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-12672/DiplanROG-neue-Stellungnahme-Hook-zurucksetzen

**Description:** This PR updates `data-cy` values:  ad to each each `subnav` item a data-cy attribute, to use it for testing purposes.

**Changes to the existing hooks (testers have been informed):**
createStatement --> listStatements:createStatement
listStatements:statementList --> listOriginalStatements:list
statement.new --> listOriginalStatements:createStatement

**Assessment table:**
assessmentTable --> assessmentTable:view
assessmentTableOriginal --> assessmentTable:viewOriginal
createStatement --> ssessmentTable:createStatement

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/base_pageheader.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/base_pageheader.html.twig
@@ -74,7 +74,7 @@
                 <li class="inline-block {{ loop.last ?: 'u-mr-0_5' }}">
                     <a
                         class="{{ item.current|default == true  ? 'o-link--active' : 'o-link--default' }} whitespace-nowrap"
-                        :data-cy="`pageHeader:item:{{ loop.index }}`"
+                        :data-cy="`pageHeader:item:{{ item.datacy }}`"
                         href="{{ item.href }}"
                         {% if item.datacy|default != '' %}data-cy="{{ item.datacy }}"{% endif %}
                         {% if item.id|default != '' %}id="{{ item.id }}"{% endif %}
--- a/composer.lock
+++ b/composer.lock
@@ -19861,21 +19861,21 @@
         },
         {
             "name": "goetas-webservices/xsd2php",
-            "version": "0.4.13",
+            "version": "0.4.14",
             "source": {
                 "type": "git",
                 "url": "https://github.com/goetas-webservices/xsd2php.git",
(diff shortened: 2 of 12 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanCore/includes/base_pageheader.html.twig
```diff
@@ -74,7 +74,7 @@
                 <li class="inline-block {{ loop.last ?: 'u-mr-0_5' }}">
                     <a
                         class="{{ item.current|default == true  ? 'o-link--active' : 'o-link--default' }} whitespace-nowrap"
-                        :data-cy="`pageHeader:item:{{ loop.index }}`"
+                        :data-cy="`pageHeader:item:{{ item.datacy }}`"
```
hwiem, line 77: What happens if `datacy` is not defined?

sakutademos: then i would be undefined or null :D but I hope it won’t happen, because they always need the same naming for testing. If I choose a different one as the default value, I think it would not make sense, because tests will crash

hwiem: but there are usages of `base_pageheader` with `subnav` without `datacy` defined for the item(s), so I think we should adjust these usages

sakutademos: ach, got you! I was too focused on other things :D You are totally right, there are some usages without datacy, so it needs to be adjusted and will take more time. I will do it a bit later. thx for spotting it!

sakutademos: done here https://github.com/demos-europe/demosplan-core/pull/4510/commits/6cb8042e3b2efbbae34393db3c2f965225e6a126

hwiem (approved):
lgtm, thanks!

PR #4502, opened 2025-03-26
feat DPLAN-15322: Implement support for multiple custom field types i…

sierrazamudiodemos:
…n the Procedure

- Updated CustomFieldList to handle different custom field types (RadioButtonField, DropdownField, etc.)
- Modified fromJson method to detect and instantiate appropriate custom field classes
- Ensured each custom field class implements CustomFieldInterface with fromJson and toJson methods

### Ticket
DPLAN-XXXXX
or
https://www.dev.diplanung.de/DefaultCollection/EfA%20DiPlanung/_workitems/edit/XXXXX

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Add/Update data-cy attributes ([conventions](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/twig_html.html#guideline-for-naming-cypress-hooks))
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Link all relevant tickets
- [ ] Move the tickets on the board accordingly
- [ ] Update changelog

PR #4511, opened 2025-03-28
feat(refs DPLAN-15338, #AB26283): add interface to create segments fields

hwiem:
### Ticket
[DPLAN-15338](https://demoseurope.youtrack.cloud/issue/DPLAN-15338/ADO-Issue-26283-Hinzufugen-konfigurierbare-Felder-auf-Abschnittsebene-Part-1-Anlegen)

Adds a new view for creating fields (form controls) for segments. These will later be translated to single selects (for now) in the interface in certain views where segments are displayed (not part of this PR).
The view is available both in a procedure and in a procedureTemplate.

### How to review/test

- Go to a procedure or procedureTemplate
- Navigate to 'Konfigurierbare Abschnittsfelder' in the side menu (on the left (procedure) or on the right (procedureTemplate))
- create a new field with options - it should work. no success message will be displayed for now (known issue)
- create a new field with the same name as an existing one - it should not work
- create a new field with two options that have identical names - it should not work

**Known issues**
- expanding the row in the data table to see all options (if there are more than 2) does not work right now

### Tasks
- [x] the backend route needs to be added (for now, an existing route is adjusted to render the new template. this has to be reverted!)
- [x] the menu label in the dashboard probably needs to be renamed
- [x] the hint displayed at the top of the new view has not been defined yet and will have to be changed
- [x] create integration branch from main
- [x] merge https://github.com/demos-europe/demosplan-core/pull/4502

### PR Checklist

- [x] Run `yarn lint`


client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -0,0 +1,401 @@
...
+          v-model="newFieldOptions[0]"
+          maxlength="250"
+          required />
+        <dp-input
+          id="newFieldOption:2"
+          class="mb-2 w-[calc(100%-26px)]"
+          data-cy="customFields:newFieldOption2"
+          v-model="newFieldOptions[1]"
+          maxlength="250"
+          required />
```
salisdemos, line 33: can't we move them into `additionalOptions`? (and then rename the var)

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -0,0 +1,401 @@
...
+          :key="`option:${idx}`">
+          <div class="w-[calc(100%-26px)] inline-block mb-2">
+            <dp-input
+              v-model="newFieldOptions[idx + 2]"
+              :id="`option:${newFieldOptions[idx + 2]}`"
+              :data-cy="`customFields:newFieldOption${idx + 2}`"
+              maxlength="250" />
+          </div>
+          <dp-button
+            class="w-[20px] inline-block ml-1"
```
salisdemos, line 46: if done like suggested above, we can then check here something like `v-if="idx >= 2"`

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -0,0 +1,401 @@
...
+          dplan.notify.confirm(Translator.trans('confirm.saved'))
+        })
+        .catch(err => {
+          console.error(err)
+        })
+        .finally(() => {
+          this.isLoading = false
+          this.resetNewFieldForm()
+          this.fetchCustomFields()
+        })
```
salisdemos, line 349: why does this lives here in the list and not in the create form? 
If this would be in the create form, we could remove the watcher and the code is where the interaction happens

client/js/components/procedure/admin/CreateCustomFieldForm.vue
```diff
@@ -0,0 +1,144 @@
+<template>
+  <div data-dp-validate="createCustomFieldForm">
+    <div
+      v-if="!isOpen"
+      class="text-right mb-4">
+      <dp-button
+        data-cy="customFields:addField"
+        @click="open"
+        :text="Translator.trans('field.add')" />
```
salisdemos, line 9: I opt  for a morge generic aproach and to use `add` here and `name` instead of `field.name` somewhere else

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -0,0 +1,401 @@
...
+      dismissible
+      :dismissible-key="helpTextDismissibleKey"
+      :message="Translator.trans('segments.fields.edit.info')"
+      type="info" />
+
+    <create-custom-field-form
+      :handle-success="isSuccess"
+      :is-loading="isLoading"
+      @save="customFieldData => saveNewField(customFieldData)">
+      <div>
```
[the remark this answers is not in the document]
hwiem: I don't fully understand it, but it seems to be needed so the name field is validated :thinking:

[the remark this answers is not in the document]
salisdemos: :-/ ok.

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
hwiem: I think the concern is very valid and we should move it. Originally the idea was to use the create form also in the places view, since the interface is very similar. In the end I decided against it though because we don't consider places to be custom fields, and the interfaces are probably going to develop into different directions over time. So the form component also doesn't need to have a slot anymore and the option inputs can be moved into it, too, but I'd prefer to refactor that after merging this PR, as it might delay the feature freeze otherwise.

client/js/components/procedure/admin/CreateCustomFieldForm.vue
hwiem: Do you mean the translation keys?

salisdemos: yes

hwiem: I didn't find `field.name`, so I changed `field.add` only 2165963e0a8dbbe1a6478dd2e19b97e25d654c24

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -0,0 +1,401 @@
...
+    /**
+     * CustomFields reduced to the format we need in the FE
+     * @return {({id: *, name: *, description: *, options: *, open: boolean}|undefined)[]}
+     */
+    customFieldsReduced () {
+      return Object.keys(this.customFields)
+        .map(key => {
+          if (this.customFields[key]) {
+            const { id, attributes } = this.customFields[key]
+            const { description, name, options } = attributes
```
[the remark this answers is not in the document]
hwiem: done in a0c25b817eee4872a5e9f3a7fa5b171d0e84337f

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -0,0 +1,401 @@
...
+      const identicalNames = options.filter(optionName => optionName === name)
+
+      return identicalNames.length <= 1
+    },
+
+    /**
+     * Fetch custom fields that are available either in the procedure or in the procedure template
+     */
+    fetchCustomFields () {
+      this.isInitiallyLoading = true
```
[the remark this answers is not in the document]
hwiem: Done in e180bd7b99fa9faa71f98d743fc92403e4b7b1b2

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -0,0 +1,401 @@
...
+    },
+
+    displayedOptions () {
+      return (rowData) => {
+        return rowData.open ? rowData.options : rowData.options.slice(0, 2)
+      }
+    },
+
+    helpTextDismissibleKey () {
+      return `${this.currentUserId}:procedureAdministrationSegmentsFieldsHint`
```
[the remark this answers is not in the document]
hwiem: Done in cc800b791d0a98ec247bfa91487d0e499dcbfeb5

client/js/components/procedure/admin/CreateCustomFieldForm.vue
salisdemos: sorry. it was`field_name` ...

hwiem: Good idea, done in a817bac869522aea9771954ebccd2c2f405693cd

PR #4514, opened 2025-03-31
style(refs DPLAN-12921,  #AB9149): fix some linting errors + warnings

hwiem:
### Ticket
[DPLAN-12921](https://demoseurope.youtrack.cloud/issue/DPLAN-12921)

- Updates eslint-plugin-vue to a version compatible with vue 3, so we get correct warnings and errors
- eslint-plugin-vue requires vue-eslint-parser, so it was installed
- Fixes errors and warnings that could be fixed automatically or easily, so a number of errors and a lot of warnings still remain

**Before:**
✖ 800 problems (82 errors, 718 warnings)

**After:**
✖ 335 problems (42 errors, 293 warnings)

### PR Checklist

- [x] Run `yarn lint`


client/js/components/procedure/admin/InstitutionTagManagement/InstitutionTagManagement.vue
```diff
@@ -26,7 +26,7 @@
       :is-active="activeTabId === 'tagList'"
       :label="Translator.trans('tag.administrate')">
       <slot>
-        <tag-list @tagIsRemoved="institutionListReset" />
+        <tag-list @tag-is-removed="institutionListReset" />
```
gruenbergerdemos, line 29: I think the emit is still `tagIsRemoved`. Does vue recoginze this with `@tag-is-removed` (same pattern as props?)?

client/js/components/user/ToggleForm.vue
```diff
@@ -46,17 +46,14 @@ export default {
     abort () {
       if (this.formId !== '') {
         document.querySelector(`form#${this.formId}`).reset()
-      } else {
-        this.$emit('form-abort')
```
gruenbergerdemos, line 50: Why were these emits deleted? Do we not need them anymore? (also see below save()).

gruenbergerdemos (changes requested):
Overall lgtm, thanks for fixing! Just two questions about emits.

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionTagManagement.vue
hwiem: It's actually the recommended way of doing it in vue 3 (see https://vuejs.org/guide/components/events + https://eslint.vuejs.org/rules/v-on-event-hyphenation.html), but I'm not sure we want to do it like that. I like it better when you can search for the event name and find all usages, so I disabled the eslint rule for now. This change was accidental, reverted in 92dba4713b5255f4495849a5f5e4d3d8b7d5bdea

client/js/components/user/ToggleForm.vue
hwiem: I could not find any usages, so I assumed they can be removed.

gruenbergerdemos (approved):

PR #4515, opened 2025-03-31
refac(DPLAN-15295): splitStatement issue: properly remove segment from the array 

sakutademos:
### Ticket
[DPLAN-15295
](https://demoseurope.youtrack.cloud/issue/DPLAN-15295/Abschnitt-kann-nicht-erstellt-gespeichert-werden-STN-aufteilen)

**Description:** The way the object was deleted from the array was incorrect: using `delete` removes the value but not the index, resulting in a `null` hole in the array. Use the `splice()` method instead to properly remove the element.

### How to review/test
Stellungnahmen > STN aufteilen > Text markieren > speichern

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/client/js/store/statement/SplitStatementStore.js
+++ b/client/js/store/statement/SplitStatementStore.js
@@ -237,7 +237,7 @@ const SplitStatementStore = {
              * (because often that are closing or opening tags)
              * and should probably not be needed in a real world scenario.
              */
-            .filter(segment => (segment.charEnd - segment.charStart) > 10)
+            .filter(segment => segment && (segment.charEnd - segment.charStart) > 10)
 
           // Check if we are getting overlapping segments from pipeline that would cause errors
           if (doUpdate) {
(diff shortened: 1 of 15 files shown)


client/js/store/statement/SplitStatementStore.js
```diff
@@ -50,7 +50,7 @@ const SplitStatementStore = {
     deleteSegment (state, id) {
       const index = state.segments.findIndex((el) => el.id === id)
       if (index >= 0) {
-        delete state.segments[index]
+        state.segments.splice(index, 1)
```
spiess-demos, line 53: This reminds me on a similar problem i was facing just a few days ago in DiplanWissen. What type is `state.segments`, Object or Array? If it is an Object, operating on its keys as if in were the numeric keys of an array would be incorrect.

After being hinted on that topic, i eventually [landed on this solution.](https://www.dev.diplanung.de/DefaultCollection/DiPlanWissen/_git/diplan-wissen/pullrequest/8447)

PR #4522, opened 2025-04-01
b(refs: DPLAN-12821) change abort route 

riechedemos:
### Ticket
DPLAN-12821

Changed abort route to to direct the user back to the site whey they came from.

### How to review/test
Try to create new stn and abort as Datenerfasser

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanAssessmentTable/DemosPlan/dhtml/v1/assessment_table_new_statement.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanAssessmentTable/DemosPlan/dhtml/v1/assessment_table_new_statement.html.twig
@@ -14,7 +14,12 @@
     } %}
 
     {% set form_action = templateVars.formAction|default(path("DemosPlan_statement_new_submitted", {'procedureId': templateVars.table.procedure.ident})) %}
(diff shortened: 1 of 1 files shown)


templates/bundles/DemosPlanCoreBundle/DemosPlanAssessmentTable/DemosPlan/dhtml/v1/assessment_table_new_statement.html.twig
```diff
@@ -14,7 +14,12 @@
     } %}
 
     {% set form_action = templateVars.formAction|default(path("DemosPlan_statement_new_submitted", {'procedureId': templateVars.table.procedure.ident})) %}
-    {% set abort_path = templateVars.abortPath|default(path("dplan_assessmenttable_view_original_table", { 'procedureId': templateVars.table.procedure.ident })) %}
+    {% if hasPermission('area_statement_data_input_orga') %}
```
hwiem, line 17: If the user is "Datenerfasser*in", should it always be this route or might it also be defined via `templateVars.abortPath` (so that it should be
```
{% set abort_path = templateVars.abortPath|default(path('DemosPlan_statement_orga_list', { 'procedureId': templateVars.table.procedure.ident })) %}
```
)?

hwiem (approved):
Lgtm!

PR #4524, opened 2025-04-02
bug (refs DPLAN-15337): Use dpDraggable

gruenbergerdemos:
### Ticket
DPLAN-15337

- Use DpDraggable instead of vue draggable. 
- Fix some linter issues

### How to review/test
Privatperson -> Öffentliches Verfahren -> Save statement as draft -> Meine Entwürfe -> Page should not be empty and the drafts should be shown

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
+++ b/client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
@@ -13,32 +13,26 @@
       v-if="transformedStatements.length === 0"
       :message="Translator.trans('statement.list.empty')"
       type="info" />
-    <dp-draggable
-      v-else
-      v-model:content-data="transformedStatements"
-      class="space-stack-m"
-      data-cy="publicStatementList"
-      :is-draggable="hasPermission('feature_statements_manualsort')"
-      @change="saveSort">
+    <div class="space-stack-m">
       <dp-public-statement
         v-for="(statement, idx) in transformedStatements"
         v-bind="statement"
         :key="idx"
-        @open-map-modal="openMapModal"
-        @open-statement-modal-from-list="(id) => $parent.$emit('open-statement-modal-from-list', id)"
         :menu-items-generator="menuItemCallback"
         :procedure-id="procedureId"
         :show-author="showAuthor"
-        :show-checkbox="showCheckbox" />
-    </dp-draggable>
-    <dp-map-modal
-      ref="mapModal"
-      :procedure-id="procedureId" />
+        :show-checkbox="showCheckbox"
+        @open-map-modal="openMapModal"
+        @open-statement-modal-from-list="(id) => $parent.$emit('open-statement-modal-from-list', id)"/>
+      <dp-map-modal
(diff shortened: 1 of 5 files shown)


hwiem:
Does drag + drop work for you? :thinking:

gruenbergerdemos:
No, unfortunately not. But at the moment nobody has this permission to drag and drop at this page and with this change the page does work again.
I was already checking out the drag and drop functionality else where. Do we have further knowledge on this? Is it a known issue?

hwiem:
> No, unfortunately not. But at the moment nobody has this permission to drag and drop at this page

Interesting, I have been wondering for a while if draggable is needed here anymore. For now I think we can leave it like this, but maybe we should discuss with POs if we can remove it.

 > I was already checking out the drag and drop functionality else where. Do we have further knowledge on this? Is it a known issue?

We had to upgrade `DpDraggable` to a vue3-compatible version, and unfortunately this usage of draggable was not updated. Other usages of `DpDraggable` should work though. If it doesn't work here I guess it's because of the permission?

gruenbergerdemos:
> We had to upgrade DpDraggable to a vue3-compatible version, and unfortunately this usage of draggable was not updated. Other usages of DpDraggable should work though. If it doesn't work here I guess it's because of the permission?

The drag and drop does not work for map layers and categories in the admin area. But then it is maybe just this instance. I realised it while trying to fix the test for DpPublicLayerListCategory.

client/js/components/statement/publicStatementLists/DpPublicStatementList.vue
```diff
@@ -13,12 +13,12 @@
       v-if="transformedStatements.length === 0"
       :message="Translator.trans('statement.list.empty')"
       type="info" />
-    <draggable
+    <dp-draggable
       v-else
-      v-model="transformedStatements"
+      v-model:content-data="transformedStatements"
```
hwiem, line 18: If I remember correctly. v-model doesn't work with the new implementation of DpDraggable. So it should just be
```
:content-data="transformedStatements"
```
While we don't need to fix draggable right now, we could still change this in case we need to fix it in the future.

gruenbergerdemos:
> Interesting, I have been wondering for a while if draggable is needed here anymore. For now I think we can leave it like this, but maybe we should discuss with POs if we can remove it.

Yes, good point. I asked PO and maybe I can already remove it here.

hwiem (approved):
Just a nitpick.. Rest looks good, and thanks for the eslint fixes :heart:

gruenbergerdemos:
> Yes, good point. I asked PO and maybe I can already remove it here.

The feedback was that we don't need it anymore, so I removed the draggable completely here. [d4d9d1e](https://github.com/demos-europe/demosplan-core/pull/4524/commits/d4d9d1ec8b894f7dae267d795498932bbbfc0f55)

hwiem (approved):
Nice, thanks!

PR #4356, opened 2025-02-18
f DPLAN-14953 escape disallowed tags :

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-14953

Description: 

- [X] Link all relevant tickets
- [X] Move the tickets on the board accordingly

PR #4528, opened 2025-04-02
chore (refs DPLAN-14953): escape disallowed chars in statement text

mussbach:
### Ticket
DPLAN-14953

This is an alternative implementation of #4356 that further improves security by using the HTMLPurifier 

### PR Checklist

- [x] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


bakridemos (changes requested):
Manually replacing < and > with ';lt' and ';gt' instead of using the correct  '<' and '>' is unfortunately not working . ';lt' and ';gt' are not valid HTML entities and will not be interpreted correctly. They will be interpreted like a normal text and displayed like this too, i mean instead of '<' or '>', ;lt' and ';gt' will be showed.

mussbach:
I do not have the time to improve this one, lets go further with the initial PR #4528

PR #4531, opened 2025-04-03
f DPLAN-15354 keep customer ID after the soft delete procedure

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15354/Customer-ID-nach-dem-Soft-Delete-Verfahren-behalten

Description: Keep customer ID after the soft delete procedure

### How to review/test
 - delete a procedure
 - check in the DB if the deleted procedure still has its customer Id

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
@@ -910,7 +910,6 @@ class ProcedureService extends CoreService implements ProcedureServiceInterface
             foreach ($procedureIds as $procedureId) {
                 $data = [
                     'ident'    => $procedureId,
-                    'customer' => null,
                     'deleted'  => true,
                 ];
 
--- a/demosplan/DemosPlanCoreBundle/Repository/ProcedureRepository.php
+++ b/demosplan/DemosPlanCoreBundle/Repository/ProcedureRepository.php
@@ -563,7 +563,6 @@ class ProcedureRepository extends SluggedRepository implements ArrayInterface, O
         }
         if (array_key_exists('deleted', $data)) {
             $procedure->setDeleted($data['deleted']);
-            $procedure->setCustomer(null);
             $procedure->setProcedureCategories([]);
             $procedure->setDeletedDate(Carbon::now());
         }


martendemos (changes requested):
I think there is another change needed in\demosplan\DemosPlanCoreBundle\Logic\Procedure\ProcedureService::calculateCopyMasterId
In case of the defaultProcedureBlueprint of a customer is deleted == true, this shouldn't be used as blueprint on creation of a new procedure.

`    public function calculateCopyMasterId(?string $incomingCopyMasterId = null): string
    {
        // use global default blueprint as default anyway:
        $masterTemplateId = $this->getMasterTemplateId();
        $incomingCopyMasterId ??= $masterTemplateId;

        // T15664: in case of globalMasterBlueprint is set,
        // use customer master blueprint if existing, (instead of global masterblueprint)
        if ($masterTemplateId === $incomingCopyMasterId) {
            $customerBlueprint = $this->customerService->getCurrentCustomer()->getDefaultProcedureBlueprint();
            if ($customerBlueprint instanceof Procedure && !$customerBlueprint->isDeleted()) {
                $incomingCopyMasterId = $customerBlueprint->getId();
            }
        }

        return $incomingCopyMasterId;
    }`

bakridemos:
> calculateCopyMasterId

honestly i don't really see where is the problem, i mean normal procedures and blueprints have already the customer id,  MasterBlueprint id is saved directly in the customer and to my knowledge users could not adjust/delete it. Blueprint are not showed or used once deleted is true, the customer id won't/should not make any difference. i tried it once more and it works. Or do you a better way to test it ?

martendemos:
> > calculateCopyMasterId
> 
> honestly i don't really see where is the problem, i mean normal procedures and blueprints have already the customer id, MasterBlueprint id is saved directly in the customer and to my knowledge users could not adjust/delete it. Blueprint are not showed or used once deleted is true, the customer id won't/should not make any difference. i tried it once more and it works. Or do you have a better way to test it ?

The case I see is that a customer blueprint with deleted = true is used on creation of a new procedure.
This happens in case of no blueprint is selected by the user on create a new procedure (or even a new blueprint)
Then a "deleted = true blueprint" will be used as blueprint and in the logic of creation of a new procedure (or blueprint) is to complex to risk a using a deleted blueprint there.
for example: I don't know if the deleted flag would be copied into the new procedure, which would be obviously a bug.

bakridemos:
> > > calculateCopyMasterId
> > 
> > 
> > honestly i don't really see where is the problem, i mean normal procedures and blueprints have already the customer id, MasterBlueprint id is saved directly in the customer and to my knowledge users could not adjust/delete it. Blueprint are not showed or used once deleted is true, the customer id won't/should not make any difference. i tried it once more and it works. Or do you have a better way to test it ?
> 
> The case I see is that a customer blueprint with deleted = true is used on creation of a new procedure. This happens in case of no blueprint is selected by the user on create a new procedure (or even a new blueprint) Then a "deleted = true blueprint" will be used as blueprint and in the logic of creation of a new procedure (or blueprint) is to complex to risk a using a deleted blueprint there. for example: I don't know if the deleted flag would be copied into the new procedure, which would be obviously a bug.
> 
> but anyways: I need see that his fixes are done for release_ewm, I need these for main and even some more. I hope it is ok that I just cherrypicked you commit to use it in my PR: #4540

i see what u means but this exactly what i meant, the User can not copy or use a deleted bluprint, once deleted is true it will be not available anymore

martendemos:
> i see what u means but this exactly what i meant, the User can not copy or use a deleted bluprint, once deleted is true it will be not available anymore

No, not the user but the system does it by its own. 
In ProcedureService::calculateCopyMasterId: In case of no blueprint is choosen by the user, the default procedure blueprint of the customer will be used, without checking if this is actually deleted.

PR #4533, opened 2025-04-03
fix(DPLAN-15276): missing parentheses in a function

sakutademos:
### Ticket
[DPLAN-15276](https://demoseurope.youtrack.cloud/issue/DPLAN-15276/Nicht-moglich-um-FAQ-Items-oder-Kategorien-zu-loschen)

**Description**: This PR fixes an issue where a function was not triggered because the function call was missing parentheses.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


hwiem (approved):
Thanks!

client/js/components/faq/DpFaqItem.vue
```diff
@@ -257,7 +257,7 @@ export default {
           })
         }
 
-        this.queue.push(deleteAction)
+        this.queue.push(deleteAction())
```
salisdemos, line 260: Are You sure? If so, it should be changed in Line 242 too. From the design it should not be added. They get added in linke 183 in the whatcher

salisdemos (changes requested):

client/js/components/faq/DpFaqItem.vue
salisdemos: could it be, that the watch does not get triggered for some reason?

sakutademos: thx for spotting this. I was not deeply involved with this topic, so you are right, the watcher was not triggered in this particular case... It was activated only in the computed variable by the `set` action, but somehow the reactivity was not detected in the `deleteFaqItem` and the watcher was not triggered...

The problem could be: _The issue might be that pushing a function reference (without calling it) doesn’t always trigger the watcher. Vue 3 should normally detect array mutations (like a change in length when you push a new element), but if you’re pushing a function reference that isn’t immediately invoked, Vue might not detect a significant enough change to trigger the watcher reliably._

Maybe a better approach would be to use a function instead called manually after adding an action to the queue: this function could take the first queued function, execute it and then remove it from the queue (same FIFO execution model) 

done here https://github.com/demos-europe/demosplan-core/pull/4533/commits/5d1e7b641e18ce1e68c804885491883808ebf970

salisdemos (approved):
I think as a bugfix Your solution is absolutely fine. 
But the more I stare at the code, the less I like it. I think it would be good to refactor it a bit, that we don't have side effects in the computed.
I can imagine to build it with a function call on change instead using v-model.
That would just be a small change but clarify that changing things triggers an action without  reading the computed where I don't expect it.

sakutademos:
> I think as a bugfix Your solution is absolutely fine. But the more I stare at the code, the less I like it. I think it would be good to refactor it a bit, that we don't have side effects in the computed. I can imagine to build it with a function call on change instead using v-model. That would just be a small change but clarify that changing things triggers an action without reading the computed where I don't expect it.

do you find it better? --> https://github.com/demos-europe/demosplan-core/pull/4533/commits/94bee2453e7e16493d6789590c90f7cd9053e51c

salisdemos (approved):
I think this is much better, Yes. Thank You.
Just to be sure. You checked if v-model and @input work well togerther and don't reveal any race condition effects? (like the function is called before the state is updated)

sakutademos:
> I think this is much better, Yes. Thank You. Just to be sure. You checked if v-model and @input work well togerther and don't reveal any race condition effects? (like the function is called before the state is updated)

yes, i forgot to change it on `value`! thx again for spotting this! :))) https://github.com/demos-europe/demosplan-core/pull/4533/commits/a5a4d0d29fbd8e5d823630ee43faf7263425a437

PR #4535, opened 2025-04-03
B dplan 15049 get config default map fields

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-15049/Karte-soll-ganz-Deutschland-zeigen-nicht-nur-SH

Adjust the ProcedureMappingResourceType:
- getBoundingBox and mapExtend from masterTemplate
- If not available, then get it from config
- I added a unit test for the boundingBox, when PR accepted, I will add for mapExtent


mussbach (approved):
one minor remark, otherwise lgtm, code review only

PR #4537, opened 2025-04-03
B dplan 15192 fix wizard btns

riechedemos:
### Ticket
DPLAN-15192

Use element.style.display to make manual btns visible

### How to review/test
login as Walter West -> Grundeinstellungen -> Assistenten starten


hwiem (commented):
Just curious, why did the previous solution not work?

hwiem (approved):
Lgtm, code review only

riechedemos:
> Just curious, why did the previous solution not work?

My best guess would be that setAttribute overwrites some styles.

PR #4541, opened 2025-04-04
fix: Require location always.

salisdemos:
This is neccessary to interact with 3rd-Party systems/addons that require this field too


elmasdemos (approved):
lgtm, code review

mussbach (approved):
lgtm, code review only, thx

PR #4554, opened 2025-04-07
fix(refs: DPLAN-1998) add aria-hidden to step two

riechedemos:
### Ticket
DPLAN-1998

Add aria-hidden to step two to not focus empty inputs.

### How to review/test
Create a Stn a citizen and use screen reader


hwiem (approved):
Interesting :) Fix lgtm

PR #4559, opened 2025-04-08
f-DPLAN-15372 set visibility group id to null when unset overlays

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15372/Produktion-BB-Bauleitplanung-Kartenlayer-lassen-sich-ggf.-nicht-einblenden-und-nicht-editieren-26426

DescriprionÄ:  set visibility group id to null when unset overlays


salisdemos (approved):
Thanks!

PR #4561, opened 2025-04-08
bug(refs DPLAN-15421): Enable saving button for documents

gruenbergerdemos:
### Ticket
DPLAN-15421

Saving button should also be enabled if we already have an element set and only change the document or paragraph. Also refactor logic for better understanding.

### How to review/test
Verfahren -> Stellungnahmen -> Detailansicht -> Bearbeiten -> Orts- und Dokumentenbezug -> select different elements, documents and paragraphs


hwiem (approved):
Lgtm + works!

PR #4564, opened 2025-04-08
fix(refs DPLAN-15379): use updated data for checks

hwiem:
### Ticket
[DPLAN-15379](https://demoseurope.youtrack.cloud/issue/DPLAN-15379/Beim-wahlen-von-ein-Planungsdokument-Uncaught-in-promise-TypeError-this.options-is-undefined)

- it seems the update of `this.values` via `v-model` is not completed in the `@input` handler, so we now use the updated value directly that is emitted by `DpMultiselect`
- before, `this.values.element.id` always contained the previously selected element id or `undefined` (if no element had been selected before)

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/assessmenttable/DpNewStatement.vue
+++ b/client/js/components/assessmenttable/DpNewStatement.vue
@@ -224,11 +224,11 @@ export default {
       }
     },
 
-    checkForParagraphsAndFiles () {
+    checkForParagraphsAndFiles (selectedElement) {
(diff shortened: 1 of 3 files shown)


riechedemos (approved):
As far as I can tell, it looks good. I have tested it, and no errors are appearing in the console anymore.

PR #4567, opened 2025-04-09
fix(refs #AB9149): remove obsolete .native modifier

hwiem:
In Vue 3, the .native event modifier has been removed (see https://v3-migration.vuejs.org/breaking-changes/v-on-native-modifier-removed).

- To be able to remove its usages, we need to emit the corresponding event from the component that we place the handler on
- Focus events were also added to allow for keyboard navigation
- The segmentation editor can't be focused, but we get an eslint vuejs-accessiblity warning due to `'vuejs-accessibility/mouse-events-have-key-events'` that I did not manage to disable for the corresponding lines, so I added the event listeners instead
- Some events have been renamed to match our conventions
- In CardPaneCard, DpButtonIcon has been replaced with DpButton, and unless it's used in an Addon (edit: it is), I think we can now remove the component from demosplan-ui (we can't)

### How to review/test

- Navigate to the split statement view and use the keyboard to focus buttons in the card pane cards - the corresponding segment and the card should be highlighted.
- Linting errors and warning should have gone from 
```
✖ 336 problems (44 errors, 292 warnings)
```
to
```
✖ 320 problems (40 errors, 280 warnings)
```

### Linked PRs
- [x] https://github.com/demos-europe/demosplan-ui/pull/1243
- [x] https://github.com/demos-europe/demosplan-core/pull/4575

### PR Checklist

- [x] Run `yarn lint`

### DIFF
--- a/client/js/components/statement/splitStatement/CardPaneCard.vue
+++ b/client/js/components/statement/splitStatement/CardPaneCard.vue
@@ -12,7 +12,11 @@
     class="card space-stack-xs"
     :class="{ 'highlighted': isHighlighted }"
     :style="{ top: offsetTop, left: (position * 5) + 'px', 'margin-top': (position * 5) + 'px' }"
-    :id="'tag_' + segment.id">
+    :id="'tag_' + segment.id"
+    @focusin="$emit('focusin')"
+    @focusout="$emit('focusout')"
+    @mouseenter="$emit('mouseenter')"
+    @mouseleave="$emit('mouseleave')">
     <div class="u-pr-0_25">
       <i
         :title="Translator.trans('tags')"
@@ -106,6 +110,10 @@ export default {
     'check-card-overlap',
     'delete-segment',
     'edit-segment',
+    'focusin',
+    'focusout',
+    'mouseenter',
+    'mouseleave',
     'segment:confirm'
   ],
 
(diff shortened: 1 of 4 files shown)


client/js/components/statement/splitStatement/CardPaneCard.vue
```diff
@@ -53,10 +57,12 @@
       </span>
     </p>
 
-    <dp-button-icon
-      icon="fa-pencil"
+    <dp-button
```
sakutademos, line 60: just a small styling nitpick: an additional space is needed here, because when these two buttons receive focus, the focus outline overlaps the adjacent button

sakutademos (approved):
thx a lot, it works. just a small nitpick

client/js/components/statement/splitStatement/CardPaneCard.vue
hwiem: Thanks for testing it :heart: I fixed it in a way that the space is only added when focusing the buttons with the keyboard, since otherwise the space looks too big: a6769c970665fc639f25f141eeeee74a0980bbe3 and https://github.com/demos-europe/demosplan-ui/pull/1243

sakutademos: unfortunately, I am not able to test it right now, but the idea is interesting, thx a lot

sakutademos (approved):

PR #4568, opened 2025-04-10
fix(DPLAN-15309): ubble menu is missing when the user highlights text to the end of a paragraph

sakutademos:
### Ticket
[DPLAN-15309](https://demoseurope.youtrack.cloud/agiles/141-63/current?settings&tab=allgemeines&issue=DPLAN-15309)

**Description:** This PR fixes an issue where the bubble menu is missing when the user highlights text to the end of a paragraph.

There are two issues: 

- the highlight worked only within the text area, so if the user clicked even slightly outside of it, the rangeCreator plugin did not react because the mouseup event was triggered only inside the editor area. To fix this, a global mouseup handler (globalHandler) was added to the document in the view() lifecycle hook. 

- view.coordsAtPos method from ProseMirror does not always return a plain object with the essential coordinates (when clicking outside the editor zone). Because of that, an additional check was added to return only the important coordinates for the tippy 

### How to review/test
STN Aufteilen -> Text von rechts nach komplett links wählen -> kein + Bubble

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/client/js/lib/prosemirror/plugins.js
+++ b/client/js/lib/prosemirror/plugins.js
@@ -336,39 +336,57 @@ const rangeTracker = (rangeTrackerKey, schema, rangeChangeCallback = () => {}) =
  */
 const rangeCreator = (pluginKey, rangeEditingKey) => {
   let tippy = null
+
+  const handleGlobalMouseup = (view, e) => {
+    const { state } = view
+    const { selection } = state
+    const { empty, from, to, $anchor, $head } = selection
+
+    if (empty) {
+      tippy?.destroy()
+      tippy = null
+      return
+    }
+
+    const lastClick = view.lastClick || view.input.lastClick
+    const hasClickLocationChanged = (e.clientX !== lastClick.x || e.clientY !== lastClick.y)
+
+    if (hasClickLocationChanged) {
+      const existingRanges = rangeEditingKey.getState(state)
+      let positionsCovered = []
+      Object.values(existingRanges).forEach(({ from, to }) => positionsCovered.push(...range(from, to)))
+      const selectedPositions = new Set(range(from, to))
+      positionsCovered = new Set(positionsCovered)
+      const isFullyCovered = isSuperset(positionsCovered, selectedPositions)
+
+      if (isFullyCovered) {
+        tippy?.destroy()
+        tippy = null
+        return
+      }
+      tippy?.destroy()
+      tippy = createCreatorMenu(view, $anchor.pos, $head.pos)
+    }
+  }
   return new Plugin({
     key: pluginKey,
-    props: {
-      handleDOMEvents: {
-        mouseup (view, e) {
-          const { state } = view
-          const { selection } = state
-          const { empty, from, to, $anchor, $head } = selection
+    view (view) {
+      const globalHandler = (event) => {
+        const isClickInsideMenuBubble = event.composedPath &&
+          event.composedPath().some(el => el.classList && el.classList.contains('editor-menububble__wrapper'))
 
-          if (empty) {
-            tippy?.destroy()
-            tippy = null
-            return
-          }
+        if (isClickInsideMenuBubble) {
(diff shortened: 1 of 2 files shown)


client/js/lib/prosemirror/plugins.js
```diff
@@ -336,39 +336,57 @@ const rangeTracker = (rangeTrackerKey, schema, rangeChangeCallback = () => {}) =
...
+      positionsCovered = new Set(positionsCovered)
+      const isFullyCovered = isSuperset(positionsCovered, selectedPositions)
+
+      if (isFullyCovered) {
+        tippy?.destroy()
+        tippy = null
+        return
+      }
+      tippy?.destroy()
+      tippy = createCreatorMenu(view, $anchor.pos, $head.pos)
```
salisdemos, line 368: You can save one line ;-) 
```suggestion
      tippy?.destroy()

      if (isFullyCovered) {
        tippy = null
        return
      }

      tippy = createCreatorMenu(view, $anchor.pos, $head.pos)
```

client/js/lib/prosemirror/plugins.js
```diff
@@ -336,39 +336,57 @@ const rangeTracker = (rangeTrackerKey, schema, rangeChangeCallback = () => {}) =
...
+
+  const handleGlobalMouseup = (view, e) => {
+    const { state } = view
+    const { selection } = state
+    const { empty, from, to, $anchor, $head } = selection
+
+    if (empty) {
+      tippy?.destroy()
+      tippy = null
+      return
```
salisdemos, line 348:
```suggestion
      tippy = null

      return
```

client/js/lib/prosemirror/plugins.js
```diff
@@ -336,39 +336,57 @@ const rangeTracker = (rangeTrackerKey, schema, rangeChangeCallback = () => {}) =
...
+    const lastClick = view.lastClick || view.input.lastClick
+    const hasClickLocationChanged = (e.clientX !== lastClick.x || e.clientY !== lastClick.y)
+
+    if (hasClickLocationChanged) {
+      const existingRanges = rangeEditingKey.getState(state)
+      let positionsCovered = []
+      Object.values(existingRanges).forEach(({ from, to }) => positionsCovered.push(...range(from, to)))
+      const selectedPositions = new Set(range(from, to))
+      positionsCovered = new Set(positionsCovered)
+      const isFullyCovered = isSuperset(positionsCovered, selectedPositions)
```
salisdemos, line 360: For me this bock is not so good to read. But If You don't like my suggestion, feel free to drop it.
```suggestion
      const existingRanges = rangeEditingKey.getState(state)
      const selectedPositions = new Set(range(from, to))
      let positionsCovered = []
      
      Object.values(existingRanges).forEach(({ from, to }) => positionsCovered.push(...range(from, to)))
      const isFullyCovered = isSuperset(new Set(positionsCovered), selectedPositions)
```

client/js/lib/prosemirror/plugins.js
```diff
@@ -336,39 +336,57 @@ const rangeTracker = (rangeTrackerKey, schema, rangeChangeCallback = () => {}) =
...
+
+      if (isFullyCovered) {
```
elmasdemos, line 362: I would like if you add a comment something like
/**
 * Prevent showing the tooltip if the user's selection is already fully covered by the selection before.
 * This avoids redundant actions like trying to create a new annotation inside an already annotated area.
 * The user can still select text within annotated ranges, but no tooltip will be shown in that case.
 */

elmasdemos (changes requested):

salisdemos (approved):
Just nitpics. Looks like a good solution. (but not tested in interface)

elmasdemos (approved):

PR #4580, opened 2025-04-16
Fix return Type of getRecords in ExcelImporterHandleImportedTagsRecordsEvent

jalalifardemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15506
Description: return should be of Type MapIterator not ArrayIterator. Because we insert CSV file


MoritzMandler (approved):
lgtm

PR #4581, opened 2025-04-16
chore: Bump a bunch of Yarn dependencies and clean dependabot.yml

salisdemos:
to avoid conflicts in the yarn.lock file, I just upddated a couple of dependencies by hand.

- Some dependencies are tidy coupled to @masterportal/masterportalapi. they should only be updates as bundle. To ensure that, I listed them in the dependabot.yml
- with that: olcs has some issues with resolving its path with our env, so I had to teak webpack a bit (again)..

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,7 @@ updates:
           - ">= 5"
 
       # These dependencies have to be updated according to the needs of @masterportal/masterportalapi
+      # on that note: we have to update the masterportalapi maually, because its hosted by bitbucket and not via npm
       - dependency-name: ol
       - dependency-name: ol-mapbox-style
       - dependency-name: olcs
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,7 +21,7 @@ updates:
         versions:
           - ">= 5"
 
-      # This dependencies have to be updated according to the needs of @masterportal/masterportalapi
+      # These dependencies have to be updated according to the needs of @masterportal/masterportalapi
       - dependency-name: ol
       - dependency-name: ol-mapbox-style
       - dependency-name: olcs
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,25 +16,17 @@ updates:
           - ">= 0.8.0"
           - "< 99"
 
-      # Jest upgrade is blocked because the vue-test-utils don't support it yet for Vue 2
-      - dependency-name: jest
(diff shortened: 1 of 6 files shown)


spiess-demos (approved):
looks reasonable (code review).

.github/dependabot.yml
```diff
@@ -16,25 +16,17 @@ updates:
...
-      - dependency-name: vue-loader
-        versions:
-          - ">= 16"
-
       # Can't be used as CommonJs but only ESM which will break our FE build atm
       - dependency-name: chalk
         versions:
           - ">= 5"
+
+      # This dependencies have to be updated according to the needs of @masterportal/masterportalapi
```
hwiem, line 24:
```suggestion
      # These dependencies have to be updated according to the needs of @masterportal/masterportalapi
```

.github/dependabot.yml
```diff
@@ -16,25 +16,17 @@ updates:
...
+      - dependency-name: ol
```
hwiem, line 25: How does this ensure that they are only updated together?

hwiem (commented):
Thanks, just a nitpick and a question

.github/dependabot.yml
spiess-demos: including them here without specifying the "versions" key leads to dependabot excluding them entirely, so this does only ensure they are never updated by dependabot.

hwiem: Can't we create a group with masterportalapi so dependabot creates updates for the group?

salisdemos: unfortunately no. masterportalapi is not provided by npm. so we have to check for updates manually (As far as I could see bitbucket is not availalble as depenabot source - https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories).
With that, we have to check for masterportal-updates manually and when we update manually, we have to adjust the dependecies too. And: with almost every masterportal update olcs comes with another issue regarding import resolution :-/ that has to be handled manually anyway

hwiem: Right :see_no_evil:  Could we add that info here (masterportalapi is not provided by npm, so we have to check for updates manually, and we need to update these dependencies along with masterportalapi)? Or is this documented somewhere else?

salisdemos: like so 40e191cff4c782456c9fc5ad97ca527778df63ea

PR #4583, opened 2025-04-16
feat (ADO-24467): use interface when dispatching events

breierDemos:
### Ticket
[ADO-24467](https://www.dev.diplanung.de/DefaultCollection/EfA%20DiPlanung/_workitems/edit/24467)

Ensure that we dispatch event for original statement and statement creation and ensure that this event s can be used in addons.

WIP: Maybe the events for copied statements will be removed. If not we have to replace them with new events. Something like StatementCopiedEvent...

### How to review/test
code review


mussbach:
@breierDemos I think it is better to create a new event like StatementCreatedEvent as you suggested

breierDemos:
This will be handled here: https://github.com/demos-europe/demosplan-core/pull/4608

PR #4584, opened 2025-04-17
bug(refs DPLAN-15488): Fix editable list

gruenbergerdemos:
### Ticket
DPLAN-15488

Editable list now emits the delete and showUpdateForm events directly and not via parent.

### How to review/test
Verfahren -> Stellungnahmen -> Detailansicht -> Bearbeiten -> Editing and deleting votes should work


salisdemos (approved):
Thank you!

PR #4587, opened 2025-04-17
bug(refs DPLAN-15495): Fix path for new statement

gruenbergerdemos:
### Ticket
DPLAN-15495

We only want to link to the import center if `feature_simplified_new_statement_create` is enabled. Otherwise, we need the old view for creating statements.

### How to review/test
Verfahren -> Stellungnahmen -> Neue Stellungnahme -> You should get to the "old" create new statement view and not the import center.

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/list_statements.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/list_statements.html.twig
@@ -1,10 +1,7 @@
 {% extends '@DemosPlanCore/DemosPlanCore/base.html.twig' %}
 {% block component_part %}
 
-    {% set path = hasPermission('feature_statements_import_excel') or
-        hasPermission('feature_import_statement_pdf') or
-        hasPermission('feature_segments_import_excel') or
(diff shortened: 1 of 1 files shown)


riechedemos (approved):
Looks good. I get the old view.

templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/list_statements.html.twig
```diff
@@ -1,10 +1,7 @@
 {% extends '@DemosPlanCore/DemosPlanCore/base.html.twig' %}
 {% block component_part %}
 
-    {% set path = hasPermission('feature_statements_import_excel') or
-        hasPermission('feature_import_statement_pdf') or
-        hasPermission('feature_segments_import_excel') or
-        hasPermission('feature_simplified_new_statement_create') ? 'DemosPlan_procedure_import' : 'DemosPlan_statement_new_submitted' %}
+    {% set path = hasPermission('feature_simplified_new_statement_create') ? 'DemosPlan_procedure_import' : 'DemosPlan_statement_new_submitted' %}
```
hwiem, line 4: I'm not sure, but shouldn't it be
`
(hasPermission('feature_simplified_new_statement_create') && (hasPermission('feature_import_statement_pdf') || hasPermission('feature_segments_import_excel'))
`
?
Or is there no case in which there's only the new statement create form and no import?

PR #4588, opened 2025-04-17
feat(refs DPLAN-15338): display customFields in StatementSegmentList

salisdemos:
### Ticket
DPLAN-15338

Adds the dropdowns for custom fields in the "Erwiderungen verfassen" view and allows saving the selected values; these should also be displayed when hovering over the '#'.

Sub-Branch to integration branch:
https://github.com/demos-europe/demosplan-core/pull/4579

### How to test
- Go to a procedure that has segments and custom fields
- Go to the statement list and select "Details und Erwiderungen" from the flyout menu of a statement that has segments
- In the "Erwiderungen verfassen" tab there should now be the option "Weitere Felder bearbeiten"
- Select values for the custom fields and save them
- If you hover over the '#', the custom fields and values should be displayed after saving


sakutademos (approved):
code review, lgtm

PR #4594, opened 2025-04-23
fix(DPLAN-15284): splitStatement: dropdown menu open direction

sakutademos:
**Description:** Vue-Multiselect automatically positions the dropdown where there is the most space. The multiselect is inside a scrollable container, and sometimes this causes the dropdown to appear above (because the dropdown menu is absolutely positioned, the container does not scroll it, but simply clips it)

I tried different approaches: adding a min-height or removing the absolute positioning, but there were always issues and none were perfect. I settled on having the first six dropdowns open downward (since the dropdowns max-height is 300 px). I know it is not the ideal solution, but I find it better than the others...

Additional PR in UI is provided

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/client/js/components/statement/splitStatement/SideBar.vue
+++ b/client/js/components/statement/splitStatement/SideBar.vue
@@ -74,6 +74,7 @@
           <tag-select
             v-for="(topic, idx) in tagTopics"
             :class="{'mb-1': idx < tagTopics.length + 1}"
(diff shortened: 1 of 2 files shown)


gruenbergerdemos (approved):
lgtm

PR #4597, opened 2025-04-23
bug(refs DPLAN-12748): Add warning if there are no places

gruenbergerdemos:
### Ticket
DPLAN-12748

We want a notification for the user if there are no places yet, because without them the statement cannot be split into segments. 

### How to review/test
Go to a procedure without places (Bearbeitungsschritte) -> Stellungnahmen -> Aufteilen -> There should be a warning with a link to create places.


hwiem (approved):
Nice! (Code review only)
I think we should merge it into `main` though, not `release`

PR #4603, opened 2025-04-28
style(refs: DPLAN:15522): adjust custom event names

rafelddemos:
### Ticket
DPLAN-15522

Changes custom event names that were not named according to the naming guidelines.

We decided to follow the Vue3 recommendations in general (camelCase when emitting events, kebab-case when listening to them), but we still allow for patterns like `entity:myEvent` (e.g. `statement:updateText`).

https://demoseurope.youtrack.cloud/articles/TECH-A-88/Javascript-Vue#custom-event-names
https://vuejs.org/guide/components/events.html

### How to review/test

When running `yarn lint .`, the result should be
```
✖ 417 problems (41 errors, 376 warnings)
```

Most of the warnings about events having to be camelCase should be gone, but there are some left, mostly because of DpSlidebar in demosplan-ui (I decided to keep the events kebab-case, because we need to replace them with a better solution anyway).

### PR Checklist


meissnerdemos (approved):
Thank you for fixing this PR :pray:

PR #4605, opened 2025-04-28
fix(DPLAN-15474): filter reset issue

sakutademos:
### Ticket
[DPLAN-15474](https://demoseurope.youtrack.cloud/agiles/141-63/current?settings&tab=allgemeines&issue=DPLAN-15474)

**Description:** This PR fixen an issue with `this.$refs[("filterFlyout" + (intermediate value))].reset is not a function`

- get all filterFlyout in the array and apply reset() to each flyout, instead of adding index to each flyout


gruenbergerdemos (approved):
lgtm and works

PR #4606, opened 2025-04-28
fix: inverse side does not exists

martendemos:
Error in symfony toolbar indicated this problem.

No migration needed, because this is doctrine only and was never modeled.


mussbach (approved):
lgtm, code review only, thx

PR #4609, opened 2025-04-29
chore: exclude addon-folder. They should be treated manually

salisdemos:
To avoid 3rd-Party errors when firing the linter


hwiem (approved):
Thanks!

PR #4625, opened 2025-05-05
F(dplan-15556): unable to edit text when split statement is active

sakutademos:
### Ticket
[DPLAN-15556](https://demoseurope.youtrack.cloud/agiles/141-63/current?settings&tab=allgemeines&issue=DPLAN-15556)

**Description:** This PR fixes an issue that prevented editing when a split statement was started but not completed. Users can now edit the text before the split process finishes and draft segments are created.

- check if the draft segments exist rather than just `segmentDraftList`
- update the warning message accordingly


gruenbergerdemos (changes requested):
I don't understand how this solves the bug. I also now get an error if I want to access the other statement with the same text to edit it (Bearbeiten):
```
 TypeError: Cannot read properties of null (reading 'data')
    at Proxy.render (StatementSegmentsList.vue?./node_modules/vue-loader/dist/templateLoader.js??ruleSet%5B1%5D.rules%5B3%5D!./node_modules/vue-loader/dist/index.js??ruleSet%5B0%5D!./node_modules/source-map-loader/dist/cjs.js:243:84)
```
at `"has-draft-segments": this.statement.attributes.segmentDraftList.data.attributes.segments.length > 0`

sakutademos:
> I don't understand how this solves the bug. I also now get an error if I want to access the other statement with the same text to edit it (Bearbeiten):
> 
> ```
>  TypeError: Cannot read properties of null (reading 'data')
>     at Proxy.render (StatementSegmentsList.vue?./node_modules/vue-loader/dist/templateLoader.js??ruleSet%5B1%5D.rules%5B3%5D!./node_modules/vue-loader/dist/index.js??ruleSet%5B0%5D!./node_modules/source-map-loader/dist/cjs.js:243:84)
> ```
> 
> at `"has-draft-segments": this.statement.attributes.segmentDraftList.data.attributes.segments.length > 0`

thx for spotting this issue: I refactored the code to use optional chaining and `Boolean()` to check whether the properties exist.

The STN text is now editable until it is actually splited. Previously, we only checked for the existence of the `segmentDraftList` (JSON object, which was created directly after the user clicked “split statement”, which could be done mistakenly for example). Now, because we specifically check the segments array in the segmentsDraftList, this issue should no longer appear.

gruenbergerdemos (approved):
Thanks, lgtm and works!

PR #4637, opened 2025-05-07
chore: remove unused component

spiess-demos:
The last usage has been [removed here](https://github.com/demos-europe/demosplan-core/commit/7994591e1ae625efb074763687cac7ba7ddf0bfc) so we can drop it!


salisdemos (approved):
Thanks for cleaning up

PR #4640, opened 2025-05-09
Fix(DPLAN-15601): rendering DpTreeList inside a DpTab that uses v-sho…

sakutademos:
### Ticket
[DPLAN-15601](https://demoseurope.youtrack.cloud/agiles/141-202/142-669?issue=DPLAN-15601)

Description: This PR fixes an issue with the Sticky-Header inside the DpTreeList component.

- the rendering DpTreeList inside a DpTab that uses v-show directive causes an error, because v-show applies 'display: none', so measurements like `offsetHeight` return zero. Instead, render the TagList component only when its tab is active and add some adjustments in UI library.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


riechedemos (approved):
Thx, tested and lgtm!

PR #4641, opened 2025-05-09
fix(refs DPLAN-14565): switch trash and edit buttons

meissnerdemos:
### Ticket
[DPLAN-14565](https://demoseurope.youtrack.cloud/issue/DPLAN-14565)

Switched edit and trash buttons in the template of TagListItem.vue so that edit appears first.


salisdemos (approved):
Thanks!

PR #4646, opened 2025-05-12
refactor(refs DPLAN-15262): Use modelvalue for autocomplete

salisdemos:
After refactoring DpAutocomplete in demosplan-ui, we now should use modelValue when using it


sakutademos (approved):
thx!!!

PR #4653, opened 2025-05-13
Fix (refs DPLAN-15423): add changelog entry

ghost:
### Ticket
DPLAN-15423

I need help to find a better solution for this: 048741e69b9054fb22eb36d6e0cdd67d3e50088b

Changes of custom fields of segments (the only one so far) are tracked now.
It shoudl be worked for all changes of custom fields of segemnts, independent how the change is made.
I see two options to change a custom fields of a segment:
1. via bluk edit
2. via single edit in the "Erwiderungen verfassen"-view

The code I have to modify was much more complex and extensive as I expected.
Especially the logic to display/outpupt the entries (history) does many extra steps and convertions to the data from the database before rendering it for in the UI.
I do not fully understand why it seems to be necessary to get the current entity on loading the history of this entity.
I tried to bypass this kind of logic because it is pretty expensive to load the customFieldConfiguration if you only have the customFieldValue which includes the ID of the customFieldConfiguration. And from my point of view it is simply not necassary.

This logic I mean is located in getContentChangeComparisonString and probably should be checked on reviewing this PR.

- [x] Create/Update tests
- [ ] Update documentation
- [x] Update changelog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ addonZip
 addonZips
 addons
 CLAUDE.md
+claude-mds
 node_modules
 yarn-error.log
 docker-compose.yml
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
@@ -437,7 +437,7 @@ export default {
       }
 
       this.listSegments({
-        include: ['assignee', 'comments', 'place', 'tag', 'assignee.orga', 'comments.submitter', 'comments.place'].join(),
+        include: ['assignee', 'comments', 'place', 'tags', 'assignee.orga', 'comments.submitter', 'comments.place'].join(),
         sort: 'orderInProcedure',
         fields: {
           Place: ['name', 'sortIndex'].join(),


sierrazamudiodemos:
@martendemos 
- I think using the Strategy Pattern here could help to simplify things. I've seen it used at Demos code, and it could help make the `EntityContentChangeService` class more manageable, and hopefully a bit more readable.

- I understand that custom fields add complexity, resulting to a specific treatment per each field type in one large service class (`EntityContentChangeService`). I believe it's fine to have different treatment for different fields—some work well with Doctrine, others don't. I think the key is organizing this in a way that prevents `EntityContentChangeService` from becoming too inflated.

- The `EntityContentChangeService` is handling the difference of 3 different fieldTypes:
-- customFields (new)
-- Object (for assignee, place) 
-- HtmlFieldChange (for text fields)
- Maybe each of them can be a strategy, so that we can remove that code from the EntityContentChangeService outside, and each fieldType strategy can be testable. That said, if in the future we need a custom contentChange adjustement, we can add a new "Strategy".

- With the help of a friend ;) this is how I think it can work:

1. Create a common interface (FieldChangeCalculatorInterface) that defines how any field type calculates changes.
2. Extract the existing custom field logic from EntityContentChangeService into a dedicated CustomFieldChangeCalculator class.
3. Create separate calculator classes for other field types like ObjectFieldChangeCalculator (for assignee, place) and HtmlFieldChangeCalculator (for text fields).
4. Build a manager class (FieldChangeCalculatorManager) that holds all the calculators and picks the right one based on field type.
5. Update the main EntityContentChangeService to use this manager instead of having all the logic mixed together.
6. Configure the calculators in the service container so they're automatically registered and available.
7. Remove the special case if ('customFields' === $propertyName) logic since each field type now has its own handler.
8. Update the field mapping configuration to be more consistent across all field types. ---> this refers to the file config/entity_content_change_fields_mapping.yml because not all of the field types have the same amount of keys on the yml
9. Write focused unit tests for each calculator class separately, making testing much simpler.
10. This approach makes it super easy to add new field types in the future - just create a new calculator class without touching existing code!

martendemos:
> @martendemos
> 
> * I think using the Strategy Pattern  ....
I see it like you; there are definetly tons of improvements and/or refactorings advisable here and it was quite challenging for me to **not** go into these while programming this feature to avoid a more extensive and complex PR and to be able to warp this up in time.

Instead I would try/suggest to implementes the improvements I have in my mind and the ones you worte, in an additional PR.
What did you think?

sierrazamudiodemos (approved):
- Change the target branch to release ;)
- Document a way to remove the custom_field_change column in the entity_change table
Thanks for the talk

PR #4657, opened 2025-05-13
f DS-82 preserve underlined and strike through text 

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/tickets/DS-82/Zip-Ex-Import-RO-EWM

Description
this xls export can be imported later and html tags have to be preserved or converted (markdown) in some cases otherwise the provided structure and formatting will be lost

 


mussbach (approved):
Lgtm, thx. Code review only

PR #4661, opened 2025-05-13
fix(refs DPLAN-15558): display safed brandings

salisdemos:
It looks like safed brandings where not loaded in the orga-list.
This Pr should fix that


salisdemos:
closed in favor of #4705

PR #4662, opened 2025-05-14
f DS-43 remove JSON_FORCE_OBJECT flag : 

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/tickets/DS-43/BOM-Fehlermeldung-bezuglich-Umgriff

Description: JSON_FORCE_OBJECT converts JSON arrays to PHP arrays, and JSON objects to associative arrays, in this case a we have to decode jsons by default weil a numeric arrays are expected and not associative arrays.


salisdemos (approved):
I am pretty sure that does the trick. And I saw it works.
I am bit concerened, that it breaks something somewhere else, but we tried to test everything. I suggest to tell QA to take a really close look at it.

mussbach (approved):
lgtm, thx

PR #4666, opened 2025-05-14
Tailwind v4 Migration

spiess-demos:
## Summary
This PR upgrades the project from Tailwind CSS v3.4 to Tailwind CSS v4.1. The update includes configuration changes, dependency updates, and adjustments to maintain compatibility with our custom styling system.

## Changes
- Update Tailwind CSS from v3.4.15 to v4.1.6, add/update needed dependencies PostCSS, @tailwindcss/postcss
- Update demosplan-ui from v0.4.14 to v0.4.17
- Modify CSS import structure with new @layer syntax introduced by Tailwind 4
- Adjusted webpack configuration for CSS processing
- Updated SCSS components for Tailwind v4 compatibility
- Update component styles to maintain consistency with the new Tailwind version
- Drop autoprefixer bc it is included in postcssPresetEnv

## Implementation details
- The `!important` keyword was removed on some Scss utilities, because when calculating which declaration takes precedence, the order in which layers are declared is inverted for declarations with `!important`. 
- When mixing @layer and Scss, some restrictions apply: while Css variables are globally visible (regardless if they have been declared "layerless" or within any layer), Scss variables are only visible inside the layer they're declared in. only Scss variables declared outside a layer are globally visible.

### How to review/test
After building, things should look as before, or even better.

### Linked PRs 
- https://github.com/demos-europe/demosplan-ui/pull/1275

### PR Checklist

The tests are failing, however idk if that was the case before.

- [x] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


client/js/components/procedure/DpSearchProcedures.vue
```diff
@@ -19,7 +19,6 @@
       }"
       @enter="search" />
     <dp-button
-      class="ml-0.5 mt-0.5"
```
salisdemos, line 22: I thought, we want to define outer spacings outside the component. Why did You remove it?

salisdemos (approved):
legit. Just a question out of curiosity

client/js/components/procedure/DpSearchProcedures.vue
spiess-demos: right! while the top margin led to some weird layout, the left margin is still needed. re-added it in 4484babb146eaa5b4a8857555a6173bf08300933

hwiem (approved):
Looks good (as far as I can tell)

PR #4667, opened 2025-05-14
Fix(dplan 15291): toggle all does not work

sakutademos:
### Ticket
[DPLAN-15291](https://demoseurope.youtrack.cloud/agiles/141-63/current?settings&tab=allgemeines&issue=DPLAN-15291)

Description: This PR fixes an issue with the toggle-all function in the segments list. The problem appeared after the Vue 3 migration. Unfortunately, using the ref with dynamic keys do not work as expected in Vue 3. Use ref="segment" within the loop instead, which sets `this.$refs.segment` to an array of component instances.

- while technically still allowed, Vue 3 does not recommend using dynamic keys like :ref="'segment' + idx' inside v-for, use a static ref name instead.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/SegmentsRecommendations.vue
+++ b/client/js/components/procedure/StatementSegmentsList/SegmentsRecommendations.vue
@@ -46,10 +46,10 @@
       <!--Segments, if there are any-->
       <div v-else>
         <statement-segment
-          v-for="(segment, idx) in segments"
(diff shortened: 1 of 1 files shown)


salisdemos (approved):
Thanks! Nice

PR #4671, opened 2025-05-15
fix(DPLAN-15627): unable to add a new comment after drawing a point on map

sakutademos:
### Ticket
[DPLAN-15609](https://demoseurope.youtrack.cloud/agiles/141-63/current?settings&tab=allgemeines&issue=DPLAN-15609)

**Description:** This PR fixes an issue where a comment could not be added after a point feature was triggered on the map. 

The problem was the incorrect handling of how relationships were removed from the store: mutating store objects directly without using Vuex mutation actions is not a good practice and causes an issue. 

- use dpApi method instead of the saveAction of vuex-json-api, as there are cases when vuex-json-api may include relationships in the PATCH request; It does not support excluding them from the payload. To avoid unintentionally sending relationship data, use a dpApi method with a manually constructed payload and update the store after a successful response

- send the customFields in the attributes only if the permission is set

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
```diff
@@ -323,6 +306,21 @@ export default {
...
+
+      segment.attributes = {
+        ...segment.attributes,
+        updatedAttributes
+      }
+
+      const storePayload = {
+        ...segment,
+        id: segment.id
+      }
```
salisdemos, line 322: with this it would be a bit more compact and we would not need the JSONpars/stringify:
```suggestion
      const storePayload = {
        ...this.segment,
        attributes: {
          ...this.segment.attributes,
          updatedAttributes
        }
        id: segment.id
      }
```

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -801,10 +802,12 @@ export default {
...
-          customFields: {
-            ...this.segment.attributes.customFields,
-            ...payload.data.attributes?.customFields
-          }
+          ...(hasCustomFields && {
+            customFields: {
+              ...this.segment.attributes.customFields,
+              ...payload.data.attributes?.customFields
+            }
+          })
```
salisdemos, line 810: just a really personal opinion, because I don't like to use && to then return something:
```suggestion
          ...(hasCustomFields ? {
            customFields: {
              ...this.segment.attributes.customFields,
              ...payload.data.attributes?.customFields
            }
          } : {})
```

salisdemos (approved):
Legit. Just none-blocking suggestions

client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
```diff
@@ -234,51 +234,34 @@ export default {
...
+        attributes
+      }
 
       /**
-       *  Comments need to be removed as updating them is technically not supported
-       *  After completing the request, they are added again to the store to be able to display them
+       * Vuex-json-api may include relationships in the PATCH request,
+       * and it does not support excluding them from the payload.
+       * To avoid unintentionally sending relationship data, use a dpApi method with a manually constructed payload
+       * and update the store after a successful response.
```
salisdemos, line 251: :-/

salisdemos (approved):

client/js/components/procedure/StatementSegmentsList/SegmentLocationMap.vue
salisdemos: I just wanted to say that its sad, that we have to do it like this.
Thanks for the comment. its good to explain why we can't use the vuex-store here

PR #4673, opened 2025-05-15
B dplan-15641 enable unset custom value in segments

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-15641/Nicht-zugewiesen-Option-wird-nicht-gespeichert-wenn-sie-ausgewahlt-wird

Allow to unset custom values already set in the segment

### How to review/test
- Go to UI and click on "Nicht zugewiesen"
- In the database, the json should not have the value anymore

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/Utils/CustomField/CustomFieldValueCreator.php
```diff
@@ -46,12 +46,18 @@ public function updateOrAddCustomFieldValues(
...
-                $existingCustomFieldValue->setValue($newCustomFieldValue->getValue());
+                $this->handleExistingCustomField($currentCustomFieldValuesList, $existingCustomFieldValue, $newCustomFieldValue);
             } else {
-                $currentCustomFieldValuesList->addCustomFieldValue($newCustomFieldValue);
+                $this->handleNewCustomField($currentCustomFieldValuesList, $newCustomFieldValue);
             }
         }
 
+        // Sort fields to ensure consistent ordering in the database
+        $currentCustomFieldValuesList->sortByFieldId();
```
martendemos, line 56: I got it; this is done to ensure the order does not change, but
at some point the user will wonder about the order, because the UUIDs are random and so the order will be.
 We may should add a creation-date or something to use it for this case. :thinking: 

But anyway: this is a bugfix, so its okay for me and just a note.

salisdemos (approved):
legit.

demosplan/DemosPlanCoreBundle/CustomField/CustomFieldInterface.php
```diff
@@ -19,6 +19,10 @@
  */
 interface CustomFieldInterface
 {
+    /**
+     * Special value used to indicate that a custom field should be removed.
+     */
+    public const UNASSIGNED = 'UNASSIGNED';
```
martendemos, line 25: how is this solved with the assignee field?
Using a string seems to be very unstablel to me.
I rather would use a flag coming from FE which indicates that the value should be unset.
Or at least using a value which is invalid and impossible for the customer to create like an empty string.

martendemos (changes requested):

demosplan/DemosPlanCoreBundle/Utils/CustomField/CustomFieldValueCreator.php
sierrazamudiodemos: hey @martendemos thanks for the feedback, I resonate with you, here are my thoughts:

I think the devs will wonder about the order of the custom fields stored in the json column: Sometimes it's ["color" => "red", "food" => "banana"] and other times it's ["banana" => "banana", "color" => "red"]. To me this inconsistency could confuse future devs since the order isn't predictable.

I think the order displayed to the end user does not rely on the order stored of the values of the custom fields, so I think we are safe there.

Adding createdAt or updatedAt timestamps could be useful for tracking changes, but we don't currently have per-column timestamps in the entity. And i think the US part 4 will help us with this tracking changes part.

My main motivation is to make the data storage predictable and eliminate any randomness to avoid confusion among devs in the future, but I also hope this line will not fire back and instead create confusion

demosplan/DemosPlanCoreBundle/CustomField/CustomFieldInterface.php
sierrazamudiodemos: @martendemos the FE is sending this string UNASSIGNED and consequently the BE will detect it.
It was not my first choice, I will try to explain why I did it like this, maybe also another more fitting idea can come:

- initially I would prefer that FE sends null when a previously stored value is unset
- if that is case, then the BE will have to handle this null value, but I wanted to avoid that: Because in order to allow null value, I would have to allow the CustomFieldsList to store null values, but it will defeat the purpose that this column should only store whenever there is actually a value. That said, I could have the option to anyway allow it in the object CustomFieldsList to allow null values, but it would have opened the door to misused the  CustomFieldsList

demosplan/DemosPlanCoreBundle/Utils/CustomField/CustomFieldValueCreator.php
martendemos: Ah okay, then I missunderstood your intention here I guess.
I thought you do this to ensure the consistency of the order in the UI, but you did it to ensure the order for us devs.

My thoughts are that we have a reasonable order in the UI for the users and thats would be by ordereing the creation date descending but I do not want to change the actual order of the json.
But I also miss the fact that we can't use the already known `@Gedmo\Timestampable(on="create")` to set a createdAt attribute at a single custom-value, because the attribute of a segment is a list of multiple values stored as json.
Even it is of course not impossible, but it would be necessary to use your own logic for this.

demosplan/DemosPlanCoreBundle/CustomField/CustomFieldInterface.php
martendemos: for refs: we talked about it and the plan is to use null or empty string as indicator to execute a deletion of the customfalue from the custom value list.

martendemos (approved):

PR #4683, opened 2025-05-16
feat(refs DPLAN-15588, #AB28108): use diplan-karten in public detail

hwiem:
### Ticket
DPLAN-15588

We want to have the possiblilty to use `init/diplan-karte` instead of open-layers in the public detail.
For now that comes with a limited featureset, but this will increase soon.

Since diplankarte can't handle nonces atm, we have to allow inline style blocks for the moment. This will be re-denied the moment we are able to set nonces there. (the team from diplankarte is on it.)

As long as we don't have a permission, we have to tweak the `isDiplanKarte` a bit. Thats  just for dev purpose and will be adjusted when the permission is created and configured.

This is the first Part for a bigger Feature. So it will be merged into an Integration branch and not without further changes come to main.

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


client/js/components/map/publicdetail/DiplanKarteWrapper.vue
```diff
@@ -0,0 +1,38 @@
...
+instance.appContext.app.use(MapPlugin, {
+  template: {
+    compilerOptions: {
+      isCustomElement: (tag) => tag === "diplan-karte",
+    }
+  }
+})
+
+const toggleStatementModal = (updateStatementPayload) => {
+  console.log('updateStatementPayload', instance.parent)
```
elmasdemos, line 35: please remove this console.log here

hwiem: Thanks, done in 0668a1494e00b9a09d0853d245c396d9dcd9852b

elmasdemos (approved):
lgtm

PR #4688, opened 2025-05-19
fix(refs: DPLAN:15188): Add spacing between btns

rafelddemos:
On smaller screens when the btns stack vertically.

### Ticket
[DPLAN-15188](https://demoseurope.youtrack.cloud/issue/DPLAN-15188/Keine-Padding-zwischen-Gruppe-anlegen-und-Abbrechen-wenn-kleinere-Window)

Adds an additional wrapper-div to the buttons to implement flexbox independently from the 'layout__item' class.


spiess-demos (approved):
lgtm!

PR #4697, opened 2025-05-20
fix(refs DPLAN-15673): add question mark to length condition

meissnerdemos:
### Ticket
[DPLAN-15673](https://demoseurope.youtrack.cloud/issue/DPLAN-15673/Abschnitt-Aktionen-sind-alle-kaputt)

The bug occurs when some segments have customFields and some don't. The question mark ensures the length is only checked if there actually are customFields.

### How to review/test

1. have a statement with some segements that have customFields and some that don't
2. navigate to "Abschnitte"
3. check that actions mentioned in the ticket work


hwiem (approved):
Code lgtm!

PR #4700, opened 2025-05-21
feat (refs DPLAN-15650): Remove error prone flushes.

breierDemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-15650/Verfahren-kann-nicht-erstellt-werden

I removed two flush calls, redundant persist calls. In copyParagraphsOfElement I think we can use the new created paragraph entities directly. Until now I can't reproduce the Deadlock exception on this branch. I can well imagine that the deadlock exception could still occur. But perhaps less frequently?

### How to review/test
run locally bin/project dp:maintenance to start the maintenance task, then create a procedure via web interface

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [] Move the tickets on the board accordingly
- [ ] Update changelog

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Repository/ParagraphRepository.php
+++ b/demosplan/DemosPlanCoreBundle/Repository/ParagraphRepository.php
@@ -489,6 +489,7 @@ class ParagraphRepository extends FluentRepository implements ArrayInterface, Ob
     {
         $paragraphIdMapping = [];
         $paragraphsToCopy = $this->getParagraphOfElement($elementToCopy);
+        $copiedParagraphs = [];
         foreach ($paragraphsToCopy as $paragraphToCopy) {
             $copiedParagraph = new Paragraph();
             $copiedParagraph->setCategory($paragraphToCopy->getCategory());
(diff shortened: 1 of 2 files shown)


demosplan/DemosPlanCoreBundle/Repository/ParagraphRepository.php
```diff
@@ -507,23 +508,17 @@ public function copyParagraphsOfElement(Elements $elementToCopy, Elements $copie
...
-        $this->getEntityManager()->flush();
 
-        $copiedParagraphs = $this->getParagraphOfElement($copiedElement);
         foreach ($copiedParagraphs as $copiedParagraph) {
             if ($copiedParagraph->getParent() instanceof Paragraph) {
                 $oldParentId = $copiedParagraph->getParent()->getId();
-                $relatedCopiedParagraph = $this->getEntityManager()->getReference(
-                    Paragraph::class, $paragraphIdMapping[$oldParentId]
-                );
+                $relatedCopiedParagraph = $paragraphIdMapping[$oldParentId];
```
mussbach, line 518: Isnt this wrong? we need the parent id, not the related paragraphId

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/Repository/ParagraphRepository.php
breierDemos: I think the code here is difficult to understand. At first, I also thought it was wrong.
When copying the paragraph, the original parent is also set as the parent for the copy. Which is wrong.
What's being done here is that we save all copies under their old ID in the $paragraphIdMapping array. Then we go through all the copies, get the parent ID, and then look in the $paragraphIdMapping array for the correct new parent, which is one of the created copies.

breierDemos:
> hey @breierDemos thanks for this! Is there a way that we can do a unit test that the exception occurs, and then with the changes of this PR, then the exception will not occur?

Not sure how we can do this.

PR #4704, opened 2025-05-21
f-DPLAN-11632 Case insensitive email address comparison

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-11632/BOB-SH-Bauleitplanung-Prod-bestimmte-Mailadresse-kann-in-einem-Verfahren-nicht-gespeichert-werden-T36900

Description: 
Case insensitive email address comparison:
 - array_diff is case-sensitive, therefore the already existing email from input could not be found and so it appears in array_diff which causes an unique constraint error when updating.
 - this is done while sorting email addresses too otherwise emails with corresponding value will be incorrectly omitted from the return which causes errors too.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Repository/EmailAddressRepository.php
+++ b/demosplan/DemosPlanCoreBundle/Repository/EmailAddressRepository.php
@@ -106,6 +106,7 @@ class EmailAddressRepository extends CoreRepository implements EmailAddressRepos
      */
     protected function sortByGivenArray(array $sortedStrings, array $unsortedEntities): array
     {
+        $sortedEmailAddresses = [];
         foreach ($sortedStrings as $sortedEmailAddressString) {
             foreach ($unsortedEntities as $unsortedEmailAddressEntity) {
                 $fullAddress = $unsortedEmailAddressEntity->getFullAddress();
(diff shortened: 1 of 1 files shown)


demosplan/DemosPlanCoreBundle/Repository/EmailAddressRepository.php
```diff
@@ -97,10 +106,14 @@ public function deleteOrphanEmailAddresses(array $emailIds): int
...
     {
-        $sortedEmailAddresses = array_fill_keys($sortedStrings, null);
-        foreach ($unsortedEntities as $emailAddressEntity) {
-            $fullAddress = $emailAddressEntity->getFullAddress();
-            $sortedEmailAddresses[$fullAddress] = $emailAddressEntity;
+        foreach ($sortedStrings as $sortedEmailAddressString) {
+            foreach ($unsortedEntities as $unsortedEmailAddressEntity) {
+                $fullAddress = $unsortedEmailAddressEntity->getFullAddress();
+                if (0 === strcasecmp($sortedEmailAddressString, $fullAddress)) {
+                    $sortedEmailAddresses[$sortedEmailAddressString] = $unsortedEmailAddressEntity;
```
mussbach, line 113: You need to initialize $sortedEmailAddresses

bakridemos: thx done in https://github.com/demos-europe/demosplan-core/pull/4704/commits/8fbe34ac131b1bd9aec21bfa721787d8abd8617a

PR #4705, opened 2025-05-22
fix(refs DPLAN-15558): Load branding/cssvars for orgas

salisdemos:
Load branding/cssvars for orgas to display the saved values

### Ticket
DPLAN-15558

The saved branding data where not displayed, so it was not possible to see what colorsettings where saved.

### How to review/test

Ass support goto orga-edit and set branding colors. see if they get displayed after saving and after reloading the page. For both: pending and "accepted" orgas


sakutademos (approved):
thx!

PR #4707, opened 2025-05-22
fix(refs: DPLAN-15664): Render correct error message

huellnerdemos:
### Ticket
[DPLAN-15664](https://demoseurope.youtrack.cloud/issue/DPLAN-15664/Nach-dem-speichern-von-alle-Pflichtfelder-und-wieder-loschen-von-ein-paar-Pflichtfelder-und-speichern-zeigen-wir-nicht-das)

### How to review/test

The Bug is fixed in the administration_edit.html.twig. In this file there is an {% include ... %} that includes the label and a tooltip. I replaced the following input element with a dp-input component. The dp-input component can have an own label (i deleted it because it is the same like the include but without the tooltip).  

### PR Checklist

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
@@ -198,13 +198,14 @@
                                     showPlainHint: true
                                 } %}
                                 {{ form_errors(form.agencyMainEmailAddress) }}
-                                <input id="{{ form.agencyMainEmailAddress.vars.id }}"
-                                       name="agencyMainEmailAddress[fullAddress]"
-                                       value="{{ form.agencyMainEmailAddress.vars.value.fullAddress }}"
-                                       data-cy="internal:agencyMainEmailAddress"
-                                       class="o-form__control-input w-full"
-                                       type="email"
-                                       required>
+                                <dp-input
+                                    id="main-email"
+                                    data-cy="agencyMainEmailAddress"
+                                    name="agencyMainEmailAddress[fullAddress]"
+                                    required
+                                    type="email"
+                                    value="{{ form.agencyMainEmailAddress.vars.value.fullAddress }}">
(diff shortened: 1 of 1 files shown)


hwiem (commented):
What is the error message that should be displayed? I'd like to test the fix in the interface and I'm not sure what the desired behavior is

hwiem (commented):
I followed the steps in the description, but the error message I get is

> Bitte korrigieren Sie Ihre Eingabe in folgenden Feldern: Urheberrecht des Bildes, Alternativtext des Bildes. Erst dann können Sie Ihre Angaben speichern.

Am I doing something wrong?

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
```diff
@@ -198,13 +198,15 @@
...
                                 {{ form_errors(form.agencyMainEmailAddress) }}
-                                <input id="{{ form.agencyMainEmailAddress.vars.id }}"
-                                       name="agencyMainEmailAddress[fullAddress]"
-                                       value="{{ form.agencyMainEmailAddress.vars.value.fullAddress }}"
-                                       data-cy="internal:agencyMainEmailAddress"
-                                       class="o-form__control-input w-full"
-                                       type="email"
-                                       required>
+                                <dp-input
+                                    id="main-email"
```
hwiem, line 202: Why did you change the value of the `id`? It should match the value of the `for` attribute of the label

hwiem (changes requested):
Just a small change request left

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/administration_edit.html.twig
huellnerdemos: Thanks for this point, i can't remember anymore why i changed the id, i think it was just for testing, i changed it back

hwiem (approved):
Nice, thanks!

PR #4712, opened 2025-05-23
fix: Fix failing test

sierrazamudiodemos:
We adjusted the way how we store custom fields. We sort first the custom field values per id, and then store them in the column custom_fields of the segment table. Consequently, this test started to failed. this pr is to fix the test


breierDemos (approved):
lgtm thanks

PR #4720, opened 2025-05-26
Dplan 15347 set visibility group id to null

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15347/BOB-SH-Bauleitplanung-Prod-Stage-Bei-Nutzung-einer-Blaupause-konnen-im-Verfahren-nicht-alle-Einstellungen-der-Grundkarte

Description: 
- adjust resource type: visibility group id has to be set to null instead empty string
- toggleDefaultVisibility will be desabled when visibilty group id is null: visibility groupd id can not anymore have empty strings as value when layers are not grouped, instead it will be set to null.
- toggleDefaultVisibility will check if visibilityGroupId is different from null instead of empty string too.
-picked changes from https://github.com/demos-europe/demosplan-core/pull/4559

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
@@ -10,7 +10,10 @@
 
 namespace demosplan\DemosPlanCoreBundle\Controller\User;
 
+use DemosEurope\DemosplanAddon\Contracts\Config\GlobalConfigInterface;
 use DemosEurope\DemosplanAddon\Contracts\Entities\RoleInterface;
+use DemosEurope\DemosplanAddon\Contracts\Logger\ApiLoggerInterface;
+use DemosEurope\DemosplanAddon\Contracts\MessageBagInterface;
 use DemosEurope\DemosplanAddon\Contracts\PermissionsInterface;
 use DemosEurope\DemosplanAddon\Controller\APIController;
 use DemosEurope\DemosplanAddon\Logic\ApiRequest\TopLevel;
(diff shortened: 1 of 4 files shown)


client/js/components/map/admin/AdminLayerListItem.vue
```diff
@@ -106,7 +106,7 @@
         <input
           type="checkbox"
           data-cy="adminLayerListItem:toggleDefaultVisibility"
-          :disabled="'' !== layer.attributes.visibilityGroupId || (true === isChildOfCategoryThatAppearsAsLayer)"
+          :disabled="null !== layer.attributes.visibilityGroupId || (true === isChildOfCategoryThatAppearsAsLayer)"
```
salisdemos, line 109: may it would be even more resiliant like so:
```suggestion
          :disabled="layer.attributes.visibilityGroupId || (true === isChildOfCategoryThatAppearsAsLayer)"
```

client/js/components/map/admin/AdminLayerListItem.vue
```diff
@@ -825,7 +825,7 @@ export default {
     toggleHasDefaultVisibility () {
       this.preventActiveFromToggeling = true
       // Can't be updated when it's a visiblityGroup
-      if ((this.layer.attributes.visibilityGroupId !== '' && this.layer.type !== 'GisLayerCategory') || this.isLoading) {
+      if ((this.layer.attributes.visibilityGroupId !== null && this.layer.type !== 'GisLayerCategory') || this.isLoading) {
```
salisdemos, line 828:
```suggestion
      if ((this.layer.attributes.visibilityGroupId && this.layer.type !== 'GisLayerCategory') || this.isLoading) {
```

salisdemos (changes requested):
I think if we check for falsy/truthy its even more reliable.So just couple of little adjustments.

PR #4724, opened 2025-05-27
chore: Update js dependencies

salisdemos:
Bump already bumbed Dependencies
Changes where Lost during some backemerge
@see ce372e6690166dae8131edd1cf5ffe01b14613ca


salisdemos:
> Do I understand correctly that this is just restoring some previously made changes?
correct.

> Maybe eslint-config-standard should be updated, too?
Will come. Would do it in a separate PR

hwiem (approved):

PR #4725, opened 2025-05-27
chore: Skip Tests in PublicDetailLayerList

salisdemos:
to get them running, the usage of $root.$on has to be removed. This is a quite complex task. And to get the test pass, we decided to skipp them for now until we managed to refactor the root.ons


hwiem (approved):
Thank you!

PR #4735, opened 2025-05-30
f DPLAN-15765 use Symfony's string component to provide Unicode handling, including whitespace and normalization functions

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15765/Abschnittsimport-Upoload-gibt-Fehlermeldung

Description: 
use Symfony's string component to provide Unicode handling, including whitespace and normalization functions:
Many "invisible" or "special" characters can find their way into Excel ".xlsx" cell data, especially with copy-paste, imports, or exports between various systems and formats which causes errors while uploading segments, that's why the tags titles strings have to be cleaned before be used and added.


mussbach (approved):
lgtm, code review only, test would be nice

bakridemos:
> lgtm, code review only, test would be nice

oh sorry its already merged, but i ll try to make tests in other pr

PR #4743, opened 2025-06-02
feat: provide a securityVoter to check if user is active

mussbach:
### Ticket
No specific ticket - security infrastructure improvement

Provide a security voter to check if a user is active, enabling proper authorization checks based on user activity status throughout the application.

This security voter integrates with Symfony's authorization framework to automatically deny access to inactive users across all protected resources, providing a centralized way to enforce user activity requirements without modifying individual controllers or services.

### How to review/test
- [ ] Review the security voter implementation for correct logic
- [ ] Test that active/inactive users get appropriate access

### PR Checklist
- [ ] Create/Update tests
- [ ] Run `yarn lint`
- [ ] Run `yarn test`

### DIFF
--- a/config/parameters_default.yml
+++ b/config/parameters_default.yml
@@ -43,6 +43,10 @@ parameters:
     # who are granted access (true)
     procedure_user_restricted_access: false
 
+    # User activity checking thresholds (in days)
+    user_activity_last_login_threshold_days: 180
+    user_activity_claimed_statements_threshold_days: 180
+
     # Route and parameters for the public index page
     #
     # - Using the controller forward notation here will result in an internal forward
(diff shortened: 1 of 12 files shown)


sierrazamudiodemos (commented):
Thanks for this Sami, I have few questions:

1, Where will the UserActivityVoter be used?
2. is the "security voter" pattern something that can be used for accessConditions in the resource types or maybe im mixing concepts?
3. I think i understood the concept of the vote itself and how it is used. Is this something that we want to use in the future, or existing functionality that we should replace on behalf of the voter?
4. can we replace the permissions approach that we ahve and instead use voters?

mussbach:
1. the securityVoter will be used in an addon but imho it is a sensible feature to have in core.
2. Yes, it should be possible albeit not tested
3. I definitely think that we could make wider use of the concept. But we need to be a bit careful not to stretch it too much. I think this use case is at the edge of the initial use case of the SecuritzVoters

sierrazamudiodemos (approved):
Thanks for the explanation. I think sonar cloud is complaining

PR #4773, opened 2025-06-06
fix DPLAN-15816 Adjust error message to import

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-15816/Falsches-Fehlermeldung-beim-Abschnitte-Import

- Add new translation key to display accurately that the step (or also called place) is missing in order to import the segments
- There is an improvement that the Exception is too general, and the message to specific. One could detect from the exception what entity was missing and then consequently display the message.
- For now this should be enough

### How to review/test
- Go to Verfahren and then Click on Import
- Then import segments
- You should see a new error message that says to add first a step before importing segments


martendemos (approved):
codereview only

PR #4796, opened 2025-06-12
fix(refs dplan-15818): Enable expanding custom fields options

rafelddemos:
### Ticket
[DPLAN-15818](https://demoseurope.youtrack.cloud/issue/DPLAN-15818/Konfigurierbar-Abschnitssfelder-Accordion-ist-kaputt-nicht-moglich-um-alle-Optionen-zu-sehen)

This PR deals with the caret-btn that should expand the custom fields options when clicked (and there are more than 2 options available). The reason why it didn't work is that the show/hideOptions handlers were trying to mutate the 'open'-property returned from the computed property 'customFieldsReduced', which broke the reactivity. This PR moves the state to the component's data, and uses it for the updates inside the handlers and the computed property.

### How to review/test

Go to 'Konfigurierbare Abschnittssfelder' --> create a custom field with at least 3 options --> save it --> it should be possible to expand the options list.

### PR Checklist

- [ x ] Move the tickets on the board accordingly


client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -360,11 +357,7 @@ export default {
     },
 
     showOptions (rowData) {
-      const field = this.customFieldsReduced.find(field => field.id === rowData.id)
-
-      if (field) {
-        field.open = true
-      }
+    this.expandedFields[rowData.id] = true
```
hwiem, line 360:
```suggestion
      this.expandedFields[rowData.id] = true
```

rafelddemos: fixed in 13fa87b8b7565fed0d6d205d63fd0f8c6e240daa

PR #4809, opened 2025-06-16
ref (DPLAN-15835): handle set phase in statement, if phase not set

jalalifardemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15835/Nicht-moglich-um-eine-Stellungnahme-zu-einreichen

Description: set Configuration as a phase if it doesn't set manually


bakridemos (approved):
lgtm, code review only

PR #4818, opened 2025-06-17
DPLAN-15822: fix draft pdf export inPublicView as Admin

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-15822/Offentlichkeitsansicht-Download-als-PDF-weiterleitet-zu-einer-Fehlerseite

Description:
In case you log in as an admin and chose the publicView of a procedure. There you navigate to stn-drafts - do not choose any - chlick downlad pdf. Ann empty array gets send for itemsToExport
Adjust the logic to handle an empty array like null - which means getAll.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/DraftStatementService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/DraftStatementService.php
@@ -893,9 +893,13 @@ class DraftStatementService extends CoreService
 
         $filenameSuffix .= '.pdf';
 
-        $selectedStatementsToExport = isset($itemsToExport)
-            ? explode(',', $itemsToExport)
-            : null;
+        $selectedStatementsToExport = null;
+        if (isset($itemsToExport) && is_string($itemsToExport)) {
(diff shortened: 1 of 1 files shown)


sierrazamudiodemos (approved):

sierrazamudiodemos (changes requested):
- Thank you for splitting the methods! I think it makes more readable each step of what the method is doing.
- Thank you as well for the test! 
- I would still keep the collection instead of using php arrays. It seems like in our code base, the collection has been used, and also there is intentionally a library that is installed for it lluminate/collections

MoritzMandler:
use the illuminate/Collection again :(
208811cca7d664b702fcc2077acff599528901ae
0747de83917019c45ea717fd77432c8e7a757d77

sierrazamudiodemos (approved):
thanks for this

PR #4820, opened 2025-06-18
feature (DPLAN-12319): Set Default date time zone

jalalifardemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12319/falsche-Zeitangabe-Versionen-der-Stellungnahme, https://demoseurope.youtrack.cloud/issue/DPLAN-15853/Falsches-Uhrzeit-beim-Verfahrensexport

Description: Set Default date/time zone


bakridemos (changes requested):
i think this can be done differently without setting the timezone in the kernel, i would like to try something else and we can decide later together which solution is better

PR #4830, opened 2025-06-19
feat(refs: DPLAN-15854): update diplan-karten from 0.6.1 to 0.7.2

huellnerdemos:
### Ticket
DPLAN-15854

The new version of diplan-karten is needed, to add the information of the local reference to a statement. The diplan-karten updates from 0.6.1 to 0.7.2 

### How to review/test

1. Open diplanfest and open a procedure
2. In the tab "Interaktive Karte" the map should be loaded


meissnerdemos (changes requested):
This merge changed DiplanKarteWrapper.vue to have two definitions for `props`.
The definition should be: 
`
const props = defineProps({
  loginPath: {
    type: String,
    required: true
  },
  styleNonce: {
    type: String,
    required: true,
  }
})
`

The definition later in the file (around line 50) needs to be removed.

PR #4835, opened 2025-06-20
fix(refs dplan-15812): Await all Promises before leaving the page when saving Layers

salisdemos:
### Ticket
DPLAN-15812

When saving layersort ls FPA, there are two ways: "save" and "save and return". On later one, it did not saved properly because the async save method did not waited untill all save actions (for each layer one request) where settled.


riechedemos (approved):
Looks legit!

PR #4872, opened 2025-07-01
refactor(DPLAN-15945): optimize tag and boilerplate import logic

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-15945/Importierte-Schlagworte-zu-thematisch-andern-Uncaught-in-promise-TypeError-el.relationships.tag-is-undefined

Description:
These changes prevent duplicate creation of tags, topics, and boilerplates during
import operations while preserving all existing functionality.
image

- Refactored TagTopic creation to check for existing topics before creating new ones
- Improved Tag creation to reuse existing tags within topics using Doctrine Collection filter
- Enhanced boilerplate handling to avoid duplicates and only attach to newly created boilerplates
- Extracted tag import logic into handleTagImport() helper method
- Extracted boilerplate import logic into handleBoilerplateImportForTag() helper method
- Improved error handling with general Exception catching
- Added proper @throws documentation to new methods
- Maintains compatibility with addon

The Excelimport can update an AddonRelation as well as core Tag/TagTopic entities within the same process.
This lead to an Event interference between the UpdateTagEvent and the ExcelImporterPrePersistTagsEvent.
Therefore the core logic within the ExcelImporterPrePersistTagsEvent was adjusted to circumvent triggering the
UpdateTagEvent during its processing.

How to review/test
 - upload Tags - delete some of them - use the same upload....
 with these changes only the new tags included in the import file will be created - the others updated.

###
- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/Logic/Statement/TagService.php
```diff
@@ -160,6 +160,32 @@ public function moveTagToTopic($tag, $newTopic): bool
...
+        $newTopic->addTag($tag);
+        $tag->setTopic($newTopic);
+
+        $tagUpdated = $this->tagRepository->updateObject($tag);
+        $topicUpdated = $this->tagTopicRepository->updateObject($newTopic);
+
+        // Skip the event dispatch to avoid dual event handling
+
+        return $tagUpdated instanceof Tag && $topicUpdated instanceof TagTopic;
+    }
```
sierrazamudiodemos, line 187: Hey @MoritzMandler thank you for handling this edge case, I see a bit of code duplication when comparing to ethe method moveTagToTopic. What about adjust the existing method moveTagToTopic and add a parameter to trigger the event? and  $dispatchEvent Whether to dispatch UpdateTagEvent (default: true)
  public function moveTagToTopic($tag, $newTopic, bool $dispatchEvent = true): bool

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementHandler.php
```diff
@@ -2501,6 +2499,113 @@ public function consolidateAdditionalAreaInformationMessages(Collection $message
...
+     *
+     * @throws Exception
+     */
+    protected function handleBoilerplateImportForTag(Tag $tag, string $title, string $text, string $procedureId): void
+    {
+        $existingBoilerplates = $this->procedureService->getBoilerplateList($procedureId);
+        $matchingBoilerplates = array_filter(
+            $existingBoilerplates,
+            fn (BoilerplateInterface $boilerplate) => $boilerplate->getTitle() === $title
+        );
```
sierrazamudiodemos, line 2589: hey, do you think it might be possible to use collections here? something like       
`$matchingBoilerplates = collect($existingBoilerplates)->filter(fn (BoilerplateInterface $boilerplate) => $boilerplate->getTitle() === $title);`

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementHandler.php
```diff
@@ -2501,6 +2499,113 @@ public function consolidateAdditionalAreaInformationMessages(Collection $message
...
+            // Create new boilerplate and attach to tag
+            $boilerplateData = ['title' => $title, 'text' => $text];
+            $boilerplate = $this->procedureService->addBoilerplate($procedureId, $boilerplateData);
+            $this->tagService->attachBoilerplateToTag($tag, $boilerplate);
+
+            return;
+        }
+
+        // Update existing boilerplate text if different
+        $boilerplate = reset($matchingBoilerplates);
```
sierrazamudiodemos, line 2601: in case we use collections, then maybe we can transform this one to      
`$boilerplate = $matchingBoilerplates->first();`
I understand that the reset gets the first element, so we can make it a bit more readable that it gets the first element using collections

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementHandler.php
```diff
@@ -2501,6 +2499,113 @@ public function consolidateAdditionalAreaInformationMessages(Collection $message
...
+     *
+     * @param string   $tagTitle    The title of the tag to create or retrieve
+     * @param TagTopic $topic       The target topic for the tag
+     * @param string   $procedureId The procedure ID for uniqueness checking
+     *
+     * @return Tag The tag entity (either existing, moved, or newly created)
+     *
+     * @throws Exception
+     */
+    protected function handleTagImport(string $tagTitle, TagTopic $topic, string $procedureId): Tag
```
sierrazamudiodemos, line 2539: Maybe we can test this method if possible? I think we do not need to make it public, it is possible to use Reflection I think to test protected and private methods

sierrazamudiodemos (changes requested):

demosplan/DemosPlanCoreBundle/Logic/Statement/TagService.php
MoritzMandler: done in 18073db7345cbad52b54620e105e1a003e363350

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementHandler.php
MoritzMandler: I found an issue with the handling of Boilerplates and changed a lot here 18073db7345cbad52b54620e105e1a003e363350

sierrazamudiodemos (approved):
Hey @MoritzMandler thanks for changes and moving to collection!

Approved! few quick thoughts:

- I think SonarCloud is complaning because of some repeated code in the tests.

1. The case-sensitive comparison (===) for title/text matching - is this intentional? Just wondering if "Hello" vs "hello" should be treated as different boilerplates.
2.  Worth considering: could we extract the tag import logic to a dedicated service? I think the StatementHandler is getting a little bigger.

Thanks for the implementation and the tests!

PR #4878, opened 2025-07-03
B dplan 15915 fix file display global news

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-15915/Bild-und-PDF-Upload-wird-nach-Speichern-nicht-ubernommen-angezeigt

The problem is that the template `news_admin_edit.html` is using the `core_file_procedure` route which requires a `procedureId` parameter. When rendering the template for the global news, procedure is null, so we need to use the `core_file` route instead.

This template is used for rendering both procedure and global news templates, that is why we should check if we use core_file_procedure or code_file to show the pdf url consequently

### How to review/test
- Create a Global News using MandatenRole, use an image and pdf
- Logout and check the global news in public view
- Image and Pdf should be rendered

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanNews/news_admin_edit.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanNews/news_admin_edit.html.twig
@@ -317,12 +317,18 @@
 
                 <p class="lbl">{{ "pdf.document"|trans }}</p>
 
+                {% if procedure is not null %}
+                    {% set filePath = path("core_file_procedure", { 'procedureId': procedure, 'hash': news.pdf|getFile('hash') }) %}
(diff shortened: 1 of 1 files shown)


bakridemos (approved):
It depends on the context, but in one context files are either linked to a procedure or not. it can't be sometimes with procedure id and sometimes without procedure id. Notices 'Mitteilung' do not need to be linked to procedure. I think the route core_file without procedure as parameter is thr right one and should fix it.

bakridemos:
i approved this pr by mistake, and this will be fixed in other pr

PR #4882, opened 2025-07-03
Feat dplan 16042 add configurable feedback control for public participation

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-16042/Ruckmeldung-zu-Stellungnahme-steuerbar

Description:  this feature adds administrative control over citizen feedback functionality in public participation procedures. Administrators can now enable or disable the feedback option that allows citizens to request notifications about statement processing results.

  - New checkbox in procedure settings to toggle feedback availability
  - Only users with feature_feedback_on_statement_controllable permission can modify this setting
  - When disabled, all feedback-related form elements (checkbox, email fields, postal options) are hidden from citizens

### How to review/test

You can now test this by toggling the publicParticipationFeedbackEnabled setting in the procedure administration form, and the feedback functionality should appear/disappear accordingly.

- [X] Update changelog


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Export Original Statements as docx in the Statement List
 - Allow filtering of institution tags in DpAddOrganizationList
 - Allow filtering of institution tags in AdminstrationMemberList / refactor twig
+- add configurable feedback control for public participation statements
```
sierrazamudiodemos, line 12: Thank you for adding the changelog! Super minor details: what about also starting with upper case? Add

demosplan/DemosPlanCoreBundle/Entity/Procedure/ProcedureSettings.php
```diff
@@ -291,6 +291,15 @@ class ProcedureSettings extends CoreEntity implements UuidEntityInterface, Proce
...
     private $allowedSegmentAccessProcedures;
 
+    /**
+     * Enable publication of Feedback possibility.
+     *
+     * @var bool
+     *
+     * @ORM\Column(name="_p_public_participation_feedback_enabled", type="boolean", nullable=false, options={"default":true})
+     */
+    protected $publicParticipationFeedbackEnabled = true;
```
sierrazamudiodemos, line 301: what about adding bool? like protected bool $publicParticipationFeedbackEnabled = true;

demosplan/DemosPlanCoreBundle/Entity/Procedure/ProcedureSettings.php
```diff
@@ -1114,4 +1123,16 @@ public function setDesignatedPublicPhaseChangeUser(?UserInterface $designatedPub
 
         return $this;
     }
+
+    public function setPublicParticipationFeedbackEnabled(bool $enabled): ProcedureSettingsInterface
+    {
+        $this->publicParticipationFeedbackEnabled = \filter_var($enabled, FILTER_VALIDATE_BOOLEAN);
```
sierrazamudiodemos, line 1129: I think we can remove \filter_var($enabled, FILTER_VALIDATE_BOOLEAN)
This method is only called from ProcedureRepository.php:808 with values that are already boolean literals (true/false) from checkbox processing in ServiceStorage.php:498-502. Since the method signature already enforces bool $enabled, the filtering is redundant. I think we can remove it safely.

sierrazamudiodemos (commented):
Hey @bakridemos thank you very much for this, I added few small stuff! it looks very good!

sakutademos (approved):
non-blocking comment, FE lgtm, code review

client/js/components/statement/publicStatementModal/formGroups/FormGroupEvaluationMailViaEmail.vue
```diff
@@ -8,7 +8,9 @@
 </license>
 
 <template>
-  <div :class="[statement.r_getFeedback === 'on' ? prefixClass('bg-color--grey-light-2') : '', prefixClass('c-statement__formblock')]">
+  <div 
+    v-if="publicParticipationFeedbackEnabled"
+    :class="[statement.r_getFeedback === 'on' ? prefixClass('bg-color--grey-light-2') : '', prefixClass('c-statement__formblock')]">
```
sakutademos, line 13: maybe it could be simplified/refactored

```suggestion
    :class="[
      prefixClass('c-statement__formblock'),
      { [prefixClass('bg-color--grey-light-2')]: statement.r_getFeedback === 'on' }
    ]"
  ```

demosplan/DemosPlanCoreBundle/Entity/Procedure/ProcedureSettings.php
sierrazamudiodemos: yep I saw that as well, I think there might be some advantages to adding bool to the property:
  - PHP will throw a TypeError if we put something other than true/false in this property, which could help us find mistakes faster.
  - Our IDE will give us better suggestions and warnings when we're working with this property because it knows exactly what type it should be.
  - From what I've seen, this seems to be how PHP properties are typically written nowadays - it might be easier to read and understand what the property is supposed to contain.

What do you think?

bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/4882/commits/af20bc39eac12818dd5eaba5060e01502020d56b

sierrazamudiodemos (approved):
Thank you!

PR #4894, opened 2025-07-07
fix(refs: DPLAN-15198): disable copy button if no item selected

huellnerdemos:
### Ticket
[DPLAN-15198](https://demoseurope.youtrack.cloud/issue/DPLAN-15198/Falsches-Notification-Reihenfolge-Bestatigungsnachricht-angezeigt-obwohl-keine-Originalstellungnahmen-selektiert-sind)

The copy button in the original statement list dialog has to be disabled, to prevent wrong notifications or a wrong order of notifications, when no item is selected.

### How to review/test

1. Open diplanbau and navigate to the original statement list dialog
2. Check if the copy button is disabled, as long as no item is selected
3. Select an item and check, if the copy button turns active

### PR Checklist

### DIFF
--- a/client/js/components/statement/originalStatementsTable/OriginalStatementsTable.vue
+++ b/client/js/components/statement/originalStatementsTable/OriginalStatementsTable.vue
@@ -76,7 +76,7 @@
       :procedure-id="procedureId" />
 
     <slot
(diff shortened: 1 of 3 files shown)


meissnerdemos (approved):
Looks good and should work once the other bug is fixed. Thanks!

PR #4899, opened 2025-07-07
feat (ref: DPLAN-15915 ) : switch routes depends on global news or not

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15915/Bild-und-PDF-Upload-wird-nach-Speichern-nicht-ubernommen-angezeigt

Description:
The Twig template news_admin_edit.html.twig always attempted to use core_file_procedure, even if procedure was empty (for Global News):

 Distinguish between news types with templateVars.isGlobal:
 - Global News (templateVars.isGlobal == true): Use core_file route no procedureId required)
 - Procedure News (templateVars.isGlobal is not set): Use core_file_procedure route (with procedureId)

 This logic ensures that:
 - Global News files are accessible via core_file (since they are not bound to procedures)
 - Procedure News files are accessible via core_file_procedure (with appropriate access control)

picked from https://github.com/demos-europe/demosplan-core/pull/4886

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanNews/news_admin_edit.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanNews/news_admin_edit.html.twig
@@ -317,18 +317,19 @@
 
                 <p class="lbl">{{ "pdf.document"|trans }}</p>
 
-                {% if procedure is not null %}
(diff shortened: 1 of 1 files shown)


mussbach:
But wait, why? Merge of release into main should suffice

PR #4902, opened 2025-07-08
fix(refs: DPLAN-15083): replace VPopover with DpTooltip

meissnerdemos:
### Ticket
[DPLAN-15083](https://demoseurope.youtrack.cloud/issue/DPLAN-15083)

Replaced outdated VPopover with DpTooltip from demosplan-ui. I also moved the logic for constructing the text out of the template and into computeds.

### How to review/test
1. Open a project that still has the AssessmentTable, ideally robobsh.
2. Check if the tooltips contain the correct text and are fit to the content.


salisdemos (approved):
Thank You. Nice solution. Just one noneblocking nitpic.

PR #4912, opened 2025-07-09
fix an error that is preventing the build process

sakutademos:
**Description:** Fix an issue in main that is preventing the build process.


hwiem (approved):
Lgtm!

PR #4916, opened 2025-07-10
fix(refs: DPLAN-16083): add new location popup style for mobile

huellnerdemos:
### Ticket
DPLAN-16083

After marking a location on map a popup opens on the marked position. This popup is too small in mobile view, so that the user cannot see the button to switch back to the statement modal. This change fix the style.

### How to review/test

1. Open diplanbau or diplanrog and login as a private user
2. Open a procedure and switch to mobile view
3. Switch to mark location tab and mark a location on the map
4. Check if the popup is shown, has at least a height of 114px and the button inside the popup is visible 

### PR Checklist


huellnerdemos:
There was is a bahavior only in mobile view, that the little triangle is not positioned at the place where the user clicked the map, the upper left corner of the popup is placed there instead. I discussed the behavior with @salisdemos. He found out a better approach to fix the behavior. Therefor i close this PR and make a new one, after i discussed the new approach with PO.

PR #4917, opened 2025-07-10
feat(refs: DPLAN-15854): update diplan-karten to 1.0.1

huellnerdemos:
### Ticket
DPLAN-15854

Updates diplan-karten to the version 1.0.1

### How to review/test

1. Login as a private user, open a procedure and check if the map is shown in the tab "Interaktive Karte"


salisdemos (approved):
legit

PR #4930, opened 2025-07-14
chore(refs DPLAN-661): Use checkresponse from dpApi

salisdemos:
### Ticket
DPLAN-661

dpApi will use it always.

With this change the response from checkResponse changes. Instead of returning `response.data`, it now returns `response`. So we have to adjust the structure on all places where we removed checkResponse in here.

This change was necessary because that is the response that would come when checkResponse is not in use. And its much harder to find all occurrences where dpApi is used but not checkResponse than check the occurrences, where we had to remove checkresponse anyway.

And: dpRPC even don't always return a `response.data` but there too we get checkResponse out of the box now.

As a hint for Reviewers:
If a checkResponse is removed. The response payload has to have been adjusted. So if You spot a line, where the checkResponse is removed and no other changes are made, check if the response body is used. If not. Its a mistake.

requires: https://github.com/demos-europe/demosplan-ui/pull/1316/

### DIFF
--- a/client/js/components/map/admin/AdminLayerList.vue
+++ b/client/js/components/map/admin/AdminLayerList.vue
@@ -436,17 +436,20 @@ export default {
       dragClass: 'o-sortablelist__drag' // Class name for the dragging item
     }
 
-    this.updateState({ key: 'draggableOptions', value: {
-      ...basicOptions,
-      ...{
-        group: {
-          name: 'treeList',
-          revertClone: false,
-          pull: ['treeList'],
-          push: ['treeList']
+    this.updateState({
+      key: 'draggableOptions',
+      value: {
+        ...basicOptions,
+        ...{
+          group: {
+            name: 'treeList',
+            revertClone: false,
+            pull: ['treeList'],
+            push: ['treeList']
+          }
         }
       }
-    }})
+    })
 
     this.updateState({ key: 'draggableOptionsForBaseLayer', value: basicOptions })
   }


client/js/store/statement/Statement.js
```diff
@@ -757,7 +751,6 @@ export default {
           Accept: 'application/vnd.api+json'
         }
       })
-        .then(this.api.checkResponse)
         .then(response => {
           let assignee = {}
           if (assigneeId === '' || assigneeId == null) {
```
hwiem, line 756: Shouldn't it be
```
assignee = { id: response.data.data.id, uId: response.data.data.id, name: response.data.data.attributes.name, orgaName: response.data.data.attributes.orgaName }
```
in the else block below?

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -954,13 +946,17 @@ export default {
 
     triggerStatementDeletion (id) {
       if (window.confirm(Translator.trans('check.statement.delete'))) {
+        // Override the default success callback to display a custom message
+        this.$store.api.successCallbacks[0] = async (success) => this.$store.api.handleResponse(success, {
+          200: { type: 'confirm', text: Translator.trans('confirm.statement.deleted') },
+          204: { type: 'confirm', text: Translator.trans('confirm.statement.deleted') }
+        })
+
```
hwiem, line 954: Is this how we need to configure custom messages for the vuex json api lib requests?

hwiem (changes requested):
Just some small change requests. Thanks for cleaning this up! :pray:

client/js/components/statement/listStatements/ListStatements.vue
salisdemos: exactly.
At least I didn't find a leaner way

client/js/components/document/DpMapSettingsPreview.vue
```diff
@@ -460,7 +460,7 @@ export default {
             }
           }
         }
-      }).then(checkResponse)
+      })
```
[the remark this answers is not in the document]
salisdemos: true 18df15fb8c2ea751346db339a4f1705bc2590027

client/js/store/statement/Statement.js
salisdemos: thanks for spotting f67ecc730d8004187e520ddf420c2fadc5277058

hwiem (approved):
Lgtm

PR #4935, opened 2025-07-15
fix(refs: DPLAN-16083): change location popup style for mobile

huellnerdemos:
### Ticket
DPLAN-16083

After marking a location on map a popup opens on the marked position. This popup is too small in mobile view, so that the user cannot see the button to switch back to the statement modal. This change fix the style.

### How to review/test

1. Open diplanbau or diplanrog and login as a private user
2. Open a procedure and switch to mobile view
3. Switch to mark location tab and mark a location on the map
4. Check if the popup is shown and you can see the all content of it including the Button
5. Check if the small triangle points nearly at the position where the location was clicked
6. Check if the Button text now explains more understandable what happens when the Button is clicked.

### PR Checklist

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_map.scss
+++ b/demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_map.scss
@@ -644,23 +644,6 @@ $map-layer-color-active-hover: $dp-token-color-brand-cta-dark;
 
         //  popup triggered with mapSingleClick()
         &__popup {
-
-            //  set popover fixed for mobile devices and handle long text-overflow
-            @include media-query('palm') {
-                inset: 2% 2% auto;
-                position: fixed;
-                z-index: 1040;
-
-                max-height: 96%;
-                min-height: 0;
-                overflow-y: auto;
-
-                //  hide the popup arrow which can not be positioned with overflow-y: auto
-                &::after {
-                    display: none;
-                }
-            }
-
             bottom: 12px;
             left: -50px;
             position: absolute;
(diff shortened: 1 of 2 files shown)


demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_map.scss
```diff
@@ -644,23 +644,6 @@ $map-layer-color-active-hover: $dp-token-color-brand-cta-dark;
...
-                max-height: 96%;
-                min-height: 0;
-                overflow-y: auto;
-
-                //  hide the popup arrow which can not be positioned with overflow-y: auto
-                &::after {
-                    display: none;
-                }
-            }
-
```
huellnerdemos, line 663: I discussed a few possibilities with Florian how to get a good mobile version, where everything, size, triangle, position is at the right place. In the end the solution is, to delete the mobile style for the popup, because this gives us what we want. The only thing is, that on very small displays, you cannot see the whole popup. But the map can be moved and the popup moves with it, so it is possible to see the whole popup when moving the map a bit even on very small displays. I discussed this point with PO and they agreed on this behavior.

templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_public_participation_detail.html.twig
```diff
@@ -427,7 +427,7 @@
                     <button id="popupAction" class="{{ 'c-map__popup__button btn btn--primary u-mt-0_5'|prefixClass }}">
                         <i :class="activeStatement ? prefixClass('fa fa-commenting') : prefixClass('fa fa-comment')" aria-hidden="true"></i>
                         <template v-if="activeStatement">
-                            {{ 'statement.form.edit.location_relation'|trans }}
+                            {{ 'statement.continue'|trans }}
                         </template>
```
huellnerdemos, line 431: This change was a suggestion from my first and closed PR for this case: 
https://github.com/demos-europe/demosplan-core/pull/4916/files
I discussed it with PO and then changed it here.

salisdemos (approved):
less is more. I like it.

hwiem (approved):
Nice :sparkles: 
(I think this should be merged into `main` now)

PR #4936, opened 2025-07-15
fix (refs: DPLAN-15999): Use all() method is designed to handle array…

sierrazamudiodemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-15999
Change get() to all() to get an array from the request

Same fix as this PR https://github.com/demos-europe/demosplan-core/pull/4852 applied to ewm


MoritzMandler (approved):
thx a lot :)

PR #4941, opened 2025-07-15
fix(refs DPLAN-16060): Add Loaded flag for mapSettings

salisdemos:
To prevent that the map get initialized before the settings are loaded, we add a loading flag. If this helps is hard to say, because the bug is not reliable reproducible. So lets give it a try.

### Ticket
DPLAN-16060


mussbach (approved):
valid try, thx

PR #4949, opened 2025-07-16
fix: resolve critical PHPStan errors

bakridemos:
Description: resolve PHPStan errors
  ## Faker Generator state() method fix
  - Problem: `$this->faker->state()` undefined method for German locale
  - Solution: Custom German states array with `randomElement()` for authentic data

  ## Mixed-type comparison operations fix
  - Problem: Unsafe comparisons between int and mixed types from HTTP requests
  - Solution: Added explicit `(int)` type casting for request parameters

  ## Doctrine ObjectManager getReference() fix
  - Problem: `getReference()` method undefined on ObjectManager interface
  - Solution: Use injected EntityManagerInterface instead of getDoctrine()->getManager()


mussbach (approved):
lgtm, code review only, but one merge conflict

PR #4952, opened 2025-07-17
fix(refs: DPLAN-16074): Move getFeatureInfo request to the Frontend

salisdemos:
by now the request was send to the be and from there fetched from the source. due to security policies we have to fetch straight from the FE.

### Ticket
DPLAN-16074


mussbach (approved):
thanks a lot! I reviewed and tried it, it worked. I added some sanitizing, pls check before merge

PR #4953, opened 2025-07-17
fix(refs DPLAN-16134): Only send valid fields when requesting Layers

salisdemos:
### Ticket
DPLAN-16134

When a request is used in different places, its important to check, that the requested fields are necessary and even more important available in that circumstance.

### DIFF
--- a/client/js/components/map/publicdetail/Map.vue
+++ b/client/js/components/map/publicdetail/Map.vue
@@ -47,7 +47,7 @@ import Feature from 'ol/Feature'
 import { getResolutionsFromScales } from '@DpJs/components/map/map/utils/utils'
 import { getTopLeft } from 'ol/extent'
 import LayerGroup from 'ol/layer/Group'
-import { mapState } from 'vuex'
+import { mapActions, mapState } from 'vuex'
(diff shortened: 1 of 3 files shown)


client/js/components/map/publicdetail/Map.vue
```diff
@@ -2014,7 +2014,44 @@ export default {
...
+          'opacity',
+          'projectionLabel',
+          'treeOrder',
+          'url',
+          'visibilityGroupId'
+        ].join()
+      }
+    }
+
+    this.$store.dispatch('Layers/get', payload).then(() => {
```
hwiem, line 2054: Could we use `mapActions` here?

client/js/store/map/Layers.js
```diff
@@ -542,18 +511,12 @@ const LayersStore = {
 
       return dpApi.patch(Routing.generate('api_resource_update', { resourceType: resource.type, resourceId: resource.id }), {}, payload)
         .then(checkResponse)
-        .then(() => {
-          dispatch('get', state.procedureId)
```
hwiem, line 546: Why was this removed?

PR #4962, opened 2025-07-22
feat (refs: DPLAN-15829): show layer preview for admin

meissnerdemos:
### Ticket
[DPLAN-15829](https://demoseurope.youtrack.cloud/issue/DPLAN-15829/Ado-Issue-30419-Live-Vorschau-und-Layerwahl-beim-Einbinden-von-Planwerken-ermoglichen-um-die-Nutzerfreundlichkeit-zu-erhohen)

This PR adds a preview map for the external layer configuration for admins. The map takes the existing procedure settings and adds the currently selected layers from an external WMS/WMTS service. This makes it easier to decide which layers are needed.

### How to review/test
1. Login as admin, select a procedure, navigate to planning documents, navigate to map layers.
2. Either add a new layer or edit an existing one.
3. You should see a small preview map.
4. Change which layers are selected - the map should change.

I used [this WMS service](https://via.bund.de/bmf/inspire/so/wms?service=wms&request=getcapabilities) for testing. If you use a different service, please check in the network tab if it actually sends data and is not blocking you because of too many requests.

### DIFF
--- a/client/js/components/map/admin/LayerSettings.vue
+++ b/client/js/components/map/admin/LayerSettings.vue
@@ -88,6 +88,12 @@
       @deselectAll="deselectAllLayers"
     />
 
+    <dp-ol-map
+      v-if="hasPreview"
+      :procedure-id="procedureId"
+      small
+    />
+
     <input
       :value="layersInputValue"
       name="r_layers"
@@ -142,11 +148,13 @@
 <script>
 import { debounce, DpCheckbox, DpInput, DpLabel, DpMultiselect, DpSelect, externalApi } from '@demos-europe/demosplan-ui'
 import { WMSCapabilities, WMTSCapabilities } from 'ol/format'
+import DpOlMap from '../map/DpOlMap.vue'
 
 export default {
   name: 'LayerSettings',
 
   components: {
+    DpOlMap,
     DpCheckbox,
     DpInput,
     DpLabel,
@@ -159,6 +167,12 @@ export default {
       type: Array,
       required: false,
       default: () => []
+    },
+
+    hasPreview: {
+      type: Boolean,
+      required: false,
+      default: false
     },
 
     initLayers: {
@@ -198,6 +212,12 @@ export default {
     },
 
     initVersion: {
+      type: String,
+      required: false,
+      default: ''
+    },
+
+    procedureId: {
       type: String,
       required: false,
       default: ''
@@ -524,6 +544,8 @@ export default {
   },
 
   mounted () {
+    console.log('hier')
+    console.log(this.currentProcedureId)
     this.getLayerCapabilities()
   }
 }
--- a/client/js/components/map/admin/LayerSettings.vue
+++ b/client/js/components/map/admin/LayerSettings.vue
@@ -88,16 +88,17 @@
       @deselectAll="deselectAllLayers"
     />
 
-    <dp-ol-map
-      v-if="hasPreview"
-      :procedure-id="procedureId"
-      small
-    />
-
     <input
       :value="layersInputValue"
       name="r_layers"
       type="hidden">
+
+    <dp-ol-map
+      v-if="hasPreview"
+      :layers="previewLayers"
+      :procedure-id="procedureId"
+      small
+    />
 
     <dp-select
       v-if="hasPermission('feature_map_wmts') && serviceType === 'wmts'"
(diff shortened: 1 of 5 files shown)


client/js/components/map/map/DpOlMapLayer.vue
```diff
@@ -125,17 +132,23 @@ export default {
...
-      this.map.getLayers().insertAt(this.order, layer)
+      this.map.getLayers().insertAt(this.order, this.layer)
     }
   },
 
   mounted () {
     this.addLayer()
   },
 
+  beforeUnmount () {
```
hwiem, line 146: Why is this needed?

templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_admin_gislayer_edit.html.twig
```diff
@@ -156,13 +156,15 @@
 
                 {% block baseSettings %}
                     <layer-settings
+                        has-preview="true"
```
hwiem, line 159:
```suggestion
                        has-preview
```

templates/bundles/DemosPlanCoreBundle/DemosPlanMap/map_admin_gislayer_new.html.twig
```diff
@@ -105,6 +105,7 @@
 
         {% block baseSettings %}
             <layer-settings
+                has-preview="true"
```
hwiem, line 108:
```suggestion
                has-preview
```

client/js/components/map/admin/LayerSettings.vue
```diff
@@ -271,6 +292,16 @@ export default {
       return this.layers.map(el => el.label).join(',')
     },
 
+    previewLayers () {
+      return [{
+        name: `preview-layers-${this.layersInputValue}`, // Force component recreation on layer change (DpOlMapLayer)
```
hwiem, line 297: When a new layer is added, all selected layers are removed and added again. Is there a way of not removing the selected layers? (Not a blocker, can also be looked into separately - but I think it could happen that a user selects a lot of layers, and maybe there will be performance issues)

hwiem (changes requested):
Nice ✨ Just 2 very small change requests, the performance issue can also be addressed separately.
In the user story it also says that errors should be handled, I guess that will happen in another PR?

client/js/components/map/admin/LayerSettings.vue
meissnerdemos: I am not sure. There is another way to do it, but I don't know if that will improve perfomance. Currently, since the layers are all coming from the same WMS service, they are all the same instance of DpOlMapLayer. Adding or removing a layer just modifies the request that is made to the service. I could make it so that each layer has its own DpOlMapLayer and sends its own request that is just asking for one layer. Then, if a layer is unselected, only that DpOlMapLayer is removed. I think it could be faster for removing a layer completely. However since every time you zoom or change the view in the map somehow, a request is sent to the service, and that would then be many separate requests for each layer, I think it would be a worse performance for that use case.
I currently can't think of another idea how to do it, since changing a layer always means changing the request to the service, but maybe you have one?

client/js/components/map/map/DpOlMapLayer.vue
meissnerdemos: Otherwise the layers stay visible in the OpenLayers map when the component is removed.

client/js/components/map/admin/LayerSettings.vue
hwiem: As discussed, I think we can leave it like this for now - in this view, users will probably not add a lot of layers, and we can always address it in a bug fix if necessary.

client/js/components/map/admin/LayerSettings.vue
```diff
@@ -142,11 +149,13 @@
 <script>
 import { debounce, DpCheckbox, DpInput, DpLabel, DpMultiselect, DpSelect, externalApi } from '@demos-europe/demosplan-ui'
 import { WMSCapabilities, WMTSCapabilities } from 'ol/format'
+import DpOlMap from '../map/DpOlMap.vue'
 
 export default {
   name: 'LayerSettings',
 
   components: {
+    DpOlMap,
```
hwiem, line 158: Since we don't always need DpOlMap here, we could use a dynamic imort instead:

```
DpOlMap: defineAsyncComponent(() => import('../map/DpOlMap')),
```

and the above import can be removed. That way DpOlMap is only imported if we actually need it.

meissnerdemos: Good idea! Did not know about this option before

hwiem (approved):
Looks good! ✨

sierrazamudiodemos (approved):
Thx

PR #4963, opened 2025-07-22
fix(refs DPLAN-16160): Add padding left to BTN in side menu

rafelddemos:
### Ticket
[DPLAN-16160](https://demoseurope.youtrack.cloud/issue/DPLAN-16160/Mitzeichnende-Toggle-ist-verrutscht-und-wenn-man-darauf-klickt-wird-das-STN-Formular-submitted)

This PR is not striclty related to the bug ticket, but deals with a side effect of changing the position of the caret icon (now after the text). This placed the text too far left, when the component is used in the side menu. 
It adds the necessary padding. It also disables border-bottom in the side menu coming from dp-accordion (to avoid double border).


hwiem (approved):
Thanks, lgtm. While you're at it, maybe you could also change the background-color of the button to white? (That's how it was before)

hwiem (approved):
Lgtm, thank you!

PR #4973, opened 2025-07-25
fix: increase search field width

hwiem:
Somehow the originally intended search field width in the procedure list got lost, it is hereby restored. :sparkles: 

Also, there was a purple notification because of the 'procedure' include in the AdministrationProcedure fetch request, so I removed it.


riechedemos (approved):
Thank you, lgtm!

PR #4989, opened 2025-07-29
Fix DPLAN-16144 Fix PHP 7.4+ typed property initialization issues

bakridemos:
Ticket

Description:  
PHP 7.4 introduced typed properties, and with them came the requirement that typed properties must be initialized before being accessed.
Initialize DateTime properties with lazy initialization in getters using isset() checks


mussbach (commented):
Not sure, could this also be fixed with adding 
$this->creationDate = new DateTime();
$this->modificationDate = new DateTime();
to the constructor?

mussbach (approved):

PR #5018, opened 2025-08-07
B dplan 16308 fix polygon drawing

meissnerdemos:
### Ticket
[DPLAN-16308](https://demoseurope.youtrack.cloud/issue/DPLAN-16308)

The issue was caused by a layer ID mismatch in `setLayerSource` - layer names have dashes removed but lookup compared against original IDs with dashes, causing lookups to fail and pass `undefined` to `createLayerSource`. Added proper ID matching and null checks to prevent the JavaScript error that interrupted the drawing process. Also adds missing layer fields that are required by source creation methods.

### How to review/test
  1. Navigate to a public procedure detail page with map
  2. Draw a polygon using the polygon tool
  3. Verify the polygon remains visible after completion and no console errors occur
  4. Test that the polygon can be saved with the statement form

### DIFF
--- a/.env
+++ b/.env
@@ -75,6 +75,8 @@ JWT_PASSPHRASE=0699e1b59a14e6fec876c06cb8279190
 ###> php-amqplib/rabbitmq-bundle ###
 # You need to define RABBITMQ_DSN in your .env.local
 RABBITMQ_DSN=amqp://:@localhost:5672/%2f
+RABBITMQ_CONNECTION_TIMEOUT=30
+RABBITMQ_READ_WRITE_TIMEOUT=60
 ###< php-amqplib/rabbitmq-bundle ###
 
 ###> sentry/sentry-symfony ###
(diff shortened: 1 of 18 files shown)


client/js/components/map/publicdetail/Map.vue
```diff
@@ -1871,9 +1871,11 @@ export default {
...
     setLayerSource (layer) {
       if (layer.getSource() === null) {
-        const layerObj = this.layers.find(el => el.id === layer.get('name'))
-        const source = this.createLayerSource(layerObj)
-        layer.setSource(source)
+        const layerObj = this.layers.find(el => el.id.replaceAll('-', '') === layer.get('name'))
+        if (layerObj) {
+          const source = this.createLayerSource(layerObj)
+          layer.setSource(source)
+        }
```
hwiem, line 1878: nitpick
```suggestion
        const layerObj = this.layers.find(el => el.id.replaceAll('-', '') === layer.get('name'))
        
        if (layerObj) {
          const source = this.createLayerSource(layerObj)
          
          layer.setSource(source)
        }
```

hwiem (approved):
Code lgtm, thank you! Just a tiny nitpick

PR #5038, opened 2025-08-12
feat(refs dplan-14969, #AB23560): Change timer format

rafelddemos:
This PR is related to [DPLAN-14969](https://demoseurope.youtrack.cloud/issue/DPLAN-14969/ADO-Issue-23560-Rechtzeitige-Ankundigung-Ausloggen-damit-gespeichert-werden-kann)

1. It changes the formatting of the timer from **01:59:45** to **01 h 59 min** (not rendering seconds anymore).
2. The text turns red 3 (and not 10) min. before logout.
3. Seconds are shown only when the timer hits 3 min.
4. Since only min. are rendered, warning MSGs now appear 11 and 4 minutes before logout (when **10 min** and **3 min** are rendered).

### How to review/test

Log in as admin, the timer should appear int he right upper corner, warning notifications in the right bottom one.

### Linked PRs (optional)

Linked PR: https://github.com/demos-europe/demosplan-core/pull/4970

### DIFF
--- a/client/js/components/shared/SessionTimer.vue
+++ b/client/js/components/shared/SessionTimer.vue
@@ -59,7 +59,7 @@ export default {
     },
 
     isWarning () {
-      return this.timeLeft <= 10 * millisecondsPerMinute
+      return this.timeLeft <= 3 * millisecondsPerMinute
     },
 
     shouldShowTimer () {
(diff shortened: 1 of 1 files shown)


client/js/components/shared/SessionTimer.vue
```diff
@@ -92,9 +92,13 @@ export default {
     formatTimeString ({ hours, minutes, seconds }) {
       const pad = (num) => String(num).padStart(2, '0')
 
-      return hours > 0
-        ? `${ hours }:${ pad(minutes) }:${ pad(seconds) }`
-        : `${ pad(minutes) }:${ pad(seconds) }`
+      if (hours > 0) {
```
hwiem, line 95: I have thought a bit about how we could make this look a little nicer because I think we should display german translations for min. and sec. Here's what I came up with:
hrs_mins
and
hrs_mins_secs
or
mins_secs
(the numbers have font-size 16px and the text should get some light gray color)
I liked the previous implementation better where we were showing hours, minutes and seconds, somehow it looks weird if we only show the remaining minutes and nothing else? What do you think?

hwiem (commented):
Code looks good, I'm just wondering if we could improve the design slightly

PR #5045, opened 2025-08-12
fix(refs dplan-16328): Make the buttons disappear after saving the changes

rafelddemos:
### Ticket
[DPLAN-16328](https://demoseurope.youtrack.cloud/issue/DPLAN-16328)

After editing a statement's text and saving the changes, the edit mode should close. This was happening only after clicking 'save' twice. The reason was that the callback FN, setting isEditing to false, was being returned, but not called in the update FN.

### How to review/test

Log in as admin --> go to Abwägungstabelle --> edit the text in a statement --> the editing mode should close the first time you click save.

### PR Checklist


huellnerdemos (commented):
I tested it locally and i could save a change on a statement, i also had to click the button just once, but to see the change i had to reload the page.

rafelddemos:
oh, are you sure, the change is not hidden behind 'mehr anzeigen'? or maybe cache clear?

PR #5047, opened 2025-08-13
fix(refs: DPLAN-16161): change internal note textarea to dp-text-area

huellnerdemos:
### Ticket
[DPLAN-16161](https://demoseurope.youtrack.cloud/issue/DPLAN-16161/Interne-Notiz-bleibt-ausgegraut-obwohl-es-editierbar-ist)

The tailwind preflight.css sets the background-color of textareas to transparent. In the "Grundeinstellungen" of a procedure there is an internal note textarea and the background of the parent element is gray. Because of the transparent background of the textarea und the dark gray text color it looks as if the textarea is disabled. I changed the textarea to dp-text-area and also changed the background-color of it to tailwind class bg-surface (this is also used for the other elements in this dialog, it is white)

### How to review/test

1. Login as admin and choose a procedure
2. Change to the "Grundeinstellungen" and open "Intern"
3. Check if the internal note text area has a white background 

### PR Checklist

### DIFF
--- a/client/js/bundles/procedure/administrationEdit.js
+++ b/client/js/bundles/procedure/administrationEdit.js
@@ -11,7 +11,7 @@
  * This is the entry point for administration_edit.html.twig
  */
 
-import { DpCheckbox, DpDateRangePicker, dpValidate } from '@demos-europe/demosplan-ui'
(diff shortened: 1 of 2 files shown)


huellnerdemos:
> Love it, thank you! But please change the target branch to main 🙏

Thank you, for the review :)

PR #5058, opened 2025-08-15
fix(refs: DPLAN-16383): prevent duplicated custom field and option names

huellnerdemos:
### Ticket
DPLAN-16383

It was possible to create custom fields with the same name, or to change the name of a field to one that already existed. It was also possible to create an option, name it like an existing option, then create another option and give it a unique name and then save all this. In edit mode you got an error form the BE and when creating a new field the duplicated option names where saved.
This PR solves the duplicated name issues for both field and option name.  

### How to review/test

1. Login as admin and open a procedure
2. Change to "Konfigurierbare Abschnittsfelder" and create a new field
3. Try to create another field with the same name
4. Check if there comes an error message popup and the duplicated field name can't be saved
5. When creating a new field, try to add two options with the same name followed by an option with a unique name
6. Check if there comes an error message popup and the duplicated options can't be saved
7. Check the same points for editing a field by changing the name to an existing name and add or rename options
8. Check if all error messages make sense for the specific case

### PR Checklist

### DIFF
--- a/client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
+++ b/client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
@@ -384,7 +384,7 @@ export default {
     checkIfNameIsUnique (name) {
       const identicalNames = Object.values(this.customFields).filter(field => field.attributes.name === name)
 
-      return identicalNames.length <= 1
+      return identicalNames.length <= 0
     },
 
     /**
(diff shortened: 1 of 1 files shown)


client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -384,7 +384,7 @@ export default {
     checkIfNameIsUnique (name) {
       const identicalNames = Object.values(this.customFields).filter(field => field.attributes.name === name)
 
-      return identicalNames.length <= 1
+      return identicalNames.length <= 0
```
meissnerdemos, line 387: Why `<=`? It should be impossible for it to be smaller than zero.

meissnerdemos (changes requested):
It works, I just have a tiny issue with the logic.

PR #5064, opened 2025-08-18
f DPLAN-16340 remove join on strings

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-16340/Organisationliste-ladt-nicht

Description: 
orgaFields is defined as an object where each property is already a joined string (e.g., Customer: orgaFieldsArrays.Customer.join()). But a join() is called again on orgaFields.Customer, which it's already a string, not an array. Same with the other calls


elmasdemos (approved):
lgtm, code review

PR #5068, opened 2025-08-19
feat(refs dplan-15846, #AB30421): mark outdated map layers

rafelddemos:
### Ticket
[DPLAN-15846](https://demoseurope.youtrack.cloud/issue/DPLAN-15846/ADO-30421-Veraltete-Layer-automatisch-erkennen-und-Nutzer-informieren)

This PR adds a check for the previously saved map layers. 
1. When the service options are fetched with getCapabilitiesRequest, the previously saved selection is compared with the available options, and the previously saved layers, that are now outdated/unavailable are marked with a warning icon (with additional info on hover).
2. Similarly, if the service request returns an error (no options), a corresponding MSG is shown on hover.
3. To mark outdated layer-tags with the warning icon, a customized tag-slot is used.

*** NOTES***
- layers && layersOptions don't have IDs --> hence comparison of labels
- from what I found, you need to recreate the whole subSlot markup if you customise it (so also the i-tag for remove-button)
- tag-slot seems the right one from the subSlots options 

### How to review/test

go to 'Planungsdokumente und Planzeichnung' of a procedure --> 'Kartenebenen definieren' --> in edit-mode, tags for the previously saved layers that are oudated should have a warning icon (and info on hover)

if the map service throws an error --> the MSG on hover next to the saved layers should refer to the service not responding
### PR Checklist


translations/messages+intl-icu.de.yml
```diff
@@ -2096,7 +2096,9 @@ last.used: "Zuletzt genutzt:"
 layer.land-useplan.public: "Planzeichnung"
 layer.land-useplan.short: "B-Plan"
 layer.land-useplan: "Bebauungsplan"
+layer.no.longer.included.in.map.service: "Der zuvor gespeicherte Layer ist im aktuellen Kartendienst nicht mehr enthalten. Möglicherweise wurde er entfernt oder umbenannt."
```
hwiem, line 2099: Maybe we could shorten the key to `layer.unavailable_in_service`, or just `layer.unavailable`? The latter is less specific, but for now it doesn't seem to collide with any other use case

client/js/components/map/admin/LayerSettings.vue
```diff
@@ -86,7 +86,30 @@
       @input="filterMatrixSetByLayers"
       @selectAll="selectAllLayers"
       @deselectAll="deselectAllLayers"
-    />
+    >
+
+      <template #tag="{ props }">
+        <span class="multiselect__tag">
+          {{ props.option.label }}
+          <dp-contextual-help
```
hwiem, line 94: I'm not sure: can this be accessed via keyboard?

hwiem (commented):
Just some small remarks

client/js/components/map/admin/LayerSettings.vue
```diff
@@ -86,7 +86,30 @@
       @input="filterMatrixSetByLayers"
       @selectAll="selectAllLayers"
       @deselectAll="deselectAllLayers"
-    />
+    >
+
+      <template #tag="{ props }">
```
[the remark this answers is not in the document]
rafelddemos: good to know, thx!
changed in 6ebb2ced09b09d991151ceb324ff8943c88c4a68

client/js/components/map/admin/LayerSettings.vue
hwiem: Which warnings?
I think we should always aim to make things accessible, also in the admin area. I didn't test it, so I wasn't sure if the help text could also be toggled via keyboard (I think it should be possible, but I don't consider this a blocker).

hwiem (approved):
Lgtm! Really nice feature :sparkles:

PR #5075, opened 2025-08-20
fix(refs DPLAN-16399): initialize custom field values on stmt segment

huellnerdemos:
### Ticket
[DPLAN-16399](https://demoseurope.youtrack.cloud/issue/DPLAN-16399/Custom-Fields-wenn-der-Option-geandert-ist-wird-es-nicht-richtig-im-Abschnitt-Erwiderungsansicht-ubergenommen)

When editing a statment segment form the statement segement list it is possible to choose values for custom field for the segment. After saving the values they where not preselected when reloading the page. This PR initializes the custom field multiselects with the last saved values.

### How to review/test

1. Login as admin and open a procedure with custom fields and with a statement, that have segments
2. Open the segment list and edit (click on the three little dots ... next to the segment) a segment
3. Click on "Erwiederung" and then check the checkbox "Weitere Felder bearbeiten"
4. Choose a value for a custom field and save
5. Leave the dialog and visit it again
6. Check if the saved value is selected in the multiselect field of the custom field  

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
@@ -737,6 +737,18 @@ export default {
       dplan.notify.notify('confirm', Translator.trans('recommendation.pasted'))
     },
 
+    getCurrentSelectedOption (fieldId) {
+      const selectedOption = this.customFields[fieldId].attributes.options?.find(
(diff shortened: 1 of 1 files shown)


rafelddemos (approved):
lgtm, thanks :)

huellnerdemos:
> lgtm, thanks :)

Thank you for your review :)

PR #5076, opened 2025-08-20
 feat(refs DPLAN-15854): update diplan-karten to 1.2.0 

huellnerdemos:
### Ticket
DPLAN-15854

After update diplan-karten to version 1.2.0 no toolbar or searchbar where visible on the map anymore. This PR enables them again.  

### How to review/test

1. Open a procedure as a guest or private user
2. Check if the map is shown and there is a searchbar and toolbar on the map


meissnerdemos (approved):
It works, thank you!

PR #5079, opened 2025-08-20
test: add missing mutation mock

hwiem:
`yarn test` failed because the `updateState` mutation was missing in `AdminLayerList.spec.js`.

### How to review / test

Run `yarn test`, there should be no errors. :rocket:


rafelddemos (approved):
lgtm, thank you :)

PR #5080, opened 2025-08-20
style: fix linting errors

hwiem:
- to fix the eslint error: 

> "Mocks should not be manually imported from a __mocks__ directory. Instead use
> `jest.mock` and import from the original module path",

some mock files that were actually just fixtures were moved to the `fixtures` directory.
- the `sessionStorage` mock could be removed as it is already implemented in jsdom and works without extra mocking

### How to review/test

- When running `yarn lint .`, the result should be
`✖ 298 problems (26 errors, 272 warnings)`
instead of
`✖ 303 problems (31 errors, 272 warnings)`
- When running `yarn test`, the tests should all pass

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`


meissnerdemos (commented):
I am not sure what is happening, but on my machine, when I run `yarn lint`, I get:
`317 problems (45 errors, 272 warnings)`
and when I run `yarn test`, I get:
`Test Suites: 4 failed, 24 passed, 28 total`
`Tests: 38 failed. 13 skipped, 121 passed, 172 total`
I executed both commands inside the container.

Some of the errors for the failing tests seem to be referring to missing files, like this one:
` Cannot find module './fixtures/DpAutofillSubmitterData.json' from 'tests/frontend/unit/DpAutofillSubmitterData.spec.js'`
I looked inside the fixtures folder - there are only 2 files, layer_json.js and layers_api_response.js
Maybe this helps find the issue?

hwiem:
> I just looked at the linting errors again and the same test file that is failing is throwing 19 linting errors, which accounts for the difference. So probably fixing that error will fix the additional linting errors?

Thank you for checking :pray: Is that a file you have only locally? I can't find it

meissnerdemos (approved):
> Thank you for checking 🙏 Is that a file you have only locally? I can't find it

Yep, it seems to be an unversioned file. No idea where that is from :see_no_evil: 
Now everything works as it should. Thank you!

PR #5081, opened 2025-08-20
style: always return value from Array.map()

hwiem:
Either return `undefined` or, if no value is returned anyway, use `.forEach()` instead.
This fixes 2 eslint errors.

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`


huellnerdemos (approved):
Thank you for this change, it makes sense to use forEach instead of map in the init function, that's very good.

PR #5094, opened 2025-08-21
fix: replace double quotes with single quotes in migration SQL

breierDemos:
Replace double-quoted string literals with single-quoted ones in Version20250723134332
to ensure proper SQL syntax regardless of ANSI_QUOTES mode settings.

Branch is already tested on develop (migration was running)


MoritzMandler (approved):
lgtm

PR #5102, opened 2025-08-22
chore: use flat config for eslint

hwiem:
since esLint 9 they restructured the config file to a "flat" way.
And eslintrc.js is replaced by eslint.config.js

The linter itself runs, but when building the fe-assets, some error occurs.

Since the webpack-hook didn't help us to reduce linting issues we decided to move the linter checks to GH and remove it from webpack. there the new format breaks the build for some reason


rafelddemos (approved):
yeah, thank you so much for this PR! :heart:
high time the trailing comma arrived :)

PR #5103, opened 2025-08-22
fix(refs dplan-16226): Add two-way data binding to email input in DpNewProcedure

rafelddemos:
### Ticket
[DPLAN-16226](https://demoseurope.youtrack.cloud/issue/DPLAN-16226/E-Mail-Verfahrenstrager-wird-geloscht-wenn-ich-Verfahrenstyp-oder-Blaupause-wahle)

This PR adds v-model to the email address DpInput in DpNewProcedure, so that the email input doesn't get reset on changing other values in the form.

### How to review/test

create a new procedure --> type in an email address:
1. change 'Verfahrenstyp' --> the typed in email should persist
2. change 'Blaupause --> if the saved Blaupause email address is not an empty string, the typed in address will be overwritten by the one from the Blaupause

### PR Checklist

- [x] Run `yarn lint`


huellnerdemos (approved):
Thank you!
Code looks good to me and worked fine in the test.

PR #5105, opened 2025-08-22
fix: use buttons for interactive elements

hwiem:
- divs should not be used as buttons (and the eslint vuejs-accessibility plugin throws an error), so they were replaced with actual buttons
- fixes the linting error: "visible, non-interactive elements should not have an interactive handler"

### How to review/test

- go the admin layer list view and make sure that when clicking a layer in the list, it is 'activated' and can be grouped with another layer in the list (ungrouping seems to be broken, but that was the case before these fixes)
- the list should look the same as before

### Linked PRs
- [ ] https://github.com/demos-europe/demosplan-core/pull/5102
- [ ] https://github.com/demos-europe/demosplan-core/pull/5104

### PR Checklist

- [x] Run `yarn lint` (`✖ 473 problems (41 errors, 432 warnings)`)
- [x] Run `yarn test`


rafelddemos (approved):
lgtm, thank you:) 
The list looks the same as before.
I didn't really manage to group the layers (but it also didn't work on main for me, so it can have sth to do with my handling of the link-icon..)

** EDIT**
I checked procedures with already grouped layers before --> added new ones now and the grouping works too :)

PR #5124, opened 2025-08-25
 f_(refs DPLAN-16179): implementing geocoder as auto suggestion addre…

kijewskidemos:
### Ticket
[DPLAN-16179](https://demoseurope.youtrack.cloud/issue/DPLAN-16179) ADO ISSUE: 33675 Adresssuche in der Karte ermöglichen - Anbindung Geocoder BKG

### Description:
Enabling Geocoder as an external API service for address search autosuggestions. It is linked to a permissions check in order to decide if the newly implemented geocoder should be used or the already existing db query search. This feature should primarily used by citizens to search for procedures in their area. 

### How to review/test
Click on any of the Verfahren, and use the address search bar in the map. This should work for logged in citizens as well as anonymous users. 

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Maps/GeodatenzentrumAddressSearchService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Maps/GeodatenzentrumAddressSearchService.php
@@ -28,7 +28,7 @@ use UnexpectedValueException;
 
 class GeodatenzentrumAddressSearchService implements GeocoderInterface
 {
-    private const GEODATENZENTRUM_ADDRESS_SEARCH = 'https://sg.geodatenzentrum.de/gdz_ortssuche__353cdae2-2c78-1654-c1f0-85192cfa13d6/geosearch?count=5';
+    private const GEODATENZENTRUM_ADDRESS_SEARCH = '';
 
     public function __construct(
         private readonly HttpClientInterface $httpClient,


breierDemos (commented):
For the sonar cloud issue: To reduce the number of returns, you can use a single result variable and return it at the end.

Reduce 6 Return Statements to 1

 ```
 public function searchAddress(string $query, int $limit = 20, ?array $maxExtent = null): array
  {
      $startTime = microtime(true);
      $logContext = [/* ... existing ... */];

      // Existing permission check (keeps throw)
      try {
          $this->currentUser->hasPermission('feature_geocoder_address_search');
      } catch (Exception $permissionException) {
          $this->logger->error(/* ... existing ... */);
          throw $permissionException;
      }

      $formattedResults = [];

      try {
          $requestOptions = [/* ... existing ... */];
          $response = $this->httpClient->request('GET', self::GEODATENZENTRUM_ADDRESS_SEARCH, $requestOptions);
          $statusCode = $response->getStatusCode();

          if (200 !== $statusCode) {
              $this->logger->error(/* ... existing logging ... */);
              // REMOVE: return [];
          }

          if (200 === $statusCode) {  // ADD this condition
              $result = $response->toArray();

              if (!isset($result['features']) || !is_array($result['features'])) {
                  $this->logger->error(/* ... existing logging ... */);
                  // REMOVE: return [];
              }

              // ADD this condition wrapper:
              if (isset($result['features']) && is_array($result['features'])) {
                  foreach ($result['features'] as $feature) {
                      $formatted = $this->formatResult($feature);
                      if (!empty($formatted['name']) || !empty($formatted['city'])) {
                          $formattedResults[] = $formatted;
                      }
                  }
              }
          }
          // REMOVE: return $formattedResults;

      } catch (TransportExceptionInterface|ClientExceptionInterface|ServerExceptionInterface|DecodingExceptionInterface $e) {
          $this->logger->error(/* ... existing ... */);
          // REMOVE: return [];
      } catch (RedirectionExceptionInterface $e) {
          // REMOVE: return [];
      } catch (Throwable $e) {
          $this->logger->error(/* ... existing ... */);
          // REMOVE: return [];
      }

      return $formattedResults; // ADD: Single return
  }
```

  - Remove 6 return [] statements
  - Add if (200 === $statusCode) wrapper
  - Add if (isset($result['features']) && is_array($result['features'])) wrapper
  - Add single return $formattedResults at end
  
  You could also think about to split the method into smaller ones :thinking:

File: demosplan/DemosPlanCoreBundle/Logic/LocationService.php [hunk 1]
```diff
@@ -24,29 +25,155 @@ class LocationService
...
+                'originalBehavior' => 'Database search supported geographic filtering via maxExtent',
+                'currentBehavior'  => 'External API returns results without geographic filtering',
+                'recommendation'   => 'Consider implementing client-side filtering if geographic bounds are critical',
+            ]);
+        }
+
         try {
-            $locations = $this->getLocationRepository()
-                ->searchCity($searchString, $limit, $maxExtent);
+            $locations = $this->geodatenzentrumAddressSearchService->searchAddress($searchString, $limit, $maxExtent);
```
MoritzMandler, line 132: This seem strange.
Is it the city or the address we are getting here.

demosplan/DemosPlanCoreBundle/Controller/Platform/LocationSearchController.php
```diff
@@ -38,7 +39,11 @@ public function searchLocationJsonAction(Request $request, LocationService $loca
             }
 
             $limit = $query['maxResults'] ?? 50;
-            $restResponse = $locationService->searchCity($query['query'], $limit);
+            if ($currentUser->hasPermission('feature_geocoder_address_search')) {
+                $restResponse = $locationService->searchAddress($query['query'], $limit);
+            } else {
+                $restResponse = $locationService->searchCity($query['query'], $limit);
+            }
             $result = $restResponse['body'] ?? [];
```
sierrazamudiodemos, line 47: @kijewskidemos maybe you can move this part to a service, and this service returns the $result -> then you encapsulate that logic into a service, and the services takes care of the if else. Maybe refrain from using the word Service in the class

demosplan/DemosPlanCoreBundle/Logic/LocationService.php
```diff
@@ -24,29 +25,155 @@ class LocationService
...
+    }
+
     /**
      * Get a City by Name or postal code.
+     *  CRITICAL CHANGE: Now uses external Geodatenzentrum API instead of database query.
      *
-     * @param string     $searchString
-     * @param int        $limit
-     * @param array|null $maxExtent
+     * @param string     $searchString Search query for city/location
```
sierrazamudiodemos, line 101: is this a breaking change? does it work with the previous search format? I do not understand the comment here ;)

demosplan/DemosPlanCoreBundle/Logic/Maps/GeodatenzentrumAddressSearchService.php
```diff
@@ -0,0 +1,210 @@
...
+            'method'       => 'searchAddress',
+            'query'        => $query,
+            'limit'        => $limit,
+            'maxExtent'    => $maxExtent,
+            'api_endpoint' => self::GEODATENZENTRUM_ADDRESS_SEARCH,
+        ];
+
+        // Check permissions before making external API call
+        try {
+            $this->currentUser->hasPermission('feature_geocoder_address_search');
```
sierrazamudiodemos, line 60: is this line throiwng an exception? I have the feling that is missing something like Assert::true. I think this line returns tur or false, but does not throw an exception in case the user does not have the permission

demosplan/DemosPlanCoreBundle/Logic/Maps/GeodatenzentrumAddressSearchService.php
```diff
@@ -0,0 +1,210 @@
...
+
+            // Extract address components with safe defaults
+            $street = $properties['strasse'] ?? '';
+            $houseNumber = $properties['haus'] ?? '';
+            $postcode = $properties['plz'] ?? '';
+            $city = $properties['ort'] ?? '';
+            $federalState = $properties['bundesland'] ?? '';
+            $longitude = $coordinates[0] ?? null;
+            $latitude = $coordinates[1] ?? null;
+            $municipalCode = $properties['ags'] ?? null;
```
sierrazamudiodemos, line 167: maybe these are good candidates for constants, because they are a bit in german..

demosplan/DemosPlanCoreBundle/Logic/Maps/GeodatenzentrumAddressSearchService.php
```diff
@@ -0,0 +1,210 @@
...
+use Symfony\Contracts\HttpClient\Exception\RedirectionExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ServerExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Throwable;
+
+class GeodatenzentrumAddressSearchService
+{
+    private const GEODATENZENTRUM_ADDRESS_SEARCH = 'https://sg.geodatenzentrum.de/gdz_ortssuche__353cdae2-2c78-1654-c1f0-85192cfa13d6/geosearch?count=5';
+
```
sierrazamudiodemos, line 29: this is super good that is move to a constant, maybe for the future not for now: we can move it to the parameters file (yml) because maybe this might be different per project, unless this is a key gdz_ortssuche__353cdae2-2c78-1654-c1f0-85192cfa13d6 and maybe we should move it to a config file

demosplan/DemosPlanCoreBundle/Controller/Platform/LocationSearchController.php
kijewskidemos: Yes good point! I should restructure the GeodatenzentrumAddressSearchService to follow the SOLID principle by creating different services.

demosplan/DemosPlanCoreBundle/Controller/Platform/LocationSearchController.php
```diff
@@ -47,10 +52,19 @@ public function searchLocationJsonAction(Request $request, LocationService $loca
...
+                            'value' => $entry['name'].' '.$entry['housenumber'].' '.$entry['postcode'].' '.$entry['city'],
+                            'data'  => $entry,
+                        ];
+                    } else {
+                        // searchCity DB suggestions
+                        $suggestions[] = [
+                            'value' => ($entry['name'] ?? '').' '.($entry['postcode'] ?? ''),
+                            'data'  => $entry,
+                        ];
+                    }
```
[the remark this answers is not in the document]
kijewskidemos: Thanks, yes I agree!

demosplan/DemosPlanCoreBundle/Logic/LocationService.php
kijewskidemos: Thanks, will be more specific with the comment!

demosplan/DemosPlanCoreBundle/Logic/Maps/GeodatenzentrumAddressSearchService.php
mussbach: It is definitely for now to move that into a param or env var!

mussbach (changes requested):
Basically good, one remark and tests are missing

PR #5127, opened 2025-08-26
fix(refs DPLAN-16049): change height of metadata menu container

huellnerdemos:
### Ticket
[DPLAN-16049](https://demoseurope.youtrack.cloud/issue/DPLAN-16049/Overflow-Fehler-Seitenleisten-Element-verlasst-Containergrenzen)

When answering a statement and switching to edit mode, the last point of the metadata menu overflows the metadata box, when scrolling to the bottom. This depends on how much data follows after this box or how large the screen is. This PR extends the height of the menu container, to prevent items from the menu from overflowing the border of the metadata box. 

### How to review/test

1. Login as admin and open a procedure with statements
2. Go to segments and edit a segment
3. Switch to edit mode and check if the menu of the metadata box does not overflow the bottom border, even if there is a lot of data following or switching to the mobile view.

### PR Checklist

- [x] Run `yarn lint`

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMeta.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementMeta.vue
@@ -21,7 +21,7 @@
     </div>
 
     <div class="flex">
-      <div class="sticky top-[70px] mt-2 basis-1/4 max-h-11">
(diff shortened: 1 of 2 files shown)


huellnerdemos:
> that seems to do the trick, thank you :) the FE build fails though

Thank you for your review

PR #5128, opened 2025-08-26
f DPLAN-15662 refactor: use updateProcedure instead of updateProcedur…

bakridemos:
### Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15662/ROBOB-Ubernahme-der-automatischen-Umstellung-in-Auswertung-bei-Fristende-wie-Regio-BLP-wird-nicht-ins-Protokoll-ubernommen

Description: refactor: use updateProcedure instead of updateProcedureObject: 
Replace direct entity manipulation with service method calls in ProcedureHandler for both internal and external procedure phase updates. The updateProcedure method generates report entries for phase changes initiated by MaintenanceController. This will ensure that automatic phase changes from the maintenance system are properly logged/tracked through report entries.


MoritzMandler (approved):
But why does it not work for the updateObject method... fixing that would be nice :)

PR #5143, opened 2025-08-27
feat: modify webpack warning behaviour

graupnerdemos:
Sass Loader warnings are now never emitted, but Vue Template Compiler warnings and all others that might appear will be printed on every build.

**Testplan**

Run the webpack build for any project, Vue Template Compiler warnings should be displayed.


hwiem (approved):
Nice, thank you!

PR #5173, opened 2025-09-02
fix: reset padding only for layout__item

hwiem:
After the tailwind 4 migration, the specificity order of styles has changed, and `c-proceduresearch__heading` was also overriding the padding for `o-page__padded`, which it was not intended to do. I restored the previous styling so the heading on the index page is aligned with the content again.


meissnerdemos (approved):
Looks much better, thank you!

PR #5178, opened 2025-09-03
fix: use ProcedureInterface

breierDemos:
use ProcedureInterface type hint in ProcedureService::updateProcedureObject()

Changes method signature from Procedure to ProcedureInterface parameter to match the interface contract, resolving compile error about incompatible declarations.


winderdemos (approved):
looks good

PR #5213, opened 2025-09-08
Fix SonarCloud quality gate violations

mussbach:
After merge of several release branches sonarcloud complains

## Summary
- Fix critical cognitive complexity issue in DemosPlanProcedureController by refactoring boilerplateListAction method
- Remove duplicate property definitions in StatementModal.vue
- Fix useless variable assignments in ProcedureHandler.php

## Test plan
- SonarCloud quality gate should pass
- Existing functionality should remain unchanged
- No breaking changes to public APIs

Resolves 5 SonarCloud violations that were causing the quality gate to fail.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
@@ -129,7 +129,7 @@ class DemosPlanProcedureController extends BaseController
 {
     private const NONE = 'none';
     private const AGGREGATION_STATUS_PRIORITY = 'status_priority';
-    
+
     private const PARAM_BOILERPLATE_DELETE_CHECKED = 'boilerplateDeleteChecked';
     private const PARAM_BOILERPLATE_DELETE = 'boilerplate_delete';
     private const PARAM_BOILERPLATE_GROUP_IDS_TO_DELETE = 'boilerplateGroupIdsTo_delete';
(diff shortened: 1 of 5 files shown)


demosplan/DemosPlanCoreBundle/Controller/Procedure/DemosPlanProcedureController.php
```diff
@@ -2367,78 +2368,93 @@ public function boilerplateListAction(
...
+            return;
         }
 
-        $templateVars = [];
-        $templateVars['list'] = $procedureService->getBoilerplateList($procedure);
-        $templateVars['boilerplateGroups'] = $procedureService->getBoilerplateGroups($procedureId);
-
-        if (!$request->isMethod('GET')) {
-            // prevent sending the same post multiple times when browser reloads the same page (F5)
-            // therefore redirect to self as a get call instead.
```
sierrazamudiodemos, line 2430: Thanks for moving this to the beginning! do you think it makes sense to add the comment to the refactored code? acually is this a case that happens in every controller?

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureHandler.php
```diff
@@ -905,7 +903,7 @@ public function switchToEvaluationPhasesOnEndOfParticipationPhase(): Collection
                 && !$endedExternalProcedure->getMaster() && !$endedExternalProcedure->isDeleted()) {
                 $data = [
                     'id'                       => $endedExternalProcedure->getId(),
-                    'publicParticipationPhase' => $internalPhaseKey,
+                    'publicParticipationPhase' => $externalPhaseKey,
```
sierrazamudiodemos, line 906: why would it be changed to externalPhaseKey? maybe I missed something on the code  changes

mussbach: Sorry, I should have mentioned this. This is a bugfix along the way, as it is the publicParticipationPhase the external phase key needs to be used. was a c&p error somewhere before

sierrazamudiodemos:
nice Thanks for the addressing my comments! Should it still be target main? or we wait until we release?

sierrazamudiodemos (approved):
Thanks!

PR #5214, opened 2025-09-08
refactor: Make webpack error handling less annoying

graupnerdemos:
Webpack used to dump the stats object at every opportunity it found. Unfortunately, the stats object is stupidly large and mostly useless for debugging.

This change adds error handling for webpack config which should drastically reduce the config object dump(ster fire) occurrences. Additionally, Some refactoring for readability and reuse were done along the way.

Overall, webpack build stability should be increased, along with debuggability especially of the config.

### How to review/test

- Make the webpack config fail, for instance by adding a non-existent variable to any of the plugin lists in moduleRules.
- Build and be happy that there is no stats object dump.


hwiem (approved):
Thanks for this! :pray: Lgtm

PR #5223, opened 2025-09-09
refactor: remove duplicated props

sakutademos:


hwiem (commented):
This has also been done in https://github.com/demos-europe/demosplan-core/pull/5213

PR #5228, opened 2025-09-09
Dplan-16586 fix automatic logout keycloak

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-16586/fix-automatic-logout-keycloak

### How to review/test
We could test it in prod.
For now, detects the parameter written in yml  `session_lifetime_seconds`
If the parameter is not set, then sets the default to 7200 seconds

### DIFF
--- a/config/parameters_default.yml
+++ b/config/parameters_default.yml
@@ -439,6 +439,9 @@ parameters:
    # uses the BKG address search for autosuggestions
     geodatenzentrum_address_search_url: ''
 
+    # session lifetime in seconds used for automatic Keycloak logout
+    session_lifetime_seconds: ''
+
     #### project specific settings
 
 


config/parameters_default.yml
```diff
@@ -439,6 +439,9 @@ parameters:
    # uses the BKG address search for autosuggestions
     geodatenzentrum_address_search_url: ''
 
+    # session lifetime in seconds used for automatic Keycloak logout
```
mussbach, line 442: to be strict this has nothing directly to do with keycloak, right?

demosplan/DemosPlanCoreBundle/Logic/User/OzgKeycloakLogoutManager.php
```diff
@@ -125,9 +114,8 @@ public function hasValidToken(SessionInterface $session): bool
         return $isValid;
     }
 
-    public function storeTokenAndExpirationInSession(SessionInterface $session, int $expirationTimestamp, array $tokenValues): void
+    public function storeTokenAndExpirationInSession(SessionInterface $session, array $tokenValues): void
     {
-        $session->set(self::EXPIRATION_TIMESTAMP, $expirationTimestamp);
```
mussbach, line 130: Is this correct to delete this? Where is the token stored in the session in prod mode?

mussbach (changes requested):
Not sure, but it looks as if the token is not stored in prod mode atm

demosplan/DemosPlanCoreBundle/Logic/User/OzgKeycloakLogoutManager.php
sierrazamudiodemos: maybe to explain it a bit: we were already injecting a custom life time for QA to test, what I did it was to use that custom injection to actually use it and store it always (not only for testing)

PR #5246, opened 2025-09-16
deps: Update caniuse db

graupnerdemos:
Out caniuse db was thoroughly outdated which lead to builds complaining and could also lead to very inconvenient and hard to track display differences between browsers.

We should consider adding caniuse as a first class dependency so that there are handled by Dependabot again as they used to. Maybe however, updating it now will do so anyway.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12,41 +12,20 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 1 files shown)


graupnerdemos:
You're right, we do, I must have skipped over it. Then it's highly likely that us not updating it just lead to dependabot ignoring further updates. Let's just merge and see what happens.

PR #5254, opened 2025-09-17
Make the container build respect project frontend dependencies

graupnerdemos:
I'm sorry I rewrote everything. Adding more flags to the build script without it being getopt-ified seemed a little pointless so I did that. While on that I noticed that we always pass `demosplan-production` therefore that could be factored out of the argument list. And finally, the Dockerfile did the yarn install before the project code was available and was therefore missing the project dependencies.

### DIFF
--- a/docker/demosplan-production/Dockerfile
+++ b/docker/demosplan-production/Dockerfile
@@ -47,14 +47,20 @@ RUN --mount=type=cache,target=/var/composer --mount=type=cache,target=/var/www,u
         composer install -o --prefer-dist --no-progress; \
     else \
         composer install -o --prefer-dist --no-dev --no-progress; \
-    fi && \
-    yarn install --immutable && \
-    yarn cache clean && \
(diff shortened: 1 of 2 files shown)


docker/demosplan-production/Dockerfile
```diff
@@ -47,14 +47,20 @@ RUN --mount=type=cache,target=/var/composer --mount=type=cache,target=/var/www,u
...
+    fi
 
 USER root
 COPY . /srv/www
 
+USER www-data
+# Now do the yarn workspace focus after files are copied
+RUN --mount=type=cache,target=/var/www,uid=33,gid=33 \
+    yarn workspaces list && \
+    yarn workspaces focus demosplan-core "$PROJECT_NAME" && \
```
mussbach, line 59: shouldnt --production be added in prod mode?

mussbach (changes requested):

PR #5266, opened 2025-09-23
feat: optimize Statement List API performance

mussbach:
## Summary
* Add segmentsCount attribute to get count without loading all segments
* Remove segments relationship from Statement list to reduce response size
* Limit File fields to metadata only (hash, filename, mimetype, size)
* Add textPreview attribute for efficient text truncation

## Test plan
- [ ] Verify Statement list loads significantly faster with large segment counts
- [ ] Confirm segments count still displays correctly in the interface
- [ ] Test that file metadata is still accessible for attachments
- [ ] Validate text preview shows correctly truncated content

## Performance Impact
This reduces response size from 24MB to under 1MB for 10 statements with large segment counts, significantly improving list performance.

### DIFF
--- a/client/js/components/statement/listStatements/ListStatements.vue
+++ b/client/js/components/statement/listStatements/ListStatements.vue
@@ -756,7 +756,6 @@ export default {
         fields: {
           Statement: statementFields.join(),
           SourceStatementAttachment: [
-            'attachmentType',
             'file'
           ].join(),
           File: [


client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -718,10 +718,10 @@ export default {
         'submitterEmailAddress',
         'text',
         'textIsTruncated',
+        'segmentsCount',
```
hwiem, line 721: nitpick: sort alphabetically

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -487,7 +487,7 @@ export default {
     items () {
       return Object.values(this.statementsObject)
         .map(statement => {
-          const segmentsCount = statement.relationships.segments.data.length
+          const segmentsCount = statement.attributes.segmentsCount
```
hwiem, line 490: nitpick; I guess the default should be 0?
```suggestion
          const { segmentsCount = 0 } = statement.attributes
```

client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -749,15 +749,21 @@ export default {
...
           'sourceAttachment.file'
         ].join(),
         fields: {
           Statement: statementFields.join(),
           SourceStatementAttachment: [
+            'attachmentType',
             'file'
+          ].join(),
+          File: [
+            'hash',
```
hwiem, line 763: Do we really need more than the `hash`?

I also noticed that when I try to open an "Original-PDF", I sometimes get an error ("kein Inhalt gefunden") - but I don't know if it's because of these changes

mussbach: you receive the pdf error in case dplan cannot find the file on your machine. I do not think that this is a regression here. you are correct about the other properties, hash is sufficient [57b6221](https://github.com/demos-europe/demosplan-core/pull/5266/commits/57b6221d31d2d923f79045af7613434b4bf452c8)

client/js/components/statement/listStatements/ListStatements.vue
mussbach: [57b6221](https://github.com/demos-europe/demosplan-core/pull/5266/commits/57b6221d31d2d923f79045af7613434b4bf452c8)

hwiem (approved):
I still have 2 small remarks, but in general it looks good and it seems to me that the statement list loads significantly faster 🎉

demosplan/DemosPlanCoreBundle/ResourceConfigBuilder/StatementResourceConfigBuilder.php
```diff
@@ -61,6 +62,7 @@
  * @property-read AttributeConfigBuilderInterface<ClauseFunctionInterface<bool>, Statement> $userPosition @deprecated Move into separate resource type (maybe StatementSubmitData resource type or something similar)
  * @property-read AttributeConfigBuilderInterface<ClauseFunctionInterface<bool>, Statement> $userState @deprecated Move into separate resource type (maybe StatementSubmitData resource type or something similar)
  * @property-read AttributeConfigBuilderInterface<ClauseFunctionInterface<bool>, Statement> $votesNum @deprecated Use relationship to {@link StatementVote} instead
+ * @property-read AttributeConfigBuilderInterface<ClauseFunctionInterface<bool>, Statement> $textPreview
```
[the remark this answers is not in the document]
mussbach: good catch, ultimately this wasn't used any more [23a9275](https://github.com/demos-europe/demosplan-core/pull/5266/commits/23a9275e3019af0ada177155a8e302bf95d7641d)

PR #5303, opened 2025-10-06
fix(refs DPLAN-16550): adjust autocomplete usages

hwiem:
### Ticket
[DPLAN-16550](https://demoseurope.youtrack.cloud/issue/DPLAN-16550/Abweichungen-vom-Sucherverhalten-von-DiPlanPortal)

- since DpAutocomplete has been refactored to use DpResettableInput under the hood in https://github.com/demos-europe/demosplan-ui/pull/1358, we now need to pass an `id` to it (for the input)
- since DpAutocomplete now has an icon-only search button integrated, we don't need the search button in DpSearchProcedureMap anymore when using autocomplete

### How to review/test

- Go to the public index page and search for a city in the search field
- try to navigate and select a suggestion using the keyboard
- Go to the public detail page and search for a city in the search field in the map
- Go to Grundeinstellungen > Verortung des Verfahrens and search for a city in the search field above the map

### Linked PRs
- [ ] https://github.com/demos-europe/demosplan-ui/pull/1358

### Tasks
- [ ] Merge https://github.com/demos-europe/demosplan-ui/pull/1358
- [ ] Release new demosplan-ui version
- [ ] Bump demosplan-ui version in demosplan
- [ ] Merge this PR

### PR Checklist

- [ ] Run `yarn lint`
- [ ] Run `yarn test`

### DIFF
--- a/client/js/components/map/map/DpOlMap.vue
+++ b/client/js/components/map/map/DpOlMap.vue
@@ -71,6 +71,7 @@
           <div :class="prefixClass('float-right')">
             <dp-autocomplete
               v-if="_options.autoSuggest.enabled"
+              id="ol_map_autosuggest"
               :class="prefixClass('u-mb inline-block w-11 bg-color--white')"
               :options="autoCompleteOptions"
               :route-generator="(searchString) => {
--- a/client/js/components/map/publicdetail/Map.vue
+++ b/client/js/components/map/publicdetail/Map.vue
@@ -11,6 +11,7 @@
   <div aria-hidden="true">
     <dp-autocomplete
       v-if="hasPermission('feature_map_search_location')"
+      id="map_autosuggest"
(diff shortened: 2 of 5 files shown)


huellnerdemos (commented):
Thank you!
Code looks good for me so far and most of the functionality worked well when i tested it, but there was one behavior, that seems to be buggy:
- Open a public detail view of a procedure
- Search for Berlin and choose it from the drop down list with the keyboard
- After the map zoomed to Berlin, delete the search entry with backspace
- Don't leave the search field and type Hamburg as the next search term
- In this case, there was no drop down list when i tested it locally
- But nonetheless, if you wait one or two seconds and press enter the map changes to Hamburg
- If you click somewhere else and enter the search field again, then also the drop down list appears again
 
So the bug is the missing drop down list on the second search try

PR #5304, opened 2025-10-06
refs DPLAN-16737: harden 404 page

mussbach:
## Ticket
DPLAN-16737

## Summary
Remove WYSIWYG filter from 404 error message to harden security by avoiding HTML rendering of the path variable.

## Changes
- Remove `|wysiwyg` filter from error message in 404.html.twig template
- Error message now displays as plain text instead of rendered HTML

## Test plan
- [ ] Verify 404 page displays correctly
- [ ] Confirm error message is shown as plain text
- [ ] Test with various invalid URLs


hwiem (approved):
Lgtm!

PR #5790, opened 2026-01-23
feat (DPLAN-17245): Add initial implementation of tracking custom fie…

sierrazamudiodemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-17245

## What this PR does
Statements now write an entry into the change-history table when their **custom fields** are edited — single-select, multi-select or text. Until now this only worked for segments; statements were missing.

### 1. New event subscriber for statement updates
- `StatementOnUpdateEventSubscriber` listens when a statement is updated through the API and calls `EntityContentChangeService::trackChanges()`, which writes one history row per changed custom field.

- The subscriber checks the `feature_statement_content_changes_save` permission first, so projects that don't enable that permission stay untouched.

### 2. New test
`testEntityContentChangeEntryOnUpdateStatementMultiSelectCustomFieldValue` in `EntityContentChangeServiceTest` mirrors the existing radio-button test but on statements with multi-select custom fields:

- creates 2 statements + 2 multi-select custom fields with 2 options each selected
- asserts that 4 history rows are written
- asserts that each row contains the joined labels (`"Option A, Option B"`) on the right custom field — using the existing
  `getPreUpdateValueOfContentChange` / `getPostUpdateValueOfContentChange` helpers so we keep one JSON-path everywhere

## How to test it

1. Login as FP admin
2. Create custom fields in the Verfahren
3. Create stn
4. Add values on that stn for custom fields
5. Change the values
6. Click on see history of changes and you will see the changes


MoritzMandler (approved):
lgtm - tests work..
I thought about renaming StatementOnUpdateEventSubscriber to something like
OnStatementApiUpdateEventSubscriber for clarity but not a blocker

sierrazamudiodemos:
@MoritzMandler done ca658af4853e82289de5594ef14c37657770cf54 thank you for the feedback!

PR #5907, opened 2026-02-17
refac(BMDS-7161): removes the preselected docx_normal file type from the statement export modal

sakutademos:
### Ticket
https://d3.ce.capgemini.com/jira/browse/BMDS-7161

**Description:** This PR removes the preselected docx_normal file type from the statement export modal. Preselected or default states can lead to confusion or unintended actions by users. By requiring an explicit selection, we ensure clearer user intent and align with our validation and UX guidelines.


hwiem:
Considering this a bit more, I'd discuss with the POs if it's ok to implement this for all the projects. I understand the point that if a radio button is preselected, users may more easily make a mistake, but I think it's also possible that docx is simply the most used export type, and having to select it every time may be annoying. Also, the consequences of having docx preselected and not noticing are not dramatic (but of course it can be annoying, too).

PR #5957, opened 2026-02-27
f_DPLAN-17226: altered docx export to also show tag topic title

kijewskidemos:
### Ticket
[DPLAN-17226](https://demoseurope.youtrack.cloud/issue/DPLAN-17226) BE Synopse: Anzeige von gefilterten Schlagworten im docx Export

### Decription: 
This PR alters the header of a docx export file. The header now includes the metadata information tag [tag topic], by which the export was filtered. If no filter was applied during the export process, the docx file stays untouched. 
This PR also cleans up the StatementExportTagFilter class and removes unused code. 

### How to review/test
Log in as admin, got to Stellungnahmen, click export on the left upper corner, select Docx Export and a few tags you want to have included in the header of the export file.


breierDemos (changes requested):
Pair review notes:
- preamble is on every page, I think it should only in the first page header
- permission 'feature_adjust_export_file_name' we should change the name because the permission does not change the name of the file, it changes only the header of the first page

PR #5966, opened 2026-03-04
feat (refs DPLAN-16766): add phase_definition_id FK migrations and update entities

breierDemos:
### Ticket
DPLAN-16766 and [DPLAN-17570](https://demoseurope.youtrack.cloud/issue/DPLAN-17570/NEU-ADO-44313-Manuelles-Erstellen-von-Custom-Verfahrensschritten)

Builds on the `ProcedurePhaseDefinition` entity (introduced in #5949) by wiring it into the full application stack: adding FK columns to `procedure_phase`, `statement`, and `draft_statement`, and replacing all legacy phase key lookups with phase definition names throughout reports, exports, and the frontend.

**Migrations**
- `Version20260304100000`: adds `phase_definition_id` FK to `procedure_phase`
- `Version20260304130000`: adds `phase_definition_id` FK to `_statement`
- `Version20260312100000`: adds `phase_definition_id` FK to `draft_statement`

**Entities**
- `ProcedurePhase`, `Statement`, and `DraftStatement` each gain a `phaseDefinition` ManyToOne FK relation. On `Statement`, `setPhaseDefinition()` also writes the phase definition name into the existing `$phase` string field (deprecated) to maintain backward compatibility with code that still reads it directly. `DraftStatement` retains its existing `$phase` string field unchanged alongside the new FK.
- `Procedure`: `$closed` property marked deprecated — closed state now derived from the phase definition's `permissionSet`

**Service / API layer**
- New `ProcedurePhaseDefinitionService` for retrieving customer-scoped phase definitions with global fallback
- `ProcedurePhaseDefinitionRepository` extended with customer fallback query
- `ServiceStorage` (procedure save): switched from phase key strings to phase definition IDs; configuration detection now uses `orderInAudience === 0` instead of `key === 'configuration'`
- All report entry factories (`Element`, `Paragraph`, `SingleDocument`, `Statement`), `ReportMessageConverter`: phase name resolution migrated to `getPhaseDefinition()->getName()` — deprecated global config lookups removed
- `ProcedureExtension` Twig `getPhase()` rewritten to read from phase definition objects
- `AdminProcedureResourceType`: exposes `phaseDefinitionName` and `publicParticipationPhaseDefinitionName`
- `StatementResourceType`, `OriginalStatementResourceType`: updated for phase definition relationships
- `GenerateCustomerCommand`: seeds default phase definitions for new customers
- `DraftStatementService`, `DocxExporter`, `StatementArrayConverter`, `PresentableOriginalStatementFactory`, `StatementHandler`, `StatementToLegacyConverter`, `ProcedureToLegacyConverter`: updated to pass/store phase definitions

**Frontend**
- `AdministrationProceduresList.vue`: phase name column reads from definition
- `SharedProcedureMethods.js` (public map index): phase names read from definition entities
- `administration_edit.html.twig`, `administration_edit_master.html.twig`: phase dropdowns use definition-based names
- LaTeX export templates: phase name from definition
- `assessment_table_new_statement.html.twig`, `assessment_statement_detail_statement_data.html.twig`: updated

**Tests & fixtures**
- New `LoadProcedurePhaseDefinitionData` fixture, `ProcedurePhaseDefinitionFactory`, updated `ProcedureFactory`, `StatementFactory`, `LoadStatementData`, `LoadProcedureData`, `LoadSegmentData`
- 13 new German translation keys

### How to review/test
Check migration status and run migrations.
**List of affected places in the application is included in the ticket description (`Betroffene/angepasste Stellen`)**

#### Backend
- Run all 3 migrations on a database with existing data; verify `phase_definition_id` is populated on `procedure_phase`, `_statement`, and `draft_statement` rows
- Export a statement as Docx/PDF — phase name should come from the phase definition, not a raw key
- Check statement reports display the correct phase name

#### Frontend
- Log in as FA admin → procedure basic settings: phase dropdowns show definition-based names; saving works correctly
- Admin procedure list: phase name column reflects definition names
- As guest/citizen: public procedure list and detail page show Mandantenadmin-defined phase names

- (**comprehensive list in the ticket description**)

### Linked PRs
- Based on entity, resource type & initial migration PR: #5949

### PR Checklist
- [ ] Create/Update tests
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureHandler.php
```diff
@@ -867,15 +860,7 @@ public function switchToEvaluationPhasesOnEndOfParticipationPhase(): Collection
         $systemUserName = $this->translator->trans('user.system.name');
 
         // internal:
-        $internalWritePhaseKeys = $this->getDemosplanConfig()->getInternalPhaseKeys('write');
-        $endedInternalProcedures = $this->procedureService->getProceduresWithEndedParticipation($internalWritePhaseKeys);
```
sierrazamudiodemos, line 871: there used to be a fallback in the old code: ` $internalPhaseKey = 'evaluating';`
now there is a new method that returns an array, do we still want to handle a fallback? is there a case where not phases are found?

sierrazamudiodemos: - ok so it seems that `GenerateCustomerCommand` is the way how to reproduce that this new method `$this->procedureService->getProceduresWithEndedParticipation();`  returns null. That is because `GenerateCustomerCommand` adds a "Konfiguration" phase for new customers — no evaluating phase is ever created.
- So when a procedure's participation deadline expires and the systemtries to auto-switch it to the evaluating phase, it finds nothing to switch to, silently skips the update, and the procedure stays stuck. This only affects customers created after this PR;
- all existing customers are fine. 
- one way to fix it is by adding also  evaluating (read/finished) phase for each audience. in the `seedPhaseDefinitions() ` but I dont know if that makes sense...

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
```diff
@@ -2687,66 +2685,23 @@ public function bulkEditStatementsAddData(StatementBulkEditVO $statementEdit): v
...
-
-        $statementId = $statement['id'] ?? null;
-        $statementObject = null !== $statementId ? $this->getStatement($statementId) : null;
-
-        if (!$statementObject instanceof Statement) {
-            $this->logger->warning('Statement with id '.($statementId ?? '').' not found.');
-
-            return '';
-        }
+        $statementObject = $this->getStatement($statement['id']);
```
sierrazamudiodemos: I think this will be performant concerning, the old code also had this challenge and I think it was addressed, at least I see a comment...
On large statement exports this reintroduces per-row DB queries

sierrazamudiodemos (changes requested):
I think the performance issue might be a blocker...

demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
breierDemos: Adjusted in ffe7bbb

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedurePhaseDefinitionResolver.php
```diff
@@ -0,0 +1,50 @@
...
+            }
+            foreach ($definitions as $definition) {
+                if (null !== $definition->getId()) {
+                    $this->idToNameMap[$definition->getId()] = $definition->getName();
+                }
+            }
+        }
+
+        return $this->idToNameMap[$id] ?? '';
+    }
```
sierrazamudiodemos, line 49: do you think it makes sense to add a log here? i think this would break later which is ok, because on the first place always an id has to have a name, oder?

huellnerdemos (commented):
Thank you for the great work! 🌞 I tested the PR and it worked:
BE: 
- after migration i found phase_definition_id in every place where it should be in db
- i created pdf from a statement and could see the phase name in it. i think it came from the phase definition, because i used a phase that i created before as a MA

FE:
- i could find the phase, that i crated before a MA, in the basic settings of the procedure
- choosing and saving the phase worked well
- i could find the names from the phase definition in the Admin procedure list
- i could find the phase i defined before in public procedure list and detail page, when i'm logged in as private/guest user
- the phase was also visible in the date tooltip in the assessment table
- the hole process worked for an existing procedure and also for a completely new one

But i also found a few small things in code, that could maybe be adjusted a bit. I will write single comments for each point in the following.

client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
```diff
@@ -964,7 +964,7 @@ export default {
 
       // Add submitted date and phase
       parts.push(`${Translator.trans('statement.date.submitted')}: ${this.statementDate(this.statement.submitDate)}`)
-      parts.push(`${Translator.trans('phase')}: ${this.statement.phase}`)
+      parts.push(`${Translator.trans('phase')}: ${this.statement.procedurePhase?.name}`)
```
huellnerdemos: maybe this is never the case, but if procedurePhase can't be loaded it will be rendered as "Phase: undefined" in the tooltip in the moment. maybe it's better to use it like you did it in DpPublicStatementList.vue:

```suggestion
      parts.push(`${Translator.trans('phase')}: ${this.statement.procedurePhase?.name ?? ' - '}`)
```

client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
```diff
@@ -76,7 +76,7 @@
             </td>
             <td v-cleanhtml="element" />
             <td>
-              {{ statement.phase }}
+              {{ statement.procedurePhase?.name }}
```
huellnerdemos, line 79: This is the same like in the DpAssessmentTableCard.vue:

```suggestion
              {{ statement.procedurePhase?.name ?? ' - ' }}
```

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
```diff
@@ -66,6 +66,8 @@ export default {
 
   provide () {
     return {
+      currentInternalPhaseDefinitionId: this.currentInternalPhaseDefinitionId,
+      currentExternalPhaseDefinitionId: this.currentExternalPhaseDefinitionId,
```
huellnerdemos, line 69: This is not in alphabetical order

client/js/components/procedure/basicSettings/AutoSwitchProcedurePhaseForm.vue
```diff
@@ -229,6 +233,12 @@ export default {
       return this.isInternal ? 'r_autoSwitch' : 'r_autoSwitchPublic'
     },
 
+    evaluatingPhaseName () {
+      const finishedPhase = Object.values(this.availableProcedurePhases)
+        .find(phase => phase.participationState === 'finished')
+      return finishedPhase?.label ?? ''
```
huellnerdemos, line 240: i think here should be a blank line before return:

```suggestion
        .find(phase => phase.participationState === 'finished')
        
      return finishedPhase?.label ?? ''
```

client/js/components/procedure/basicSettings/ParticipationPhases.vue
```diff
@@ -166,7 +160,7 @@ export default {
     },
 
     isInParticipation () {
-      return this.participationPhases.includes(this.selectedPhase)
+      return this.phaseOptions.find(option => option.value === this.selectedPhase)?.permissionset === 'write'
```
huellnerdemos: isn't it better to use .some() in this case?

```suggestion
      return this.phaseOptions.some(option => option.value === this.selectedPhase && option.permissionset === 'write')
```

huellnerdemos (changes requested):

client/js/components/statement/originalStatementsTable/OriginalStatementsTableItem.vue
rafelddemos: Here, I think it is not necessary:
 - the other columns in the table don't use `' - '` fallback either
 - Vue {{ }} interpolation  renders `undefined` as empty, so the word "undefined" won't be shown

client/js/components/statement/assessmentTable/DpAssessmentTableCard.vue
rafelddemos: thanks for the review, Lars! :-)

good catch, thanks! Since it is a tooltip (other data is also only rendered if there's a value), I wrapped it in an `if-statement`, so that the phase is skipped altogether should there be no value for it.

changed in ff41d5a97fd049bbb76931529f9e15bda0d109fb

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
rafelddemos: yes, thanks! changed in ff41d5a97fd049bbb76931529f9e15bda0d109fb

client/js/components/procedure/basicSettings/AutoSwitchProcedurePhaseForm.vue
rafelddemos: yes, thanks! changed in ff41d5a97fd049bbb76931529f9e15bda0d109fb

client/js/components/procedure/basicSettings/ParticipationPhases.vue
rafelddemos: you're right, thanks! changed in ff41d5a97fd049bbb76931529f9e15bda0d109fb

client/js/bundles/procedure/publicIndexNoMap.js
```diff
@@ -13,24 +13,30 @@
...
- *  Results for combined filters
- *  all in participation
+const countMatchingProcedures = function (phaseIds) {
+  const listElements = document.getElementsByClassName('c-procedurelist__item')
+  let count = 0
+  for (const listElement of listElements) {
+    const phaseSpan = listElement.getElementsByClassName('phase')[0]
+    const phaseExtSpan = listElement.getElementsByClassName('phaseExt')[0]
+    const phaseValue = phaseSpan ? phaseSpan.innerHTML.trim() : ''
+    const phaseExtValue = phaseExtSpan ? phaseExtSpan.innerHTML.trim() : ''
```
huellnerdemos: The two ternaries used here repeat a null guard that optional chaining already handles. You could change it like this:

```suggestion
    const phaseValue = phaseSpan?.innerHTML.trim() ?? ''
    const phaseExtValue = phaseExtSpan?.innerHTML.trim() ?? ''
```

client/js/bundles/procedure/publicIndexNoMap.js
```diff
@@ -13,24 +13,30 @@
...
+
+/**
+ * Splits a space-separated option value into individual phase ids.
  */
-const externalParticipationPhases = document.querySelectorAll('[data-phase-key="participation"], [data-phase-key="externalphase2"], [data-phase-key="externalphase3"]')
-const internalParticipationPhases = document.querySelectorAll('[data-phase-key="participation"], [data-phase-key="internalphase2"], [data-phase-key="internalphase3"]')
-//  Results for preparation
-const preparationPhase = document.querySelectorAll('[data-phase-key="preparation"]')
+const splitOptionValue = function (value) {
+  return value.split(/\s+/).filter(v => v.trim().length > 0)
```
huellnerdemos: The variable name v is not very descriptive, maybe change it to a more descriptive one

client/js/components/statement/listOriginalStatements/ListOriginalStatements.vue
```diff
@@ -654,6 +652,13 @@ export default {
       return paragraph ? paragraph.attributes.title : '-'
     },
 
+    getProcedurePhaseName (originalStatementId) {
+      const originalStatement = this.originalStatements[originalStatementId]
+      const procedurePhase = originalStatement.relationships.procedurePhase?.data ? originalStatement.relationships.procedurePhase.get() : null
+
+      return procedurePhase ? procedurePhase.attributes.name : ''
```
huellnerdemos: This ternary can be simplified, by using optional chaining:

```suggestion
      return procedurePhase?.attributes.name ?? ''
```

client/js/components/procedure/publicindex/SharedProcedureMethods.js
```diff
@@ -52,13 +52,13 @@ const Methods = {
...
   phaseName () {
-    const { internalPhaseTranslationKey, externalPhaseTranslationKey } = this.procedure
-    if (!!internalPhaseTranslationKey && !!externalPhaseTranslationKey) {
-      return `${Translator.trans(internalPhaseTranslationKey)} / ${Translator.trans(externalPhaseTranslationKey)}`
-    } else if (externalPhaseTranslationKey) {
-      return Translator.trans(externalPhaseTranslationKey)
-    } else if (internalPhaseTranslationKey) {
-      return Translator.trans(internalPhaseTranslationKey)
+    const { internalPhaseDefinitionName, externalPhaseDefinitionName } = this.procedure
+    if (!!internalPhaseDefinitionName && !!externalPhaseDefinitionName) {
```
huellnerdemos: is the double negation really needed here? isn't the behavior the same without it?

huellnerdemos (commented):
Just found a few more small points.

client/js/bundles/procedure/publicIndexNoMap.js
rafelddemos: good catch, thanks. changed in 16ddc3a993451f41df7d27fe967a98a115fead2e

client/js/bundles/procedure/publicIndexNoMap.js
rafelddemos: changed in 16ddc3a993451f41df7d27fe967a98a115fead2e

client/js/components/statement/listOriginalStatements/ListOriginalStatements.vue
rafelddemos: thx, changed in 16ddc3a993451f41df7d27fe967a98a115fead2e

client/js/components/procedure/publicindex/SharedProcedureMethods.js
rafelddemos: no, actual Boolean is not needed here, thx. changed in 16ddc3a993451f41df7d27fe967a98a115fead2e

huellnerdemos (commented):
Looks good now from my frontend point of view. Thank you! 🎉

rafelddemos:
closing this PR in favor of https://github.com/demos-europe/demosplan-core/pull/6310, which combines changes from two separate branches (https://github.com/demos-europe/demosplan-core/pull/5949 and https://github.com/demos-europe/demosplan-core/pull/5966)

PR #5971, opened 2026-03-06
f_DPLAN-17311: segment grouping in excel export

kijewskidemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17311/BE-Synopse-Gruppierung-nach-Abschnitten-im-Excel-Export

### Description: 
This PR alters the Statement Excel Export. Now, a segment only triggers a new line in the data sheet once, instead of triggering the creation each time it appears. Tags belonging to the segments are now also listed in one cell. Tag Topic Titles are now added in brackets behind each tag. Purpose of this PR is it to make the excel file more convenient to use. 

### How to review/test
Log in as admin, go to a procedure, then to Stellungnahmen, click export and select the xlsl option. Select tags and click apply. Export the file and see if segments now only appear once in the data sheet, with several tags listed in one cell belongig to it. Also check if each tag has its topic title added in brackets behind.


MoritzMandler (approved):
just a few nitpicks - otherwise looks good - thx :)

PR #5984, opened 2026-03-10
fix (refs DPLAN-17458): return early on screenshot save failure, return null on missing file hash

breierDemos:
### Ticket
DPLAN-17458

When saving a map screenshot fails (e.g. due to a permission error on the files directory), `createMapScreenshot()` was silently continuing with an empty hash, persisting a broken `mapFile` value like `Map_xxx.png:` to the statement. During a later docx export, `getScreenshot()` would return `''` instead of `null`, causing `fileExists('')` to be called on the S3 adapter, which throws an `InvalidArgumentException` because AWS requires a non-empty key.

**Changes:**
- `MapService::createMapScreenshot()`: return early with `''` when the screenshot file cannot be saved, preventing a broken `mapFile` from being persisted
- `DocxExporter::getScreenshot()`: return `null` instead of `''` when file lookup fails, so the screenshot row is correctly skipped during export

### How to review/test
- Trigger a docx export on a procedure that has a statement with map data but a missing/inaccessible screenshot file
- Export should complete without the `InvalidArgumentException: [Key] expected string length to be >= 1` error

### PR Checklist
- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Update changelog


mussbach:
@breierDemos what about this PR?

PR #6020, opened 2026-03-19
F_DPLAN-17480_bmds key cloak - solved conflicts

MoritzMandler:
Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-17480/Umsetzung-Refresh-Token

Description:
Solved conflicts with multi-orga multi-customer implementations against token-refresh and request-buffer implementation

Initial PR: https://github.com/demos-europe/demosplan-core/pull/5924
This PR will replace the one mentioned above in order to be able to review the solved merge conflicts.
Some further implementations are needed to handle the - redirect back pattern (to the page the user was working on after "quick" relog now with also persisted orga context).
Also missing Test coverage will be addressed here since the conflicts are now resolved.

How to test:
For multi-orga users you can test the redirect pattern if you append
```?keycloakTest=multi-org-user&simulateBufferedData=1```
to any project Url. This dev-mode only authenticator will setup fake buffered request data as well as a fake pending_page Url to /faq. The redirect to the pending_page will only happen if the same orga has been chosen after the reAuth then before - within the static-authenticator the before orga will be set to: ```Test Organisation Beta - Abfallwirtschaft```

For single-orga-users you can test the redirect pattern if you append
```?keycloakTest=fachplaner-admin&simulateBufferedData=1```
to any project Url. If the user is only a member of a single orga the redirect to pending_page Url will happen all the time.

For the real keyCloak authenticator the pending_request and pending_page will be stored for min 1 and max 2 hours.
After this period a fresh login will lead you to the projects landing page.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### CODE REVIEW

File: demosplan/DemosPlanCoreBundle/Logic/OAuth/PendingRequestCacheService.php [hunk 1]
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * The encrypted body is decrypted before filling the value object.
+     * Returns null if no entry exists or it has expired.
+     */
+    public function retrieve(string $userId): ?PendingRequestData
+    {
+        try {
+            $cacheKey = self::CACHE_KEY_PREFIX.$userId;
+
+            $cached = $this->cache->get($cacheKey, function (): string {
```
mussbach, line 105: `CacheInterface::get()` computes-and-stores on miss — so `retrieve()` creates an empty cache entry (`''`) for every non-existing key, then immediately deletes it. There's a race window where a concurrent request reads back the empty string. Same pattern in `store()`: `delete()` then `get()` with a callback has a TOCTOU race — if the entry reappears between the two calls the callback won't fire and the old data survives.

Consider injecting `Psr\Cache\CacheItemPoolInterface` and using `getItem()` + `isHit()` for reads, and explicit `save()` for writes:

```php
// Read
$item = $this->pool->getItem($cacheKey);
if (!$item->isHit()) {
    return null;
}
$cached = $item->get();

// Write
$item = $this->pool->getItem($cacheKey);
$item->set($data);
$item->expiresAfter(self::CACHE_TTL_SECONDS);
$this->pool->save($item);
```

demosplan/DemosPlanCoreBundle/Security/Authentication/Authenticator/AbstractOzgKeycloakAuthenticator.php
```diff
@@ -0,0 +1,251 @@
...
+                $this->currentOrganisationService->setCurrentOrganisation($user, $singleOrga);
+            }
+        }
+
+        $this->messageBag->add('confirm', 'confirm.session.renewed');
+
+        // Not yet implemented: if pendingRequest has a buffered POST, redirect to pending request review page
+
+        return new RedirectResponse($pendingRequest->getPageUrl());
+    }
```
mussbach, line 205: Potential open redirect. The `OAuthToken` entity has an `Assert\Regex` constraint requiring `^/` on `pendingPageUrl`, but `storePendingPageUrl()` does not call the validator — so a URL like `//evil.com/steal-token` (protocol-relative, passes `^/`) could be stored and redirected to. Same applies in `OrganisationSelectionController`.

Consider adding an explicit check before redirecting:

```php
if (str_starts_with($pendingRequest->getPageUrl(), '//') || !str_starts_with($pendingRequest->getPageUrl(), '/')) {
    return new RedirectResponse($this->router->generate('core_home_loggedin'));
}
```

Or validate via `parse_url($url, PHP_URL_HOST) === null` to ensure it's truly a relative path.

demosplan/DemosPlanCoreBundle/Logic/OAuth/TokenEncryptionService.php
```diff
@@ -0,0 +1,137 @@
...
+
+            // Decrypt with authentication verification — false indicates tampered or corrupt data
+            $plaintext = sodium_crypto_secretbox_open($ciphertext, $nonce, $this->encryptionKey);
+
+            if (false === $plaintext) {
+                throw new TokenEncryptionException('Decryption failed - data may be corrupted or tampered with');
+            }
+
+            return $plaintext;
+        } catch (SodiumException $e) {
```
mussbach: The `catch (SodiumException $e)` block only catches Sodium-level errors. The `TokenEncryptionException` thrown at lines 110, 117, and 125 (invalid base64, invalid nonce, tampered data) will bubble up uncaught — so those failure paths skip the "Failed to decrypt" log entry and propagate a different exception type than the `SodiumException` path.

Consider catching both:

```php
} catch (SodiumException|TokenEncryptionException $e) {
    $this->logger->error('Failed to decrypt OAuth token', [
        'exception' => $e->getMessage(),
    ]);

    throw new TokenEncryptionException('Token decryption failed', 0, $e);
}
```

Or restructure so all early returns use a plain `throw` and only the outer catch wraps + logs.

demosplan/DemosPlanCoreBundle/Repository/OAuthTokenRepository.php
```diff
@@ -0,0 +1,114 @@
...
+     * @param int $olderThanMinutes Delete entries with pending data older than this many minutes (default: 60 = 1 hour)
+     *
+     * @return int Number of deleted entries
+     */
+    public function clearOutdated(int $olderThanMinutes = 60): int
+    {
+        $timezone = new DateTimeZone(OAuthToken::TIMEZONE);
+        $threshold = new DateTime("-{$olderThanMinutes} minutes", $timezone);
+
+        $qb = $this->createQueryBuilder('t');
```
mussbach: Optional: `clearOutdated()` fetches all matching tokens into memory and removes them one-by-one. With many stale tokens this is unnecessarily expensive. A bulk DQL DELETE avoids hydration entirely:

```php
$qb = $this->getEntityManager()->createQueryBuilder();
return $qb->delete(OAuthToken::class, 't')
    ->where('t.pendingRequestTimestamp IS NOT NULL')
    ->andWhere('t.pendingRequestTimestamp < :threshold')
    ->setParameter('threshold', $threshold)
    ->getQuery()
    ->execute();
```

demosplan/DemosPlanCoreBundle/ValueObject/TokenData.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Fill the value object with decrypted token data from an OAuthToken entity.
+     *
+     * @throws TokenEncryptionException if decryption fails
+     */
+    public function fill(OAuthToken $oauthToken, TokenEncryptionService $encryptionService): void
+    {
+        // Decrypt required tokens (guaranteed by validation on storage)
+        $this->accessToken = $encryptionService->decrypt($oauthToken->getAccessToken());
```
mussbach, line 70: `getIdToken()` can return `null` at the entity level (the column is nullable). While `storeTokens()` validates id_token is present, data corruption or manual DB edits could leave it null — passing `null` to `decrypt()` (which expects `string`) would cause a `TypeError`.

Consider adding a null guard like the one used for `refreshToken` below:

```php
if (null !== $oauthToken->getIdToken()) {
    $this->idToken = $encryptionService->decrypt($oauthToken->getIdToken());
}
```

demosplan/DemosPlanCoreBundle/Entity/User/OAuthToken.php
```diff
@@ -0,0 +1,411 @@
...
+ *         @ORM\Index(name="idx_access_expires", columns={"access_token_expires_at"}),
+ *         @ORM\Index(name="idx_pending_timestamp", columns={"pending_request_timestamp"})
+ *     }
+ * )
+ *
+ * @ORM\Entity(repositoryClass="demosplan\DemosPlanCoreBundle\Repository\OAuthTokenRepository")
+ */
+class OAuthToken implements EntityInterface
+{
+    public const TIMEZONE = 'Europe/Berlin';
```
mussbach: Hardcoded `Europe/Berlin` creates a latent bug if the DB server timezone differs, and will behave incorrectly during DST transitions (token expiry comparisons off by 1 hour twice a year).

Consider making this configurable via an env var + parameter, defaulting to `UTC`:

`.env`:
```
OAUTH_TOKEN_TIMEZONE=UTC
```

`parameters_default.yml`:
```yaml
oauth_token_timezone: '%env(OAUTH_TOKEN_TIMEZONE)%'
```

Then inject the parameter here instead of the hardcoded constant. That way token expiry math is timezone-safe by default and deployments that need `Europe/Berlin` can set it explicitly.

config/parameters_default.yml
```diff
@@ -393,6 +393,23 @@ parameters:
     # https://login.microsoftonline.com/<tenant>/oauth2/v2.0/logout
     oauth_azure_logout_route: ''
 
+    # Encryption key for storing OAuth tokens securely (AES-256-GCM via Sodium)
```
mussbach: Comment says "AES-256-GCM via Sodium" but the implementation uses `sodium_crypto_secretbox` which is XSalsa20-Poly1305. The CHANGELOG and `TokenEncryptionService` class docblock correctly say XSalsa20-Poly1305.

```suggestion
    # Encryption key for storing OAuth tokens securely (XSalsa20-Poly1305 via Sodium)
```

demosplan/DemosPlanCoreBundle/Logic/OAuth/KeycloakTokenRefreshService.php
```diff
@@ -0,0 +1,206 @@
...
+     *
+     * IMPORTANT: KeyCloak rotates ALL tokens on refresh (access token, refresh token, ID token).
+     * The old refresh token becomes invalid immediately after use.
+     *
+     * @param string $userId The user ID (UUID)
+     *
+     * @return bool True if tokens were successfully refreshed, false on any error
+     */
+    public function refreshTokensForUser(string $userId): bool
+    {
```
mussbach, line 131: `refreshTokensForUser` is `public` but only called from `tryRefreshTokens()` (same class) and `TokenRefreshTerminateListener`. Consider whether the terminate listener should go through `tryRefreshTokens()` instead — that way it would benefit from the refresh-token-expired check and this method could be made `private`.

mussbach (changes requested):
See inline comments — the cache race condition and open redirect should be addressed before merge.

demosplan/DemosPlanCoreBundle/Logic/OAuth/TokenEncryptionService.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6020/commits/e99e904c2b241d6f182663ad3cecf3d8065a0f6f

TokenEncryptionService has been replaced entirely by the       
existing SecretEncryptor (from main). In SecretEncryptor::decrypt(), all early validation   
failures (invalid base64, short nonce, tampered data) throw CryptoException consistently,   
and the try/catch around sodium_crypto_secretbox_open() wraps the SodiumException into 
CryptoException as well. So there's no longer a mixed exception type issue — every failure
path produces the same exception type.

demosplan/DemosPlanCoreBundle/Security/Authentication/Authenticator/AbstractOzgKeycloakAuthenticator.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6020/commits/26760fac1527dc3e104e7aadfc29034ec87087fb

storePendingPageUrl() was missing the $this->validator->validate() call that storePendingRequest() already had. Added it with error logging. The existing Assert\Regex('^/') constraint on the entity now rejects protocol-relative URLs like //evil.com/steal-token on both storage paths.

demosplan/DemosPlanCoreBundle/Logic/OAuth/PendingRequestCacheService.php
MoritzMandler: As discussed - the cache is only build after a successful re-authentication at keyCloak to bridge the gap between the multi-orga-selection page and the actual usage of the buffered-data after same orga select.

This is not really a race condition scenario

demosplan/DemosPlanCoreBundle/Repository/OAuthTokenRepository.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6020/commits/2a1949d09d80d5c6f6a7af5b0976848ae3c8da1a
replaced with a bulk DQL DELETE that avoids hydration entirely. Single query regardless of how many stale entries exist

demosplan/DemosPlanCoreBundle/ValueObject/TokenData.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6020/commits/898fcb2b0885cd63f954d799bdec782521463fa2
added a null guard matching the existing refreshToken pattern. Property and  @method annotation are now nullable too

demosplan/DemosPlanCoreBundle/Entity/User/OAuthToken.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6020/commits/3cf1de98b2e7127a26c097588227301634580a46

the timezone is now configurable via oauth_token_timezone parameter (env var OAUTH_TOKEN_TIMEZONE, defaults to Europe/Berlin). The OAuthToken::TIMEZONE constant has been removed and all services receive the timezone via constructor injection. Repository methods accept a DateTimeZone parameter from their callers. This eliminates the DST drift risk for deployments where the DB server uses a different timezone

config/parameters_default.yml
MoritzMandler: gone due to https://github.com/demos-europe/demosplan-core/pull/6020/commits/e99e904c2b241d6f182663ad3cecf3d8065a0f6f

mussbach:
The validator call in `storePendingPageUrl()` is a good addition, but the `^/` regex still accepts protocol-relative URLs like `//evil.com/steal-token` — it starts with `/`, so the constraint passes. The redirect in `handleSingleOrgaReauth()` and `OrganisationSelectionController` would follow it.

Either tighten the regex to `^/[^/]` or add an explicit guard before redirecting.

demosplan/DemosPlanCoreBundle/Logic/OAuth/KeycloakTokenRefreshService.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6020/commits/396662d8676d704e071efb292fee67827a5ab6fc
refreshTokensForUser() is now private. The terminate listener calls tryRefreshTokens() after its own accessTokenNeedsRefresh() check, so it benefits from the refresh-token-expired guard. Tests use reflection via a helper method to reach the private method for lock/concurrency tests

MoritzMandler:
> The validator call in `storePendingPageUrl()` is a good addition, but the `^/` regex still accepts protocol-relative URLs like `//evil.com/steal-token` — it starts with `/`, so the constraint passes. The redirect in `handleSingleOrgaReauth()` and `OrganisationSelectionController` would follow it.
> 
> Either tighten the regex to `^/[^/]` or add an explicit guard before redirecting.

done in https://github.com/demos-europe/demosplan-core/pull/6020/commits/06a3ce4bcb6a9f8fcdb97e64736abf76e0d16da8
tightened both regex constraints from ^/ to ^/[^/], which rejects //evil.com 
while still accepting all valid relative paths like /verfahren/123

mussbach:
**Token cleanup gap: rows without pending requests are never purged**

`OAuthTokenRepository::clearOutdated()` only deletes rows where `pendingRequestTimestamp IS NOT NULL` and older than 60 minutes. This means rows where:

- The user's access + refresh tokens are both expired
- But `pendingRequestTimestamp` is null (no pending request was buffered)
- And the user never explicitly logs out (just closes the browser)

...will accumulate indefinitely in the `oauth_tokens` table.

Consider extending `clearOutdated()` to also delete rows where `accessTokenExpiresAt` is older than some threshold (e.g. 24h) regardless of pending request state:

```php
->orWhere($qb->expr()->andX(
    't.accessTokenExpiresAt IS NOT NULL',
    't.accessTokenExpiresAt < :tokenThreshold'
))
```

mussbach:
@MoritzMandler this looks well now, only the minor comment from today remains. I would suggest to set feature_keycloak_used_for_login_only to true in core for all projects so that we can merge this without risk and test the feature later on in real keycloak environment, wdyt?

MoritzMandler:
> **Token cleanup gap: rows without pending requests are never purged**
> 
> `OAuthTokenRepository::clearOutdated()` only deletes rows where `pendingRequestTimestamp IS NOT NULL` and older than 60 minutes. This means rows where:
> 
>     * The user's access + refresh tokens are both expired
> 
>     * But `pendingRequestTimestamp` is null (no pending request was buffered)
> 
>     * And the user never explicitly logs out (just closes the browser)
> 
> 
> ...will accumulate indefinitely in the `oauth_tokens` table.
> 
> Consider extending `clearOutdated()` to also delete rows where `accessTokenExpiresAt` is older than some threshold (e.g. 24h) regardless of pending request state:
> 
> ```
> ->orWhere($qb->expr()->andX(
>     't.accessTokenExpiresAt IS NOT NULL',
>     't.accessTokenExpiresAt < :tokenThreshold'
> ))
> ```

thx, done in https://github.com/demos-europe/demosplan-core/pull/6020/commits/096f7c93d7b4291d5ca32225fbd37bb2f9cd0cf9 clearOutdated() now handles two cases:
 - stale pending requests (existing, 60  - min threshold) 
 - abandoned sessions where the user never logged out (new, 8h threshold = 6h max session lifetime + 2h buffer) 

Both run on the existing hourly schedule via the same handler.

MoritzMandler:
> @MoritzMandler this looks well now, only the minor comment from today remains. I would suggest to set feature_keycloak_used_for_login_only to true in core for all projects so that we can merge this without risk and test the feature later on in real keycloak environment, wdyt?

Agreed — replaced the permission with an ```oauth_keycloak_login_only``` parameter in 
parameters_default.yml (defaults to true). This means the merge is safe with zero behaviour change.
Projects opt in to full token management by setting ```oauth_keycloak_login_only: false``` in their ```parameters_default_project.yml```. As a side effect, isKeycloakLoginOnly() is now a cheap parameter read
instead of a permission check, so it moved into the early-return path of both listeners (before any user entity lookups), and CurrentUserService was removed from OzgKeycloakSessionManager. https://github.com/demos-europe/demosplan-core/pull/6020/commits/e747901f70fde611e99e3450b643743732066e30

mussbach (approved):
lgtm, login works, thanks for this hard one

PR #6030, opened 2026-03-20
fix: handle promise rejections in DpDashboardTaskCard

mussbach:
## Summary
- Add `.catch()` handlers to all API calls in `DpDashboardTaskCard.vue` to prevent unhandled promise rejections on the procedure dashboard
- Chain inner `dpApi.patch()` rejection to the outer promise via `return` so errors propagate to a single `.catch()`
- Component gracefully degrades to its defaults (0 segments, no filter link) on failure

## Test plan
- [ ] Open a procedure dashboard (`/verfahren/{id}/uebersicht`)
- [ ] Verify the task card loads correctly with assigned segments
- [ ] Simulate a network failure and confirm no unhandled rejection appears in the console


riechedemos (approved):
Lgtm and works, thank you!

PR #6058, opened 2026-03-27
F dplan 12697 new exchange format transformer adjustments

MoritzMandler:
Ticket

BE implementation for https://github.com/demos-europe/demosplan-core/pull/4473

Description: handle new segment exchange format in BE transformer
  Parse segment text from <segment-mark> elements in textualReference
  instead of relying on charStart/charEnd positions. Segments metadata
  (tags, assignee, place) is matched by ID from the segments array.

How to review/test:
The FE will auto convert the old draftsInfoJson format whenever an in between segmentation step gets saved (like splitting an additional segment from the statement-text or setting a tag, assignee ect.)
Or it will be converted on finalization when the be attempts to transform the draftsInfoJson into actual segments.

So the BE will only get the new exchange format - so you can attempt to transform any non segmented statement and that should work.

- [x] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


mussbach (commented):
## Suggestion: Use `Masterminds\HTML5` instead of `DOMDocument::loadHTML()`

`DOMDocument::loadHTML()` uses libxml2's HTML4 parser, which doesn't understand custom elements like `<segment-mark>`. It can silently mangle the DOM structure or mishandle such elements. The current code works around this with `libxml_use_internal_errors(true)`, which also leaks global state (suppresses libxml errors for other code running in the same request).

`masterminds/html5` is already a direct dependency in `composer.json` and is used in the codebase for the same kind of work — see `HTMLFragmentSlicer` (`demosplan/DemosPlanCoreBundle/Services/HTMLFragmentSlicer.php:149`):

```php
$parser = new HTML5();
$this->dom = $parser->loadHTMLFragment($this->originalFragment);
```

Using it in `SegmentMarkParser` would:
- Properly handle `<segment-mark>` as a custom element per the HTML5 spec
- Handle UTF-8 natively (no `<?xml encoding="UTF-8">` wrapper hack)
- Remove the need for `libxml_use_internal_errors` / `libxml_clear_errors`
- Be more resilient to varied HTML from the frontend

The `parse()` method could be simplified to roughly:

```php
$parser = new \Masterminds\HTML5();
$dom = $parser->loadHTMLFragment($html);
$xpath = new \DOMXPath($dom);
$nodes = $xpath->query('//segment-mark');
// ... rest stays the same
```

No new dependency needed — just a more appropriate tool for the job that's already available.

mussbach (changes requested):

MoritzMandler:
> ## Suggestion: Use `Masterminds\HTML5` instead of `DOMDocument::loadHTML()`
> 
> `DOMDocument::loadHTML()` uses libxml2's HTML4 parser, which doesn't understand custom elements like `<segment-mark>`. It can silently mangle the DOM structure or mishandle such elements. The current code works around this with `libxml_use_internal_errors(true)`, which also leaks global state (suppresses libxml errors for other code running in the same request).
> 
> `masterminds/html5` is already a direct dependency in `composer.json` and is used in the codebase for the same kind of work — see `HTMLFragmentSlicer` (`demosplan/DemosPlanCoreBundle/Services/HTMLFragmentSlicer.php:149`):
> 
> ```
> $parser = new HTML5();
> $this->dom = $parser->loadHTMLFragment($this->originalFragment);
> ```
> 
> Using it in `SegmentMarkParser` would:
> 
>     * Properly handle `<segment-mark>` as a custom element per the HTML5 spec
> 
>     * Handle UTF-8 natively (no `<?xml encoding="UTF-8">` wrapper hack)
> 
>     * Remove the need for `libxml_use_internal_errors` / `libxml_clear_errors`
> 
>     * Be more resilient to varied HTML from the frontend
> 
> 
> The `parse()` method could be simplified to roughly:
> 
> ```
> $parser = new \Masterminds\HTML5();
> $dom = $parser->loadHTMLFragment($html);
> $xpath = new \DOMXPath($dom);
> $nodes = $xpath->query('//segment-mark');
> // ... rest stays the same
> ```
> 
> No new dependency needed — just a more appropriate tool for the job that's already available.

changed in https://github.com/demos-europe/demosplan-core/pull/6058/commits/8ce7fc043c7b8ab89184cb2a43eabaffc9a4054b
Switched from DOMDocument::loadHTML() to Masterminds\HTML5 — proper HTML5 custom element handling, native UTF-8, no more libxml error suppression hacks.

mussbach (approved):
lgtm, thx, code review only

PR #6080, opened 2026-03-31
fix(DPLAN-17599): show logout button on IDP error page to terminate Keycloak session

mussbach:
DPLAN-17599

## Summary
- When IDP authentication fails (e.g. missing roles), the user lands on the error page as `AnonymousUser` — the header incorrectly showed "Anmelden" even though a Keycloak session exists
- The controller now resolves the Keycloak logout URL via `OzgKeycloakLogoutManager` and passes it to the template
- Both userbox templates (`base_header_userbox` and `base_oeb_header_userbox`) check for `idpLogoutUrl` and render "Abmelden" linking to the Keycloak logout endpoint

## Test plan
- [ ] Log in via IDP with a user that has missing roles/permissions
- [ ] Verify the error page shows "Abmelden" in the top right instead of "Anmelden"
- [ ] Click "Abmelden" and verify the Keycloak session is terminated
- [ ] Verify normal login/logout flow is unaffected (no `idpLogoutUrl` is set on other pages)


demosplan/DemosPlanCoreBundle/Controller/Platform/EntrypointController.php
```diff
@@ -152,16 +153,29 @@ public function index(
...
+        $idpLogoutUrl = null;
+        try {
+            $logoutRoute = $ozgKeycloakLogoutManager->getEffectiveLogoutRoute();
+            if (null !== $logoutRoute) {
+                $idpLogoutUrl = $ozgKeycloakLogoutManager->getLogoutUrl($logoutRoute, null);
+            }
+        } catch (Exception $e) {
+            // If we cannot build the logout URL, fall back to no logout link
+        }
+
```
sierrazamudiodemos, line 173: @mussbach i have a question regarding this one: so if tehre is not idpLogoutUrl route, it iwll be null and the template will decide to show based if it is null or not right? that is why there are changes in twig

sierrazamudiodemos: maybe it is worth writting a log for this case?

mussbach: Can be maybe improved by a log entry, yes. No improvement compared to current state is imho ok when no logoutUrl can be found

PR #6085, opened 2026-04-01
feat(DPLAN-17442): Redesign user management page

meissnerdemos:
### Ticket
DPLAN-17442

### Description
Redesigned the user management page: the create-user form now opens in a modal instead of an accordion. User selection was reworked to support "select all" across pages by tracking deselections rather than individual selections. An expandable section shows which users are currently selected. Action buttons (invite/delete) display the selection count. Added an info notification about the automated invitation email.

### How to review/test
1. Log in as Masteruser
2. Go to Nutzende
3. Test creating a new user via the modal
4. Test selecting/deselecting individual users and "select all"
5. Expand the selection list to verify selected users are shown
6. Test invite and delete actions with selections across multiple pages

### PR Checklist
- [x] Run `yarn lint`
- [x] Run `yarn test`


client/js/components/user/DpUserList/DpUserFormFields.vue
```diff
@@ -150,10 +150,15 @@
       </dp-multiselect>
     </div>
   </div>
+  <dp-inline-notification
+    class="mt-4"
```
hwiem:
```suggestion
    class="mb-4"
```

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -14,87 +14,127 @@
...
-            :text="deleteSelectedUsersLabel"
-            @click="deleteItems(selectedItems)"
+            :text="Translator.trans('delete') + ` (${selectedItemsCount})`"
+            @click="deleteItems"
           />
+          <dp-button
+            class="p-1"
+            color="primary"
+            data-cy="userList:manageUsers"
+            type="button"
```
hwiem: Here and above: `type="button"` is not needed

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -177,16 +222,35 @@ export default {
       totalPages: 'totalPages',
     }),
 
+    allOnPageSelected () {
+      const itemIds = Object.keys(this.items)
+      return itemIds.length > 0 && itemIds.every(id => this.currentPageSelections[id])
```
hwiem:
```suggestion
      const itemIds = Object.keys(this.items)

      return itemIds.length > 0 && itemIds.every(id => this.currentPageSelections[id])
```

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -14,87 +14,127 @@
...
+        <div>
+          <input
+            id="select_all"
+            type="checkbox"
+            data-cy="allSelected"
+            :checked="allOnPageSelected"
+            @change="toggleAll(!allOnPageSelected)"
+          >
+          <label
+            v-if="hasPermission('feature_user_delete') || true"
```
sakutademos: hm, i am just wondering weather we really need such kind of permission check, because it returns true all the time

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -297,30 +375,155 @@ export default {
...
+    async fetchAllUsersForSelectionMap () {
+      const allUsers = {}
+      const filter = this.buildUserFilter()
+      const include = ['roles', 'orga', 'department', 'orga.allowedRoles'].join()
+      let page = 1
+      let totalPages = 1
+
+      while (page <= totalPages) {
+        // Store not used so we stay on selected page
+        const response = await dpApi.get('/api/2.0/AdministratableUser', {
```
sakutademos: I am a bit concerned about this implementation, especially if there are 1000 users... this could cause a memory leak when a user selects all, deselects one, and then opens the user list. Even with 70 users it needs some time for 7 requests. Maybe in this case we should fetch all users in a single request instead.

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -297,30 +375,155 @@ export default {
...
+    },
+
+    async fetchAllUserIds () {
+      const allIds = []
+      const filter = this.buildUserFilter()
+      const include = ['roles', 'orga', 'department', 'orga.allowedRoles'].join()
+      let page = 1
+      let totalPages = 1
+
+      while (page <= totalPages) {
```
sakutademos: to  fetch all users, you do not need to use a while loop and the page property. You can simply remove the page property from the request, and you will get all users in the response

client/js/components/user/DpCreateItem.vue
```diff
@@ -33,27 +45,35 @@
           v-bind="dynamicComponentProps"
           :is="dynamicComponent"
           ref="formFields"
+          class=""
```
sakutademos: remove me

client/js/components/user/DpCreateItem.vue
```diff
@@ -17,14 +17,26 @@
...
+      </h3>
+      <dp-button
+        variant="outline"
+        :text="Translator.trans('add')"
+        @click="toggleModal"
+      />
+    </div>
+    <dp-modal
+      ref="createItemModal"
+      content-classes="u-2-of-3"
```
sakutademos: we can use tailwind classes instead, like width

sakutademos (changes requested):
thx, it generally works fine and the code looks goo, I just have a few concerns. Also, as I can see, the design has changed and there should be an accordion (as before) instead of a modal

client/js/components/user/DpCreateItem.vue
meissnerdemos: 4c9e5e34e347cacebf6ea95074a93f8649896566

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: you're right, the fallback to true is unnecessary e8e498b313ed5296d9ccc2330eb400e60f3edc22

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: Right! I just assumed it had to have a page parameter because of what was there already d41d100cf222e7dd7e85d1c7d184f1d41d857103

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: I changed it to fetch all in  a single request and I removed the include to hopefully reduce the data size. d41d100cf222e7dd7e85d1c7d184f1d41d857103

client/js/components/user/DpUserList/DpUserFormFields.vue
```diff
@@ -8,8 +8,8 @@
 </license>
 
 <template>
-  <div class="whitespace-nowrap">
-    <div class="inline-block w-1/2 pr-3 my-3">
+  <div class="whitespace-nowrap p-4">
```
sakutademos: i think we can only set the vertical spacing here 

Image

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -177,22 +228,38 @@ export default {
...
+        this.toggledItems.length
+    },
+
+    selectedUsersForDropdown () {
+      return Object.values(this.selectedUsersMap)
     },
 
     tooltipContent () {
-      return '<h3 class="u-mt color--white">' + Translator.trans('search.options') + '</h3>' +
+      return '<h3 class="mt-4 text-on-dark">' + Translator.trans('search.options') + '</h3>' +
```
sakutademos: i think it would be a bit more readable if we use template literals instead of the `<h3> + Translator.trans() + </h3>`

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -297,30 +392,144 @@ export default {
...
+    },
+
+    toggleAll (status) {
+      this.allUsersFetched = false
+      this.selectedUsersMap = {}
+      this.trackDeselected = status
+      this.toggledItems = []
+    },
+
+    toggleOne (id) {
```
sakutademos, line 528: this method is a bit complex. maybe we could split it into more descriptive methods instead of adding a comments

sakutademos (changes requested):
thx for the refactoring, I tested it and the code looks good, just a few nitpicks!
While testing, I get an issue when selecting all users, deselecting one and then clicking on _Auswahl anzeigen_, could you please check it

client/js/components/user/DpUserList/DpUserFormFields.vue
meissnerdemos: Yes, that looks better 177ec1173862d9dbb101cfe76e449487cdc6068b

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: changed it in b4edd74c28e2a80618de9a9c260a93e65dce4111

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: I split it up in f070a61e2071443ae1028ab748554b35f5e2e647

meissnerdemos:
> While testing, I get an issue when selecting all users, deselecting one and then clicking on _Auswahl anzeigen_, could you please check it

There was a problem with the routing. It didn't have the `app_dev.php` part. This should fix it: b4edd74c28e2a80618de9a9c260a93e65dce4111

hwiem:
> > While testing, I get an issue when selecting all users, deselecting one and then clicking on _Auswahl anzeigen_, could you please check it
> 
> There was a problem with the routing. It didn't have the `app_dev.php` part. This should fix it: [b4edd74](https://github.com/demos-europe/demosplan-core/commit/b4edd74c28e2a80618de9a9c260a93e65dce4111)

I tested it and it works :tada:

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -14,87 +14,133 @@
...
-      <div class="layout__item u-1-of-1 flex">
-        <dp-search-field
-          data-cy="search:currentSearchTerm"
-          :placeholder="Translator.trans('searchterm')"
-          @search="handleSearch"
-          @reset="handleReset"
-        />
-        <dp-contextual-help :text="tooltipContent" />
-      </div>
+    <div class="flex mt-4">
```
hwiem: We could remove the `flex` class, because we don't display tooltips aligned to the right anymore, but right next to the label they belong to.

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -14,87 +14,133 @@
...
-            @change="dpToggleAll(!allSelected, items)"
-          >
-          <label
-            v-if="hasPermission('feature_user_delete') || true"
-            for="select_all"
-            class="cursor-pointer btn-icns inline-block"
-          >
-            {{ Translator.trans('select.all.on.page') }}
-          </label>
+    <div v-if="false === isLoading">
```
hwiem: nitpick - we don't use yoda style (I know you didn't introduce it, but it's something that can be easily cleaned up while touching the code)
```suggestion
    <div v-if="!isLoading">
```

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -14,87 +14,133 @@
...
+        v-if="isUserSelected"
+        class="shadow-sm rounded-lg mt-2 py-4 px-6"
+      >
+        <div class="flex items-center justify-between border-b border-neutral py-2">
+          <span>{{ Translator.trans('users.selected', { count: selectedItemsCount }) }}</span>
+          <dp-button
+            :disabled="trackDeselected && toggledItems.length === 0"
+            :icon="showSelectionList ? 'caret-up' : 'caret-down'"
+            :text="Translator.trans('selection.show')"
+            type="button"
```
hwiem: nitpick: `'button'` is the default for the `type` prop, so we can omit it

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -14,87 +14,133 @@
...
+            @click="toggleSelectionList"
+          />
         </div>
+        <dp-transition-expand>
+          <div v-show="showSelectionList">
+            <dp-loading
+              v-if="isLoadingSelectionList"
+              class="mt-1"
+            />
+            <div
```
hwiem: I am not sure if `v-else` and `v-for` should be used on the same element - at least for `v-if` and `v-for`, this is [not recommended](https://vuejs.org/guide/essentials/list.html#v-for-with-v-if). To me it seems cleaner to wrap the `div` in a `template` and place the `v-else` on it.

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -14,87 +14,133 @@
...
-            :text="deleteSelectedUsersLabel"
-            @click="deleteItems(selectedItems)"
+            :text="Translator.trans('invite') + ` (${selectedItemsCount})`"
+            @click="inviteItems"
           />
         </div>
       </div>
+      <div class="mt-4 flex items-center justify-between">
+        <!-- 'Select all'-Checkbox -->
+        <div>
```
hwiem: Just curious: Was there a reason not to use `DpCheckbox` here?

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -177,23 +228,36 @@ export default {
...
+      const itemIds = Object.keys(this.items)
+      return itemIds.length > 0 && itemIds.every(id => this.currentPageSelections[id])
+    },
+
+    currentPageSelections () {
+      const toggledSet = new Set(this.toggledItems)
+      return Object.keys(this.items).reduce((acc, id) => {
+        const isInToggled = toggledSet.has(id)
+        acc[id] = this.trackDeselected ? !isInToggled : isInToggled
+        return acc
```
hwiem: nitpick
```suggestion
      const toggledSet = new Set(this.toggledItems)

      return Object.keys(this.items).reduce((acc, id) => {
        const isInToggled = toggledSet.has(id)

        acc[id] = this.trackDeselected ? !isInToggled : isInToggled

        return acc
```

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -216,30 +280,77 @@ export default {
...
+    addToSelection (id) {
+      this.toggledItems = [...this.toggledItems, id]
+      this.selectedUsersMap = { ...this.selectedUsersMap, [id]: this.items[id] }
+    },
+
+    buildUserFilter () {
+      const searchTerms = this.searchValue.split(' ').filter(Boolean)
+      if (searchTerms.length === 0) {
+        return {}
+      }
```
hwiem, line 290: nitpick
```suggestion
      const searchTerms = this.searchValue.split(' ').filter(Boolean)
      
      if (searchTerms.length === 0) {
        return {}
      }
```

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -270,57 +414,61 @@ export default {
...
       this.userList({
         page: {
           number: page ?? 1,
         },
-        filter: userFilter,
+        filter: this.buildUserFilter(),
         include: ['roles', 'orga', 'department', 'orga.allowedRoles'].join(),
       })
-        .then(() => {
+        .then((response) => {
```
hwiem: nitpick
```suggestion
        .then(response => {
```

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -270,57 +414,61 @@ export default {
...
-    hasOwnProp (obj, prop) {
-      return hasOwnProp(obj, prop)
+    async inviteItems () {
+      let ids
+      try {
+        ids = await this.resolveSelectedIds()
+      } catch (error) {
+        console.error('Failed to resolve selected user ids for invite:', error)
+        dplan.notify.notify('error', Translator.trans('error.api.generic'))
+        return
```
hwiem, line 441: nitpick
```suggestion
        dplan.notify.notify('error', Translator.trans('error.api.generic'))
        
        return
```

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -270,57 +414,61 @@ export default {
...
+      }
+      const form = this.$el.closest('form')
+      const currentPageIds = new Set(Object.keys(this.items))
+
+      // Add hidden inputs only for users not on the current page (those already have checkboxes)
```
hwiem, line 446: We have a helper function to add hidden form fields that could be used here: `addFormHiddenField` (`client/js/lib/core/libs/FormActions.js`)

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -270,57 +414,61 @@ export default {
...
-      this.searchValue = ''
+    handleSearch (term) {
+      this.searchValue = term
+      this.resetSelection()
       this.getFilteredItems()
     },
 
-    hasOwnProp (obj, prop) {
-      return hasOwnProp(obj, prop)
+    async inviteItems () {
```
hwiem: I think we should move from "items" to "users" in this component (I think the naming is currently inconsistent), but that's just a side note for the future :)

client/js/components/user/DpCreateItem.vue
```diff
@@ -17,43 +18,64 @@
...
+    >
+      <h3
+        class="mb-0 font-semibold"
+      >
+        {{ Translator.trans(itemTitle) }}
+      </h3>
+      <dp-button
+        variant="subtle"
+        :icon-after="isFormOpen ? 'caret-up' : 'caret-down'"
+        :text="Translator.trans('add')"
```
hwiem, line 33: I wonder if we should hide this button once the form is displayed. It's a bit confusing that the button says "add", but what happens when you click it is that the form is toggled (closed). But it's not a blocker for this PR.

client/js/components/user/DpCreateItem.vue
```diff
@@ -17,43 +18,64 @@
...
+          <component
+            v-bind="dynamicComponentProps"
+            :is="dynamicComponent"
+            ref="formFields"
+            @[dynamicEvent]="update"
+            @reset:complete="shouldResetForm = false"
+          />
+
+          <!-- Save/Abort buttons   -->
+          <div
```
hwiem: I think we could now use DpButtonRow again, because the button order and styling was adjusted there - but it's not a blocker.

client/js/components/user/DpCreateItem.vue
```diff
@@ -17,43 +18,64 @@
...
       </div>
-    </div>
-  </dp-accordion>
+    </dp-transition-expand>
+  </div>
 </template>
 
 <script>
-import { DpAccordion, DpButtonRow, dpValidateMixin } from '@demos-europe/demosplan-ui'
+import { DpButton, DpButtonRow, DpTransitionExpand, dpValidateMixin } from '@demos-europe/demosplan-ui'
```
hwiem, line 72: If we don't use DpButtonRow above, can this import be removed?

client/js/components/user/DpCreateItem.vue
```diff
@@ -325,8 +350,17 @@ export default {
         .finally(() => this.reset())
     },
 
-    toggleItem (open) {
-      this.isOpen = open
+    toggleForm () {
+      this.isFormOpen = !this.isFormOpen
+      if (this.isFormOpen) {
+        this.showHeaderBottomBorder = false
```
hwiem, line 353: nitpick
```suggestion
      this.isFormOpen = !this.isFormOpen
      
      if (this.isFormOpen) {
        this.showHeaderBottomBorder = false
```

client/js/components/user/DpUserList/DpUserList.vue
```diff
@@ -249,18 +360,51 @@ export default {
...
+      this.toggledItems = [...this.toggledItems, id]
+      this.selectedUsersMap = Object.fromEntries(
+        Object.entries(this.selectedUsersMap).filter(([key]) => key !== id),
+      )
+    },
+
+    async fetchAllUserIds () {
+      const filter = this.buildUserFilter()
+      const params = Object.keys(filter).length > 0 ? { filter } : {}
+      const url = Routing.generate('api_resource_list', { resourceType: 'AdministratableUser' })
```
hwiem: Can't we use sparse fieldsets here to only fetch the ids (as the method name indicates)? Or better yet, can we merge `fetchAllUserIds` and `fetchAllUsersForSelectionMap` into one method that fetches only `id`, `firstname`, `lastname` and `email` via sparse fieldsets? While this would mean fetching `firstname`, `lastname` and `email` even if the user doesn't toggle the "Auswahl anzeigen" content, no extra request would be necessary when they do. In general this would reduce request size - I am also worried about performance in case there are a lot of users in this list. I am not sure what numbers are realistic, but I think it's generally a good idea to optimize for performance.
I also noticed - and maybe that's an edge case we can ignore - that
when I select a user and toggle the "Auswahl anzeigen" content, then adjust `firstname`, `lastname` or `email` in the actual user list, that change is not reflected in the "Auswahl anzeigen" area. (Either way not a blocker)

hwiem (commented):
Most of my concerns are nitpicks that are optional, but I'd like to see if we can use sparse fieldsets to reduce the request size. Apart from that, I think the user list now looks a lot better. :)

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: Just removing `flex` put it below the search field in a weird way, but `inline-flex` put it right next to the search field, which I think is what you wanted. Have another look :slightly_smiling_face: : 4720ca250ce1352098c8e3ea19e27696da404547

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: I didn't know it was called yoda style :smile: 911f176e3ca66462b6cc396afd8dc31f9742eeb3

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: :see_no_evil:  911f176e3ca66462b6cc396afd8dc31f9742eeb3

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: changed it in 911f176e3ca66462b6cc396afd8dc31f9742eeb3

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: The reason is that my brain has "code that is already there" bias and it was a checkbox before when it only selected all on page... changed it 4720ca250ce1352098c8e3ea19e27696da404547

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: 911f176e3ca66462b6cc396afd8dc31f9742eeb3

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: 911f176e3ca66462b6cc396afd8dc31f9742eeb3

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: 911f176e3ca66462b6cc396afd8dc31f9742eeb3

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: 911f176e3ca66462b6cc396afd8dc31f9742eeb3

client/js/components/user/DpCreateItem.vue
meissnerdemos: Yes, this annoyed me too, but I thought making the button disappear would be weird UX also. I changed it so that the button becomes only a caret icon once the section is expanded. So you still have the button to collapse the section in addition to the "cancel" button. 827455fc562aa3ad8ffbed77983c2465afc6cae4

client/js/components/user/DpCreateItem.vue
meissnerdemos: 911f176e3ca66462b6cc396afd8dc31f9742eeb3

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: Thank you! changed it f86a2fdf5d464e6ad4cbe38604021468fadfd2b4

client/js/components/user/DpCreateItem.vue
meissnerdemos: yes :slightly_smiling_face: c20f40d88d27808904810cf736c1e6c1e528ddf0

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: I did it now :) b34507ad7fd180c27cbe6015ad16fcc8faef4723

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: made a small error (missed a `[]`), that I fixed in f99cad3d0a490e64db501d2a7b7fb747252245d9

client/js/components/user/DpUserList/DpUserList.vue
meissnerdemos: I refactored the two methods to one `fetchAllUsers` with sparse fieldsets f99cad3d0a490e64db501d2a7b7fb747252245d9

meissnerdemos: To the other point: I think its fine to leave this edge case uncovered. If I change the user, unselect it, then select it again, the new name shows. It just doesn't change _immediately_. I don't think it's worth it.

client/js/components/user/DpUserList/DpUserList.vue
hwiem: perfect! :heart:

client/js/components/user/DpCreateItem.vue
hwiem: Good idea, thanks :)

client/js/components/user/DpUserList/DpUserList.vue
hwiem: > If I change the user, unselect it, then select it again, the new name shows. It just doesn't change immediately. I don't think it's worth it.

Thanks for checking it, I didn't realize that. I agree that it's not worth fixing. :+1:

hwiem (approved):
Thanks for fixing all my concerns so meticulously ❤️ Lgtm + works

PR #6095, opened 2026-04-07
Add commands to list/detach organisations from customers

mussbach:
## Summary
- Add `dplan:customer:list-organisations` command to interactively list all orgas for a selected customer
- Add `dplan:customer:detach-organisation` command to remove orga-customer relationships without deleting the organisation
- Fix dangerous `cascade={"remove"}` on `OrgaStatusInCustomer` ManyToOne relations that would cascade-delete Orga, OrgaType, and Customer entities (separate commit, cherry-pickable)
- Both commands skip the default citizen organisation (Privatperson)
- Extracted shared `CustomerSelectionTrait` for interactive customer selection

## Test plan
- [x] 10 unit tests pass (4 list, 6 detach including citizen org protection)
- [ ] Manually verify `dplan:customer:list-organisations` shows correct orgas
- [ ] Manually verify `dplan:customer:detach-organisation` removes relationship without deleting the org
- [ ] Verify Privatperson is not listed or selectable in either command
- [ ] Verify graceful handling when only Privatperson exists for a customer

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Command/Data/DetachOrganisationFromCustomerCommand.php
+++ b/demosplan/DemosPlanCoreBundle/Command/Data/DetachOrganisationFromCustomerCommand.php
@@ -12,11 +12,11 @@ namespace demosplan\DemosPlanCoreBundle\Command\Data;
 
 use demosplan\DemosPlanCoreBundle\Command\CoreCommand;
 use demosplan\DemosPlanCoreBundle\Command\Helpers\CustomerSelectionTrait;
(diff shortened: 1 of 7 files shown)


bakridemos (approved):
as discussed the following adjustment are made :  
- Skip soft-deleted organisations in both list and detach commands                                                                                                                          
- Sort output alphabetically by organisation name                                                                                                                                           
- Show '(no name)' placeholder for orgs with empty/null names

code review and tested too, lgtm

PR #6108, opened 2026-04-09
fix(refs DPLAN-16987, DS-487): allow overriding demosplan-ui tailwind defaults

hwiem:
### Ticket
DPLAN-16987

The css properties used in demosplan-ui are now mapped to the scss variables used in the projects for overrides so that the overrides now actually take effect for demosplan-ui components that are using tailwind classes, like DpButton.
This is a temporary fix to restore interface consistency (without it, components like DpButton use the defaults defined in demosplan-ui, while buttons not using DpButton have the correct project styles).
Long-term, project theming should be handled directly by demosplan-ui tailwind config.

### How to review/test
Buttons and links should now always use the project colors instead of demosplan-ui defaults.

### PR Checklist

- [x] Run `yarn test`


sakutademos (approved):
lgtm, thx!

PR #6119, opened 2026-04-13
F dplan 17129 poc api library

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-17129/poc-api-library-ADO-50717
Description: 
This PR introduces API Platform as a second API framework alongside the existing EDT (Entity Definition Toolkit). The long-term goal is to gradually migrate all JSON:API endpoints from EDT to API Platform. This PR sets up only the infrastructure — no concrete API resources are exposed yet.                                                                                      
  
  What this PR does                                                                                                                                                                           
                  
  New dependencies

  - api-platform/core ^3.4 — The all-in-one API Platform metapackage. We use core instead of split packages (api-platform/symfony, api-platform/doctrine-orm, etc.) because the JSON:API      
  serializer classes (ApiPlatform\JsonApi\Serializer\EntrypointNormalizer, ItemNormalizer, CollectionNormalizer) only exist inside core. Split packages cause missing class errors at
  container compilation.                                                                                                                                                                      
  - nelmio/cors-bundle ^2.6 — CORS support for API Platform endpoints. Configured to allow localhost by default.
  - doctrine/persistence bumped from ^2.0 to ^2.0 || ^3.0 — Required by api-platform/doctrine-orm (bundled in core). This is safe because all other packages in the dependency tree           
  (doctrine/orm 2.20.9, doctrine/doctrine-bundle 2.13.3, etc.) already accept both v2 and v3. Only demosplan's own constraint was holding it back.                                            
  - demosplan-addon bumped to ^0.69 — The addon now includes three new classes for API Platform support:                                                                                      
    - PlainIdJsonApiNormalizer — Decorates API Platform's JSON:API item normalizer to strip IRI prefixes from id fields, returning plain UUIDs (550e8400-...) instead of IRIs                 
  (/api/3.0/Resource/550e8400-...). This is required for compatibility with EDT and the existing frontend, which both expect plain UUIDs. This class is permanent infrastructure.             
    - ApiPlatformRelationshipConfig — A bridge class that fakes being an EDT DoctrineResourceType while holding API Platform metadata (state provider, resource class, type name). This allows
   EDT resources to declare relationships to API Platform resources during the migration period. Temporary — will be removed when EDT is fully replaced.                                      
    - ExtendedDynamicTransformer — Extends EDT's DynamicTransformer to intercept relationship serialization and route API Platform relationships through their StateProviders instead of EDT.
  Temporary — will be removed when EDT is fully replaced.                                                                                                                                     
                  
  Bundle registration (config/bundles.php)                                                                                                                                                    
                  
  ApiPlatformBundle and NelmioCorsBundle are registered. The file was also reformatted by Symfony Flex (whitespace alignment removed).                                                        
  
  API Platform configuration (config/packages/api_platform.yaml)                                                                                                                              
                  
  - mapping.paths points to demosplan/DemosPlanCoreBundle/ApiResources — API Platform scans this directory for #[ApiResource] DTO classes. The directory is currently empty (.gitkeep), ready 
  for future resources.
  - formats: [jsonld, json, jsonapi] — JSON:API format (application/vnd.api+json) is enabled to stay consistent with EDT's response format.                                                   
  - use_symfony_listeners: false — Prevents API Platform from registering its own Symfony event listeners, which would conflict with DemosPlan's existing listeners (e.g.,                    
  AccessProcedureListener, ConfigurePermissionsListener).                                                                                                                                     
  - defaults.stateless: false — API Platform defaults to stateless mode, which prevents session usage. DemosPlan uses session-based authentication, so this must be false. Without it, every  
  request to an API Platform endpoint fails with "Session was used while the request was declared stateless".                                                                                 
                  
  Route registration (config/routes/api_platform.yaml)                                                                                                                                        
                  
  API Platform routes are registered under the /api prefix. Concrete resources will define their own routePrefix: '/3.0', so endpoints will be available at /api/3.0/{ResourceName}, clearly  
  separated from EDT's /api/2.0/ namespace.
                                                                                                                                                                                              
  Security firewall (config/packages/security.yaml)

  A new api_platform firewall is added before the existing api firewall:                                                                                                                      
  
  api_platform:                                                                                                                                                                               
      pattern: ^/api/3\.0
      context: main
      provider: dplan_default_user_provider
      lazy: true
      custom_authenticators: '%demosplan_user_authenticator_service%'
      entry_point: '%demosplan_user_authenticator_service_entry_point%'                                                                                                                       
  
  This is necessary because API Platform routes (/api/3.0/...) would otherwise be caught by the existing api firewall (pattern: ^/api/\d.+), which is stateless and expects JWT               
  authentication. Since we need session-based authentication (same as the main app), the dedicated firewall intercepts these routes first.
                                                                                                                                                                                              
  context: main is critical — it shares the security session with the main firewall, so users already logged in via the normal UI are automatically authenticated for API Platform endpoints  
  without requiring a separate login.
                                                                                                                                                                                              
  Service overrides (config/services.yml)

  Two service registrations are added:                                                                                                                                                        
  
  1. api_platform.openapi.command — Overrides the class to ApiPlatform\OpenApi\Command\OpenApiCommand. This is a workaround for a bug in api-platform/symfony 3.4.17: the bundle's openapi.xml
   registers this command service pointing to the old namespace ApiPlatform\Symfony\Bundle\Command\OpenApiCommand, which is a stub file. The real class was moved to the api-platform/openapi
  sub-package. Without this override, DebugClassLoader throws on every bin/console command in dev mode.                                                                                       
  2. PlainIdJsonApiNormalizer — Registered as a decorator for api_platform.jsonapi.normalizer.item, wiring the addon's normalizer to intercept all JSON:API item serialization.
                                                                                                                                                                                              
  AccessProcedureListener fix (AccessProcedureListener.php)
                                                                                                                                                                                              
  Added an is_array() check before accessing $controllerEvent->getController()[0]. Standard Symfony controllers return an array [ControllerInstance, 'methodName'], but API Platform's        
  MainController returns a callable object. Without this check, every API Platform request crashes with a type error.
                                                                                                                                                                                              
  CORS configuration (.env, config/packages/nelmio_cors.yaml)

  Default CORS configuration generated by Symfony Flex. Allows localhost origins, standard HTTP methods, and Content-Type/Authorization headers.                                              
  
  Test configuration (phpunit.xml)                                                                                                                                                            
                  
  CORS_ALLOW_ORIGIN environment variable added (auto-generated by Symfony Flex).                                                                                                              
  
  What this PR does NOT do                                                                                                                                                                    
                  
  - No concrete API resources are exposed yet — the ApiResources directory is empty
  - No EDT resources are migrated or modified
  - No frontend changes                                                                                                                                                                       
  - No bridge wiring between EDT and API Platform (the bridge classes exist in the addon but are not connected yet)
                                                                                                                                                                                              
  Next steps      
                                                                                                                                                                                              
  1. Create the first API Platform resource (e.g., AdminProcedureResource with StateProvider and StateProcessor)                                                                              
  2. Wire the EDT-to-API Platform bridge for cross-system relationships (Claim resource)
  3. Update frontend to consume /api/3.0/ endpoints                                                                                                                                           
  4. Add filtering, sorting, and pagination to API Platform providers                                                                                                                         
                                                                                                                                                                                              
  Test plan                                                                                                                                                                                   
                  
  - Application boots without errors (bin/console commands work)                                                                                                                              
  - Existing EDT endpoints (/api/2.0/...) continue to work unchanged                                                                                                                   
  - No regressions in existing functionality (session auth, permissions, procedure access)

- [x] Update changelog


config/packages/api_platform.yaml
```diff
@@ -0,0 +1,23 @@
+api_platform:
+    title: DemosPlan API
```
mussbach:
```suggestion
    title: demosplan API
```

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ## UNRELEASED
 
+## v4.34.0 (2026-04-13)
+-  introduces API Platform as a second API framework alongside the existing EDT (Entity Definition Toolkit). The long-term goal is to gradually migrate all JSON:API endpoints from EDT to API Platform. This PR sets up only the infrastructure — no concrete API resources are exposed yet.
```
mussbach: you should add that new env CORS_ALLOW_ORIGIN is needed

config/services.yml
```diff
@@ -1392,3 +1392,19 @@ services:
     DemosEurope\DemosplanAddon\Contracts\Services\SegmentServiceInterface: '@demosplan\DemosPlanCoreBundle\Logic\Segment\SegmentService'
 
     DemosEurope\DemosplanAddon\Contracts\Services\SegmentTransformerInterface: '@demosplan\DemosPlanCoreBundle\Transformers\Segment\DraftsInfoToSegmentTransformer'
+
+    # API Platform: Override buggy service definition from api-platform/symfony 3.4.17.
```
mussbach: Is this still needed then?

composer.json
```diff
@@ -36,14 +36,14 @@
...
-    "demos-europe/demosplan-addon": "^0.68",
+    "demos-europe/demosplan-addon": "^0.69",
     "doctrine/annotations": "^2.0",
     "doctrine/common": "^3.2",
     "doctrine/doctrine-bundle": "^2",
     "doctrine/doctrine-fixtures-bundle": "^3",
     "doctrine/doctrine-migrations-bundle": "^3.2",
     "doctrine/orm": "^2.0",
-    "doctrine/persistence": "^2.0",
+    "doctrine/persistence": "^2.0 || ^3.0",
```
mussbach, line 49: is dplan compatible with ^3.0?

bakridemos: Yes, it's safe. The bump from ^2.0 to ^2.0 || ^3.0 is needed because api-platform/doctrine-orm (bundled inside api-platform/core) requires doctrine/persistence ^3.2.
                                                                                                                                                                                              
  The key point is that demosplan's own composer.json was the only thing holding it back. Every other package in the dependency tree already accepts both v2 and v3:
                                                                                                                                                                                              
  - doctrine/orm 2.20.9 requires doctrine/persistence ^2.4 || ^3                                                                                                                              
  - doctrine/doctrine-bundle 2.13.3 requires doctrine/persistence ^2.2 || ^3                                                                                                                  
  - doctrine/doctrine-migrations-bundle requires doctrine/persistence ^2.2 || ^3                                                                                                              
  - stof/doctrine-extensions-bundle requires doctrine/persistence ^2.2 || ^3                                                                                                                  
   
  Composer resolves the actual installed version based on all constraints combined. With ^2.0 || ^3.0, Composer will pick the version that satisfies all packages — in practice it resolves to a v3.x version since api-platform/doctrine-orm requires ^3.2. The || keeps backward compatibility in case the constraint is ever needed without API Platform.
                                                                                                                                                                                              
  The doctrine/persistence v3 release was a minor cleanup — it removed some deprecated interfaces and tightened type hints, but the public API that demosplan uses (ObjectManager,            
  ObjectRepository, ManagerRegistry) is unchanged between v2 and v3. doctrine/orm 2.20.x was specifically designed to work with both versions.

config/packages/api_platform.yaml
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/6119/changes/777d7a507344c233c94bee1912a3e5ec5d474e44

config/services.yml
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/6119/changes/498edff94c4d0b02bf60d3610385ee1ea162f1ed. The problem last week was likely during the phase when i was using split packages (api-platform/symfony + api-platform/doctrine-orm) instead of api-platform/core. With split packages, the real class file (src/OpenApi/Command/OpenApiCommand.php with the class_alias on line 89) wasn't installed at all — only the stub existed.                                                                                                                                                             
                                                                                                                                                                                              
  Once i switched to api-platform/core (which contains everything), the real class with the alias came along. So the override became unnecessary at this point, but it was never removed because it was already working.

CHANGELOG.md
bakridemos: thx done in https://github.com/demos-europe/demosplan-core/pull/6119/changes/7d081b68d517eb8650efda525b55e7e12535ce97

mussbach: This is a bit too verbose....

composer.json
mussbach: This is not true imho. With a composer install doctrine/persistence is bumped to version 3 which is a major change. This might work, but as it is a major change we should separate it from this PR and do it beforehand and test it

sierrazamudiodemos (approved):
thx for this

composer.json
bakridemos: this done in a seperat pr https://github.com/demos-europe/demosplan-core/pull/6270

PR #6133, opened 2026-04-14
DPLAN-17244: add filter capability to sourceEntity and targetEntity

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17244/ADO-50218-Konfigurierbare-Felder-in-Stellungnahmen-Part-4.1-Custom-Fields-fur-Abschnitte-und-Statements-zusammenfuhren-Bau-Rog

**Until now** a procedure had custom fields for either `Statement` or `Segment`, never both. Querying by `sourceEntityId` alone was enough.

**Now** a procedure can have custom fields for both `Statement` and `Segment` simultaneously, and more source types beyond `Procedure` are coming — meaning multiple rows can share the same `sourceEntityId`. Without filtering by `targetEntity` and `sourceEntity`, a query for `Statement` custom fields would also return `Segment` ones.

**What changed**

- Added `setFilterable()` to `targetEntity` and `sourceEntity` in `CustomFieldResourceType`
- Added `setAliasedPath` for both since the API names (`targetEntity`, `sourceEntity`) differ from the entity property names (`targetEntityClass`, `sourceEntityClass`) — without this the DQL query fails
- Added a composite index `(source_entity_id, source_entity_class, target_entity_class)` on the table — since every lookup filters by all three columns, without an index MySQL would scan the entire table on each request


demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
```diff
@@ -123,8 +123,8 @@ protected function getResourceConfig(): ResourceConfigInterface
             ->addPathCreationBehavior()
             ->addPathUpdateBehavior();
         $configBuilder->description->setReadableByPath()->addPathCreationBehavior()->addPathUpdateBehavior();
-        $configBuilder->targetEntity->addPathCreationBehavior()->setReadableByPath();
-        $configBuilder->sourceEntity->addPathCreationBehavior();
+        $configBuilder->targetEntity->setAliasedPath(['targetEntityClass'])->addPathCreationBehavior()->setReadableByPath()->setFilterable();
+        $configBuilder->sourceEntity->setAliasedPath(['sourceEntityClass'])->addPathCreationBehavior()->setFilterable();
```
breierDemos: sourceEntity has setFilterable() but no setReadableByPath(), meaning it can be filtered by but not returned in API responses. Is this intentional?

breierDemos (changes requested):
Hey I have two questions.

The logic in CustomFieldConfigurationRepository::copy() (line 99) hardcodes targetEntity = 'SEGMENT':

  $customFieldsConfigurations = $this->findCustomFieldConfigurationByCriteria(
      'PROCEDURE_TEMPLATE',                                                                                                                                                                
      $sourceProcedureId,
      'SEGMENT' // only copies Segment custom fields                                                                                                                                       
  );                         

  With this PR enabling procedures to have custom fields for both Statement and Segment simultaneously, this method still only copies SEGMENT configurations. Is this intentional, or      
  should Statement custom fields also be copied when applying a procedure template?

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
sierrazamudiodemos: changed

sierrazamudiodemos:
> Hey I have two questions.
> 
> The logic in CustomFieldConfigurationRepository::copy() (line 99) hardcodes targetEntity = 'SEGMENT':
> 
> $customFieldsConfigurations = $this->findCustomFieldConfigurationByCriteria( 'PROCEDURE_TEMPLATE', $sourceProcedureId, 'SEGMENT' // only copies Segment custom fields );
> 
> With this PR enabling procedures to have custom fields for both Statement and Segment simultaneously, this method still only copies SEGMENT configurations. Is this intentional, or should Statement custom fields also be copied when applying a procedure template?

good catch... i will adjust and let u know

sierrazamudiodemos:
> > Hey I have two questions.
> > The logic in CustomFieldConfigurationRepository::copy() (line 99) hardcodes targetEntity = 'SEGMENT':
> > $customFieldsConfigurations = $this->findCustomFieldConfigurationByCriteria( 'PROCEDURE_TEMPLATE', $sourceProcedureId, 'SEGMENT' // only copies Segment custom fields );
> > With this PR enabling procedures to have custom fields for both Statement and Segment simultaneously, this method still only copies SEGMENT configurations. Is this intentional, or should Statement custom fields also be copied when applying a procedure template?
> 
> good catch... i will adjust and let u know

@breierDemos I added a couple of things as well, to make sure it it possible to allow to create customFields on procedure to for STATEMENT and SEGEMENTS

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
```diff
@@ -114,11 +115,22 @@ protected function getResourceConfig(): ResourceConfigInterface
 
         $configBuilder->id->setReadableByPath();
         $configBuilder->name->setReadableByPath(DefaultField::YES)->addPathCreationBehavior()->addPathUpdateBehavior();
-        $configBuilder->isRequired->setReadableByPath()->addPathCreationBehavior(OptionalField::YES)->addPathUpdateBehavior();
+        $configBuilder->isRequired
+            ->setReadableByCallable(
+                static fn (CustomFieldInterface $customField): ?bool =>
+                    (new ReflectionClass($customField))->hasProperty('isRequired')
```
breierDemos: CustomFieldInterface doesn't declare getRequired() / getOptions(), so reflection is used to probe for the properties at   
  runtime. This is a sign the interface contract is incomplete rather than a legitimate use of reflection.                  
   
  The approach is fragile — renaming $isRequired or $options in any concrete class silently breaks serialization with no    
  static analysis warning.
                                                                                                                            
  Suggestion: Add getRequired(): ?bool and getOptions(): array to CustomFieldInterface, or introduce sub-interfaces         
  (CustomFieldWithOptionsInterface etc.) and use instanceof. Either approach expresses the capability distinction in the
  type system where it belongs.

breierDemos (changes requested):
Thank you. I have one thing that I think we should adjust:

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
sierrazamudiodemos: totally makes sense

sierrazamudiodemos: fixed:
9474571847aae429a8557b73da1865c3364ca94e
1d283bd69f82c4f6486375e87fe611004d8bbe24

breierDemos (approved):
lgtm

PR #6137, opened 2026-04-14
feat: GDPR audit logging for personal data changes

mussbach:
## Summary
- Adds automatic audit trail for changes to personal data entities (User, Orga, Address, StatementMeta, DraftStatement, ProcedurePerson)
- Doctrine onFlush/postFlush listener writes changes via DBAL to avoid ORM recursion
- Sensitive fields (passwords) masked as `***`, null-to-empty-string changes skipped
- Listener disabled during GDPR data removal to prevent noise entries
- RemoveUserDataCommand anonymizes both userName and stored pre/post values

## Test plan
- [ ] Run `PersonalDataAuditListenerTest` (5 tests: email change, password masking, blank skip, untracked field, disable toggle)
- [ ] Verify existing `EntityContentChange` tests still pass
- [ ] Run migration on dev DB, change a user email, verify row in `personal_data_audit_log`
- [ ] Change submitter name on a Stellungnahme, verify audit entry for StatementMeta


demosplan/DemosPlanCoreBundle/Command/Data/RemoveUserDataCommand.php
```diff
@@ -150,7 +153,10 @@ public function configure(): void
     public function execute(InputInterface $input, OutputInterface $output): int
     {
         $this->output = $output;
-        $this->totalProgress = $this->initializeProgressBar(22, 22);
+        $this->totalProgress = $this->initializeProgressBar(23, 23);
+
```
sierrazamudiodemos: - I think 23 is a magic number
- and 23 steps is not accurate right now. I think there are 21 steps

I was wondering if we make it more generic, and rather put like a number not related to the amount if steps, and we can put it under a const

Something like this:
```
private const PROGRESS_BAR_WIDTH = 40;
  $this->initializeProgressBar(count($this->getWipeSteps()), self::PROGRESS_BAR_WIDTH);
```

what do you think?

demosplan/DemosPlanCoreBundle/Command/Data/RemoveUserDataCommand.php
sierrazamudiodemos: I know this is not related to this PR, but taken advantage of touching this, maybe we could also move this as const?

demosplan/DemosPlanCoreBundle/Entity/PersonalDataAuditLog.php
```diff
@@ -0,0 +1,237 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of the package demosplan.
+ *
```
sierrazamudiodemos, line 7: it should use attributes and not comments

sierrazamudiodemos:
it looks good because it is isolated, but I am concern that this opens another way to make an audit. and I wonder if we could actually make some parts of this general... or adjust it in a way that not every entity has its way of doing audit?
I am putting the question to the table and would like to hear ur opinion

demosplan/DemosPlanCoreBundle/Entity/PersonalDataAuditLog.php
mussbach: [c0c187d](https://github.com/demos-europe/demosplan-core/pull/6137/commits/c0c187d8eed78afa461215930dcdbf6be9d0ebb0)

demosplan/DemosPlanCoreBundle/Command/Data/RemoveUserDataCommand.php
mussbach: yes [c0c187d](https://github.com/demos-europe/demosplan-core/pull/6137/commits/c0c187d8eed78afa461215930dcdbf6be9d0ebb0)

demosplan/DemosPlanCoreBundle/Command/Data/RemoveUserDataCommand.php
mussbach: indeed, thank you [fa11d9e](https://github.com/demos-europe/demosplan-core/pull/6137/commits/fa11d9e70a6e188dfddc6fed6b5b99db8c588f2e)

mussbach:
> it looks good because it is isolated, but I am concern that this opens another way to make an audit. and I wonder if we could actually make some parts of this general... or adjust it in a way that not every entity has its way of doing audit? I am putting the question to the table and would like to hear ur opinion

Good question. These are intentionally two mechanisms with different jobs rather than two ad-hoc audits:
- EntityContentChange — user-facing history/rollback, invoked explicitly from services, permission-gated, shown in the UI. Domain feature.
- PersonalDataAuditLog (this PR) — automatic GDPR compliance trail via a Doctrine onFlush listener, written through DBAL to avoid ORM recursion, no UI/rollback. Infrastructure concern.

Folding compliance logging into the versioning feature would couple it to that feature's permissions and UI, which we don't want. It's also now consistent with PersonalDataLogEntry that just landed on main — its own docblock calls it the "sister mechanism to PersonalDataAuditLog", so the codebase is deliberately treating these as a small family of complementary log types, not one generic audit.

sierrazamudiodemos:
@mussbach but do we want to go generic audit? or each entity has its own audit?

mussbach:
@sierrazamudiodemos I'd suggest that we have a distinct audit for PII/Gdpr-Data (this one) and then decide from case to case how other entities are logged. Might be something generic like doctrine loggable or something generic, we need to discuss that in detail

sierrazamudiodemos:
@mussbach oki.. the thing is that this challenge of auditing is coming already now for another entity: procedure phases audit challenge.. and we were facing 3 alternatives:
1. use a new audit log entity (like the GDPR)
2. use existing report_entry
3. use existing entity_content_change
I am up for the3 option, as it seems that it was built for generic purposes as well. I understood this PR more like related to GDPR and that is why it makes sense to isolated. but I guess existing entities can already use existing audti mechnism

mussbach:
@sierrazamudiodemos it depends on where the information is needed. 
1) is not needed as it is no pii
2) in case it should be in procedure report and be exportable via pdf as an event. This fits best in this case when i understand it correctly, as planners need to prove that and when a phase changed
3) in case it should appear in the version history of a segment

But let us discuss this better in person

sierrazamudiodemos:
@mussbach yep, so the changes of the vefahrneshit in the procedure ar logged in the report_entry.
but the changes it self of the verfahrenshit: for example name, or permission, where should it be stored? if in report_entry, then it is for all procedures. or rather in the entity_content change

breierDemos:
@mussbach @sierrazamudiodemos I made a pr for procedure phase definitions here: [#6396](https://github.com/demos-europe/demosplan-core/pull/6396). I save only on report entry per change, but read it on every procedure report list.

PR #1474, opened 2023-06-05
Fix Language for PhpWord generated Docx exports

graupnerdemos:
**Ticket:** https://yaits.demos-deutschland.de/T31218

This consolidates the language settings to be used on more PhpWord
    instances throughout the codebase. In addition, documents are
    generated with a clean proof read and grammar state to improve
    loading performance in MS Word.

### PR Checklist

Delete the checkbox if it doesn't apply/isn't necessary.

- [x] Link all relevant tickets
- [x] Move the tickets on the board accordingly

PR #6143, opened 2026-04-15
feat(refs: DPLAN-17244): enable multiple custom field types on same project

huellnerdemos:
### Ticket
DPLAN-17244

This PR enables the creation and use of multiple custom field types in the same project. The different dialogs call the custom fields depending on their context.

### How to review/test

Admin:
1. If both field types are available in a project, create a procedure and create some Segment and some Statement custom fields in the custom field dialog of the procedure.
2. It should only be possible to create single select fields for Segments and Multiselect fields for Statements
3. If the first statement was created for the procedure, statement fields can't be selected anymore as the target context. The statement fields are also not editable from this moment on.
4. The custom fields should be listed under their corresponding tab depending on their target context
5. Check the custom fields in assessment table if available and also check it in original statement
6. In assessment table there should be a stylistic separating gray border left to every custom field 
7. Create segments if possible and check if custom fields are still settable on them

Private User and Guest:
1. Create Statements and Draft Statements to check if Statement Custom Fields are shown and usable
2. Check if there are any Segment Custom fields in a Statement Custom Field context (usually Statement and Segment Custom Fields should not be found together in one dialog, expect the creation dialog)
3. Also check if there are any Statement Fields in a Segment context 

### Linked PRs
demosplan-ui PR: [#1474](https://github.com/demos-europe/demosplan-ui/pull/1474)

### PR Checklist

- [x] Run `yarn lint`
- [x] Update changelog


riechedemos:
Thank you! Overall is seem to work fine am still testing, but when you create custom field, it will be display in the tab that was opened while creating. E.g here tab segments was open then I created a statement custom field and the content of the statement tab is displayed is segment tab. 
display

huellnerdemos:
> Thank you! Overall is seem to work fine am still testing, but when you create custom field, it will be display in the tab that was opened while creating. E.g here tab segments was open then I created a statement custom field and the content of the statement tab is displayed is segment tab. display

Thanks for this Point! :pray:  I will change this.

client/js/composables/useCustomFieldDefinitions.js
```diff
@@ -0,0 +1,240 @@
...
+      return pendingFetches.get(cacheKey)
+    }
+
+    const url = Routing.generate('api_resource_list', {
+      resourceType: 'CustomField',
+    })
+
+    const params = {
+      fields: {
+        CustomField: ['name', 'description', 'options', 'fieldType', 'isRequired'].join(),
```
riechedemos, line 98: The fields list for CustomField includes name, description, options, fieldType, isRequired, but not targetEntity. Since the Statement/Segment tab assignment can now differ per field, targetEntity needs to be part of the response for the list to categorize correctly. Could this be the root cause of the display bug I reported (field shows up in the tab that was open during creation)?

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -359,20 +368,27 @@ export default {
...
+    targetOptions: {
+      type: Object,
+      required: true,
+    },
   },
 
   data () {
     return {
+      activeTabId: Object.keys(this.targetOptions ?? {})[0] || '',
+      allDefinitions: [],
```
riechedemos: allDefinitions is misleading — this array only contains the definitions for the currently active tab (since the server filters by targetEntity). A name like currentTabDefinitions would better reflect the actual scope and prevent assumptions about cross-tab data.

client/js/components/customFields/SingleselectCustomField.vue
```diff
@@ -38,39 +38,40 @@ All rights reserved
...
     <dp-label
+      v-if="showLabel"
       :class="prefixClass('mb-2')"
       :for="`custom-field-${field.id}`"
       :required="field.attributes.isRequired"
       :text="field.attributes.name"
       :tooltip="field.attributes.description || ''"
     />
-    <dp-select
+    <dp-multiselect
```
riechedemos, line 52: SingleselectCustomField now uses DpMultiselect instead of DpSelect. Could you add a short comment explaining why? The component name suggests a single-select widget, but the implementation uses a multi-select component, which is confusing without context.

client/js/components/customFields/CustomFieldsList.vue
```diff
@@ -440,6 +484,14 @@ export default {
         })
     },
 
+    findDefinitionForField (fieldId) {
```
riechedemos: This doesn't seem to be used anywhere, could we get rid of it?

client/js/components/customFields/CustomFieldsList.vue
```diff
@@ -24,41 +24,47 @@ All rights reserved
...
+          :definitions="definitions"
           :enable-toggle="enableToggle"
           :fields="fieldsToRender"
+          :fields-with-definitions="fieldsWithDefinitions"
           :mode="mode"
           :resource-id="resourceId"
           :resource-type="resourceType"
         >
           <div :class="layoutClasses">
-            <custom-field
```
riechedemos, line 34: This PR had to update three near-identical <custom-field> render blocks in parallel (one inside dp-details, one in fieldset, one standalone). The duplication isn't introduced here, but it's worth flagging — a future extraction into a reusable slot or sub-component would reduce the risk of inconsistency.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -962,36 +916,33 @@ export default {
...
-
-          this.toggleAssignableUsersSelect()
-          this.$nextTick(() => {
-            if (this.$refs.recommendationContainer) {
-              this.$refs.imageModal.addClickListener(this.$refs.recommendationContainer.querySelectorAll('img'))
-            }
+          /*
+           * Custom fields are saved via a separate PATCH using the composable's updateCustomFields,
+           * which bypasses the vuex-json-api diff mechanism (unreliable for array attributes)
+           * and properly invalidates the composable's value cache.
```
riechedemos, line 963: The save flow now issues two sequential PATCHes (segment + custom fields) instead of one. If the segment save succeeds but the custom fields save fails, the state becomes inconsistent — the segment is partially updated with no rollback. The code comment explains the reason (vuex-json-api unreliable for array attributes), but could we at least flag this in error handling (e.g. specific error message telling the user custom fields weren't saved)?

templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/includes/listentry_content.html.twig
```diff
@@ -149,6 +149,7 @@
         mode="readonly"
         resource-id="{{ statement.ident }}"
         resource-type="DraftStatement"
+        target-entity="STATEMENT"
```
riechedemos, line 153: listentry_content.html.twig only passes target-entity="STATEMENT" but not source-entity, while every other <custom-fields-list> consumer in this PR passes both. Is this intentional, or should it be source-entity="PROCEDURE" for consistency? Without it, the composable omits the sourceEntity filter, which could surface fields from other source entities (e.g. procedure templates).

riechedemos (changes requested):
This looks really good and is working for fest and rog.

I can create and edit custom fields (segment and statement). I can delete segments custom field - everything is happening in one place. All fields are being displayed. But if I assign a custom field to a segments the hover grey box is only updated after a refresh.

client/js/composables/useCustomFieldDefinitions.js
huellnerdemos: Thanks for the suggestion! The fetch in AdministrationCustomFieldsList already passes targetEntity directly to the server-side filter, so the response only ever contains fields for the active tab. No client-side categorization by targetEntity is needed. 

After investigating the display bug: the actual root cause is in DpTabs. It manages its own internal activeTabId ref initialized once in onMounted, with no watch on the activeId prop. When this.activeTabId = targetEntity is set programmatically after creation, the content area switches correctly (via v-if in the parent), but DpTabs' internal state stays on the old tab, so the tab header still highlights the wrong tab. This needs to be fixed in demosplan-ui and i will have a look at how much effort that would be, maybe there is a quick solution for that.

client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
huellnerdemos: Thanks for this Point! Changed it in 1e59a749a2cb2396f2e426315614ae79281309ac

client/js/components/customFields/SingleselectCustomField.vue
huellnerdemos: You are right, an explanation could be helpful here. Thank you! Added a comment in d6747d18247704be779ea885ae36303c1fab51aa

client/js/components/customFields/CustomFieldsList.vue
huellnerdemos: Oh that's right, i didn't see that. Thanks for this Point. Deleted the method in fcf3c8f4972a31e9f791fc916de2386a9cf64226

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
huellnerdemos: Thanks for that! I made a little change, if the custom fields PATCH fails after a successful segment save, the user now sees a specific error message listing the affected field names. Changed it in be27152092eff1731c8397fe6fe94764d9123ccb

templates/bundles/DemosPlanCoreBundle/DemosPlanStatement/includes/listentry_content.html.twig
huellnerdemos: Oh, i think i forgot it at this place :hushed:  Thanks for this point! Added it in 45fc4f57625866ca31dd98ad6ca55b26a238e417

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
huellnerdemos: I had to adjust this a bit because of a sonarcloud nesting error. adjusted it in 558f19319e3fb3ce6816994f0c6f0eef492ee865

riechedemos:
Thank you for the fixes :)
Now you just need to update your outdated branch and fix the DpTabs issue in demosplan-ui.

huellnerdemos:
> Thank you for the fixes :) Now you just need to update your outdated branch and fix the DpTabs issue in demosplan-ui.

Thanks for your review! I created a separate PR for the demosplan-ui issue: [1474](https://github.com/demos-europe/demosplan-ui/pull/1474)

client/js/components/customFields/CustomFieldsList.vue
huellnerdemos: Thanks for this point! You are right, as long as the content is the same, this can be optimized. Tried to optimize it a bit by reducing the repeating component entries by one, so that there are only two left. Did that change in 7386948f2164af393f72b006231dd0e36bc2d644

riechedemos (approved):
yesss, you did it  🎉

huellnerdemos:
> yesss, you did it 🎉

Hey thanks for your review! :tada: :confetti_ball: :balloon:

PR #6146, opened 2026-04-15
DPLAN-17244 Display custom fields in pdf export in draft statements

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17244/ADO-50218-Konfigurierbare-Felder-in-Stellungnahmen-Part-4.1-Custom-Fields-fur-Abschnitte-und-Statements-zusammenfuhren-Bau-Rog

**Description**
- Custom fields filled in on a statement (e.g. radio buttons, checkboxes) now appear in the PDF export. 
-   A new CustomFieldDisplayResolver service was added to translate stored option IDs into human-readable labels  each field type is responsible for formatting its own value. The PDFtemplates were updated to render those labels, and a warning is logged if a field definition no longer exists in the database.

### How to review/test
- As a FP Admin create a new Verfahren and add custom fields for stn
- As a private citzen go to that Verfarehn and add stn (put it on draft)
- And then download the PDF, the chosen custom fields should be there

### DIFF
--- a/demosplan/DemosPlanCoreBundle/CustomField/RadioButtonField.php
+++ b/demosplan/DemosPlanCoreBundle/CustomField/RadioButtonField.php
@@ -84,7 +84,7 @@ class RadioButtonField extends AbstractCustomField
             return '';
         }
 
-        return $this->getCustomOptionValueById($value)?->getLabel() ?? $value;
+        return $this->getCustomOptionValueById($value)?->getLabel() ?? '';
     }
 
     protected function validateFieldSpecific(array $options): void
--- a/demosplan/DemosPlanCoreBundle/CustomField/RadioButtonField.php
+++ b/demosplan/DemosPlanCoreBundle/CustomField/RadioButtonField.php
@@ -78,6 +78,15 @@ class RadioButtonField extends AbstractCustomField
         return null;
     }
 
+    public function formatValueForDisplay(mixed $value): string
+    {
+        if (!is_string($value) || '' === $value) {
+            return '';
+        }
+
+        return $this->getCustomOptionValueById($value)?->getLabel() ?? $value;
+    }
+
     protected function validateFieldSpecific(array $options): void
     {
         if (count($options) < 2) {
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/DraftStatementService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/DraftStatementService.php
@@ -873,16 +873,7 @@ class DraftStatementService
                 $filenameSuffix = $this->translator->trans('export.filenames.statement.single.suffix');
                 // Die Stellungnahme kann nicht übergeben werden, rufe sie ab
                 $statement = $this->getSingleDraftStatement($itemsToExport[0]);
-                if ($this->currentUser->hasPermission('feature_statements_custom_fields')
-                    && $statement['customFields'] instanceof CustomFieldValuesList
-                ) {
-                    $statement['resolvedCustomFields'] = $this->customFieldDisplayResolver->resolveForDisplay(
-                        $statement['customFields'],
-                        CustomFieldSupportedEntity::procedure,
-                        $procedureId,
-                        CustomFieldSupportedEntity::statement
-                    );
-                }
+                $statement = $this->attachResolvedCustomFields($statement, $procedureId);
                 $templateVars['statement'] = $statement;
                 $itemsToExport = null;
                 break;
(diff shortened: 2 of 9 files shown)


mussbach (commented):
Nice separation of concerns — each field type owning its own formatting is the right call. A few things worth addressing before merge, inline.

demosplan/DemosPlanCoreBundle/CustomField/RadioButtonField.php
```diff
@@ -78,6 +78,15 @@ public function getCustomOptionValueById(string $customFieldOptionValueId): ?Cus
         return null;
     }
 
+    public function formatValueForDisplay(mixed $value): string
+    {
+        if (!is_string($value) || '' === $value) {
+            return '';
+        }
+
+        return $this->getCustomOptionValueById($value)?->getLabel() ?? $value;
```
mussbach: Inconsistent fallback vs. `MultiSelectField::formatValueForDisplay`: multi-select silently drops unresolvable option IDs, radio returns the raw UUID (`?? $value`). End-users will see garbage-looking identifiers in the PDF when an option has been deleted. Align both — probably drop the `?? $value` here and return `''` (or a translated placeholder), matching the multi-select behavior.

demosplan/DemosPlanCoreBundle/Logic/Statement/DraftStatementService.php
```diff
@@ -2157,6 +2168,17 @@ function (array $statementArray) use ($procedureId) {
                 $statementArray['documentlist'] = $this->paragraphService
                     ->getParaDocumentObjectList($procedureId, $statementArray['elementId']);
 
+                if ($this->currentUser->hasPermission('feature_statements_custom_fields')
+                    && $statementArray['customFields'] instanceof CustomFieldValuesList
+                ) {
+                    $statementArray['resolvedCustomFields'] = $this->customFieldDisplayResolver->resolveForDisplay(
```
mussbach: N+1: this runs inside the per-statement `map()` in `addContentToStatementsArrayForPdfProcess`, and `resolveForDisplay` re-fetches `getCustomFieldsByCriteria` every call. `sourceEntityId` (procedureId) and `targetEntity` are constant across the list, so for a 200-statement export that's 200 identical DB queries. Hoist the definition lookup out of the loop (pass the collection in) or memoize it in `CustomFieldProvider`/the resolver keyed on `(sourceEntityId, targetEntity)`.

demosplan/DemosPlanCoreBundle/Logic/Statement/DraftStatementService.php
```diff
@@ -870,9 +872,18 @@ public function generatePdf($draftStatementList, $type, $procedureId, $itemsToEx
 
                 $filenameSuffix = $this->translator->trans('export.filenames.statement.single.suffix');
                 // Die Stellungnahme kann nicht übergeben werden, rufe sie ab
-                $templateVars['statement'] = $this->getSingleDraftStatement(
-                    $itemsToExport[0]
-                );
+                $statement = $this->getSingleDraftStatement($itemsToExport[0]);
+                if ($this->currentUser->hasPermission('feature_statements_custom_fields')
```
mussbach: This permission + `instanceof` guard is duplicated verbatim in `addContentToStatementsArrayForPdfProcess` (line ~2171). Worth extracting into a private helper like `attachResolvedCustomFields(array $statement, string $procedureId): array` so the check lives in one place and the two call sites stay in sync.

demosplan/DemosPlanCoreBundle/Utils/CustomField/CustomFieldDisplayResolver.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $resolved = [];
+        foreach ($values->getCustomFieldsValues() as $fieldValue) {
+            $customFieldDefinition = $customFieldDefinitions->filter(
+                fn (CustomFieldInterface $field) => $field->getId() === $fieldValue->getId()
+            )->first();
+
+            if (!$customFieldDefinition instanceof CustomFieldInterface) {
+                $this->logger->warning(
+                    'CustomFieldDisplayResolver: no definition found for field value ID "{fieldValueId}". '
```
mussbach: PSR-3 `{fieldValueId}` placeholders are only interpolated if `PsrLogMessageProcessor` is attached to the monolog handler — otherwise the braces end up literal in the log line. Safer to `sprintf` the human-readable part and keep the context array for structured fields. Also: on a procedure with orphaned values this fires once per affected value per export; consider dropping to `info` or debouncing per resolver instance to avoid log spam.

demosplan/DemosPlanCoreBundle/CustomField/RadioButtonField.php
sierrazamudiodemos: thanks! 2f070b26fd314df0c09661c0afd228ecd4c926d0

demosplan/DemosPlanCoreBundle/Logic/Statement/DraftStatementService.php
sierrazamudiodemos: true that 9dab82bca12f1b807e98e20f534c647a24f09fd6

demosplan/DemosPlanCoreBundle/Logic/Statement/DraftStatementService.php
sierrazamudiodemos: makes sense. i added this 2d73aaf89fb997157e3eabafdad9605e428571b5

demosplan/DemosPlanCoreBundle/Utils/CustomField/CustomFieldDisplayResolver.php
sierrazamudiodemos: makes sense. done 51739fa72d9096879f6630fb3c63b277499b9677

mussbach (approved):
lgtm, code review only

PR #6150, opened 2026-04-15
feat: add retention period for deleted procedure purging

mussbach:
## Summary
- Introduce `purge_deleted_retention_period_days` parameter (default: 60) so deleted procedures are kept for a configurable grace period before permanent purging
- Filter purge query by `deletedDate` to only purge procedures older than the retention period
- Thread retention days from message handler through handler to service via `ParameterBagInterface`

## Test plan
- [x] Existing `PurgeDeletedProceduresMessageHandlerTest` updated and passing (4/4)
- [ ] Verify on a environment with `purge_deleted_procedures: true` that only procedures deleted > 60 days ago are purged

### DIFF
--- a/config/parameters_default.yml
+++ b/config/parameters_default.yml
@@ -43,6 +43,8 @@ parameters:
     procedure_public_list_default_sort_key: "publicParticipationEndDate"
     # Purge procedures if deleted
     purge_deleted_procedures: false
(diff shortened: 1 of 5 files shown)


breierDemos (approved):
lgtm thanks.
Belongs that to a ticket?

PR #6156, opened 2026-04-16
Fix DPLAN-17645 Elasticsearch keyword parsing error when removing use…

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-17645/Kein-Autobahnnutzer-darf-die-Rolle-Fachplanung-Masteruser-haben

Description: 

- getCustomers() used toArray() which preserves Collection array keys. When removeElement() removes a middle element, it leaves a gap in keys (e.g. 0,1,2,3,5). PHP json_encode treats non-sequential keys as a JSON object instead of array, causing Elasticsearch to reject the document with "failed to parse field [customers] of type [keyword]".
- Replaced toArray() with getValues() which re-indexes keys sequentially.


mussbach (approved):
lgtm, thx for this hard one

PR #6157, opened 2026-04-16
fix: delete physical files from storage when purging procedures

mussbach:
## Summary
- Delete files from Flysystem storage before removing DB records during procedure purge
- Add `deleteFilesFromStorage()` to handle physical file deletion with error handling
- Cover both per-entity file cleanup and the catch-all remaining-files sweep
- Fix bug in `processImportEmailAttachments` passing wrong array shape to `deleteFiles`

## Test plan
- [x] Added `testDeleteProcedureRemovesFilesFromStorage` — verifies files are deleted from disk
- [x] Added `testDeleteProcedureDryRunKeepsFilesInStorage` — verifies dry-run leaves files intact


bakridemos (approved):
lgtm and test work thx , code review

PR #6170, opened 2026-04-17
chore(deps): remove 10 unused direct composer dependencies

mussbach:
### Ticket
_No ticket — dependency housekeeping._

Drops 10 direct composer requires that have zero application-code references. Transitive installs remain where something else still needs them (e.g. `myclabs/php-enum` survives as a transitive dep of `timeweb/phpstan-enum`).

### Removed

**Runtime**
| Package | Reason |
|---|---|
| `composer/package-versions-deprecated` | Deprecated upstream; zero refs. Also dropped from `config.allow-plugins`. |
| `goetas-webservices/xsd2php-runtime` | Used only by the `xbeteiligung-async` addon, which declares its own require. |
| `khanamiryan/qrcode-detector-decoder` | Zero refs (only `endroid/qr-code` is used for generation). |
| `myclabs/php-enum` | Zero refs; project uses native PHP 8.1 enums. |
| `symfony/polyfill-php80` | Project requires PHP `^8.2 \|\| ^8.3` — polyfill is dead code. |

**Dev**
| Package | Reason |
|---|---|
| `escapestudios/symfony2-coding-standard` | Zero refs; `php-cs-fixer` + `phpstan` cover linting. |
| `goetas-webservices/xsd2php` | Same story as runtime variant — addon declares its own require. |
| `php-parallel-lint/php-console-highlighter` | Zero refs. |
| `squizlabs/php_codesniffer` | Zero config / script refs. |
| `wsdltophp/packagegenerator` | Zero refs. |

### Side effects
- `symfony.lock` entries for the above (and their transitive deps) removed by Flex.
- `.gitignore`: the `squizlabs/php_codesniffer` recipe block is removed.
- `composer.lock` shrinks by ~1100 lines as transitive-only deps drop out.

### How to review/test
- CI (BE + FE).
- Locally: `bin/diplanbau --version` boots cleanly; `RouterTest` (still using Prophecy) passes.

### Follow-ups
- Migrate `tests/backend/core/Core/Unit/Utilities/Twig/RouterTest.php` off Prophecy and drop `phpspec/prophecy` + `phpspec/prophecy-phpunit` — separate PR.
- Decide whether Psalm stack (`vimeo/psalm`, `psalm/plugin-symfony`, `weirdan/doctrine-psalm-plugin`) is still part of the QA pipeline.

### PR Checklist
- [x] Link all relevant tickets — n/a

PR #6171, opened 2026-04-17
chore(deps): migrate RouterTest off Prophecy, drop phpspec/prophecy

mussbach:
### Ticket
_No ticket — dependency housekeeping follow-up._

`RouterTest` was the only file using Prophecy. This PR ports its two mocks to PHPUnit's native `createMock()` and removes both Prophecy packages from `require-dev`.

### Changes
- `tests/backend/core/Core/Unit/Utilities/Twig/RouterTest.php`:
  - Removed `Prophecy\PhpUnit\ProphecyTrait` import and `use ProphecyTrait;`.
  - `$this->prophesize(...)` → `$this->createMock(...)`.
  - `->willReturn()` on a specific-argument expectation collapsed to `->method('generate')->willReturn(...)` (any-args) since `getSut()` stubbed all three `ReferenceType` variants with the same return value.
  - `$mock->reveal()` → `$mock`.
- `composer.json`: drop `phpspec/prophecy` and `phpspec/prophecy-phpunit`.

### How to review/test
- `phpunit tests/backend/core/Core/Unit/Utilities/Twig/RouterTest.php` — 15/15 pass (1 pre-existing CI-skip).
- CI BE.

### Depends on
- #6170 (unused composer deps cleanup) — independent, can merge either order.

### PR Checklist
- [x] Link all relevant tickets — n/a


sierrazamudiodemos (approved):
thx, it looks good. I understood actually the reason why prophecy existed, before it was difficult to mock using php unit. now it is not.
It looks like this prophecy library is now maintained, like the community is migrating to using mocking php unit.. also i fix the conflicts and i will merge as well

PR #6172, opened 2026-04-17
fix: harden LaTeX PDF export against ligatures and babel gaps

mussbach:
## Summary
- Emit `\textquotedbl` (textcomp, babel-independent) instead of `\dq` (babel-german-only) so PDF export still works when `ngerman.ldf` is missing on the TeX server
- Apply NFKC normalization + small replacement map in `LatexExtension::latexFilter` to collapse typographic ligatures (`ﬁ ﬂ ﬀ ﬃ ﬄ` → plain ASCII) and compatibility whitespace, strip zero-width chars, and rewrite arrows to math-mode (portable across TeX Live versions)
- Add `ext-intl` to composer.json — already installed in both Dockerfiles (`php8.3-intl`)

## Context
A real pdflatex run (TeX Live 2013) aborted with 100 cascading errors: root cause was missing `ngerman.ldf`, which made every `\dq` explode, compounded by unsupported Unicode chars (ligatures, narrow spaces, arrows) in user-pasted content.

## Test plan
- [x] `LatexExtensionTest` — 71/71 pass (updated `"` assertion, added cases for ligatures, narrow/thin/zero-width spaces, arrows, plus three direct `normalizeUnicode()` tests)
- [ ] Paste a snippet containing ligatures, German quotes, narrow spaces, arrows into a procedure statement and export to PDF; verify the PDF renders and the pdflatex log contains no `Unicode char` / `Undefined control sequence` entries


sierrazamudiodemos:
I guess there is no way to update to the latest TeX Live version?

sierrazamudiodemos (approved):
it works

PR #6175, opened 2026-04-20
fix(refs: DPLAN-17626) fix tooltip for not visible column in datatable

riechedemos:
### Ticket
[DPLAN-17626](https://demoseurope.youtrack.cloud/issue/DPLAN-17626/Abschnittsliste-Tooltip-fur-unsichtbare-Column-ist-sichtbar)

Tooltips appeared for column headers that were scrolled behind the sticky  flyout column. The flyout had `pointer-events: none` so the adjacent column's resize handle (which protrudes 21px into the flyout area via `right: -21px`) could be grabbed through it. As a side effect, hover events passed through to the occluded column headers and triggered their `v-tooltip`.                                                                                              
                                                                                                        
Raise the resize handle above the flyout via `z-index` so it can be grabbed directly, and switch its hidden state from `visibility: hidden`  to `opacity: 0` so the mouse can still reach it while it's invisible. The flyout can now block events normally, so hidden columns no longer receive hovers.  

### How to review/test
  1. Open segmentsList (Abschnittsliste in rog,ewm or fest)                                                                      
  2. Scroll the table horizontally so columns disappear behind the flyout                                   
  3. Hover the flyout area — no tooltip should appear for occluded columns                                  
  4. Hover a visible column header — its tooltip should appear normally                                     
  5. Hover the right edge of the last visible column (in the flyout area) - the resize handle should become visible and be grabbable    

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Update changelog


huellnerdemos (approved):
Thank you! Looks right for me, tested it and it worked well 🎉

PR #6176, opened 2026-04-20
Introduce symfony/runtime with public/ front-controller layout (BC)

mussbach:
## Summary

- Adopts the modern Symfony entry-point layout: single `public/index.php` + `public/app_dev.php` driven by `symfony/runtime`.
- Webpack output moves from `projects/\$name/web/` → `public/`. Prod `Dockerfile` and `nginx.conf.template` serve from `public/` with `index index.php`. `framework.http_cache: true` handles the `HttpCache` wrap `when@prod`.
- Adds closure-returning `FrontController::webRuntime()` / `::consoleRuntime()` alongside the legacy `::web(string)` / `::console(string)` — legacy signatures are **byte-for-byte unchanged** so per-project `web/app.php` entry points (in separate repos) keep working until each project migrates.

## composer.lock note

`composer update symfony/runtime` additionally evicts 13 transitive deps already orphaned by #6169–#6171 (phpspec/prophecy, goetas-webservices/xsd2php*, wsdltophp, squizlabs/php_codesniffer, composer/package-versions-deprecated, …). Bundled here because the natural graph re-solve is cleaner than hand-crafting two separate lock updates.

## Follow-up

Remove `demosplan/DemosPlanCoreBundle/Logic/HttpCache.php` once no caller of legacy `::web()` remains.

## Test plan

- [ ] `composer install` runs clean, no unintended recipe side-effects
- [ ] `bin/console <cmd>` boots the kernel via runtime; `--env=<x> --no-debug` argv flags propagate
- [ ] Static assets rebuild with `./fe build <project>`; `/js/legacy/jquery.js`, `/favicon.*`, `/manifest.json`, `/robots.txt` resolve 200
- [ ] `curl -I http://<project>.dplan.local/` returns dynamic response (not served as static PHP source)
- [ ] `curl -I http://<project>.dplan.local/dplan/login` returns 200
- [ ] Tus upload (PATCH `_tus/upload/…`) completes (writes under `public/uploads`)
- [ ] Legacy `projects/*/web/app.php` callers (external repos) still compile against the unchanged `FrontController::web(string, bool)` signature
- [ ] `bin/phpunit` kernel-boot tests pass in `test` env


docker/demosplan-production/nginx.conf.template
```diff
@@ -107,13 +106,13 @@ server
...
     location /
     {
-        try_files $uri $uri/ /app.php$is_args$args;
+        try_files $uri $uri/ /index.php$is_args$args;
     }
 
-    location ~ ^/app\.php(/|$)
+    location ~ ^/index\.php(/|$)
     {
         fastcgi_split_path_info ^(.+\.php)(/.*)$;
```
sierrazamudiodemos, line 139: @mussbach hello... thanks for this change! does it mean that the app.php per each project is outdated and could be deleted safely after merging this pr?

mussbach: Yes — `projects/*/web/app.php` and `app_dev.php` become dormant after this PR merges. Nginx root switches to `public/`, so they're no longer reachable, and the legacy `FrontController::web(string)` / `::console(string)` signatures are kept byte-for-byte so any remaining caller still compiles. Each project repo can delete the two files in a follow-up PR at its own pace once prod deploys are confirmed healthy on the new layout — no coordination needed.

(The other tracked files under `web/` — `favicon.*`, `manifest.json`, `robots.txt` — have already been copied into each project's `app/Resources/DemosPlanCoreBundle/public/`, so `web/` itself will be empty after the `app.php` removal.)

public/app_dev.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+use demosplan\DemosPlanCoreBundle\Application\FrontController;
+
+require_once dirname(__DIR__).'/vendor/autoload_runtime.php';
+
+$_SERVER['APP_DEBUG'] = '1';
+$_SERVER['APP_ENV'] = 'dev';
+
+return FrontController::webRuntime();
```
sierrazamudiodemos, line 18: is the plan to delete this file as well? to me it looks like and index.php for dev, and what i understand form symfony is that all the calls (even the ones from dev/local) would also be used in index,php... I know there will be some side effects from deleting this

sierrazamudiodemos: I guess if we delets this file and we manage to set those APP_DEBUG and APP_ENV somewhere, we can also delete or adjust the nginx.conf.template-dev

mussbach: I build it like this to not break DX as in the "new" way devs have to fiddle with the env vars. Once we are all used to the new workflow this file can be indeed be deleted.

demosplan/DemosPlanCoreBundle/EventListener/ApcuClearEventSubscriber.php
```diff
@@ -48,7 +48,7 @@ public function onControllerRequest(ControllerEvent $event)
             return;
         }
 
-        $cacheScheduleFile = DemosPlanPath::getProjectPath(CacheClearCommand::APCU_CLEAR_SCHEDULE_FILE);
+        $cacheScheduleFile = DemosPlanPath::getPublicPath(CacheClearCommand::APCU_CLEAR_SCHEDULE_FILE);
```
sierrazamudiodemos, line 52: this method `getPublicPath` was introduced intentionally so that it is centralized the detection of the public path, but also there are other way to get the public path, like $this->parameterBag->get('kernel.project_dir').'/public' --> I was curious that we have 2  ways to detect that public path... maybe we can have one, and I think u thought this case already, i was curious to hear what u think

mussbach: We created DemosPlanPath to ease the general path handling and they should be used when ever possible. We cannot restrict the usage of $this->parameterBag->get('kernel.project_dir').'/public' technically and sometimes it is not possible to use DemosPlanPath

demosplan/DemosPlanCoreBundle/Command/CacheClearCommand.php
sierrazamudiodemos, line 109: maybe we can delete this if? that is because in the old code: 
- it was  $file = `DemosPlanPath::getProjectPath(self::APCU_CLEAR_SCHEDULE_FILE);` --> that is /`srv/src/projects/THE_PROJECT/web/uploads/scheduled-apcu-clear`
- and now after the change:  `$file = DemosPlanPath::getPublicPath(self::APCU_CLEAR_SCHEDULE_FILE);` --> then this is : `/srv/www/public/uploads/scheduled-apcu-clear`


so because we do not use getProjectPath, then the `projects/THE_PROJECT/` will not be used.. consequently the if will not be executed... what do you think?

mussbach: good catch. we cannot remove it but we need to adjust the path [2bafdfc](https://github.com/demos-europe/demosplan-core/pull/6176/commits/2bafdfced5aceed4759f0bc44531436bd6fbd588)

sierrazamudiodemos:
@mussbach  i commited this 8f07dd00c235d97b6ce6ad3a28abc9dfba556ff3 and this cde07570cf02eb1524adc60ee20bcc292352c0fc

it is about also using the public path for detecting the status of the documentHandler, it seemed like an old bug
what do you think?

sierrazamudiodemos:
@mussbach maybe not relevant, but I guess migrating to vite is on the roadmap?

php wise it looks good I think, I would only say that the PR is blocked so you can take a look at the commits that i did (the last 2)

sierrazamudiodemos (changes requested):
Not requesting changes, more like taken a look at the 2 commits that I did

mussbach:
@sierrazamudiodemos yes vite is on the roadmap, then all those config files needs to be adjusted anyways. Thank you for your commits, they are totally valid

sierrazamudiodemos (approved):
PHP wise is approved

sierrazamudiodemos (approved):
@mussbach FE wise i think it looks ok.. I dont see anything weird

PR #6182, opened 2026-04-21
F dplan-17619 add text field

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17619/ADO-52327-Erweiterung-der-InstitutionsVerschlagwortung-um-konfigurierbare-Textfelder-Part-1-Erstellen

**refs: DPLAN-17619**
                                                                                                                                                                                            
**What changed**               
                                                                                                                                                                                            
This PR adds a new text field type to the custom fields system. Previously, custom fields only supported types like checkboxes or select lists. Now admins can create a plain text input    
field.
                                                                                                                                                                                            
**New TextField type**         

- Stores a free-text value (string)                                                                                                                                                         
- Supports an optional "required" flag — validation will reject empty values if the field is marked as required
- No predefined options (unlike select fields)                                                                                                                                              
                           
**Security: customer scope in access conditions**                                                                                                                                               
                           
When loading custom fields through the API, the system now checks the current customer. If a custom field belongs to a customer source entity, only fields from the current customer are returned. This prevents one customer's data from leaking to another.
                                                                                                                                                                                            
**Server-side customer ID injection on creation**

When creating a custom field with sourceEntity: customer, the backend now automatically sets the sourceEntityId to the current customer's ID. The frontend no longer needs to send this value, which also prevents a user from creating a field for a different customer.


### How to review/test

### Linked PRs (optional)
- https://github.com/demos-europe/demosplan-core/pull/6220


demosplan/DemosPlanCoreBundle/CustomField/TextField.php
```diff
@@ -0,0 +1,76 @@
...
+    public function getFieldType(): string
+    {
+        return 'text';
+    }
+
+    public function fromJson(array $json): void
+    {
+        $this->fieldType = $json['fieldType'];
+        $this->name = $json['name'];
+        $this->description = $json['description'];
```
mussbach, line 29: No key guards on `fieldType`/`name`/`description`. Partial or malformed JSON from DB throws `ErrorException`. `isRequired` already null-coalesces — be consistent: either guard all four or trust all four.

demosplan/DemosPlanCoreBundle/CustomField/TextField.php
```diff
@@ -0,0 +1,76 @@
...
+{
+    protected string $fieldType = 'text';
+    protected bool $isRequired = false;
+
+    public function getFieldType(): string
+    {
+        return 'text';
+    }
+
+    public function fromJson(array $json): void
```
mussbach, line 25: Missing test for malformed JSON entering `fromJson` (e.g. legacy row without `description`). Add a unit test asserting graceful behaviour or explicit failure.

demosplan/DemosPlanCoreBundle/CustomField/TextField.php
```diff
@@ -0,0 +1,76 @@
...
+namespace demosplan\DemosPlanCoreBundle\CustomField;
+
+class TextField extends AbstractCustomField
+{
+    protected string $fieldType = 'text';
+    protected bool $isRequired = false;
+
+    public function getFieldType(): string
+    {
+        return 'text';
```
mussbach: `getFieldType()` hardcodes `'text'`, ignoring `$fieldType` property set by `setFieldType()`/`fromJson`. Property + setter become effectively dead. Either drop the property or return `$this->fieldType` (siblings do the latter).

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
```diff
@@ -92,7 +97,31 @@ public function getTypeName(): string
...
+        /*
+         * For CUSTOMER source the row must belong to the current customer.
+         * Non-CUSTOMER sources (PROCEDURE etc.) keep their existing behaviour
+         * — their sourceEntityId is still supplied by the caller as a filter.
+         */
+        return [
+            $this->conditionFactory->anyConditionApplies(
+                $this->conditionFactory->propertyHasNotValue(
+                    CustomFieldSupportedEntity::customer->value,
+                    $this->sourceEntityClass
```
mussbach, line 111: NULL semantics: in SQL `sourceEntityClass != 'CUSTOMER'` is UNKNOWN when the column is NULL, so rows with NULL `sourceEntityClass` are filtered out under both branches. Confirm no legacy rows have NULL there, otherwise they become invisible.

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
```diff
@@ -92,7 +97,31 @@ public function getTypeName(): string
...
     {
-        return [];
+        $customerId = $this->customerService->getCurrentCustomer()->getId();
+
+        /*
+         * For CUSTOMER source the row must belong to the current customer.
+         * Non-CUSTOMER sources (PROCEDURE etc.) keep their existing behaviour
+         * — their sourceEntityId is still supplied by the caller as a filter.
+         */
+        return [
```
mussbach, line 107: Security-critical scoping with no test. Add: (a) read returns empty for fields belonging to a different customer, (b) `resolveAttributes` ignores client-supplied `sourceEntityId` and overwrites with current customer's ID.

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
```diff
@@ -283,4 +313,20 @@ public function deleteEntity(string $entityIdentifier): void
...
     }
+
+    /**
+     * @param array<non-empty-string, mixed> $attributes
+     *
+     * @return array<non-empty-string, mixed>
+     */
+    private function resolveAttributes(array $attributes): array
+    {
+        if (CustomFieldSupportedEntity::customer->value !== $attributes['sourceEntity']) {
```
mussbach, line 324: Unguarded `$attributes['sourceEntity']` — undefined-index notice if EDT layer doesn't guarantee the key. Use `($attributes['sourceEntity'] ?? null)`.

demosplan/DemosPlanCoreBundle/DataGenerator/Factory/CustomFields/TextFieldFactory.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace demosplan\DemosPlanCoreBundle\DataGenerator\Factory\CustomFields;
+
+use demosplan\DemosPlanCoreBundle\CustomField\TextField;
+use Zenstruck\Foundry\ObjectFactory;
+
+/**
+ * @extends ObjectFactory<TextField>
+ */
+final class TextFieldFactory extends ObjectFactory
```
mussbach: Class name collides with `Utils/CustomField/Factory/TextFieldFactory` (runtime factory). Different namespaces but identical name — confusing on grep/IDE jumps. Suggest renaming this fixture class to `TextFieldStubFactory` or `TextFieldFixtureFactory`.

mussbach (commented):
Inline notes on correctness, style, security tests, and missing test coverage.

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
sierrazamudiodemos: `$attributes['sourceEntity']` has to be always present in creation, EDT makes sure it is there. that is because in this line `$configBuilder->sourceEntity->setAliasedPath(['sourceEntityClass'])->addPathCreationBehavior()->setReadableByPath()->setFilterable();` the part of `addPathCreationBehavior()` if no parameter is passed, then it requires the field on the payload

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
sierrazamudiodemos: every column must have a value on sourceEntityClass, it cannot be nullable

demosplan/DemosPlanCoreBundle/CustomField/TextField.php
sierrazamudiodemos: done 6f9779f77478cf2b16b5c01ce5503ba6e41f217b

demosplan/DemosPlanCoreBundle/CustomField/TextField.php
sierrazamudiodemos: done 6f9779f77478cf2b16b5c01ce5503ba6e41f217b

demosplan/DemosPlanCoreBundle/CustomField/TextField.php
sierrazamudiodemos: done c3d3b4f6d3cf10c51df1239c8ccdf378ae3ca64d

demosplan/DemosPlanCoreBundle/DataGenerator/Factory/CustomFields/TextFieldFactory.php
sierrazamudiodemos: done aebff19abd16318c2263454445554d0a81177899

demosplan/DemosPlanCoreBundle/ResourceTypes/CustomFieldResourceType.php
sierrazamudiodemos: - No dedicated test for `getAccessConditions` exists yet in any of the resourceTypes. For `resolveAttributes` method the method already guarantees that when sourceEntity is CUSTOMER, the sourceEntityId is always injected from the authenticated session — the client-supplied value is silently overwritten, so there is no attack surface there. 
- Architecturally, injecting session data inside a ResourceType is not ideal, and I'd like to move that logic closer to the domain. 
- However, CUSTOMER is currently the only source type whose ID is determined server-side, and a second one is expected soon —   at that point there will be enough context to design a clean solution without overengineering. For now, leaving it here keeps the change minimal and reviewable.

mussbach (approved):
lgtm, code review only, thx

PR #6184, opened 2026-04-22
fix (DPLAN-17601): enforce single visible base layer for global layer

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17601/Globale-Grundkarten-mehr-als-1-anzeigbar

**Description:**
When a global base map ("Grundkarte") had "Karte beim Laden anzeigen" activated, the mutual-exclusivity constraint was not enforced — other global base maps kept their                     
defaultVisibility=true, causing multiple base maps to be shown simultaneously. The same constraint already worked correctly for procedure-specific layers but was explicitly skipped for
global layers via an early return in BaseLayerVisibilityValidator.                                                                                                                          
                           
**What changed:**

- Removed the early return that skipped global layers (procedureId = null/'') in BaseLayerVisibilityValidator::ensureOnlyOneBaseLayerIsVisible(). Global layers now go through the same     
enforcement path as procedure-specific layers.
- Added MapHandler::disableDefaultVisibilityForOtherGlobalBaseLayers() — a single DQL UPDATE query that turns off defaultVisibility on all other global base layers except the one being    
saved. This replaces what would have been a per-entity loop that triggers updateRelatedGis(), which cascades to all procedure copies (one UPDATE per procedure) and causes DB lock wait     
timeouts in environments with many procedures.
                                                                                                                                                                                            
**How to review/test**                                                                                                                                                                          

1. As Master-Admin, go to /gislayer/ and create or edit a global base map                                                                                                                   
2. Activate "Karte beim Laden anzeigen" and save
3. Verify only that base map has the flag active — all other global base maps should have it turned off                                                                                     
4. Log in as Fachplaner-Admin, open any procedure map view and verify only one base map loads by default


demosplan/DemosPlanCoreBundle/Logic/Map/MapHandler.php
```diff
@@ -378,6 +379,29 @@ public function updateGis($data)
...
+            ->update(GisLayer::class, 'g')
+            ->set('g.defaultVisibility', ':false')
+            ->where('g.procedureId = :procedureId')
+            ->andWhere('g.type = :type')
+            ->andWhere('g.id != :exceptId')
+            ->andWhere('g.defaultVisibility = :true')
+            ->setParameter('false', false)
+            ->setParameter('true', true)
+            ->setParameter('procedureId', '')
+            ->setParameter('type', BaseLayerVisibilityValidator::BASE_LAYER_TYPE)
```
winderdemos: Is there a reason BASE_LAYER_TYPE is used rather than GisLayerInterface::TYPE_BASE?

sierrazamudiodemos: true that, I fixed it here: f78ab27ebf7d79cb5f5ead8faa82e747f15e53ef
there was a duplicated const

winderdemos (approved):
LGTM :) Is it possible to unit test 🤔

PR #6185, opened 2026-04-22
perf(refs DS-428): speed up procedure phase resolution

mussbach:
## Summary

Reduces per-statement overhead in `StatementProcedurePhaseResolver` and `StatementService::getProcedurePhaseNameFromArray`, which dominated the cost of assessment-table list and export paths when a procedure contained statements with phase keys no longer defined in the phase config.

- Memoize the available phase list and a precomputed `key → ProcedurePhaseVO` map in `StatementProcedurePhaseResolver`; `getProcedurePhaseVO` becomes O(1) and stops rebuilding the list per call.
- Skip the per-statement `getStatement($id)` Doctrine round-trip in `getProcedurePhaseNameFromArray` when `publicStatement` is already on the array (ES- and JSON-sourced statement arrays carry it). The fallback path to the entity is retained for callers that pass a minimal array.
- Downgrade `logger->error` → `logger->warning` for the "Undefined phase for key" path: legacy phase values on statements caused this to flood the error channel during large exports without representing an actionable runtime fault.

## Ticket

DS-428

## Test plan

- [ ] `tests/backend/core/Statement/Functional/StatementServiceTest.php` passes (47/47 green locally, 22 skipped)
- [ ] `tests/backend/core/AssessmentTable/` passes (38/38 green locally, 2 skipped)
- [ ] Manually export an assessment table on a procedure that carries legacy phase values and confirm no "Undefined phase" entries appear on the error channel (warning channel is expected)
- [ ] PHPStorm inspections on both files: no new warnings introduced

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
@@ -2698,16 +2698,27 @@ class StatementService implements StatementServiceInterface
      */
     public function getProcedurePhaseNameFromArray(array $statement): string
     {
-        $statementObject = $this->getStatement($statement['id']);
+        // Fast path: large exports feed thousands of statement arrays through here,
+        // so avoid the per-statement getStatement() round-trip when publicStatement
+        // is already present on the array (ES- and JSON-sourced statements have it).
+        if (isset($statement['publicStatement'])) {
+            return $this->getProcedurePhaseName(
+                $statement['phase'] ?? '',
+                StatementInterface::EXTERNAL === $statement['publicStatement']
+            );
+        }
+
+        $statementId = $statement['id'] ?? null;
+        $statementObject = null !== $statementId ? $this->getStatement($statementId) : null;
 
         if (!$statementObject instanceof Statement) {
-            $this->logger->error('Statement with id '.$statement['id'].' not found.');
+            $this->logger->warning('Statement with id '.($statementId ?? '').' not found.');
 
             return '';
         }
 
         return $this->getProcedurePhaseName(
-            $statement['phase'],
+            $statement['phase'] ?? '',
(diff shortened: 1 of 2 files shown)


demosplan/DemosPlanCoreBundle/Logic/Statement/StatementService.php
```diff
@@ -2698,16 +2698,27 @@ public function bulkEditStatementsAddData(StatementBulkEditVO $statementEdit): v
...
     {
-        $statementObject = $this->getStatement($statement['id']);
+        // Fast path: large exports feed thousands of statement arrays through here,
+        // so avoid the per-statement getStatement() round-trip when publicStatement
+        // is already present on the array (ES- and JSON-sourced statements have it).
+        if (isset($statement['publicStatement'])) {
+            return $this->getProcedurePhaseName(
+                $statement['phase'] ?? '',
+                StatementInterface::EXTERNAL === $statement['publicStatement']
+            );
```
sierrazamudiodemos, line 2708: hellooo
I think there is an issue here... the original code had       
```
return $this->getProcedurePhaseName(
          $statement['phase'],
          $statementObject->isSubmittedByCitizen()  // ← always correct: checks org UUID
      );    
```
now we are comparing to  StatementInterface::EXTERNAL === $statement['publicStatement'] to check if it is citizen or orga... but what i undersand is that all the STN are set as external when they are given from the public view (even if they are from orga or private citizen) consequently, getting with this in the PR external phases for orga stn.

i guess it can be changed to: `UserInterface::ANONYMOUS_USER_ORGA_ID === ($statement['oId'] ?? '')` or maybe there is another way to detect it

does it make sense what I write?

sierrazamudiodemos (changes requested):
i have one concern

mussbach:
superseded by DPLAN-17570

PR #6187, opened 2026-04-22
Feat(DPLAN-17499): enable spellcheck

sakutademos:
### Ticket
[DPLAN-17499](https://demoseurope.youtrack.cloud/issue/DPLAN-17499) FE: Rechtschreibprüfung im Tiptap Editor

**Description:**  This PR adds spellcheck to DpEditor and adds a new error class.

### PR Checklist
- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_tiptap.scss
```diff
@@ -218,6 +218,10 @@
 
             font-size: .9em;
 
+            .lt-error {
```
hwiem, line 221: I'd suggest to use tailwind classes for this: `border-b border-status-failed`. That way, we also use the red color from our design system.

hwiem (changes requested):

demosplan/DemosPlanCoreBundle/Resources/client/scss/components/_tiptap.scss
sakutademos: ah, I totally forgot to refactor it and use (for example) the `$dp-color-status-failed-border` color instead of just red.. https://github.com/demos-europe/demosplan-core/pull/6187/commits/30763bfa4e662cbf5cd45959c97b6c29dcdf6417
 but what do you mean by tailwind classes? do you mean using it directly here?

hwiem: No, instead of adding the `lt-error` class in `src/components/DpEditor/libs/languageTool/LanguageToolExtension.js`, we could add tailwind classes directly. or am i missing something?

sakutademos: ahh, no, I do not think it is a better approach, because the `lt-error `class is also used as a selector. So we still need a selector, and I do not think that is a bad thing, especially since the tiptap styling is defined in the `tiptap.scss` file. So I think it makes sense to keep it there..

hwiem (approved):
Lgtm!

PR #6189, opened 2026-04-23
fix(refs: DPLAN-17672): change initial state handling for statements

huellnerdemos:
### Ticket
DPLAN-DPLAN-17672

Pre-selected default values (e.g. the pre-selected county) were applied to the statement form via setStatementData, which does not update the initForm baseline. This caused the form to be considered unsaved immediately on open, before the user made any changes, resulting in the wrong processing status being set for new statements.             

A new applyInitialDefaults Vuex mutation is introduced that updates both statement and initForm at the same time, so  the form baseline correctly reflects the applied defaults. Additionally, the r_useName enforcement for non-anonymous procedures was moved inside the fetchCustomFields().then() callback to ensure it runs after async initialization.

### How to review/test

1. Open a procedure as a public user where anonymous statements are not allowed
2. Open the Statement modal without making any changes                                                                           
3. Check if the button that opens the modal does not immediately change its text after opening the modal
4. Fill in a field and check, if the button changes its text now to "Stellungnahme fortsetzen..."

### PR Checklist

- [x] Run `yarn lint`

### DIFF
--- a/client/js/components/statement/publicStatementModal/StatementModal.vue
+++ b/client/js/components/statement/publicStatementModal/StatementModal.vue
@@ -2141,9 +2141,11 @@ export default {
               this.counties.find(el => el.selected)?.value :
               '',
           }
+
           if (this.isNameUsageRequired) {
             initialStatementDefaults.r_useName = '1'
           }
+
           this.applyInitialDefaults(initialStatementDefaults)
         }
       } else {
--- a/client/js/components/statement/publicStatementModal/StatementModal.vue
+++ b/client/js/components/statement/publicStatementModal/StatementModal.vue
@@ -1202,6 +1202,10 @@ export default {
       return this.unsavedDrafts.includes(this.draftStatementId)
     },
 
+    isNameUsageRequired () {
+      return !this.allowAnonymousStatements && this.formData.r_useName !== '1'
+    },
+
     personalDataFormDefinitions () {
       return this.personalDataFormFields.map(el => {
         this.availableFormComponents[el.name].width = this.availableFormComponents[el.name].width || 'u-1-of-2'
@@ -1257,6 +1261,7 @@ export default {
 
     ...mapMutations('PublicStatement', [
       'addUnsavedDraft',
+      'applyInitialDefaults',
       'clearDraftState',
       'removeStatementProp',
       'removeUnsavedDraft',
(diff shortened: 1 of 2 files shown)


client/js/components/statement/publicStatementModal/StatementModal.vue
```diff
@@ -2123,26 +2128,29 @@ export default {
...
+          const initialStatementDefaults = {
             r_county: this.counties.some(el => el.selected) ?
               this.counties.find(el => el.selected)?.value :
               '',
-          })
+          }
+          if (this.isNameUsageRequired) {
+            initialStatementDefaults.r_useName = '1'
+          }
+          this.applyInitialDefaults(initialStatementDefaults)
```
hwiem, line 2149: nitpick
```suggestion
          }
          
          if (this.isNameUsageRequired) {
            initialStatementDefaults.r_useName = '1'
          }
          
          this.applyInitialDefaults(initialStatementDefaults)
```

client/js/components/statement/publicStatementModal/StatementModal.vue
```diff
@@ -2123,26 +2128,29 @@ export default {
...
           })
         } else {
-          this.setStatementData({
+          const initialStatementDefaults = {
             r_county: this.counties.some(el => el.selected) ?
               this.counties.find(el => el.selected)?.value :
               '',
-          })
+          }
+          if (this.isNameUsageRequired) {
```
hwiem, line 2145: Should this not be done if `draftStatementId !== ''`? (Before your changes, it was done outside the `if/else` block)

huellnerdemos: Thank you for this Point! I thought about this.
Before this change the if statement was called synchronously with the rest of the mounted()-code. so it was called before there a draftStatementId was read from the sessionStorage and it was called always independent from the draftStatementId. Now it is in the asynchronous context. Since the draft statements can only be accessed by logged-in private users, who always submit with their name, the r_useName enforcement should not be needed for drafts.

client/js/components/statement/publicStatementModal/StatementModal.vue
huellnerdemos: Thanks for that! Changed it in 80b6423ff299989d6489e5900eb7eabdf8931c7c

hwiem (approved):
Lgtm + works!

PR #6190, opened 2026-04-23
fix(refs: DPLAN-17671): Initial fix for performance

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17671/Globale-Kartenebenen-Speichern-verursacht-Timeout-bei-vielen-Verfahren

**DPLAN-17671**                                                                                                                                                                                 
                                                                                                                                                                                            
Saving a global GIS layer updates all procedure-specific copies of that layer. The previous implementation loaded each copy as a Doctrine entity and called flush() individually per copy —  resulting in N SQL UPDATE statements and N Doctrine flush cycles, one per procedure. On instances with many procedures this caused request timeouts.                                        
                                                                                                                                                                                            
`updateRelatedGis()` now executes a single DQL bulk UPDATE (UPDATE GisLayer SET … WHERE gId = :globalId) regardless of how many procedure copies exist, reducing database round-trips from N  
to 1.
                                                                                                                                                                                            
**How to review/test**                                                                                                                                                                          
 
1. As master admin open an existing global GIS layer at /gislayer/                                                                                                                          
2. Change any field (e.g. the name) and save
3. The page should return immediately without a long load or HTTP 500


mussbach (commented):
Inline comments on the bulk-update refactor. Approach is right; flagging blockers around `clear()` and projection pairing.

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
```diff
@@ -333,27 +333,55 @@ private function updateGisFromHash(GisLayer $gis, $data)
...
+                }
 
-            foreach ($listToUpdate as $layer) {
-                $this->updateGisFromHash($layer, $dataWithoutPId);
+                $qb->getQuery()->execute();
             }
+
+            // DQL bulk UPDATE bypasses Doctrine's identity map. Clear it so the caller
+            // and any subsequent loads return the updated state from the database.
+            $this->getEntityManager()->clear();
```
mussbach: Blocker: `$em->clear()` wipes the entire identity map for ALL entity types. Any caller above us in the request that holds a managed entity (Procedure, User, ContextualHelp, …) gets it detached, and a later flush in the same request can silently drop changes.

Replace with a scoped refresh of the global entity — copies aren't returned anywhere, so only `$item` needs re-syncing:
```php
$this->getEntityManager()->refresh($item);
return $item;
```
This also removes the extra `SELECT` in `$this->get(...)` on line 384.

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
```diff
@@ -333,27 +333,55 @@ private function updateGisFromHash(GisLayer $gis, $data)
...
 
-            $listToUpdate = $this->findBy(['gId' => $item->getIdent()]);
+            // Scalar properties safe to bulk-update on all procedure copies.
+            // Excluded: procedureId/globalGisId/ident (identifiers), categoryId (each copy
+            // belongs to its own procedure root category), contextualHelpText (side-effects).
+            $updates = array_intersect_key($data, array_flip([
+                'name', 'type', 'url', 'isMiniMap', 'layers', 'layerVersion', 'legend',
+                'opacity', 'print', 'capabilities', 'defaultVisibility', 'tileMatrixSet',
+                'scope', 'serviceType', 'order', 'enabled', 'deleted', 'bplan', 'xplan',
+                'treeOrder', 'userToggleVisibility', 'projectionLabel', 'projectionValue',
```
mussbach: Pairing lost: `updateGisFromHash()` only sets `projectionLabel`/`projectionValue` if **both** keys are present (see line 319-323 of the existing method). The bulk path here writes either independently → copies can end up with mismatched label/value.

Suggest dropping both from the whitelist and handling them with the same paired guard, or filter them out unless `isset($data['projectionLabel'], $data['projectionValue'])`.

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
```diff
@@ -333,27 +333,55 @@ private function updateGisFromHash(GisLayer $gis, $data)
...
+                $data['enabled'] ??= $data['visible'];
+            }
+            if (array_key_exists('isMinimap', $data)) {
+                $data['isMiniMap'] ??= $data['isMinimap'];
             }
 
-            $listToUpdate = $this->findBy(['gId' => $item->getIdent()]);
+            // Scalar properties safe to bulk-update on all procedure copies.
+            // Excluded: procedureId/globalGisId/ident (identifiers), categoryId (each copy
+            // belongs to its own procedure root category), contextualHelpText (side-effects).
```
mussbach: Behavior change worth confirming: pre-PR, every procedure copy got `contextualHelpText` written through `updateGisFromHash`. Post-PR, only the global is updated and copies drift. If the product expectation is that copies mirror the global help text, this needs a separate code path (or a shared `ContextualHelp` row). Please document the decision in the PR body.

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
```diff
@@ -333,27 +333,55 @@ private function updateGisFromHash(GisLayer $gis, $data)
...
+            $updates = array_intersect_key($data, array_flip([
+                'name', 'type', 'url', 'isMiniMap', 'layers', 'layerVersion', 'legend',
+                'opacity', 'print', 'capabilities', 'defaultVisibility', 'tileMatrixSet',
+                'scope', 'serviceType', 'order', 'enabled', 'deleted', 'bplan', 'xplan',
+                'treeOrder', 'userToggleVisibility', 'projectionLabel', 'projectionValue',
+            ]));
```
mussbach: Whitelist drift risk: any new scalar field added to `GisLayer` will silently fail to propagate to copies, with no test catching it. Two options:
- derive the list from `ClassMetadata::getFieldNames()` minus a small exclusion set, or
- add a unit test asserting whitelist ⊇ entity scalar fields minus the documented exclusions.

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
```diff
@@ -333,27 +333,55 @@ private function updateGisFromHash(GisLayer $gis, $data)
...
+
+                foreach ($updates as $property => $value) {
+                    $qb->set('g.'.$property, ':p_'.$property)
+                        ->setParameter('p_'.$property, $value);
+                }
 
-            foreach ($listToUpdate as $layer) {
-                $this->updateGisFromHash($layer, $dataWithoutPId);
+                $qb->getQuery()->execute();
             }
```
mussbach: Minor: when `$updates` is empty, no SQL fires but `clear()` (line 382) still runs. Move the cache invalidation inside the `if (!empty($updates))` block so a no-op call is truly free.

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
```diff
@@ -333,27 +333,55 @@ private function updateGisFromHash(GisLayer $gis, $data)
      *
      * @throws Exception
      */
-    private function updateRelatedGis($item, $data)
+    private function updateRelatedGis($item, $data): GisLayer
```
mussbach: PHPDoc above doesn't reflect the new return type. Add `@return GisLayer`.

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
```diff
@@ -333,27 +333,55 @@ private function updateGisFromHash(GisLayer $gis, $data)
...
+                'name', 'type', 'url', 'isMiniMap', 'layers', 'layerVersion', 'legend',
+                'opacity', 'print', 'capabilities', 'defaultVisibility', 'tileMatrixSet',
+                'scope', 'serviceType', 'order', 'enabled', 'deleted', 'bplan', 'xplan',
+                'treeOrder', 'userToggleVisibility', 'projectionLabel', 'projectionValue',
+            ]));
+
+            if (!empty($updates)) {
+                $qb = $this->getEntityManager()->createQueryBuilder()
+                    ->update(GisLayer::class, 'g')
+                    ->where('g.gId = :globalId')
```
mussbach, line 376: Confirm `gId` has a DB index (`_gis_layer.g_id`). With many global layers + many procedures the bulk UPDATE will scan otherwise.

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
mussbach: No tests added for the new path. At minimum: alias→canonical mapping unit test, and a repo test asserting one DQL UPDATE covers N copies (e.g. SQL logger count or row-count check).

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
sierrazamudiodemos: - It is safe to exclude the `contextualHelpText` in this global layer context because it is not a field on the global layer form (`map_admin_gislayer_global_edit.html.twig`). That said, `contextualHelpText` is never present in `$data` when `updateRelatedGis` is called. 
- Editing the gis layer in the  procedure-level form does have the field `contextualHelpText`, but it always submits the copy's own UUID as r_ident, so `isGlobal()` returns false and `updateRelatedGis` is never reached from that path either. 

I would keep it as in the PR, maybe it is not so easily readable?
the thing is that `contextualHelpText` info is stored in another table..

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
sierrazamudiodemos: - yes, there is an index: `@ORM\Table(name="_gis", indexes={@ORM\Index(name="_g_global_id", columns={"_g_global_id"})})` in the class `\demosplan\DemosPlanCoreBundle\Entity\Map\GisLayer`
- The doctrine attribute `gId` has the column name `_g_global_id`

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
sierrazamudiodemos: done 834e17cd6

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
sierrazamudiodemos: done 97c840f3d

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
sierrazamudiodemos: done 52da6d663

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
sierrazamudiodemos: done 52da6d663

demosplan/DemosPlanCoreBundle/Repository/MapRepository.php
sierrazamudiodemos: done:
2647a1009 + 0f0a557c4

mussbach (approved):
lgtm, code review only

PR #6191, opened 2026-04-23
fix(ADO-54112): drop stale UNIQUE index on _procedure.customer

mussbach:
Refs: ADO 54112

## Problem

Creating a procedure fails on some environments with:

```
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry
'<customer-uuid>' for key 'UNIQ_D1A01D0281398E09'
```

`UNIQ_D1A01D0281398E09` is the leftover unique index on `_procedure.customer` from when the relationship was 1:1. The Procedure entity is now `@ManyToOne`, so multiple procedures per customer should be valid.

## Root cause

`Version20240105113406` (Jan 2024) dropped that index inside a `try { ... } catch (Exception) {}` block. On environments where the drop threw (e.g. the FK still held the index at that moment), the error was swallowed and the unique constraint remained. The later `Version20250409185422` only renamed `fk_...` → `IDX_...` and did not touch a surviving `UNIQ_...`.

## Fix

New migration that, only if `UNIQ_D1A01D0281398E09` still exists:
1. Creates a non-unique `IDX_D1A01D0281398E09` on `customer` so the foreign key has a backing index.
2. Drops `UNIQ_D1A01D0281398E09`.

No FK churn, no moment without FK enforcement. No-op on environments where the Jan-2024 migration already succeeded.

## Test plan

- [ ] `dplan:migrations:diff` shows no drift on `_procedure` after the migration runs
- [ ] On an affected DB: creating a second procedure for the same customer succeeds
- [ ] On a clean DB: migration logs "did not result in any SQL statements" and produces no changes

### DIFF
--- a/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/04/Version20260423104500.php
+++ b/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/04/Version20260423104500.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of the package demosplan.
+ *
+ * (c) 2010-present DEMOS plan GmbH, for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Schema\Schema;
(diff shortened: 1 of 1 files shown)


sierrazamudiodemos (approved):
I understand that some migrations were run in the past, but one index was not removed. This migration is about deleting that index in case it is present in the database. I run locally the migrations in a couple of projects and it works (they do not have the index consequently the migration is not executed) i do not have locally a project that does have the index.

PR #6195, opened 2026-04-24
feat(DPLAN-17492): add segment lock via workflow place

MoritzMandler:
### Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-17492/BE-FE-Abgeschlossene-Abschnitte-Sperren-Entsperren-durch-FPA

### Description:
Workflow places can now be flagged as **locked**. Segments sitting on a locked place become read-only for everyone except holders of the new ```feature_administrate_segment_lock``` permission (initially ```PLANNING_AGENCY_ADMIN```). The whole feature is gated per project by the ```segment_lock_by_workflow_place``` config parameter (default `false`); only `ewm` opts in.                                                                              
                                                                                              
Enforcement runs on every write surface a segment can flow through: 
  - **JSON:API PATCH** — a `BeforeResourceUpdateFlushEvent` subscriber                        
    reads the *original* place from the UoW (so a non-admin can't escape
    the lock by including a place change in the payload) and rejects with                     
    a `SegmentLockedException` (HTTP 403) + an `error.segment.locked.by.place`                
    toast.                                                                                    
  - **`segment.bulk.edit` RPC** — pre-validates the batch via                                 
    `SegmentBulkEditorService::assertBatchEditable`, which runs a single                      
    procedure-scoped JOINed DQL (`SegmentRepository::findLockedByIds`) to                     
    identify locked IDs before any entity hydration. Rolls back the whole                     
    batch on lock; FPA "unlock" flow is just a bulk edit by a                                 
    permission-holder that moves a segment off a locked place + reassigns.                    
  - **`PlaceResourceType` PATCH** of `locked` — admin toggles emit one                        
    `entity_content_change` row per affected segment (mass case) so the                       
    Versionsverlauf shows a "Gesperrt" / "Entsperrt" entry per segment.                       
                                                                    
Lock state propagates from blueprint to clone via `ProcedureService::copyPlaces()`. Excel import is treated as a system 
operation and is intentionally not enforced.                                                
                                                                                              
### How to review/test                                                                      
                                                                                              
**Tests** (22 backend tests)                                                                                     

**Manual smoke test** (against `ewm`, where the feature is on):                                                                                              
 1. As FPA, mark a workflow place as locked via the place admin (PATCH on `PlaceResourceType.locked`). Confirm the Versionsverlauf side panel shows a "Gesperrt" entry for every segment currently on that place.
                   FE NOT YET IMPLEMENTED - manually lock an existing ```workflow_place``` as locked in your database
 2. As Sachbearbeiter, try to edit any field of a segment on the locked place. Expect a hard reject with the red "Der Abschnitt ist gesperrt" toast.
 3. As Sachbearbeiter, do a bulk edit that includes one or more locked segments. Expect the whole batch to be rejected with the `error.segment.bulk.contains.locked` toast carrying the actual count.
 4. As FPA, do a bulk edit moving one locked segment to an unlocked place + reassigning. Expect success and a "Schritt" + "Bearbeiter" + "Entsperrt" trio in that segment's Versionsverlauf.
 5. Create a new procedure from a blueprint that has a locked place — confirm the clone's place is also locked.

### Linked PRs                                                                                                                                                                            
 - Addon bump: [`demos-europe/demosplan-addon`](https://github.com/demos-europe/demosplan-addon/pull/172) ^0.68.1 adds `PlaceInterface::isLocked()` / `setLocked()` and the matching `BasePlaceResourceConfigBuilder` annotation.

### PR Checklist                                                                            
                                                                                              
  - [x] Create/Update tests
  - [x] Update changelog (entries already in `## UNRELEASED`)
  - [ ] Update documentation (in-code docblocks updated; no separate doc page)                
  - [ ] Add/Update data-cy attributes — N/A (BE only)                                         
  - [ ] Run `yarn lint` — N/A (BE only)                                                       
  - [ ] Run `yarn test` — N/A (BE only)                                                       


sierrazamudiodemos:
- `locked` is registered in `entity_content_change_fields_mapping.yml` like a normal tracked field..
- but its audit rows are written by two dedicated methods in `EntityContentChangeService` (`createSegmentLockedChangeEntryOnPlaceChange` and `createSegmentLockedChangeEntriesForPlaceToggle`) instead of the generic method in the same service (`calculateChangesOfStandardFieldsOfPreUpdateArrayAndPostUpdateObject` / `calculateChangesOfStandardFieldsOfPreUpdateObjectAndPostUpdateArray`), which has to skip the `locked` attribute in 3 places (2 in EntityContentChangeService and 1 in EntityContentChangeDisplayService::getContentChangeComparisonString. 

I think this inconsistency may confuse future devs. I also wonder if there is another way to solve it so that it is more visivle that the locked attribute is different than the others.. or to squeeze it in those 2 generic methods?

demosplan/DemosPlanCoreBundle/EventDispatcher/SegmentLockEnforcementSubscriber.php
```diff
@@ -0,0 +1,105 @@
...
+        // getOriginalEntityData returns the field values as loaded from the
+        // database at the time the entity was hydrated, independent of
+        // whether the Doctrine change set has been computed yet. Using the
+        // change set here fails because BeforeResourceUpdateFlushEvent can
+        // fire before computeChangeSets runs, which leaves the change set
+        // empty and makes the fallback read the already-mutated place.
+        $originalData = $this->entityManager->getUnitOfWork()->getOriginalEntityData($segment);
+        if (isset($originalData['place']) && $originalData['place'] instanceof Place) {
+            return $originalData['place'];
+        }
```
sierrazamudiodemos, line 100: coulld you elaborate a bit more on this one? I try to undrstand why this is needed

MoritzMandler: We enforce against the segment's original place, not its current one. A single PATCH can change the segment's content and its place in one request — if we checked the new place (already unlocked), a non-admin could move a locked segment out and edit it in one shot, bypassing the lock.

We can't get the old place from Doctrine's change set: this event fires before computeChangeSets runs, so the change set is empty and the entity's place is already the mutated one. getOriginalEntityData() returns the values as hydrated from the DB regardless, so it gives us the true pre-PATCH place. The instanceof/fallback just handles entities not yet UoW-tracked (unmutated, so current == original).

sierrazamudiodemos (approved):
thx for doing this! lets cehck in a technix the strategy pattern to solve customFields, locked fields with entity change

PR #6201, opened 2026-04-27
fix login-list cache invalidation

mussbach:
## Summary
- `UserRepository::invalidateCachedLoginList()` was calling `cache->delete(self::LOGIN_LIST_CACHE_DURATION)`, passing the integer `43200` (the TTL) as the cache key. The actual entries are stored under `login_testuser_list{customer}` and `login_testuser_list_osi{customer}`, so the delete was a silent no-op and changes to users took up to 12h to surface on the alternative login page.
- Switch `UserRepository` and `DemosPlanUserAuthenticationController` to `TagAwareCacheInterface`. Tag both write sites with a new `LOGIN_LIST_CACHE_TAG` constant. Replace the broken `delete()` with `invalidateTags()`.

## Test plan
- [ ] After creating a user, the new user appears in the alternative login list on next page load (no 12h wait).
- [ ] After deleting a user, the user is removed from the alternative login list on next page load.
- [ ] `tests/backend/core/User/Functional/UserServiceTest.php` and `tests/backend/core/Security/ApiAuthenticatorTest.php` still pass.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanUserAuthenticationController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanUserAuthenticationController.php
@@ -43,8 +43,8 @@ use Symfony\Component\RateLimiter\RateLimiterFactory;
 use Symfony\Component\Routing\Attribute\Route;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Symfony\Component\Security\Http\Authentication\UserAuthenticatorInterface;
-use Symfony\Contracts\Cache\TagAwareCacheInterface;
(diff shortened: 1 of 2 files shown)


bakridemos (commented):
Tests are working, but after creating a user, the new user does not appears in the login list on next page load. It is not a new Problem but i thought with these changes this will be fixed (i did however reset and repopulate the users index and cleaned the cache but still not there.)

mussbach:
@bakridemos even after you confirmed the double opt-in and set the correct password?

bakridemos (approved):
Verified locally again and with the right password and TagAwareCacheInterface + invalidateTags() correctly evicts the login test-user list on both user creation and deletion.

PR #6205, opened 2026-04-27
fix: real CSRF validation in CsrfSubscriber + IP-keyed anon limiter

mussbach:
## Summary
- `CsrfSubscriber` was treating the submitted token *value* as its *id* and calling `getToken($value)` followed by `isTokenValid()` on the just-fetched token, which always returned true. The central CSRF guard performed no real validation. Now compares the submitted value against the well-known `'csrf'` token id (matching the value emitted by `csrf_token('csrf')` in twig templates), so invalid tokens correctly log "invalid".
- Anonymous statement submission's rate limiter was keyed on `$request->getSession()->getId()`. Switched to `$request->getClientIp()`, matching `userRegisterLimiter` and the password-recovery limiter. Requires `PROXY_IP` to be correctly configured behind the load balancer for per-visitor behaviour — already a deploy-time invariant relied on by the other limiters.

## Notes for reviewers
- Real CSRF blocking still happens at the controller level (`isCsrfTokenValid` in `registerCitizen`, `createOrgaRegister`) and at the firewall (`CsrfTokenBadge` in `LoginFormAuthenticator`). The subscriber is defense-in-depth logging only — that part is unchanged.
- After this change, `dev.log` will start surfacing "CSRF token invalid" entries on admin form posts whose `_token` field carries a Symfony form-component CSRF (different token id, e.g. `boilerplate`, `procedure`). The form component re-validates correctly with the right id and accepts the request — the log entry is noise, not a functional break. Worth knowing if you grep dev logs.

## Test plan
- [ ] POST a form (e.g. login, public statement submit) with a valid `csrf_token('csrf')` — request passes through.
- [ ] POST the same form with a tampered or missing `_token` / `x-csrf-token` — `dev.log` records "CSRF token invalid" or "CSRF token missing".
- [ ] Verify the controller-level CSRF checks still block on bad tokens (citizen registration, orga registration, login).
- [ ] Hit `/verfahren/{procedure}/stellungnahmen/public/neu/ajax` more than `RATELIMIT_ANONYMOUS_STATEMENT_LIMIT` times from the same IP — first N pass, then HTTP 429.
- [ ] On stage, confirm `PROXY_IP` is set to the LB CIDR (smoke: register two accounts from two different external IPs within a minute; both should succeed).

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
@@ -2538,8 +2538,7 @@ class DemosPlanStatementController extends BaseController
         path: '/verfahren/{procedureId}/stellungnahmen/beteilugengsimport',
         name: 'DemosPlan_statement_participation_import',
         options: ['expose' => true],
-        methods: [Request::METHOD_POST])
-    ]
+        methods: [Request::METHOD_POST])]
     public function importParticipationStatements(
         FileService $fileService,
         ProcedureService $procedureService,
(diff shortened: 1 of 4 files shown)


mussbach:
on a second thought this is not worth it and no real gain

PR #6209, opened 2026-04-29
chore: pin d3-color to ^3.1.0 via yarn resolution

hwiem:
Resolves transitive [ReDoS vulnerability](https://github.com/demos-europe/demosplan-core/security/dependabot/4) in d3-color < 3.1.0, pulled in through d3-sankey-diagram → d3-interpolate@1 / d3-transition@1. The 3.x API surface used by these v1 consumers (rgb, interpolateRgb) is backward-compatible.

d3-sankey-diagram has not been maintained for a very long time, and we should either replace it with the official d3-sankey package or implement a new design for displaying the moved statements in the dashboard (I'd prefer the latter, but am not sure how much the feature is used and if it's worth the effort). While this is still being discussed, we can pin the resolution as a quick fix.

### How to review/test
- Open a project that has the `permission feature_statement_move_to_procedure` enabled
- Navigate to the dashboard of a procedure where statements have been moved into or from (i.e., the sankey diagram is displayed)
- Everything should work as before

### DIFF
--- a/package.json
+++ b/package.json
@@ -191,7 +191,8 @@
     "prosemirror-model": "1.25.2",
     "prosemirror-view": "1.40.1",
     "prosemirror-state": "1.4.3",
-    "prosemirror-transform": "1.10.4"
+    "prosemirror-transform": "1.10.4",
+    "d3-color": "^3.1.0"
   },
   "packageManager": "yarn@4.2.2"
 }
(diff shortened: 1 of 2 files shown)


meissnerdemos (approved):
Lgtm, thanks!

PR #6210, opened 2026-04-29
feat(refs: DPLAN-15992): switch input order

meissnerdemos:
### Ticket
DPLAN-15992

Switched the order of two inputs because that makes more sense from a UX perspective.

### How to review/test
Login as admin, go to STN, click details, scroll to submitter, check that the institution input is to the left of the department input.


riechedemos (approved):
Lgtm (code review only)

PR #6217, opened 2026-04-30
perf(HDDP-15): stabilize placeholder collections in StatementMover

mussbach:
## Ticket
HDDP-15

## Summary
- `clone $statement` is shallow: original and clone shared every `PersistentCollection` instance, making placeholder creation timing-dependent and causing full hydration of large `mappedBy` collections (counties, municipalities, priorityAreas) via inverse-side `contains()`.
- Add `Statement::resetCollectionsForPlaceholder()` that reassigns fresh empty `ArrayCollection` instances directly. Severs the shared ref before persist/flush so join-table writes are deterministic and cheap.
- Replace per-element `setCounties([])` / `setMunicipalities([])` / `setPriorityAreas([])` / `setTags([])` / `setVotes([])` / `setFragments([])` / `setFiles([])` chain in `StatementMover::createPlaceholderStatement` with the new helper.

## Why
Previous setter chain iterated the shared collection while calling `$relation->removeStatement($this)` on each element. The remove path hit `$this->statements->contains($statement)` on inverse-side `mappedBy` collections, triggering full hydration of every Statement linked to each related entity. Side effect: order-dependent join-table inserts on flush -- placeholder could end up with leaked refs from the moved statement depending on UoW commit order.

## Test plan
- [ ] CI runs new `testPlaceholderCollectionsAreIndependentFromMovedStatement` -- pins three invariants: distinct collection instances, placeholder collections empty by policy, mutating placeholder does not bleed into moved statement
- [ ] Existing `testMoveStatementCopyToProcedure` still green (already asserts placeholder counts == 0 and moved statement counts preserved)
- [ ] Manual: move a statement with municipalities/counties/priorityAreas; verify placeholder row has zero rows in `_statement_municipality` / `_statement_county` / `_statement_priority_area`; moved statement keeps its full set

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Entity/Statement/Statement.php
+++ b/demosplan/DemosPlanCoreBundle/Entity/Statement/Statement.php
@@ -1067,6 +1067,25 @@ class Statement extends CoreEntity implements UuidEntityInterface, StatementInte
         $this->statementsCreatedFromOriginal = new ArrayCollection();
     }
 
+    /**
+     * Reset collections to fresh empty ones without triggering inverse-side
+     * collection initialization. Used after `clone` for placeholder creation,
+     * where the cloned entity is not yet present in any inverse collection
+     * and the per-element setter -> remove* -> inverse-contains() chain would
+     * needlessly hydrate large mappedBy collections (counties, municipalities,
+     * priorityAreas) on each related join entity.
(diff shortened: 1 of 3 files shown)


bakridemos (approved):
lgtm, tested too thx

PR #6219, opened 2026-05-04
feat(refs: DPLAN-17656): add new heading field

meissnerdemos:
### Ticket
DPLAN-17656

### Description
Frontend for the individual DOCX export header field. Moves the input below the tag filter, switches legacy classes to Tailwind, adds a contextual hint, a 200-character limit with live counter, and a dynamic placeholder that uses the procedure name (full vs. partial export depending on whether tags are filtered).

### How to review/test
1. Log in as admin
2. Open a procedure → Statements list → click Export
3. Choose DOCX export → the "Kopfzeile anpassen" field appears below the tag filter
4. Verify placeholder switches between full/partial export wording when tags are filtered
5. Verify 200-char limit and counter

### PR Checklist
- [x] Run `yarn lint`
- [x] Run `yarn test`


client/js/components/statement/StatementExportModal.vue
```diff
@@ -207,6 +193,35 @@
           </li>
         </ul>
       </fieldset>
+      <fieldset
```
hwiem: While you didn't introduce it, I think it's a good moment to fix this: For a single input, we shouldn't really use a fieldset (because we're not grouping anything); instead, a normal DpInput with label would be fine, and we can also pass the contextual help text as `tooltip` in the `label` prop. DpInput also has a `dataCounter` prop that you can use instead of the custom counter.

client/js/components/statement/StatementExportModal.vue
```diff
@@ -260,15 +275,23 @@ export default {
       type: Boolean,
       default: false,
     },
+
```
hwiem, line 269: ❤️

translations/messages+intl-icu.de.yml
```diff
@@ -964,8 +964,10 @@ docx.export.file_name: "Dateinamen einzelner DOCX-Dateien"
 docx.export.file_name.hint: "Um den Dateinamen dynamisch zu erstellen, verwenden Sie folgende Platzhalter:<br>'{ID}' - die ID der Stellungnahme in EWM<br>'{NAME}' - der Name der einreichenden Person oder Organisation<br>'{EINGANGSNR}' - die vergebene Eingangsnummer<br>Wenn Sie weder '{ID}', '{NAME}' noch '{EINGANGSNR}' verwenden, wird dem Dateinamen eine einzigartige Nummer angehängt."
 docx.export.file_name.placeholder: "'{ID}'_'{NAME}'_'{EINGANGSNR}'"
 docx.export.filtered: "<p>Angewendete Schlagwortfilter: </p> <br>"
-docx.export.header.custom: "Individueller Kopfzeilentext"
-docx.export.header.custom.placeholder: "Bitte geben Sie hier Ihren individuellen Titel ein"
+docx.export.header.custom: "Kopfzeile anpassen"
+docx.export.header.custom.hint: "Der individuelle Kopfzeilentext erscheint in der DOCX-Datei. Maximal 200 Zeichen."
```
hwiem: For "Spaltennamen in DOCX-Exporten", we display the contextual help text "Um die Spaltennamen auf die Ursprungswerte zurückzusetzen, entfernen Sie die Einträge in den Textfeldern." I wonder if we should add something like that here, too?

translations/messages+intl-icu.de.yml
```diff
@@ -964,8 +964,10 @@ docx.export.file_name: "Dateinamen einzelner DOCX-Dateien"
...
+docx.export.header.custom.placeholder.full: "Export Synopse {procedureName}"
```
hwiem: Note: We could do something like this to have only one translation key for the two cases:
```
docx.export.header.custom.placeholder |
  {isPartialExport, select,     
      true {Teilexport Synopse {procedureName}}                                 
      other {Export Synopse {procedureName}}
  }
```
and use it like this:

```
customHeaderPlaceholder () {
      return Translator.trans('docx.export.header.custom.placeholder', { isPartialExport: this.selectedTagIds.length > 0, procedureName: this.procedureName })
},
```
Both approaches have their advantages, so you can also leave it as it is - just wanted to mention it.

tests/frontend/unit/StatementExportModal.spec.js
```diff
@@ -158,6 +159,23 @@ describe('StatementExportModal', () => {
     expect(exportEvent).toEqual(payload)
   })
 
+  it('emits export event with customHeaderText when set', () => {
```
hwiem, line 162: Thanks for updating the test. ❤️ 
I don't want to be greedy 😅 but I think we could also test if the correct placeholder is displayed (partial/full) and if the input is only displayed if `active === 'docx_normal' && !isSingleStatementExport && hasPermissionAdjustPreamble`.

hwiem (changes requested):
Thanks for using tailwind classes ❤️ I think we should fix the fieldset issue, the rest are minor improvement suggestions

client/js/components/statement/StatementExportModal.vue
meissnerdemos: Thank you for the hint about the props! changed it in 40d8abebc7c50f8e69a88cf0e97d636967093d83

translations/messages+intl-icu.de.yml
meissnerdemos: yes! changed it in c16c86ad49ca699382ced5c4d7247ef2735ed154

translations/messages+intl-icu.de.yml
meissnerdemos: I like your version c16c86ad49ca699382ced5c4d7247ef2735ed154

tests/frontend/unit/StatementExportModal.spec.js
meissnerdemos: I tried with claude e5ed38608db298356adf6f7f940c0bed2b5a5ba4

hwiem: Thanks :) Just some thoughts on that: I think it's cleaner to have only one (or two related) assertions in each test case; that way, if it fails, it's clear what failed. Here, it could be that the input is not displayed even though the permission is enabled, or the input is displayed even though xlsx export is selected, or the input is displayed even though single statement export is selected. 
Also, I'd test the actual translation that is displayed to the user, rather than which parameter the Translator has been called with, but I'm not sure one is really more valid than the other, so it's not a change request, just a thought. :)

hwiem (approved):
Lgtm + works

PR #6220, opened 2026-05-04
feat(refs: DPLAN-17619): add text custom field to institution tag management

huellnerdemos:
### Ticket
DPLAN-17619

This PR adds the text custom field type and the create custom field definition dialog as a tab to the institution tag management. It makes it possible to create text custom field definitions in this tab and list the created definitions. 

### How to review/test

1. Login as MA
2. Switch to institution tag management dialog
3. Choose tab "Freitextfelder definieren"
4. Click on button "Neues Textfeld" and create a new text field definition
5. Check if the new definition can be saved and is listed in the "Institutionsfelder" list

### Linked PRs

BE-PR: [6182](https://github.com/demos-europe/demosplan-core/pull/6182)

### PR Checklist

- [x] Run `yarn lint`
- [x] Update changelog


client/js/components/procedure/admin/AdministrationCustomFieldsList.vue
```diff
@@ -364,15 +366,37 @@ export default {
       default: false,
     },
 
-    procedureId: {
-      type: String,
-      required: true,
+    definitionSourceId: {
+      type: [String, null],
+      required: false,
+      default: null,
```
riechedemos: type: [String, null] is not a valid Vue prop type. null is not a constructor; in Vue 3 this is effectively interpreted as "no type check",  which silently disables type validation. Correct form:

  type: String,                                                                                                                                                                               
  required: false,
  default: null,

riechedemos (approved):
Thank you it works just a small non blocking thing. You introduce edit and delete button (but it is not in the scope of this story). Maybe it would be good to add a comment why it is there but disabled.

huellnerdemos:
> Thank you it works just a small non blocking thing. You introduce edit and delete button (but it is not in the scope of this story). Maybe it would be good to add a comment why it is there but disabled.

Thanks for your review! :tada:  The editing and deleting are other parts of the same epic, so it will come in the next stories. Since i reused the component for listing the custom field definition here, and it already has this buttons, i just disabled it for now. I added a comment for it in 9b27994a797f610fc718c846ecd6efabbe16ca9f
Thanks for this point.

PR #6228, opened 2026-05-06
chore(db): add unique index on customer.subdomain

mussbach:
## Summary
- Add UNIQUE index `uniq_customer_subdomain` on `customer._c_subdomain` to enforce subdomain uniqueness at the DB level and speed up `CustomerRepository::findCustomerBySubdomain` lookups.
- Migration pre-flight checks for duplicate subdomain values and aborts with a clear message if any exist, so duplicates can be resolved before the index is created.

## Test plan
- [ ] `bin/console d:m:diff` shows no further drift after running the migration
- [ ] `bin/console d:m:migrate` applies cleanly on a fresh DB
- [ ] On a DB with a synthetic duplicate `_c_subdomain`, migration aborts with the duplicate-detection message
- [ ] After migration, attempting to insert a second customer with an existing subdomain fails with a unique-constraint violation


sierrazamudiodemos:
I adjusted this test: 669852ddcd620ae9cbb9c3b298aa62a0c984a047 because it was colliding with the unique index introduced here (the test was using an existing subdomain which now should be unique)

sierrazamudiodemos (approved):
I adjusted that was failing because of the subdomain being unique. now PR should be ready to merged

PR #6235, opened 2026-05-07
feat(refs: DPLAN-17697): add inactivity-based account deletion workflow

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17697/Loschkonzept-Automatischer-Userlosch-WF-bei-Inaktivitat

### Description
A daily cron walks users whose `lastLogin` (or `createdDate` for legacy users) lies past a  configured threshold and
progresses them through a three-step cascade:
 - first warning mail → 
 - second warning mail → 
 - soft-delete + final notification.

firstWarningMail
lastWarningMail
AccountDeleted

Each stage gates on the previous mail having been queued so the cascade never skips a step.
Exceptions are legacy users who are past the project configured warning period plus both mail-action periods → deletion without warning mails.
The next successful login removes any tracking row and resets the user back to "active".
Disabled by default — enable per project by setting `account_deletion.first_warning_days`.

### Configuration                                                                           
 - `account_deletion.first_warning_days` — inactivity threshold; unset disables the feature
 - `account_deletion.warning_step_days` (default 30) — gap between first warning, second warning and deletion                                                                        
 - `account_deletion.additional_protected_user_ids` — IDs the cron must never touch (anonymous user and AI API user are always protected)                                       
 - `ACCOUNT_DELETION_HOMEPAGE_URL_TEMPLATE` env var — optional URL inserted into warning mails; `${subdomain}` placeholder for multi-tenant deployments 

### How to review/test
 - Run unit + functional tests: `phpunit
tests/backend/core/User/Unit/AccountDeletionStepsTest.php 
tests/backend/core/User/Unit/LastLoginActivityCheckerTest.php 
tests/backend/core/MessageHandler/AccountDeletionRunMessageHandlerTest.php 
tests/backend/core/Core/Functional/UserLoginSubscriberTest.php`
 - Apply migration on a project DB and verify `account_deletion_tracking` exists with FK constraints
 - Try out the test-command `bin/<project> dplan:account-deletion:prepare-test` and confirm it works as intended.
```bin/<project> messenger:consume scheduler_maintenance --no-debug```
```bin/<project> messenger:consume scheduler_daily_maintenance```
both need to run for the immediate path of the command to work. then check MailPit http://localhost:8025/
 - Relog with a user who's last login lies within the firstWarningMail and the userDeletion window and verify its Tracking entity got wiped and the removal process has reset.
---- 

- [x] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### CODE REVIEW

File: demosplan/DemosPlanCoreBundle/DataFixtures/ORM/ProdData/LoadMailTemplateData.php [hunk 1]
```diff
@@ -115,6 +115,24 @@ public function load(ObjectManager $manager): void
                 'title'    => 'Abgelehnte Freischaltung als ${orga_type} für das Bundesland ${customer} auf bauleitplanung-online.de',
                 'content'  => "Sehr geehrte/r \${firstname} \${lastname},\r\n\r\nSie haben vor kurzer Zeit eine Freischaltung Ihrer Organisation, \${orga_name}, als \${orga_type} für das Bundesland \${customer} beantragt. Ihre Anfrage wurde durch die dort zuständige Stelle geprüft und abgelehnt.\r\n\r\nSollten Sie vermuten, dass hier ein Fehler vorliegt, wenden Sie sich bitte an den Support (support@bauleitplanung-online.de).\r\n\r\nViele Grüße,\r\n\r\nIhr Team von bauleitplanung-online.de\r\n",
             ],
+            [ // row #16
```
mussbach: Typo in Supportteam and MailTemplates are deprecated. You should better use some generic template like dm_stellungnahme albeit this does not match the naming and inject subject and body via translation message

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/05/Version20260506150528.php
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+class Version20260506150528 extends AbstractMigration
```
mussbach: see above, i would drop this

demosplan/DemosPlanCoreBundle/Entity/User/AccountDeletionTracking.php
```diff
@@ -0,0 +1,118 @@
...
+use DateTime;
+use DemosEurope\DemosplanAddon\Contracts\Entities\UuidEntityInterface;
+use demosplan\DemosPlanCoreBundle\Entity\MailSend;
+use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Mapping\Annotation as Gedmo;
+
+/**
+ * @ORM\Entity(repositoryClass="demosplan\DemosPlanCoreBundle\Repository\AccountDeletionTrackingRepository")
+ */
+class AccountDeletionTracking implements UuidEntityInterface
```
mussbach, line 22: Wha do you need this table for? Wouldnt it enough to just calculate the dates by last login?

demosplan/DemosPlanCoreBundle/EventSubscriber/UserLoginSubscriber.php
```diff
@@ -16,32 +16,57 @@
...
-use Symfony\Component\Security\Core\Event\AuthenticationSuccessEvent;
+use Symfony\Component\Security\Http\Event\LoginSuccessEvent;
 
 class UserLoginSubscriber extends BaseEventSubscriber
 {
-    public function __construct(private readonly UserService $userService)
-    {
+    public function __construct(
+        private readonly UserService $userService,
+        private readonly AccountDeletionTrackingRepository $trackingRepository,
```
mussbach, line 28: Without tracking table not needed

mussbach (changes requested):
Before I go on with the review: What do you need the AccountDeletionTracking Entity and table for? It seems a bit overengineered. You could calculate the dates from last_login and save as report that mail is sent. But I am curious about your reasons

demosplan/DemosPlanCoreBundle/Entity/User/AccountDeletionTracking.php
MoritzMandler: In principle, this could be solved without, but only at significant cost:
 - the daily job would have to query the _mail_send table on every run, parse the warning date out of each found mail to confirm it isn't a stale historical send, and reconstruct the user's progression state from the matching  mails.
 - And it falls apart entirely once _mail_send entries are only retained for a configurable period of time.
As soon as a user's Warning1 mail ages out of the table, the cron loses the only signal that warning 1 ever happened.
A dedicated tracking row sidesteps all of that.

mussbach: I agree that scanning _mail_send is a bad idea, but why not just use User::last_login?

MoritzMandler: last_login only tells us how stale a user is — it doesn't tell us what the cron has already done about that  staleness.
We would rely on daily cron jobs running without any complications ever to just use 30Days exactly 60Days exactly as reason to deliver the mail. (Only ever 1 attempt)

demosplan/DemosPlanCoreBundle/DataFixtures/ORM/ProdData/LoadMailTemplateData.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6235/commits/61fae8590f95c242fe3179a2508de465d2afff2f

demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/05/Version20260506150528.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6235/commits/61fae8590f95c242fe3179a2508de465d2afff2f

demosplan/DemosPlanCoreBundle/Entity/User/AccountDeletionTracking.php
mussbach: hm, true that. It is more resilient

demosplan/DemosPlanCoreBundle/Logic/User/LastLoginActivityChecker.php
```diff
@@ -78,4 +75,130 @@ public function setDayThreshold(int $dayThreshold): void
...
+
+        if (!$secondWarningSent && $daysInactive >= $secondWarningDays) {
+            $steps[] = AccountDeletionStep::SendSecondWarning;
+        }
+
+        if ($daysInactive >= $deletionAfterDays) {
+            $steps[] = AccountDeletionStep::Delete;
+        }
+
+        return $steps;
```
mussbach: Probably only the latest step should win. A user that gets the account deleted should not get two emails that doesn't change anything as the user cannot log in to avoid deleting. Using a string instead of an array should already do the trick

demosplan/DemosPlanCoreBundle/Logic/User/LastLoginActivityChecker.php
```diff
@@ -78,4 +75,130 @@ public function setDayThreshold(int $dayThreshold): void
...
+    }
+
+    private function isProtectedSystemUser(UserInterface $user): bool
+    {
+        if (UserInterface::ANONYMOUS_USER_ID === $user->getId()) {
```
mussbach: you can use $user->isDefaultGuestUser() instead

demosplan/DemosPlanCoreBundle/MessageHandler/AccountDeletionRunMessageHandler.php
```diff
@@ -0,0 +1,420 @@
...
+        $this->initializeAnonymousUserPermissions();
+
+        try {
+            $firstWarningDays = $this->readIntParam('account_deletion.first_warning_days');
+
+            if (null === $firstWarningDays) {
+                return;
+            }
+
+            $candidates = $this->trackingRepository->findInactivityDeletionCandidates(
```
mussbach: Shouldnt this method live in UserRepository as it queries users?

demosplan/DemosPlanCoreBundle/Logic/User/LastLoginActivityChecker.php
```diff
@@ -78,4 +75,130 @@ public function setDayThreshold(int $dayThreshold): void
...
+        ?AccountDeletionTracking $tracking,
+        int $firstWarningDays,
+    ): array {
+        $stepDays = $this->readIntParam('account_deletion.warning_step_days') ?? 30;
+        $secondWarningDays = $firstWarningDays + $stepDays;
+        $deletionAfterDays = $firstWarningDays + 2 * $stepDays;
+
+        $lastLogin = $user->getLastLogin();
+
+        if (!$lastLogin instanceof DateTimeInterface) {
```
mussbach, line 117: for a smooth transition of old entries, could you please add a check that also DeleteWithoutWarnings users that have a lastLogin date and $deletionAfterDays would immediately kick? We would avoid strange mails for that users and newer users would have the new workflow

demosplan/DemosPlanCoreBundle/Logic/User/LastLoginActivityChecker.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6235/commits/93e2820c582e5bd55e70f313e755a43d32274dd5

 - evaluateInactivitySteps → evaluateInactivityStep (singular), returns `?AccountDeletionStep;` rollout-era users emit a single most-advanced step per cron run instead of a cascade of three emails the user can't act on.
 - cascading ifs weakest-to-strongest, last assignment wins; the W1 branch is guarded by `!$secondWarningSent` so a user who jumped straight to W2 on day N (no prior tracking) doesn't regress to W1 on day N+1
 - processCandidate becomes a single match on $step; dispatch methods return void (no chaining via the tracking row)
 - rewrite the two compression tests in AccountDeletionStepsTest to assert single-step jump /(SendSecondWarning, Delete); add a regression test that locks down the day-N+1 "no regress to W1" case

demosplan/DemosPlanCoreBundle/Logic/User/LastLoginActivityChecker.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6235/commits/1492b54d3d76a5a69694ff76e84b872f2e1a7176

demosplan/DemosPlanCoreBundle/MessageHandler/AccountDeletionRunMessageHandler.php
MoritzMandler: Yes, thats a better fit: https://github.com/demos-europe/demosplan-core/pull/6235/commits/023446bad5c237e4540a970ced5129676e36f9de

demosplan/DemosPlanCoreBundle/Logic/User/LastLoginActivityChecker.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6235/commits/7525a5e10787d8fa53662e8e58cbd452f06e2c6f

CHANGELOG.md
```diff
@@ -46,6 +47,9 @@
 - LaTeX PDF exports rendered more reliably (ligature and babel-related gaps)
 
 ### Deployment notes
+- **Migration**: creates `account_deletion_tracking` table and seeds three mail templates (`account_deletion_warning_first`, `account_deletion_warning_second`, `account_deletion_completed`) — run `doctrine:migrations:migrate`
```
mussbach: This is now misleading after the changes

demosplan/DemosPlanCoreBundle/MessageHandler/AccountDeletionRunMessageHandler.php
```diff
@@ -0,0 +1,412 @@
...
+            return '';
+        }
+
+        $url = str_replace('${subdomain}', $subdomain, $template);
+
+        // Wrap the URL in an <a> tag so it renders as a clickable link in the HTML
+        // mail variant. MailService runs nl2br() on the content for the HTML body
+        // and falls back to a markdown-stripped plain-text variant; both keep the
+        // URL intact and most plain-text clients auto-linkify bare URLs anyway.
+        return sprintf("\n\nSie erreichen unsere Plattform unter:\n<a href=\"%s\">%s</a>", $url, $url);
```
mussbach: please use a trans key

mussbach:
**Could `ACCOUNT_DELETION_HOMEPAGE_URL_TEMPLATE` be dropped and the link derived instead?**

The handler already injects `GlobalConfigInterface` and already resolves the per-user subdomain in `resolveCustomerSubdomain()`. `GlobalConfig` exposes everything the link needs:

- `getUrlScheme()` → `https`
- `getProjectDomain()` → e.g. `demos-plan.de`
- `getUrlPathPrefix()` → optional path

And there's already an established convention for building the host this exact way — `customer_settings.html.twig` does `[customer.subdomain, projectDomain]|join('.')`, fed by `DemosPlanCustomerController` via `getProjectDomain()`.

So `computeLinkSection()` could build the URL directly:

```php
$domain = $this->globalConfig->getProjectDomain();
if ('' === $domain || null === $subdomain) {
    return ''; // same "omit link" behaviour as today
}
$url = sprintf(
    '%s://%s.%s%s/',
    $this->globalConfig->getUrlScheme(),
    $subdomain,
    $domain,
    $this->globalConfig->getUrlPathPrefix(),
);
```

That removes the env var, the `account_deletion.homepage_url_template` parameter, the `${subdomain}` placeholder and the `str_replace`. The scheme/domain values are deployment-wide and already have to be correct for every other absolute-URL mail and the customer-settings page, so there's nothing extra to configure — and the link can't drift out of sync with the real login host.

The only case the template buys us is pointing the mail at a *different* homepage than the app host, but since the mail's purpose is "log in to keep your account," the login host (`{subdomain}.{projectDomain}`) is what we want anyway. Empty `project_domain` is a misconfiguration rather than a supported state, and returning `''` there preserves the current behaviour of just omitting the link.

mussbach (changes requested):
looks mainly very well some smaller concerns

CHANGELOG.md
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6235/commits/367dce9b7ca9dbf0bfef21e7276b33cde039a535

demosplan/DemosPlanCoreBundle/MessageHandler/AccountDeletionRunMessageHandler.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6235/commits/367dce9b7ca9dbf0bfef21e7276b33cde039a535

MoritzMandler:
> **Could `ACCOUNT_DELETION_HOMEPAGE_URL_TEMPLATE` be dropped and the link derived instead?**
> 
> The handler already injects `GlobalConfigInterface` and already resolves the per-user subdomain in `resolveCustomerSubdomain()`. `GlobalConfig` exposes everything the link needs:
> 
>     * `getUrlScheme()` → `https`
> 
>     * `getProjectDomain()` → e.g. `demos-plan.de`
> 
>     * `getUrlPathPrefix()` → optional path
> 
> 
> And there's already an established convention for building the host this exact way — `customer_settings.html.twig` does `[customer.subdomain, projectDomain]|join('.')`, fed by `DemosPlanCustomerController` via `getProjectDomain()`.
> 
> So `computeLinkSection()` could build the URL directly:
> 
> ```
> $domain = $this->globalConfig->getProjectDomain();
> if ('' === $domain || null === $subdomain) {
>     return ''; // same "omit link" behaviour as today
> }
> $url = sprintf(
>     '%s://%s.%s%s/',
>     $this->globalConfig->getUrlScheme(),
>     $subdomain,
>     $domain,
>     $this->globalConfig->getUrlPathPrefix(),
> );
> ```
> 
> That removes the env var, the `account_deletion.homepage_url_template` parameter, the `${subdomain}` placeholder and the `str_replace`. The scheme/domain values are deployment-wide and already have to be correct for every other absolute-URL mail and the customer-settings page, so there's nothing extra to configure — and the link can't drift out of sync with the real login host.
> 
> The only case the template buys us is pointing the mail at a _different_ homepage than the app host, but since the mail's purpose is "log in to keep your account," the login host (`{subdomain}.{projectDomain}`) is what we want anyway. Empty `project_domain` is a misconfiguration rather than a supported state, and returning `''` there preserves the current behaviour of just omitting the link.

nice :+1:  done in https://github.com/demos-europe/demosplan-core/pull/6235/commits/367dce9b7ca9dbf0bfef21e7276b33cde039a535

mussbach (approved):
lgtm, code review only. Thank you for this nice feature

PR #6245, opened 2026-05-11
fix(refs ADO-54897): replace all {currentYear} placeholders in map copyright

mussbach:
## Summary
Public procedure map attribution only substituted the first `{currentYear}` occurrence in the configured copyright string, so a second use of the placeholder (e.g. when citing multiple map sources) stayed unreplaced. Switched to `replaceAll` to match the sibling `DpOlMapLayer` behavior.

Ref: ADO 54897

## Test plan
- [ ] Configure a procedure copyright/source text containing two `{currentYear}` occurrences
- [ ] Open the public procedure detail map and confirm both occurrences render the current year


hwiem (approved):
Lgtm, code review only

PR #6246, opened 2026-05-11
feat(refs DPLAN-17565): render pdf_importer_image anchors inline in statement text

mussbach:
## Summary
DPLAN-17565

- PI emits image references in segment text as `<a class="pdf_importer_image" href="…">label</a>` so DOCX export can extract them via the anchor class selector. Until now the frontend showed them as plain link text and the umwandeln editor stripped them entirely.
- Frontend swaps anchor → `` at display time (`inlineImageAnchors` helper) and routes the umwandeln DpEditor through it; DpEditor's `imageButton` is enabled so `CustomImage` actually parses the resulting `` instead of dropping it as unknown.
- SegmentationEditor (drafts-list) gets a ProseMirror `markView` for the `link` mark that renders the anchor as inline `` while keeping the text label in a hidden `contentDOM` span — PM doc size and segment char offsets unchanged.
- Backend: `HtmlHelper` now accepts both the legacy anchor form (still produced by PI) and any `` in segment text. Class is only required to disambiguate `<a>` from real hyperlinks; for `` it's redundant.

## Test plan
- [x] Backend: `HtmlHelperTest` (8/8) and `ImageLinkConverterTest` (6/6) including new img-form cases
- [x] Frontend: jest `inlineImageAnchors.test.js` (9/9)
- [x] PHPStan level 5 clean on changed PHP
- [x] Manual: image renders inline in einwendungen list, abschnitte (segment views), drafts-list, and umwandeln editor; round-trip save preserves image

### DIFF
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -355,7 +355,7 @@
         v-model="values.text"
         :aria-label="Translator.trans('statement.text.short')"
         :procedure-id="procedureId"
-        :toolbar-items="{ linkButton: true }"
+        :toolbar-items="{ linkButton: true, imageButton: true }"
         required
         hidden-input="r_text"
       />
--- a/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
+++ b/client/js/components/procedure/DpSimplifiedNewStatementForm.vue
@@ -356,6 +356,7 @@
         :aria-label="Translator.trans('statement.text.short')"
         :procedure-id="procedureId"
         :toolbar-items="{ linkButton: true, imageButton: true }"
+        :tus-endpoint="dplan.paths.tusEndpoint"
         required
         hidden-input="r_text"
       />
--- a/client/js/components/statement/splitStatement/SegmentationEditor.vue
+++ b/client/js/components/statement/splitStatement/SegmentationEditor.vue
@@ -131,6 +131,30 @@ export default {
           doc: parsedContent,
           plugins: rangePlugin.plugins,
         }),
+        markViews: {
+          link: (mark) => {
+            const className = mark.attrs.class || ''
+            const anchor = document.createElement('a')
+            anchor.setAttribute('href', mark.attrs.href)
+            if (className) {
+              anchor.setAttribute('class', className)
+            }
+
+            if (!className.split(/\s+/).includes('pdf_importer_image')) {
+              return { dom: anchor, contentDOM: anchor }
+            }
+
+            const img = document.createElement('img')
+            img.setAttribute('src', mark.attrs.href)
+            img.setAttribute('alt', '')
+            img.setAttribute('loading', 'lazy')
+            const label = document.createElement('span')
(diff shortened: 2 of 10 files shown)


tests/frontend/unit/inlineImageAnchors.test.js
```diff
@@ -0,0 +1,77 @@
+/**
```
hwiem, line 1: Nice test. ❤️

client/js/components/procedure/DpSimplifiedNewStatementForm.vue
```diff
@@ -355,7 +355,7 @@
         v-model="values.text"
         :aria-label="Translator.trans('statement.text.short')"
         :procedure-id="procedureId"
-        :toolbar-items="{ linkButton: true }"
+        :toolbar-items="{ linkButton: true, imageButton: true }"
```
hwiem, line 358: When using `DpEditor` with `imageButton: true`, we also need to pass a `tusEndpoint`. Otherwise, when you click the image button, nothing happens and there's an error in the console. So we need to add this:
`:tus-endpoint="dplan.paths.tusEndpoint"`

client/js/components/statement/splitStatement/SegmentationEditor.vue
```diff
@@ -131,6 +131,30 @@ export default {
...
+            const anchor = document.createElement('a')
+            anchor.setAttribute('href', mark.attrs.href)
+            if (className) {
+              anchor.setAttribute('class', className)
+            }
+
+            if (!className.split(/\s+/).includes('pdf_importer_image')) {
+              return { dom: anchor, contentDOM: anchor }
+            }
+
```
hwiem, line 146: If I understand correctly, we can't avoid placing the image inside an `<a>` element. But I'd suggest to open the link in a new tab if the user clicks it:  
```
anchor.setAttribute('target', '_blank')
anchor.setAttribute('rel', 'noopener noreferrer')
```

client/js/components/statement/splitStatement/SegmentationEditor.vue
```diff
@@ -131,6 +131,30 @@ export default {
...
+            const img = document.createElement('img')
+            img.setAttribute('src', mark.attrs.href)
+            img.setAttribute('alt', '')
+            img.setAttribute('loading', 'lazy')
+            const label = document.createElement('span')
+            label.hidden = true
```
hwiem: To make it accessible, we shouldn't just hide the label, but make it available to screen readers:

```suggestion
            const label = document.createElement('span')
            label.className = 'sr-only'
```

hwiem (changes requested):
1 small issue and 2 suggestions, FE code review only. Thanks for adding the test, it's highly appreciated :)

client/js/components/statement/splitStatement/SegmentationEditor.vue
mussbach: thank you for questioning that, this can be simplfied c4f8fc580

hwiem (approved):
Lgtm, thank you!

PR #6249, opened 2026-05-12
fix(DPLAN-17759): unsaved changes modal triggers after save

meissnerdemos:
### Ticket
DPLAN-17759

### Description
`hasUnsavedChanges` in `StatementEntry`, `StatementSubmitter` and `StatementMetaAttachments` compared the full attribute object via `JSON.stringify`. After a save, the store re-builds the statement resource (different key order, server-canonicalized values, added fields like `modifyDate`) while the local clone stays untouched, so the comparison flagged a diff even though every edited field matched. Editing the text editor below `StatementMeta` mutated unrelated attributes on the same record and produced the same false positive. Switched each component to compare only the fields it actually edits.

### How to review/test
1. Log in as admin
2. Open a procedure → Stellungnahmen → open a statement → Details
3. Edit a submitter field (e.g. email), click the form save button
4. Click a menu link to navigate away — modal must NOT appear
5. Edit the field again, do NOT save, click a link — modal MUST appear
6. Repeat for the Eintrag section (intern ID, memo, dates, submit type)
7. Repeat for attachments: add file → save → navigate (no modal); mark for delete → save → navigate (no modal)
8. Edit text in the editor below `StatementMeta`, save it, navigate — modal must NOT appear

### PR Checklist
- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementEntry.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementEntry.vue
@@ -245,13 +245,29 @@ export default {
         return false
       }
 
-      const initialAttributes = this.deepCloneSerializable(this.statement.attributes)
-      initialAttributes.authoredDate = this.getFormattedDate(initialAttributes.authoredDate)
-      initialAttributes.submitDate = this.getFormattedDate(initialAttributes.submitDate)
-
+      const initialAttributes = this.statement.attributes
       const currentAttributes = this.localStatement.attributes
 
-      return JSON.stringify(currentAttributes) !== JSON.stringify(initialAttributes)
+      if (this.getFormattedDate(currentAttributes.authoredDate) !== this.getFormattedDate(initialAttributes.authoredDate)) {
+        return true
+      }
+      if (this.getFormattedDate(currentAttributes.submitDate) !== this.getFormattedDate(initialAttributes.submitDate)) {
+        return true
+      }
+      if ((currentAttributes.submitType ?? '') !== (initialAttributes.submitType ?? '')) {
+        return true
+      }
+      if ((currentAttributes.internId ?? '') !== (initialAttributes.internId ?? '')) {
+        return true
+      }
+      if (hasPermission('field_statement_phase') && (currentAttributes.procedurePhase?.key ?? '') !== (initialAttributes.procedurePhase?.key ?? '')) {
+        return true
+      }
+      if (hasPermission('field_statement_memo') && (currentAttributes.memo ?? '') !== (initialAttributes.memo ?? '')) {
+        return true
+      }
+
+      return false
     },
 
     isStatementManual () {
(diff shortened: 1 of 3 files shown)


client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementEntry.vue
```diff
@@ -245,13 +245,29 @@ export default {
...
 
-      const initialAttributes = this.deepCloneSerializable(this.statement.attributes)
-      initialAttributes.authoredDate = this.getFormattedDate(initialAttributes.authoredDate)
-      initialAttributes.submitDate = this.getFormattedDate(initialAttributes.submitDate)
-
+      const initialAttributes = this.statement.attributes
       const currentAttributes = this.localStatement.attributes
 
-      return JSON.stringify(currentAttributes) !== JSON.stringify(initialAttributes)
+      if (this.getFormattedDate(currentAttributes.authoredDate) !== this.getFormattedDate(initialAttributes.authoredDate)) {
```
sakutademos: thx a lot for fixing it! I just think we could improve the readability a bit, because its a bit difficult to read at first sight. for example, we could use an isDifferent helper function, something like this:
const isDifferent = (a, b) => (a ?? '') !== (b ?? '')
and then return something like this:
```
return [
        isDifferentDate(current.authoredDate, initial.authoredDate),
        isDifferentDate(current.submitDate, initial.submitDate),
        isDifferent(current.submitType, initial.submitType),
        isDifferent(current.internId, initial.internId),
        hasPermission('field_statement_phase') &&
        isDifferent(current.procedurePhase?.key, initial.procedurePhase?.key),
        hasPermission('field_statement_memo') &&
        isDifferent(current.memo, initial.memo)
      ].some(Boolean)
```

or similar to what you did in the StatementSubmitter.vue component with `stringFields`

sakutademos (commented):
thx a lot for fixing it, just a few nitpicks, overall works nicely!!!

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementEntry.vue
meissnerdemos: Good idea! I used your suggestion 5b4ee819d18a9db9be70a94a25df3ccac60b4e27

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementSubmitter.vue
```diff
@@ -244,7 +244,23 @@ export default {
...
+        'authorName',
+        'submitName',
+        'submitterEmailAddress',
+        'initialOrganisationStreet',
+        'initialOrganisationHouseNumber',
+        'initialOrganisationPostalCode',
+        'initialOrganisationCity',
+      ]
+
+      if (stringFields.some(field => (currentAttributes[field] ?? '') !== (initialAttributes[field] ?? ''))) {
```
sakutademos: ok, just a thought, but maybe we could use the same approach (`isDifferent` helper function) here as well? (:

sakutademos (approved):

client/js/components/procedure/StatementSegmentsList/StatementMeta/StatementSubmitter.vue
meissnerdemos: Good idea, applied in b3916ee7d

PR #6233, opened 2026-05-07
feat(refs: DPLAN-17223) manual tag ordering within topics in TagsList

riechedemos:
### Ticket
[DPLAN-17223](https://demoseurope.youtrack.cloud/issue/DPLAN-17223/Ado-issue-48739-Sortierte-Schlagwortliste)

### Description
Enable manual sorting of tags within and across TagTopics via drag and drop in the tag administration view. Reordering is persisted through the new `tagList.reorder` RPC (introduced by backend PR #6250), replacing the previous behaviour where tags were appended to the end when moved between topics.

Approach:
- Both intra-topic and cross-topic moves commit via DpTreeList's `@end` event with optimistic store update + RPC call.
- Cross-topic moves derive the target topic from `event.to.id` (DOM id of the Sortable container, set by DpDraggable's `groupId` binding).
- `Tag.sortIndex` is added to the JSON:API sparse fieldset so the local store stays in sync with the backendnd.

### How to review/test
1. Open the tag administration (Schlagworte) view of a procedure.
2. Drag a tag within the same TagTopic to a new position → reorder persists after reload.
3. Drag a tag into a different TagTopic, drop precisely between two existing tags → tag lands at that position and persists.
4. Drop on a collapsed TagTopic header → tag is appended at the end (expected limitation when the target list is hidden).

### Linked PRs
- Backend: #6250
- Demosplan-ui: https://github.com/demos-europe/demosplan-ui/pull/1488 

### PR Checklist
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [x] Link DPLAN-17679 ticket
- [x] Move ticket to PR Review

PR #6250, opened 2026-05-12
feat(DPLAN-17223): manual tag ordering within topics: Tag.sortIndex +

winderdemos:
### Ticket
DPLAN-17223

**Description:**

Backend slice for sortable tags. Adds a persistent `sortIndex` on the `Tag` entity so tags within a topic can be ordered manually instead of being sorted alphabetically by title. A new JSON-RPC method `tagList.reorder` moves a tag to a new position within its topic or to a different topic and re-indexes the affected siblings server-side. As a side effect, the segment-tag facet and every other consumer of `TagTopic::$tags` now sorts by `sortIndex` because the Doctrine `OrderBy` on that collection was switched from `title ASC` to `sortIndex ASC`.

Tag topics themselves stay alphabetically ordered — only tags within a topic are sortable.

**Highlights:**

- New column `_t_sort_index` on `_tag` (INT NOT NULL DEFAULT 0); migration backfills by `createDate ASC` partitioned per topic so existing data keeps its current visible order.
- `Tag::$sortIndex` getter/setter; `TagTopic::$tags` `OrderBy` switched to `sortIndex ASC`.
- `TagRepository::getNextSortIndex(topicId)` — used by `TagService::createTag` and by the JSON:API creation behaviour so newly created tags always append to the end of their topic.
- `TagTopicRepository::copy()` preserves the source `sortIndex` when cloning tags between procedures.
- `TagResourceConfigBuilder` extends the autogenerated base builder to declare `sortIndex`; `TagResourceType` exposes it as readable / sortable / filterable.
- `RpcTagListReorderer` implements `tagList.reorder`, gated by `area_admin_statements_tag`, non-transactional, handles same-topic and cross-topic moves, returns updated `sortIndex` and `topicId` for every affected tag.
- JSON schema at `config/json-schema/rpc-tag-list-reorder-schema.json`.

### How to review/test

1. Apply migrations:
   ```
   bin/<project> doctrine:migrations:migrate
   ```
   Confirm `_tag._t_sort_index` exists and contains contiguous `[0, n−1]` values per topic.

2. **JSON:API read.** `GET /api/2.0/Tag?fields[Tag]=title,sortIndex&sort=sortIndex` returns tags ordered by `sortIndex`. Listing tags for a topic via `TagTopic.tags` also returns them in `sortIndex` order out of the box.

3. **JSON:API create.** `POST /api/2.0/Tag` with a `topic` relationship — confirm the new tag's `sortIndex` is `max(existing in topic) + 1`. CSV import via `/verfahren/{procedureId}/schlagworte/import/csv` does the same.

4. **RPC reorder — same topic.**
   ```
   POST /rpc/2.0
   {"jsonrpc":"2.0","method":"tagList.reorder","id":"1",
    "params":{"tagId":"<id>","topicId":"<sameTopicId>","newIndex":2}}
   ```
   Response is a flat dict keyed by tag ID with the new `sortIndex` and `topicId` for every tag whose position changed.

5. **RPC reorder — cross-topic.** Same call with a different `topicId`. The source topic is re-indexed; the destination accepts the tag at `newIndex`.

6. **Edge cases.** `newIndex: null` appends to the end. Out-of-range `newIndex` is clamped to `[0, n]`. A no-op (same topic, same index) returns an empty result and writes nothing.

7. **Permission.** Hit the RPC as a user without `area_admin_statements_tag` → 403.

8. **Procedure copy.** Create a procedure with tagged topics, copy it (e.g. as template), and confirm tags in the copy preserve the source's `sortIndex` order.

### Linked PRs

- Frontend (drag-and-drop UI in `TagsList.vue`): https://github.com/demos-europe/demosplan-core/pull/6233

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Update changelog


sierrazamudiodemos:
it looks good, and it works very well also importing new tags and allocating them at the end (because they are not sorted)
i would like to go deeper into the review

demosplan/DemosPlanCoreBundle/Logic/Statement/RpcTagListReorderer.php
```diff
@@ -0,0 +1,215 @@
...
+            $previousTopic->removeTag($tag);
+            $this->reindexAndCollect($this->siblingsWithout($previousTopic, $tag), $changed);
+            $tag->setTopic($newTopic);
+        }
+
+        array_splice($targetSiblings, $newIndex, 0, [$tag]);
+        $this->reindexAndCollect($targetSiblings, $changed, $sameTopic ? null : $tag);
+
+        return array_values($changed);
+    }
```
sierrazamudiodemos, line 150: Hey Richard! thanks for this!
I was wondering if we could make the method more readable? 
maybe we could split it like this? what do you think?

```
  private function reorder(Tag $tag, TagTopic $previousTopic, TagTopic $newTopic, ?int $newIndex): array
  {
      return $previousTopic->getId() === $newTopic->getId()
          ? $this->reorderWithinTopic($tag, $newTopic, $newIndex)
          : $this->moveBetweenTopics($tag, $previousTopic, $newTopic, $newIndex);
  }
```

demosplan/DemosPlanCoreBundle/Logic/Statement/RpcTagListReorderer.php
```diff
@@ -0,0 +1,215 @@
...
+        $newIndex = $this->clampIndex($newIndex, count($targetSiblings));
+
+        if ($sameTopic && $previousIndex === $newIndex) {
+            return [];
+        }
+
+        $changed = [];
+
+        if (!$sameTopic) {
+            $previousTopic->removeTag($tag);
```
sierrazamudiodemos, line 141: there is a unique index constrain here `demosplan/DemosPlanCoreBundle/Entity/Statement/Tag.php:34` that title and tag should be unique.
But i also wonder that it is not even possible to create same tag under 2 different topics... I got suggested something like this (but I wanted to hear ur opinion if it is worth it

```
      foreach ($newTopic->getTags() as $existing) {
          if ($existing->getId() !== $tag->getId()
              && $existing->getTitle() === $tag->getTitle()) {
              throw DuplicatedTagTitleException::createFromTitleAndProcedureId(
                  $newTopic,
                  $tag->getTitle()
              );
          }
      }
```

sierrazamudiodemos:
Hey! thank u very much for the PR
My 2 remaining concerns:
1. Do u think is worth a test for the reorder? 
2. What about moving the reoder logic to a dedicated service instead of putting it on the Rpc Request? or is it like an overkilled?

PR #6252, opened 2026-05-12
fix(refs HDDP-18): add command to repair cross-procedure file references

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/HDDP-18/BOBSH-FM-beim-Download-von-Dokumenten-bei-Verfahren-aus-Blaupause

Description:

Adds `dplan:files:fix-procedure-mismatches` to repair historical cross-procedure file references (HDDP-18). The old blueprint-clone code copied the references in `_elements._e_file` and
  `_single_doc._sd_document` but did **not** duplicate the underlying `_files` row, so the referenced file's `procedure_id` still points at the source procedure.
  `FileController::fileProcedureAction` enforces a per-procedure check and 404s those downloads when the file is requested through a derived procedure's slug.
  
  Newer clone code (`ProcedureService::copyProcedureRelatedFilesFromBlueprint`, `ElementsService::copyDocumentRelatedFiles`, `ElementsService::copyElementRelatedFiles`) already does this
  right — this is a historical data issue, not a live code regression.

  ## What the command does

  - **Dry-run by default**: prints per-table mismatch counts and a top-30 breakdown of owning procedures with broken refs. No DB writes, no file work.
  - **With `--apply`**: repairs each broken reference by calling `FileService::createCopyOfFile($oldFileString, $owningProcedureId)` — the same method the current blueprint-clone path uses.
  Result per reference: a fresh `_files` row with new `_f_ident`/`_f_hash`, a physical blob copy via the configured Flysystem storage, and the entity's reference string rewritten through  `Elements::setFile` / `SingleDocument::setDocument`.

  ### Options

  | Option | Effect |
  |---|---|
  | `--apply` | Actually perform the fix (default: dry-run). |
  | `--procedure=<id>` | Restrict scope to one owning procedure. |
  | `--include-deleted-owners` | Also process refs whose owning procedure is soft-deleted. |
  | `--limit=<n>` | Process at most N references per table (staged rollout). |

  ### Safety

  - Default is dry-run; `--apply` is an explicit opt-in.
  - Interactive runs prompt for confirmation before writing.
  - Each reference is wrapped in its own try/catch — one failure logs and the run continues.
  - `createCopyOfFile()` returning `null` (source blob missing on disk) is recorded as **skipped**, not an error.
  - Flush every 50 entities to keep memory bounded.

  ## Out of scope

  - **`file_container`** references are not touched. Its owner is polymorphic (`entity_id` + `entity_class`); current blueprint-clone code doesn't process it either. Separate follow-up if audit surfaces mismatches there.
  - **Search-engine-cached URLs** that embed the old `_f_ident` with a derived procedure's slug will continue to 404 — the old `_f_ident` row is still owned by the source procedure. After  the fix, URLs newly generated by the application use the new ident and work.

  ## Test plan

  - [ ] Run the audit against a recent prod-style DB dump and cross-check the discovery counts against the bug report.
  Sql:
`SELECT
    f.procedure_id             AS source_procedure_id,
    src._p_name        AS source_procedure_name,
    src._p_deleted     AS source_deleted,
    src._p_master      AS source_is_blueprint,
    COUNT(*)           AS affected_single_docs
  FROM _single_doc sd
  JOIN _files f
    ON f._f_ident = SUBSTRING_INDEX(SUBSTRING_INDEX(sd._sd_document, ':', -3), ':', 1)
  JOIN _procedure p
    ON p._p_id = sd._p_id          -- owning (derived) procedure
  LEFT JOIN _procedure src
    ON src._p_id = f.procedure_id          -- source procedure that the file row belongs to
  WHERE sd._sd_deleted = 0
    AND f._f_deleted = 0
    AND p._p_deleted  = 0          -- same filter as the fix
    AND sd._p_id <> f.procedure_id 
  GROUP BY f.procedure_id , src._p_name, src._p_deleted, src._p_master
  ORDER BY affected_single_docs DESC;`
  - [ ] `bin/console dplan:files:fix-procedure-mismatches --procedure=<id>` against a single known-affected procedure and verify the per-procedure breakdown contains only that procedure.
  - [ ] `bin/console dplan:files:fix-procedure-mismatches --apply --limit=1` on local; re-run the audit; expect the count to drop by one.
  - [ ] After full `--apply`, confirm a previously-failing download URL under a derived procedure's slug returns 200.
  - [ ] Unit tests green: `vendor/bin/phpunit tests/backend/core/Core/Unit/Command/FixProcedureFileMismatchesCommandTest.php`

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Command/FixProcedureFileMismatchesCommand.php
+++ b/demosplan/DemosPlanCoreBundle/Command/FixProcedureFileMismatchesCommand.php
@@ -12,11 +12,14 @@ namespace demosplan\DemosPlanCoreBundle\Command;
 
 use demosplan\DemosPlanCoreBundle\Entity\Document\Elements;
 use demosplan\DemosPlanCoreBundle\Entity\Document\SingleDocument;
+use demosplan\DemosPlanCoreBundle\Entity\File;
 use demosplan\DemosPlanCoreBundle\Logic\FileService;
 use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection;
 use Doctrine\ORM\EntityManagerInterface;
 use Psr\Log\LoggerInterface;
+use RuntimeException;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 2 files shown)


mussbach (approved):
lgtm, thx

PR #6259, opened 2026-05-13
feat(DPLAN-17492): lock segment by workflow place — frontend

meissnerdemos:
### Ticket
DPLAN-17492

### Description
Frontend for segment locking. Segments whose workflow place is marked locked become read-only for non-admins; admins can unlock via the new `SegmentUnlockModal` (sets a different place + assignee).

- `SegmentsList.vue`: prohibit icon w/ tooltip on locked rows; checkbox locked for non-admins; bulk-edit restricted to assignee + place when selection contains locked segments (info hint shows `{lockedCount} von {totalCount}`).
- `SegmentsBulkEdit.vue`: reads `hasLocked` / `lockedCount` from lscache; hides tag / recommendation / custom-field actions when flagged.
- `StatementSegmentsEdit.vue`: claim button replaced by prohibit icon + tooltip for locked segments; editor disabled.
- `StatementSegment.vue`: lock badge with tooltip; admin badge opens unlock modal; `isSaving` state to keep UI consistent during save.
- `SegmentUnlockModal.vue` (new): place + assignee multiselect, emits `unlock`.

### Related PRs
demosplan-ui: [https://github.com/demos-europe/demosplan-ui/pull/1487](https://github.com/demos-europe/demosplan-ui/pull/1487)
backend: #6195.

### How to review/test
1. Link [demosplan-ui branch](https://github.com/demos-europe/demosplan-ui/pull/1487) 
2. Log in as admin
3. Verfahren → Bearbeitungsschritte: mark one place as 'abgeschlossen' if not already
4. go to Abschnitte: rows with locked place show prohibit icon w/ tooltip
5. Select mix of locked/unlocked → "Abschnitte bearbeiten" → only assignee + place actions visible
6. go to Erwiderung tab: locked segment shows "Gesperrt" badge; click opens unlock modal; pick place + assignee → save
7. go to Details tab: locked segment shows prohibit icon, editor hidden
8. Log in as Sachbearbeiter
9. checkbox + edit disabled in all views, tooltip explains contact admin

### PR Checklist
- [x] Run `yarn lint`
- [x] Run `yarn test`


client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -72,7 +72,34 @@
         </template>
       </v-popover>
 
+      <div
+        v-if="isLocked"
+        class="flex space-x-1 mt-1 mr-2"
+      >
+        <dp-icon
```
hwiem: I think the icon needs `margin-top: 4px` to be centered vertically, and the badge seems to need a `padding-top: 4px` so the text is centered vertically.
before:
Image
after:
Image

translations/messages+intl-icu.de.yml
```diff
@@ -3249,9 +3251,11 @@ segment.claim.fail: "Der Abschnitt konnte Ihnen nicht zugewiesen werden."
 segment.confirm.suggestion: "Vorschlag als Abschnitt übernehmen"
 segment.edit: "Abschnitt bearbeiten"
 segment.external_id: "Abschnitts-ID"
+segment.lock.hint: "Dieser Abschnitt ist als abgeschlossen gekennzeichnet und kann daher nicht mehr bearbeitet werden. Für eine Freigabe wenden Sie sich bitte an den Fachplaner Administrator."
```
hwiem: We should always use gender-neutral language:
```suggestion
segment.lock.hint: "Dieser Abschnitt ist als abgeschlossen gekennzeichnet und kann daher nicht mehr bearbeitet werden. Für eine Freigabe wenden Sie sich bitte an die Fachplanung-Administration."
```

translations/messages+intl-icu.de.yml
```diff
@@ -3249,9 +3251,11 @@ segment.claim.fail: "Der Abschnitt konnte Ihnen nicht zugewiesen werden."
...
+segment.lock.hint.admin: "Dieser Abschnitt ist derzeit gesperrt, weil der festgelegte Bearbeitungsschritt als \"abgeschlossen\" markiert ist. Um ihn wieder zur Bearbeitung freizugeben, wählen Sie bitte einen anderen Bearbeitungsschritt und eine*n Bearbeiter*in aus."
+segment.lock.hint.admin.change.view: "Dieser Abschnitt ist derzeit gesperrt, weil der festgelegte Bearbeitungsschritt als \"abgeschlossen\" markiert ist. Um ihn wieder zur Bearbeitung freizugeben, wechseln Sie bitte zur Erwiderungsansicht und wählen Sie dort einen anderen Bearbeitungsschritt"
```
hwiem:
```suggestion
segment.lock.hint.admin.change.view: "Dieser Abschnitt ist derzeit gesperrt, weil der festgelegte Bearbeitungsschritt als \"abgeschlossen\" markiert ist. Um ihn wieder zur Bearbeitung freizugeben, wechseln Sie bitte zur Erwiderungsansicht und wählen Sie dort einen anderen Bearbeitungsschritt."
```

translations/messages+intl-icu.de.yml
```diff
@@ -3269,6 +3273,8 @@ segment.save: "Abschnitt speichern"
 segment.tags: "Schlagworte"
 segment.tags.none: "Diesem Abschnitt wurden keine Schlagworte zugewiesen."
 segment.text: "Abschnittstext"
+segment.unlock: "Abschnitt zur Bearbeitung freigeben"
+segment.unlock.click.hint: "Dieser Abschnitt ist gesperrt. Als Administrator können Sie den Abschnitt wieder zur Bearbeitung freigeben. Klicken Sie dazu auf dieses Badge."
```
hwiem:
```suggestion
segment.unlock.click.hint: "Dieser Abschnitt ist gesperrt. Als Administration können Sie den Abschnitt wieder zur Bearbeitung freigeben. Klicken Sie dazu auf dieses Badge."
```

client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
```diff
@@ -22,6 +22,12 @@
     <!-- Step 1 - Chose action -->
     <template v-slot:step-1>
       <div class="border-between-vertical">
+        <dp-inline-notification
```
hwiem, line 25: According to the design, this inline notification should be dismissible (add `dismissible` prop and `dismissble-key` prop to store its value in `localStorage`)

client/js/components/procedure/admin/AdministrationPlaces.vue
```diff
@@ -392,6 +392,11 @@ export default {
           solved: this.newPlace.solved,
         },
       }
+
+      if (hasPermission('feature_segment_lock_by_workflow_place') && hasPermission('feature_administrate_segment_lock')) {
+        payload.attributes.locked = this.newPlace.solved
```
hwiem, line 397: Why do we need to send a `locked` attribute in addition to the `solved` attribute?

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -607,16 +623,56 @@ export default {
       ]
     },
 
+    canUnlock () {
+      return hasPermission('feature_administrate_segment_lock')
```
hwiem, line 639: Do we also need to check `hasPermission('feature_segment_lock_by_workflow_place')` here?

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -607,16 +623,56 @@ export default {
...
+    },
+
+    // Overrides tableSelectAllItems mixin to exclude locked segments from selection for users without unlock permission
+    currentlySelectedItems () {
+      const toggledIds = new Set(this.toggledItems.map(item => item.id))
+      let selected
+
+      if (this.trackDeselected === false) {
```
hwiem:
```suggestion
      if (!this.trackDeselected) {
```

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -768,6 +825,7 @@ export default {
           ].join(),
           Place: [
             'name',
+            ...(hasPermission('feature_segment_lock_by_workflow_place') ? ['locked'] : []),
```
hwiem, line 845: Do we also need to check `hasPermission('feature_administrate_segment_lock')` here? (There are more cases below where only `hasPermission('feature_segment_lock_by_workflow_place')` is checked, I'd like to extend the question to  them 😅 )

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
```diff
@@ -0,0 +1,93 @@
+<template>
```
hwiem, line 1: I discussed the design for the modal with Svenja, and she posted a slightly updated design in the user story ticket (you probably saw it already). Basically we thought that the hint should be a dismissible inline-notification (type info), the "Bearbeiter * in" field does not need to be mandatory if we set "Nicht zugewiesen" as a default, both the "* Pflichtfeld" and label hints are not really needed, and the heading should be slightly adjusted to match the pattern we used for similar modals (see screenshot in the ticket)

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
```diff
@@ -0,0 +1,93 @@
...
+      {{ Translator.trans('segment.lock.hint.admin') }}
+    </p>
+    <p class="mb-2">
+      {{ '* ' + Translator.trans('field.required') }}
+    </p>
+    <dp-label
+      class="mb-0 mt-4"
+      :text="Translator.trans('place')"
+      :hint="Translator.trans('place.set')"
+      :bold="false"
```
hwiem: Why is `bold` set to false? 🤔

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
```diff
@@ -0,0 +1,93 @@
...
+    <dp-button-row
+      primary
+      secondary
+      @primary-action="save"
+      @secondary-action="toggle"
+    />
+  </dp-modal>
+</template>
+
+<script setup>
```
hwiem, line 65: Nice ✨

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
```diff
@@ -0,0 +1,93 @@
...
+
+const localAssignee = ref(null)
+const localPlace = ref(null)
+const unlockModal = ref()
+
+const save = () => {
+  emit('unlock', { assignee: localAssignee.value, place: localPlace.value })
+  toggle()
+}
+function toggle () { unlockModal.value.toggle() }
```
hwiem: nitpick: I'd declare this above its usage, move `defineExpose({ toggle })` to the bottom and turn `toggle` into an arrow function, too

```
const localAssignee = ref(null)
const localPlace = ref(null)
const unlockModal = ref()

const toggle = () = > unlockModal.value.toggle()
const save = () => {
  emit('unlock', { assignee: localAssignee.value, place: localPlace.value })
  toggle()
}

defineExpose({ toggle })
```

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -72,7 +72,34 @@
...
+          class="text-interactive"
+          icon="prohibit"
+          weight="fill"
+        />
+        <dp-tooltip :text="badgeTooltipText">
+          <dp-badge
```
hwiem: When I tested this in the interface, I did not find it very intuitive that you have to click the badge to open the modal, and I discussed it with Svenja. We decided that it might be more intuitive to use the prohibit icon as the trigger for the modal and that the badge should be purely informative. The tooltip should then also be placed on the icon. Sorry that this came up only now. :(
It would then also make sense to be able to open the modal also from the other places where the prohibit icon is displayed, i.e. the segments list and the statement detail view

client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
```diff
@@ -354,9 +379,20 @@ export default {
     },
 
     isAssigneeEditable (segment) {
+      if (this.isSegmentLocked(segment)) {
+        return false
+      }
       return segment?.relationships?.assignee?.data?.id === this.currentUser.id
```
hwiem, line 447:
```suggestion
      }
      
     return segment?.relationships?.assignee?.data?.id === this.currentUser.id
```

client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
```diff
@@ -354,9 +379,20 @@ export default {
...
     },
 
+    isSegmentLocked (segment) {
+      if (!hasPermission('feature_segment_lock_by_workflow_place')) {
+        return false
+      }
+      const placeId = segment?.relationships?.place?.data?.id
+      return !!this.placeItems[placeId]?.attributes?.locked
```
hwiem, line 457:
```suggestion
      }
      
      const placeId = segment?.relationships?.place?.data?.id
      
      return !!this.placeItems[placeId]?.attributes?.locked
```

client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
```diff
@@ -38,7 +38,18 @@
         class="px-1 hover:bg-interactive-secondary-subtle-hover"
       >
         <div class="inline-block w-[5%]">
+          <dp-tooltip
+            v-if="isSegmentLocked(segment)"
```
hwiem: I think the icon should be aligned with the claim icons and the text, right now it looks like this
Image
I hope the lines clarify what I mean 😅

hwiem (changes requested):
I have mostly design concerns; I discussed them with Svenja and we decided on a few changes. Sorry that it comes after you already implemented it :(

client/js/components/procedure/admin/AdministrationPlaces.vue
meissnerdemos: They are two different things in other projects. `solved` influences whether a place shows a segment as "Abgeschlossen" in the statistics on the procedure overview, and `locked` locks the segment with that place. Projects that don't have `locked` still have `solved`. The backend implemented `locked` as its own column.

client/js/components/procedure/SegmentsList/SegmentsList.vue
meissnerdemos: Yes :see_no_evil:  At first there was only one permission and the feature was toggled via a yml entry but now it has two permissions. I'll add the checks.

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
meissnerdemos: Will do :) It was either/or and since we haven't used the composition api much I wasn't sure which was better. I don't really have strong feelings on how to do it as long we do it the same everywhere.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
meissnerdemos: Hmm okay do we then still need the `DpBadge` update? Since the badge no longer needs to be clickable? Svenja said that she would like to use the pattern in the future so I am not sure.

client/js/components/procedure/SegmentsList/SegmentsList.vue
meissnerdemos: Oh wait, I got confused with the two permissions. The `feature_administrate_segment_lock` permission is explicitly just for admins and allows the different behavior depending on user. So actually just checking whether the whole feature is enabled here is correct, since we need to fetch the information for all users.

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
meissnerdemos: It was to make the hint text not bold. But now its no longer needed since the labels don't have the hint anymore.

meissnerdemos:
**Update after review feedback**

Admins can now open the unlock modal from three places — the segments list (`SegmentsList.vue`), the recommendations tab (`SegmentsRecommendations.vue`), and the details/edit view (`StatementSegmentsEdit.vue`). Each one was duplicating the same logic, so I pulled it into a new composable, `useSegmentUnlock.js`. It exposes:

- `unlockModal` — template ref to bind on the `SegmentUnlockModal`
- `openUnlockModal(segment)` — remembers the segment and opens the modal
- `unlockSegment(payload, onSuccess?)` — sends the PATCH, shows the confirm/error notification, then runs the caller's refetch

Each consumer just binds the ref, wires `@unlock` to `unlockSegment`, and passes its own component-specific refetch as `onSuccess`.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
meissnerdemos: Fixed in 02febca68.

translations/messages+intl-icu.de.yml
meissnerdemos: Fixed in 40961c42c.

translations/messages+intl-icu.de.yml
meissnerdemos: Fixed in 40961c42c.

translations/messages+intl-icu.de.yml
meissnerdemos: Fixed in 40961c42c.

client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
meissnerdemos: Added in 3fb8b2013.

client/js/components/procedure/SegmentsList/SegmentsList.vue
meissnerdemos: Fixed in 88e2c778a.

client/js/components/procedure/SegmentsList/SegmentsList.vue
meissnerdemos: Changed in 8b3a51f07.

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
meissnerdemos: Implemented in 497142d5f (header style in ad94ebb0d).

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
meissnerdemos: Changed in 497142d5f.

client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
meissnerdemos: Fixed in 8b3a51f07.

client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
meissnerdemos: Fixed in 8b3a51f07.

client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
meissnerdemos: Tried my best :sweat_smile: in 0a0347ba7 and a6f7149eb.

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
meissnerdemos: Removed in 497142d5f.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
meissnerdemos: Implemented: the prohibit icon is now the trigger and the badge is purely informative; the modal can be opened from the recommendations view (829b975b5), the segments list (77ce71097) and the statement detail view (22753d1fb). The shared logic was extracted into the useSegmentUnlock composable (95a1d45ad).

client/js/components/procedure/SegmentsList/SegmentsList.vue
hwiem: If I understand your comment below correctly, actually we only need to check `feature_administrate_segment_lock` here, too, right? :thinking: :see_no_evil:

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -203,6 +205,16 @@
                   />
                 </template>
               </v-popover>
+              <dp-button
+                v-if="canUnlock && rowData.isPlaceLocked"
+                :text="Translator.trans('segment.unlock.click.hint')"
+                class="text-interactive"
```
hwiem: I noticed that if you hover over the segment, the background of the button remains white. Was that intentional? I think it looks better if it also turns gray, and it matches the behavior of the claim icon
```suggestion
                class="text-interactive bg-transparent! border-transparent! hover:bg-interactive-subtle-hover!"
```

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -572,6 +616,12 @@ export default {
       }
     },
 
+    badgeTooltipText () {
```
hwiem: Should we rename this now that it is used on the button/icon, not the badge?

client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
```diff
@@ -38,7 +38,31 @@
         class="px-1 hover:bg-interactive-secondary-subtle-hover"
       >
         <div class="inline-block w-[5%]">
+          <template v-if="isSegmentLocked(segment)">
+            <dp-button
+              v-if="hasPermission('feature_administrate_segment_lock')"
+              :text="lockTooltip"
+              class="text-interactive inline-block ml-0.5 align-middle"
```
hwiem: Again I'd suggest to make the background gray when the segment is hovered
```suggestion
              class="text-interactive inline-block ml-0.5 align-middle bg-transparent! border-transparent! hover:bg-interactive-subtle-hover!"
```

translations/messages+intl-icu.de.yml
```diff
@@ -3249,9 +3251,10 @@ segment.claim.fail: "Der Abschnitt konnte Ihnen nicht zugewiesen werden."
 segment.confirm.suggestion: "Vorschlag als Abschnitt übernehmen"
 segment.edit: "Abschnitt bearbeiten"
 segment.external_id: "Abschnitts-ID"
+segment.lock.hint: "Dieser Abschnitt ist als abgeschlossen gekennzeichnet und kann daher nicht mehr bearbeitet werden. Für eine Freigabe wenden Sie sich bitte an die Fachplanung-Administration."
```
hwiem: Should it be `\"abgeschlossen\"`  here, too? Or just `abgeschlossen` below

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
```diff
@@ -0,0 +1,86 @@
...
+const emit = defineEmits(['unlock'])
+
+// Default the assignee to "not assigned"
+const localAssignee = ref(assignableUsers.find(user => user.id === 'noAssigneeId'))
+const localPlace = ref(null)
+const unlockModal = ref()
+
+const toggle = () => unlockModal.value.toggle()
+
+const save = () => {
```
hwiem, line 91: I noticed that if I select nothing (i.e., no place) and click save in the modal, nothing happens (there is an error in the console though). 
The ideal solution would probably be to create a dpValidate-composable that we can use in `<script setup>`, but in order to keep this small I think we could also add a normal `<script>` block and import `dpValidateMixin` there:

```
<script>
import { dpValidateMixin } from '@demos-europe/demosplan-ui'

export default {
  mixins: [dpValidateMixin],
}
</script>
```
then use it in the template:

```
<div data-dp-validate="unlockSegment">
  <dp-multiselect id="segmentUnlockPlace" v-model="localPlace" required ... />
  ...
  <dp-button-row
    primary
    secondary
    @primary-action="dpValidateAction('unlockSegment', save, false)"
    @secondary-action="toggle"
  />
</div>
```
(didn't test it)

client/js/composables/useSegmentUnlock.js
```diff
@@ -0,0 +1,73 @@
+/**
```
hwiem, line 1: Very nice! 🎉

hwiem (changes requested):
Thank you for adjusting this! 🙏 I still found a few small issues, the most important one being the validation in the modal. The rest are really teeny tiny changes :hamsti:

client/js/components/procedure/SegmentsList/SegmentsList.vue
meissnerdemos: Fixed in 74041203e.

client/js/components/procedure/SegmentsList/SegmentsList.vue
meissnerdemos: Right :see_no_evil: fe0208e03

client/js/components/procedure/StatementSegmentsList/StatementSegmentsEdit.vue
meissnerdemos: Fixed in 1db57db7c.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
meissnerdemos: Renamed to `lockTooltipText` in 8979323c8.

translations/messages+intl-icu.de.yml
meissnerdemos: Fixed in f0e4087d1.

client/js/components/procedure/StatementSegmentsList/SegmentUnlockModal.vue
meissnerdemos: Applied it and it works! Didn't know you could use both `<script>` and `<script setup>`. I think the next time this comes up though we should make the composable. bc4283dcc

meissnerdemos: Hmm, Sonarcloud is now complaining about the two imports from demosplan-ui. Is this really an issue or unavoidable here?

meissnerdemos: I mean I could rewrite it to Options API but I really don't want to at this pint :sweat_smile:

hwiem: I would select "Accept" for the issues because it's just about readability. Once we have the composable we can replace this.

hwiem (approved):
Nice, I think now it's really polished! 💅 ❤️ Thanks for addressing all of my concerns, and bonus points for the composable + composition API 🚀

PR #6260, opened 2026-05-13
feat(ref: DPLAN-17783): Display custom fields and add permision

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17783/ADO-55068-Erweiterung-der-InstitutionsVerschlagwortung-um-konfigurierbare-Textfelder-Part-4-Felder-fur-Fachplanungs-Admin

**Summary**

Exposes `organisation custom field values` on the `InvitablePublicAgency` resource so they can be displayed, protected by a **new permission.**

**Changes**

- New permission `field_organisations_custom_fields` — controls whether an organisation's own custom field values are readable via the API. **Check project PR**
- `InvitablePublicAgencyResourceType` — when the permission is enabled, exposes customFields as a readable attribute (serialised via `Orga::getCustomFields()?->toJson()`).
- `InvitablePublicAgencyResourceConfigBuilder` — now extends the local `OrgaResourceConfigBuilder` instead of `BaseOrgaResourceConfigBuilder`, so it inherits the `customFields` property definition.

**Test plan**

- [ ] Grant field_organisations_custom_fields to a role and request the InvitablePublicAgency resource — confirm customFields is included in the response
- [ ] Remove the permission — confirm customFields is no longer returned


config/permissions.yml
```diff
@@ -2701,6 +2701,11 @@ field_organisation_phone:
     label: 'Telefon des Verfahrensträger'
     loginRequired: false
     parent: field_organisation
+field_organisations_custom_fields:
```
breierDemos, line 2716: Do the frontend need this permission? I see its only used in the backend, right? So we dont need `expose: true` here.

breierDemos (approved):
lgtm code review only
just one question about exposing the permission to the frontend

config/permissions.yml
sierrazamudiodemos: holi, yep the permission is needed in the FE

breierDemos: Hej, where is it used? I thought here: https://github.com/demos-europe/demosplan-core/pull/6313 but I cant see it.

sierrazamudiodemos: yep. Lars updted the file and you cna find the permission there

PR #6261, opened 2026-05-13
feat(refs: DPLAN-17739): make institution text custom field definitions editable

huellnerdemos:
### Ticket
DPLAN-17739

This pr makes it possible to edit the definition of text custom fields, that where defined in institution tag management.

### How to review/test

1. Login as MA-Role and open the institution tag management
2. In the "Freitextfelder definieren" tab, create a new text field definition
3. Click the edit button at the definition in the definitions list
4. Change name and description of the definition and submit
5. A modal appears, that tells us, that the field is changed everywhere where it is used
6. Click the confirm button in the modal, to save the definition change
7. Also check if the process can be aborted by clicking the corresponded abort buttons

Since there are some structural changes for the custom fields in this pr, maybe check a few other places, if the fields still works well and are rendered correctly

### PR Checklist

- [x] Run `yarn lint`
- [x] Update changelog


client/js/composables/useCustomFieldTypes.js
```diff
@@ -0,0 +1,83 @@
...
+  const enrich = FIELD_TYPE_REGISTRY[fieldType]?.enrichValue
+
+  return enrich ? enrich(rawValue, options) : {}
+}
+
+function fieldTypeSupportsOptions (fieldType) {
+  return FIELD_TYPE_REGISTRY[fieldType]?.supportsOptions ?? false
+}
+
+function getComponentForFieldType (fieldType) {
```
riechedemos, line 61: The fallback to SingleselectCustomField for unknown field types is silent — if the backend ever returns a type that isn't registered, users get a broken select with no hint that something's off. Could we add a console.warn here, in the same style as the existing one in CustomField.vue (line 297)?

```
 if (!component) {
   console.warn(`Unknown custom field type "${fieldType}"; falling back to singleSelect renderer.`)
   return SingleselectCustomField
 }

```
Same thought for getDefaultFieldTypeForTarget — an unknown target currently leaves the field type empty without any signal.

riechedemos (changes requested):
Tested locally and works as expected. Nice refactor, the type registry is much cleaner than the scattered maps from before 🎉
Left one small non-blocking suggestion inline.

client/js/composables/useCustomFieldTypes.js
huellnerdemos: Very good point, thank you! Changed it in b3ff2835b4dda5ee95df7734f6d887e1122a68d6

riechedemos (approved):
Thank you :)

PR #6268, opened 2026-05-15
chore(deps): update symfony/framework-bundle recipe + migrate Gedmo to attributes

mussbach:
## Summary

Bumps `symfony/framework-bundle` recipe (5.4 → 6.4) and migrates Gedmo Doctrine extension behaviors from docblock annotations to PHP attributes.

### Recipe changes (`config/packages/framework.yaml`)

| Setting | Before | After | Why |
|---|---|---|---|
| `handle_all_throwables` | — | `true` | Symfony 6 default; existing `kernel.exception` listeners already call `$event->getThrowable()` |
| `validation.enable_annotations` | `true` | `validation.enable_attributes: true` | `enable_annotations` is deprecated in 6.4 (gets renamed to `enable_attributes` internally). Docblock constraint annotations continue to work because `framework.annotations: true` still wires the docblock reader. Empirically verified by loading `@StatementIdsInProcedureConstraint` via the validator metadata factory |
| `annotations: false` | — | commented hint | Documented as a follow-up. Blocked by 2105 `@ORM\*` docblock annotations and the validator's two remaining `@Constraint` docblocks |

### Gedmo annotation → attribute migration

Run via Rector (`DoctrineSetList::GEDMO_ANNOTATIONS_TO_ATTRIBUTES`):
- 74 entities, 145 `@Gedmo\Timestampable(...)` → `#[Gedmo\Timestampable(...)]`
- All occurrences use only `on="create"` or `on="update"` — no other Gedmo behaviors in the codebase

To make the migration actually take effect, `config/doctrine_extensions.yml` now wires the listener with `Gedmo\Mapping\Driver\AttributeReader`:

```yaml
services:
    Gedmo\Mapping\Driver\AttributeReader: ~

    gedmo.listener.timestampable:
        class: Gedmo\Timestampable\TimestampableListener
        tags:
            - { name: doctrine.event_subscriber }
        calls:
            - [ setAnnotationReader, [ "@Gedmo\\Mapping\\Driver\\AttributeReader" ] ]
```

The previous `@annotation_reader` call routed through a docblock-only reader and would silently miss the migrated attributes — the move and the wiring change must land together.

### Incidental cleanups

Rector's active sets also produced 9 FQCN → `use` statement cleanups in `Logic/`, `Repository/`, `Transformers/`. Kept per scope decision.

## Test plan

- [x] `app_dev.php` returns 200
- [x] `/login` returns 302
- [x] `/logout` returns 302
- [x] Empirical: `Gedmo\Mapping\Driver\AttributeReader::getPropertyAnnotations()` returns `Timestampable on=create` for migrated entities
- [x] Empirical: validator metadata loads `@StatementIdsInProcedureConstraint` docblock constraint
- [ ] CI green
- [ ] Verify Timestampable fires on entity persist/update in a real flow

## Follow-ups

- Migrate 2105 `@ORM\*` annotations to `#[ORM\*]` attributes → then enable `framework.annotations: false`
- Migrate the two `@StatementIdsInProcedureConstraint` / `@ValidCssVarsConstraint` docblock constraints to attributes


sierrazamudiodemos (approved):
Hello! 
thanks for this! I tested locally and it works. 
I have an small concern out of the scope of this PR (it is not a blocker)

In this line of the addon: 
`addonDev/demosplan-addon-xbeteiligung-async/vendor/demos-europe/demosplan-addon/src/Controller/APIController.php:246`
- There is a default case that when there is an error, then return `Response::HTTP_BAD_REQUEST;` which is 400 error
- But after this PR change, also errors coming from the code crashing will also be into this fallback. In that case, I guess a 500 error would be more suitable. 
- Maybe we could add this case?

```
  case $exception instanceof Error:
      $status = Response::HTTP_INTERNAL_SERVER_ERROR;
      $message = 'error.api.generic';
      break;
  default:
      $message = 'error.api.generic';
      break;
```

mussbach:
closed in favor of #6297

PR #6275, opened 2026-05-18
fix DS-539: strip XML-illegal chars before PhpWord parses statement HTML

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/tickets/DS-539/Fehlender-Stellungnahmetext-beim-Export

Description: 

Statement texts pasted from PDFs can carry raw `\x02` bytes inside words (a soft-hyphen artifact produced by some PDF readers, e.g. `ge\x02planten` for "geplanten"). `\x02` and a few other codepoints are forbidden in XML 1.0, so when `DocxExporter::addHtml` hands the HTML to PhpWord, libxml rejects it. The catch-fallback then either prints the raw HTML tags (locally) or yields an empty cell (on prod — depending on the libxml build) instead of the actual statement text.

Reported symptom: on the procedure export (DOCX), 8 institution statements appeared with an empty text cell on prod even though the platform displayed them correctly. Each one contained a raw `\x02` byte; no other statement did.

  ### Solution
  Strip XML-1.0-illegal codepoints before PhpWord parses the HTML, in a dedicated `DocxExporter::stripXmlIllegalChars()` method.
  Stripped: C0 controls except `\t \n \r`, unpaired surrogates (`U+D800–U+DFFF`), and the two non-characters `U+FFFE` / `U+FFFF`.
  Kept: all printable text, C1 controls, the rest of the BMP, and the astral planes (so emoji and rare CJK survive).

  PDF export is unaffected — different rendering path and not impacted by this XML constraint.

- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php
@@ -942,6 +942,7 @@ class DocxExporter
         if ('' === $text) {
             return '';
         }
+        $text = $this->stripXmlIllegalChars($text);
         try {
             $text = self::replaceTags($text);
             Html::addHtml($cell, $text, false);
@@ -956,6 +957,23 @@ class DocxExporter
         }
 
         return '';
+    }
+
+    /**
+     * Drops codepoints that XML 1.0 forbids before PhpWord parses the HTML.
+     *
+     * Stripped: C0 controls except \t \n \r (e.g. \x02, the soft-hyphen marker
+     *   produced by PDF copy-paste), unpaired surrogates (U+D800-U+DFFF) and
+     *   the two non-characters U+FFFE / U+FFFF.
+     * Kept:     all printable text, C1 controls, the full BMP minus the holes
+     *   above, and the astral planes (so emoji and rare CJK survive).
+     *
+     * Without this, PhpWord's HTML parser either dumps raw tags via the catch-
+     * fallback or yields empty cells, depending on the libxml build.
+     */
+    public function stripXmlIllegalChars(string $text): string
+    {
+        return preg_replace('/[^\x09\x0A\x0D\x20-\x{D7FF}\x{E000}-\x{FFFD}\x{10000}-\x{10FFFF}]/u', '', $text);
     }
 
     private static function replaceTags(string $text): string
--- a/demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php
@@ -962,17 +962,24 @@ class DocxExporter
     /**
      * Drops codepoints that XML 1.0 forbids before PhpWord parses the HTML.
      *
+     * Invalid UTF-8 sequences are first normalized via mb_scrub so the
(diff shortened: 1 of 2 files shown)


demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php
```diff
@@ -958,6 +959,23 @@ public function addHtml($cell, $text, $styles)
...
+     *   the two non-characters U+FFFE / U+FFFF.
+     * Kept:     all printable text, C1 controls, the full BMP minus the holes
+     *   above, and the astral planes (so emoji and rare CJK survive).
+     *
+     * Without this, PhpWord's HTML parser either dumps raw tags via the catch-
+     * fallback or yields empty cells, depending on the libxml build.
+     */
+    public function stripXmlIllegalChars(string $text): string
+    {
+        return preg_replace('/[^\x09\x0A\x0D\x20-\x{D7FF}\x{E000}-\x{FFFD}\x{10000}-\x{10FFFF}]/u', '', $text);
```
mussbach, line 983: preg_replace can return null, signature promises string
  With invalid UTF-8 input (very plausible from PDF paste), preg_replace with /u returns null and emits a PCRE warning. The current return type would then throw a TypeError and bring down
  the export — replacing one failure mode with another. Two safe options:
  return preg_replace($pattern, '', $text) ?? $text;
  or scrub first:
  $text = mb_scrub($text, 'UTF-8');
  return preg_replace($pattern, '', $text);
  A targeted test feeding "\xC3\x28" (invalid UTF-8) would lock this down.

tests/backend/core/Export/Unit/DocxExporterTest.php
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * All rights reserved
+ */
+
+namespace Tests\Core\Export\Unit;
+
+use demosplan\DemosPlanCoreBundle\Logic\Export\DocxExporter;
+use Tests\Base\UnitTestCase;
+
+class DocxExporterTest extends UnitTestCase
```
mussbach, line 18: - No assertion that \x0B/\x0C/DEL (\x7F is actually allowed under the regex — sanity check this matches intent; XML 1.0 does allow \x7F).
  - No invalid-UTF-8 test (ties into issue above).
  - No test that the C1 range (\x80–\x9F) survives — the docblock claims it should.

mussbach (changes requested):

demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/6275/changes/e437462db181fcddd85c7bf7db5b8913f32fdb65 , thx

tests/backend/core/Export/Unit/DocxExporterTest.php
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/6275/changes/3eb8f69a9d2d9a5b365c293db79c627587c9a394 , thx

mussbach (approved):
lgtm, thx

PR #6276, opened 2026-05-18
fix(refs: DPLAN-17794) change fetch of custom fields in segments list and bulk edit

huellnerdemos:
### Ticket
DPLAN-17794

This PR changes the fetch behavior for the custom fields in segments list and bulk edit. It was necessary because the old fetch mechanism didn't get the values of the fields anymore.

### How to review/test

1. Login as admin and go to a procedure with custom fields
2. In the procedure, go to segments list
3. Try to add the custom field columns to the list
4. Mark one or more segments and change to bulk edit
5. Custom fields can be rendered and used in both places 

### PR Checklist

- [x] Run `yarn lint`

### DIFF
--- a/client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
+++ b/client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
@@ -355,6 +355,7 @@ import DpBoilerPlateModal from '@DpJs/components/statement/DpBoilerPlateModal'
 import lscache from 'lscache'
 import RecommendationModal from '../Shared/RecommendationModal'
 import SelectedTagsList from '@DpJs/components/procedure/SegmentsBulkEdit/SelectedTagsList'
+import { useCustomFields } from '@DpJs/composables/useCustomFields'
 
 export default {
   name: 'SegmentsBulkEdit',
@@ -433,6 +434,7 @@ export default {
       },
       assignableUsers: [],
       busy: false,
+      customFieldDefinitions: [],
       hasRecommendationTabs: false,
       isLoading: true,
       places: [],
@@ -444,10 +446,6 @@ export default {
   },
 
   computed: {
-    ...mapState('CustomField', {
-      customFieldItems: 'items',
-    }),
-
     ...mapState('Tag', {
       tagsItems: 'items',
     }),
@@ -571,10 +569,6 @@ export default {
   },
 
   methods: {
-    ...mapActions('AdminProcedure', {
-      getAdminProcedureWithFields: 'get',
-    }),
-
     ...mapActions('StatementSegment', {
       getSegment: 'get',
     }),
@@ -588,7 +582,7 @@ export default {
     }),
 
     addCustomFieldsToActions () {
-      Object.values(this.customFieldItems).forEach(customField => {
+      this.customFieldDefinitions.forEach(customField => {
         this.actions.customFields.push({
           checked: false,
           id: customField.id,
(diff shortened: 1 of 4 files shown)


client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
```diff
@@ -685,26 +679,15 @@ export default {
...
-            'description',
-            'options',
-          ].join(),
-        },
-        include: ['segmentCustomFields'].join(),
-      }
-
-      return this.getAdminProcedureWithFields(payload)
-        .catch(err => console.error(err))
+    loadSegmentCustomFields () {
```
rafelddemos, line 681: this method and getCustomFields in SegmentsList.vue do the same thing, maybe they could also have the same name then?

client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
```diff
@@ -685,26 +679,15 @@ export default {
...
+      return useCustomFields().fetchCustomFields(this.procedureId, {
```
rafelddemos: nitpick for consistency: maybe this method could also use desctructuring for fetchCustomFields, just like getCustomFields in SegmentsList.vue?

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -852,28 +843,17 @@ export default {
...
-            'options',
-          ].join(),
-        },
-        include: [
-          'segmentCustomFields',
-        ].join(),
-      }
-
-      this.getCustomFieldsForProcedure(payload)
+      const { fetchCustomFields } = useCustomFields()
```
rafelddemos, line 852: shouldn't this method also use `return` and` catch`, like `loadSegmentCustomFields` in `SegmentsBulkEdit.vue`?
Also - as they do the same thing, maybe they could have the same name too?

client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
```diff
@@ -685,26 +679,15 @@ export default {
     },
 
     /**
-     * Fetch custom fields that are available either in the procedure or in the procedure template
+     * Fetch custom fields that are available either in the procedure or in the procedure template.
```
rafelddemos: I find this comment a bit misleading (it was probably from the start): The procedure template custom fields are not really being fetched here, are they? They're probably rather copied from the blueprint..
Or is something else meant by it? &#x1F914;

rafelddemos (commented):
I tested it and it works now, thanks for fixing it! :) 
Left just a few small comments - since the two methods seem to do the same thing, maybe they could be adjusted to be more consistent.

client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
huellnerdemos: Thank you! i changed it in 618be9a980807ea72be0a7d1a68c7bdd13a99c23

client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
huellnerdemos: Yes this is maybe a bit better readable. Thank you! Changed it in a286b5370da5f5c38c2f7ed72ab5711c8ddeb6b9

client/js/components/procedure/SegmentsList/SegmentsList.vue
huellnerdemos: Yes a catch and return should be used here. Changed it in 1fee1c7df6881c3fde7ed36179acf723c5965f02

client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
huellnerdemos: Thanks for this point! Since its obvious what the method does and there shouldn't be explanations for obvious things, i deleted this comment in 4a7998abd9c54ff4f9728715a42d092eefbe58fe

rafelddemos (approved):
Thx!

PR #6284, opened 2026-05-19
fix(refs: DPLAN-17707): add dropdown to choose layer in global layer creation

huellnerdemos:
### Ticket
DPLAN-17707

In the global layer creation dialog, there was only an normal input field send the layer, but there should be a dropdown, to choose the layer like in the layer creation dialog of the procedures. Since the LayerSettings component is used for this dialog, this pr also adds the same component to the global layer creation dialog.

### How to review/test

1. Login as MA-Role and go to Plattformtools -> Globale Kartenebenen
2. Click on Neue globale Kartenebene anlegen
3. Check if you can create a new global map layer 

### PR Checklist

- [x] Run `yarn lint`


sierrazamudiodemos (approved):
BE part approved. thx

rafelddemos (approved):
FE looks good too, thanks for fixing it!

huellnerdemos:
Thanks for the reviews! :pray:

PR #6287, opened 2026-05-20
fix: allow skipping build step for addons

hwiem:
for addons that don't have a build step, we should skip it, so no error is thrown


mussbach (approved):
lgtm, code review only

PR #6290, opened 2026-05-20
refactor: Delete unused resource type

sierrazamudiodemos:
Revised body:
## Summary
- Remove `EmailAddressResourceType` and its `getEmailAddressResourceType()` accessor / constructor slot in `ResourceTypeStore`. The class had no callers anywhere in the codebase (core,
addons, frontend) and was unreachable via JSON:API anyway since both `isGetAllowed()` and `isListAllowed()` returned `false`.
- First cleanup step ahead of migrating ResourceTypes to the API Platform library — pruning dead types now reduces the migration surface.

## Test plan
- [ ] `grep -rn "EmailAddressResourceType" demosplan/ config/ client/ tests/` returns zero hits (excluding the `AgencyEmailAddressResourceType` substring matches, which is a different,
used class)


mussbach:
@sierrazamudiodemos is this PR still valid?

sierrazamudiodemos:
> @sierrazamudiodemos is this PR still valid?

yes, this resource type is not used at all

breierDemos (approved):
lgtm

PR #6301, opened 2026-05-22
fix(refs: DPLAN-17835): delete procedure_phase rows on procedure deletion

breierDemos:
### Ticket
DPLAN-17835

### Description
`ProcedureDeleter` uses raw SQL with foreign-key checks disabled, so the Doctrine cascade declared on `Procedure->ProcedurePhase` is never invoked. Each procedure deleted via `dplan:procedure:delete` therefore leaves two orphan rows in `procedure_phase` (one for `phase_id`, one for `public_participation_phase_id`).

This PR:
- collects the two phase ids per procedure before deleting `_procedure`, then deletes those rows from `procedure_phase`
- adds a core migration that removes rows already orphaned by previous runs of the command
- adds two functional test cases (real deletion + dry-run)

### How to review/test
- Inspect `ProcedureDeleter::deleteProcedures()` — new `collectProcedurePhaseIds()` / `deleteProcedurePhases()` helpers.
- Run `./vendor/bin/phpunit tests/backend/core/Procedure/Functional/ProcedureDeleterTest.php` — all 7 tests green.
- The cleanup migration was verified end-to-end against a real DB with injected orphan rows; only orphans were removed, referenced phases were untouched.

### PR Checklist
- [x] Create/Update tests
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 ### Added
 - Add text custom field definition editing
 
+### Fixed
+- `dplan:procedure:delete` now also removes the two associated `procedure_phase` rows; previous runs left orphan rows behind, which are cleaned up by a one-shot migration
+
 ## v4.42.0 (2026-05-21)
 ## v4.40.1 (2026-05-21)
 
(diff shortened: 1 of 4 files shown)


mussbach (approved):
lgtm, code review only, thx for this

PR #6251, opened 2026-05-12
feat(ref: DPLAN-17780): Add custom field attribute to organisation en…

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17780/ADO-54692-Erweiterung-der-InstitutionsVerschlagwortung-um-konfigurierbare-Textfelder-Part-3-Use-the-text-field-and-insert-values

PR Description

## Summary

Enables PATCHing text custom field values on institutions via JSON:API.
The text-field schema for `CUSTOMER → ORGA` already existed (DPLAN-17619) — this PR adds the storage column on `Orga` and the read/write attribute on `InvitableInstitution`.

## Changes

- `Orga` entity: new `customFields` JSON column + getters/setters.
- `InvitableInstitutionResourceType`: exposes `customFields` (read + update), protected by `feature_organisations_custom_fields` permission
- `OrgaResourceConfigBuilder` (new): declares the `customFields` property for the autogenerated config-builder pattern.
- Per-project Doctrine migration adds `_orga.custom_fields LONGTEXT DEFAULT NULL`.
- Test: `CustomFieldValueCreatorTest::testAddNewTextValueSuccessfully`.

## API example

```http
PATCH /api/2.0/InvitableInstitution/{orga-uuid}
Accept: application/vnd.api+json
Content-Type: application/vnd.api+json

{
  "data": {
    "id": "{orga-uuid}",
    "type": "InvitableInstitution",
    "attributes": {
      "customFields": [
        { "id": "{custom-field-config-uuid}", "value": "free text" }
      ]
    }
  }
}

Test plan

- Run the per-project migration via bin/{project} d:m:execute
- PATCH against a real institution, verify 200 and stored value
- phpunit --filter testAddNewTextValueSuccessfully green

PR #6303, opened 2026-05-22
feat(refs: DPLAN-17780): add function to save values to institution text custom fields

huellnerdemos:
### Ticket
DPLAN-17780

This PR adds the functionality to add values to text custom fields in the institution management list

### How to review/test

1. Login as MA-Role and create at least one text custom field definition for the institution management
2. Switch to the group institution list and add the custom field as a column
3. Click the edit button, add a text to the custom field and save it (maybe check a text with line breaks)
4. Reload the page and check if the values were saved correctly 

### Linked PRs

BE PR: https://github.com/demos-europe/demosplan-core/pull/6251

### PR Checklist

- [x] Run `yarn lint`
- [x] Update changelog


client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
```diff
@@ -171,26 +194,35 @@ import {
...
     DpLoading,
+    DpMultiselect,
     DpSearchField,
     DpSlidingPagination,
   },
 
   mixins: [tableScrollbarMixin],
 
   props: {
+    customerId: {
```
rafelddemos: this prop seems unsused

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
```diff
@@ -526,10 +550,90 @@ export default {
         })
     },
 
+    extractCustomFieldValues () {
```
rafelddemos: nitpick: this method should go before getCategoryTags

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
```diff
@@ -526,10 +550,90 @@ export default {
...
+            return null
+          }
+
+          return useCustomFields().updateCustomFields('InvitableInstitution', id, customFieldsPayload)
+        })
+        .then(() => {
+          dplan.notify.confirm(Translator.trans('confirm.saved'))
+        })
+        .catch(error => {
+          this.restoreInstitutionFromInitial(id)
```
rafelddemos: what happens if saving of the CF fails, and that of tags (in updateInvitableInstitution) succeeds? won't there be an error and the tags will be reset locally, although they were saved on the server? &#x1F914;
But I am not sure how likely the case is (that saving CF fails, and that of tags not)

rafelddemos (commented):
Tested it and it works nicely, thanks! &#127881;
A few small comments and one issue I noticed: 
- when you successfully define a new text field and then switch the tab to the group institution list --> the new text field is not in the list for column options (you need to manually reload the page for it to appear)

huellnerdemos:
> Tested it and it works nicely, thanks! A few small comments and one issue I noticed:
> 
>     * when you successfully define a new text field and then switch the tab to the group institution list --> the new text field is not in the list for column options (you need to manually reload the page for it to appear)

Thanks for your review! :pray:  I'll check the comments and solve everything. :eyes:

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
huellnerdemos: Oh, i forgot to delete this. Thank you! Deleted it in 708765210cfb7b45da338bcd7d1ad98feaa83d9c

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
huellnerdemos: Thanks for that! Sorted it in 808d881c8f70a06287b6b411b9dc22011f03b88d

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
huellnerdemos: Good point, thank you! :pray:  Changed the error handling in 887b438dfe9e917ddf749fcd75603730110ee93e

huellnerdemos:
> Tested it and it works nicely, thanks! 🎉 A few small comments and one issue I noticed:
> 
>     * when you successfully define a new text field and then switch the tab to the group institution list --> the new text field is not in the list for column options (you need to manually reload the page for it to appear)

Good catch, thank you! :shipit:  Added reloading of definitions on tab change in aa58c846dd71873690095a04fe64da35e9cf88d6

rafelddemos (commented):
nice! the new definition is now loaded into the options list:) 
I just commented the listing of all the custom fields in the error message - guess that might be confusing and could be skipped?

client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
```diff
@@ -526,10 +559,87 @@ export default {
...
+            this.restoreInstitutionFromInitial(id)
+          }
+
+          this.customFieldValuesByInstitutionId = {
+            ...this.customFieldValuesByInstitutionId,
+            [id]: previousCustomFieldValues,
+          }
+          const errorMessage = areTagsSaved ?
+            Translator.trans('error.custom_fields.save', {
+              fields: (this.customFieldDefinitions ?? []).map(definition => definition.attributes?.name).filter(Boolean).join(', '),
```
rafelddemos: nice, thanks for adjusting :) I am just not sure about listing all the custom fields defined for the user in the error message (it might be confusing to get the whole list if they only tried to edit one or two)? Guess `error.custom_fields.save` would be enough here. Otherwise comparing the changes and only listing the edited ones.

huellnerdemos: That's maybe right. But `error.custom_fields.save` is not a good candidate here, because it needs a parameter, otherwise the sentence gets just even more confusing than with a long list of fields. Because of this, i created i new translation that fits better in 8e8d9160a79c285b3632800bf0780fe79489d3cd and 31cd5774d2ebfa7d34eb501e2fb9f573410d10c2
Thank you for this point!

rafelddemos (approved):
Lgtm, thank you! :)

huellnerdemos:
> Lgtm, thank you! :)

Thanks for your review! :pray:

PR #6317, opened 2026-05-27
fix(refs DPLAN-17866): add missing tagTopics to request

hwiem:
### Ticket
DPLAN-17866

### The cause and the fix
- in the case that only one segment was selected in the bulk edit flow, the tagTopics were not fetched, and the store data was overwritten and missing the tagTopics, which led to an error in the multiselects in the add/delete tag actions
- in addition, we're now filtering out multiselect options (= tags) that don't have a tagTopic in their relationships

### How to review/test

- Navigate to the segments list
- select **exactly one** segment
- open the bulk edit view ("Abschnitte bearbeiten")
- check the "Schlagworte hinzufügen" and the "Schlagworte entfernen" checkboxes - the multiselect should be displayed and you should be able to select tags

### PR Checklist

### DIFF
--- a/client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
+++ b/client/js/components/procedure/SegmentsBulkEdit/SegmentsBulkEdit.vue
@@ -527,7 +527,7 @@ export default {
           title: topic.attributes.title,
           id: topic.id,
           tags: this.tags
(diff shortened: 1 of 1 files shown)


sakutademos (approved):
code revie, lgtm, thx!

PR #6320, opened 2026-05-28
chore: bump dependencies to fix high security alerts

hwiem:
- @fullhuman/postcss-purgecss 7 → 8 (drops glob 11.x)
- css-minimizer-webpack-plugin 7 → 8 (brings minimatch 10.2.4)
- terser-webpack-plugin 5.4 → 5.6
- pin glob 10.x to 10.5.0 via resolutions
- adapt purgecss default-export import (v8 breaking change)

resolves the following dependabot alerts:
- https://github.com/demos-europe/demosplan-core/security/dependabot/126
- https://github.com/demos-europe/demosplan-core/security/dependabot/128
- https://github.com/demos-europe/demosplan-core/security/dependabot/152


rafelddemos (approved):
lgtm, thank you! :)

PR #6321, opened 2026-05-28
fix(refs DPLAN-17830): add table-cell borders to ODT export of assess…

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-17830/BImSchG-LibreOffice-Ausdruck-aus-Abwagungstabelle-hat-keine-sichtbaren-Linien

Description: 
## Problem

Assessment-table export (`verfahren/abwaegung/export/<procedureId>` with format `odt`) produced ODT files with no visible cell borders. The DOCX path rendered borders correctly. Two defects in PhpWord's stock ODText writer cause this:

- `Writer/ODText/Style/Table` emits only alignment and column widths — never a style:family="table-cell"` style with border properties.
- `Writer/ODText/Element/Table::writeRow` writes `<table:table-cell>` nodes with no `table:style-name` attribute, so even a registered cell style wouldn't be referenced.

The dispatch (`Part/Content::writeAutoStyles`, `getAutoStyles`, the `$autoStyles` array, and `Element/Table::writeRow`) is all `private`, so subclassing cannot reach it.

## Solution

Wrap PhpWord's ODText writer rather than subclass it: `OdtBorderedWriter` produces the `.odt` via the stock writer, then opens the ZIP and patches `content.xml`.

- Injects one automatic style `DPlanBorderedCell` (`fo:border="0.5pt solid #000000"`, `fo:padding="0.05cm"`) into `<office:automatic-styles>`.
- Stamps `table:style-name="DPlanBorderedCell"` on every `<table:table-cell>` that does not already carry a style. Negative-lookahead regex keeps the patch idempotent and preserves any pre-existing cell styling.

Wired via `DocumentWriterSelector::createWriter()` so the wrapper activates only on ODT exports. The four call sites in `DocxExporter` (`createDocxGrouped`, `createDocxCondensed`, `createDocxUngrouped`, `createDocxGroupedBobHH`) now
obtain their writer from the selector. DOCX path is unchanged.

### Streaming to the browser

When the caller passes `php://output`, the patched bytes are streamed out via `fopen` + `fpassthru` + `fclose`. `readfile` was the obvious choice but is on PHP-FPM's `disable_functions` list; `fpassthru` is not and gives identical chunked streaming straight to the response body without buffering the whole
file in RAM.

## Files changed

- `demosplan/DemosPlanCoreBundle/Logic/Export/Odt/OdtBorderedWriter.php` (new)
- `demosplan/DemosPlanCoreBundle/Logic/Export/DocumentWriterSelector.php`
  (added `createWriter()`)
- `demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php` (four call
  sites routed through the selector; unused `IOFactory` import removed)

## Test plan

- [x] Export an assessment table via `verfahren/abwaegung/export/<procedureId>`
      with format `odt` — open the file in LibreOffice and confirm every table
      cell has visible borders.
- [x] Repeat with `format=docx` — confirm the DOCX output is unchanged.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Export/DocumentWriterSelector.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Export/DocumentWriterSelector.php
@@ -13,6 +13,10 @@ declare(strict_types=1);
 namespace demosplan\DemosPlanCoreBundle\Logic\Export;
 
 use DemosEurope\DemosplanAddon\Contracts\PermissionsInterface;
+use demosplan\DemosPlanCoreBundle\Logic\Export\Odt\OdtBorderedWriter;
+use PhpOffice\PhpWord\IOFactory;
+use PhpOffice\PhpWord\PhpWord;
+use PhpOffice\PhpWord\Writer\WriterInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
 
 class DocumentWriterSelector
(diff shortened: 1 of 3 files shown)


mussbach (approved):
I dislike very much that it needs to be in such a hacky way, but currently this is how it is

bakridemos:
> I dislike very much that it needs to be in such a hacky way, but currently this is how it is

i know i tried to do differently but it is something which may be we need to refactor in an extra pr

mussbach:
We cannot do anything here, this is PHPWord that needs to be configurable

PR #6322, opened 2026-05-28
Block password login for IdP users on configured subdomains

mussbach:
DPLAN-17900

## Summary
- New `idp_only_subdomains` parameter (array of customer subdomains, empty by default).
- `LoginFormAuthenticator` now rejects users with `providedByIdentityProvider = true` when the current customer's subdomain is listed, throwing a generic `AuthenticationException`.
- The guard is scoped to the password authenticator: `UserService::getValidUser()` and other authenticator paths are untouched, so user listings and SSO flows behave as before.
- Customer lookup is short-circuited when the list is empty and falls back safely if `CustomerNotFoundException` is thrown.

## Why
Without this gate any IdP-provisioned user whose `_u_password` column contains a hash (admin reset, migration, legacy local account flipped to IdP, etc.) can log in via the form and bypass the configured identity provider.

## Test plan
- [x] `tests/backend/core/Security/LoginFormAuthenticatorTest` — 5 cases (reject on match, allow on non-match, allow non-IdP user, empty config short-circuits customer lookup, customer lookup failure does not block).
- [ ] Manual: with `idp_only_subdomains: ['<subdomain>']`, attempt form login as an IdP-flagged user on that subdomain and confirm rejection; verify a non-flagged user on the same subdomain and an IdP user on a different subdomain still log in normally.

### DIFF
--- a/config/parameters_default.yml
+++ b/config/parameters_default.yml
@@ -378,6 +378,10 @@ parameters:
     azure_sso_enabled: false
     # and if, which customers?
     azure_sso_customers: []
+    # Subdomains of customers that have completed the IdP migration. For those subdomains the
+    # password/form login rejects users flagged providedByIdentityProvider, so they can only
+    # log in via IdP. Leave empty during the transition to keep form login working.
+    idp_only_subdomains: []
     # knpu_oauth2_client settings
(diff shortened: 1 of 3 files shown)


bakridemos (approved):
ltgm, code review only

PR #6324, opened 2026-05-28
F dplan-17814 delete text field custom field

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17814/ADO-55211-Erweiterung-der-InstitutionsVerschlagwortung-um-konfigurierbare-Textfelder-Part-5-Loschen-der-Felder-als-MA

**Description**

- Allow deleting organisation custom fields (text type)
- When a field is deleted, remove its stored values from all affected organisations

**How to test it**

0. Log in as a **Mandanten-Administration**
1. Go to Verwalter → Institutionen verschlagworten > Freitextfelder definieren 
2. Create a new text custom field (e.g. "Test notes")
3. Go to the Institutionen gruppieren  tab, open any institution, and fill in a value for that field
4. Repeat step 3 for a second institution
5. Go back to the Freitextfelder definieren  tab
6. Click the delete icon next to "Test notes" and confirm the deletion
7. Verify: The field no longer appears in the custom fields list
8. Verify: Open both institutions from step 3/4 — the field and its value should no longer be visible


demosplan/DemosPlanCoreBundle/Repository/OrgaRepository.php
```diff
@@ -806,4 +806,24 @@ public function updateObject($organisation): Orga
...
     }
+
+    /**
+     * Find all orgas that have a custom field with the given ID.
+     *
+     * @return array<Orga>
+     */
+    public function findOrgasWithCustomField(string $customFieldId): array
+    {
+        $escapedCustomFieldId = str_replace(['\\', '"'], ['\\\\', '\\"'], $customFieldId);
```
breierDemos: Could you explain why you added this line? Is $customField not a uuid?

breierDemos (changes requested):
Lgtm, I tested locally and it works.

demosplan/DemosPlanCoreBundle/Repository/OrgaRepository.php
sierrazamudiodemos: Hello loki, thanks for the feedback!!
since` $customFieldId` is always a `UUID` (this is make sure due to `UuidV4Generator` on the `CustomFieldConfiguration` entity), it can never contain " or \, so the escaping is unnecessary. I removed it in this commit 2cf1363a38a13ef2b16e38cd028de51144ef5cef

breierDemos (approved):
thank you!

PR #6328, opened 2026-05-29
fix: prevent control characters in procedure location fields breaking admin edit page

mussbach:
DPLAN-17910

## Problem

The procedure administration edit page (`administration_edit.html.twig`) crashed with a JavaScript `SyntaxError: JSON.parse: bad control character in string literal`, leaving the whole page unusable. The offending data was a stray control character — a leading tab in the procedure's municipal code — that was rendered unescaped into a `JSON.parse('...')` string literal.

## Root cause

The `:procedure-location` binding used the filter order `|e('js')|json_encode`, the reverse of every other binding on the page. With that order, `json_encode`'s control-char escapes (e.g. `\t`) get collapsed back into raw control characters by the JS string parser, which then break `JSON.parse`.

## Changes

- Correct the filter order to `|json_encode|e('js', 'utf-8')` in the procedure-location binding so escaped control characters survive into `JSON.parse`.
- Strip ASCII control characters in the `Procedure` location setters (`locationName`, `locationPostCode`, `municipalCode`, `ars`) so invalid values can no longer be persisted in the first place.
- Add unit tests covering all four setters (including the exact production data that triggered the bug).

## Notes

Existing rows that already contain a control character still need a one-off data cleanup; this PR prevents recurrence and makes the page resilient to such data.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Entity/Procedure/Procedure.php
+++ b/demosplan/DemosPlanCoreBundle/Entity/Procedure/Procedure.php
@@ -1277,7 +1277,7 @@ class Procedure extends SluggedEntity implements ProcedureInterface
      */
     public function setLocationName($locationName)
     {
-        $this->locationName = $locationName;
+        $this->locationName = $this->stripControlCharacters($locationName);
 
         return $this;
     }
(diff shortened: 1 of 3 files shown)


sierrazamudiodemos (approved):
thx

PR #6332, opened 2026-06-01
feat(refs: DPLAN-17455): widen _procedure.extern_id to VARCHAR(255)

breierDemos:
### Ticket
[DPLAN-17455](https://demoseurope.youtrack.cloud/issue/DPLAN-17455
)
The XSD defines no length limit for the XBeteiligung `planID`. In the dev environment the planID was observed to be composed of a work-title prefix plus a UUID, which can grow long. The planID is persisted into `_procedure.extern_id`, which was capped at `VARCHAR(50)`. A planID longer than 50 characters would cause procedure creation (0401) to fail. The observed value currently sits just under the limit (49 chars), so this is a latent problem rather than an active one.

This widens the column to `VARCHAR(255)` and adds a matching `Assert\Length(max: 255)` constraint on `Procedure::$xtaPlanId`.

### Changes
- `Procedure.php`: `extern_id` column length `50 → 255` + `Assert\Length(max: 255)`
- New migration `Version20260601060448`: `ALTER TABLE _procedure CHANGE extern_id extern_id VARCHAR(255)`. `down()` is intentionally a no-op — narrowing back to 50 would abort (strict mode) or truncate (non-strict) any planID already stored beyond 50 chars.

### How to review/test
- Apply migration, confirm `_procedure.extern_id` is `varchar(255)`.
- Re-running `dplan:migrations:diff` produces no `extern_id` change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@
 - Add text custom field definition editing
 - Add customer admin interface for managing procedure phases - displaying and creating new phases
 
+### Changed
+- Widen `_procedure.extern_id` from `VARCHAR(50)` to `VARCHAR(255)` so XBeteiligung planIDs longer than 50 characters can be stored (DPLAN-17455)
+
 ### Fixed
 - `AccessProcedureListener` now checks for array controller before accessing index, preventing crashes on API Platform routes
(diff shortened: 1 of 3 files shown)


mussbach (approved):
lgtm, thx

PR #6333, opened 2026-06-01
DPLAN-17832: render inline images with visible link in statement views

meissnerdemos:
### Ticket
DPLAN-17832

### Description
- Inline images in statement and segment text now render as a wrapper holding the image plus a visible link below it
- The `inlineImageAnchors` helper was extended to cover bare `` tags (e.g. those produced by `DpEditor`'s image button) in addition to the existing `pdf_importer_image` anchors.
- Transform is idempotent and falls back through alt text → filename → `Bild öffnen` for the link label. 
- The segmentation editor's ProseMirror node spec now emits the same wrapper structure at edit time so the editor and the rendered views match.
- File routes `core_file`/`core_file_procedure` support opt-in `?disposition=inline` for raster images (png/jpeg/gif/webp, no SVG) - image links open in the browser instead of downloading; all other requests keep `attachment` + `application/octet-stream`
- Editing contexts (PDF import confirm, editor) render plain images without the link wrapper - display markup must not be persisted into statement text

### How to review/test
0. You can find a PDF with an image to upload in the YT ticket
1. Log in as admin, open procedure.
2. Import a PDF statement with an image.
3. Check the image renders in all statement/segment views with a link below it.
4. Add a new statement manually with an image.
5. Separate it into segments and check that the image renders there with a link.
6. Repeat step 3.

### PR Checklist
- [x] Create/Update tests
- [x] Run `yarn lint`
- [x] Run `yarn test`


client/js/components/procedure/StatementSegmentsList/SegmentComment.vue
```diff
@@ -20,12 +20,13 @@
...
       </span>
     </div>
-    <p v-cleanhtml="comment.attributes.text" />
+    <p v-cleanhtml="displayText" />
   </div>
 </template>
 
 <script>
 import { CleanHtml, formatDate } from '@demos-europe/demosplan-ui'
+import { inlineImageAnchors } from '@DpJs/lib/shared/inlineImageAnchors'
```
hwiem: Why do we need this here? Is it possible to insert images into comments? I think it's also not defined in the ticket, or am I missing something?

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -614,7 +615,8 @@ export default {
 
     visibleRecommendation () {
       const shortText = this.segment.attributes.recommendation.length > 40 ? this.segment.attributes.recommendation.slice(0, 40) + '...' : this.segment.attributes.recommendation
-      return this.isCollapsed ? shortText : this.segment.attributes.recommendation
+      const text = this.isCollapsed ? shortText : this.segment.attributes.recommendation
```
hwiem: Similar question as above, should it be possible to add images to recommendation text? Shouldn't this be done for the segment text instead?

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,11 +1,99 @@
...
+const WRAPPER_UTILITY_CLASSES = `${WRAPPER_CLASS} inline-block text-center`
+const IMAGE_UTILITY_CLASSES = 'block'
+const LINK_UTILITY_CLASSES = 'pdf-importer-image-link block mt-1'
+const DEFAULT_FALLBACK_LABEL_KEY = 'image.open'
+
+const filenameFromSrc = (src) => {
+  if (typeof src !== 'string' || src === '' || src.startsWith('data:')) {
+    return ''
+  }
+  try {
```
hwiem, line 15: nitpick: here and below - empty lines are missing in this file after const declarations and if blocks and before return statements

```suggestion
  }
  
  try {
```

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,11 +1,99 @@
...
+  anchor.className = LINK_UTILITY_CLASSES
+  anchor.setAttribute('href', href)
+  anchor.setAttribute('target', target)
+  anchor.setAttribute('rel', rel)
+  anchor.textContent = label
+  return anchor
+}
+
+const isInsideWrapper = (element) =>
+  element.closest(`.${WRAPPER_CLASS}`) !== null
```
hwiem: Maybe it's just me, but I find this a lot more readable, especially if there is a line break:
```suggestion
const isInsideWrapper = (element) => {
  return element.closest(`.${WRAPPER_CLASS}`) !== null
 }
```

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,11 +1,99 @@
...
+  }
+  try {
+    const path = src.split(/[?#]/)[0]
+    const slashIndex = path.lastIndexOf('/')
+    if (slashIndex === -1) {
+      return ''
+    }
+    const segment = path.substring(slashIndex + 1)
+    if (segment === '' || !segment.includes('.')) {
+      return ''
```
hwiem: I think in demosplan we should reserve the name segment for segment entities and anything that's related to them.

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -17,11 +105,60 @@ export function inlineImageAnchors (html, className = DEFAULT_CLASS) {
   }
 
   root.querySelectorAll(`a.${className}[href]`).forEach((anchor) => {
-    const img = doc.createElement('img')
-    img.setAttribute('src', anchor.getAttribute('href'))
-    img.setAttribute('alt', anchor.textContent.trim())
-    img.setAttribute('loading', 'lazy')
-    anchor.replaceWith(img)
+    if (isInsideWrapper(anchor)) {
```
hwiem, line 123: I'd say we probably want to move the two `forEach` loops into separate functions, or, if that doesn't really make the code more readable, we should at least add some comments explaining why we are doing this. E.g. I think it's not exactly self-explanatory why we're handling the `orphanLink` below

tests/frontend/unit/inlineImageAnchors.test.js
```diff
@@ -7,46 +7,62 @@
  * All rights reserved
  */
 
+globalThis.Translator = { trans: jest.fn(key => key === 'image.open' ? 'Bild öffnen' : key) }
```
hwiem:
```suggestion
global.Translator = { trans: jest.fn(key => key === 'image.open' ? 'Bild öffnen' : key) }
```

tests/frontend/unit/inlineImageAnchors.test.js
```diff
@@ -73,5 +89,85 @@ describe('inlineImageAnchors', () => {
 
     expect(result).toContain('<img')
     expect(result).toContain('src="x"')
+    expect(result).toMatch(/<a[^>]*href="x"[^>]*>L<\/a>/)
+  })
+
+  it('wraps a bare img tag with a link below using its alt as label', () => {
```
hwiem: Shouldn't it be `'adds a link below a bare img tag using its alt as label'` or something like that?

tests/frontend/unit/inlineImageAnchors.test.js
```diff
@@ -73,5 +89,85 @@ describe('inlineImageAnchors', () => {
...
+    const html = '<p><img src="http://example.com/photo.jpg" alt="My photo" loading="lazy"></p>'
+    const result = inlineImageAnchors(html)
+
+    expect(result).toContain('class="pdf-importer-image-wrapper inline-block text-center"')
```
hwiem: I'd say we don't want to assert that these classes exist in the result, because it makes the test brittle. If e.g. we removed `'inline-block'`, the functionality would be unchanged, but the test would fail.

hwiem (changes requested):
When I tested these changes, I ran into a few issues (I guess they must be local issues, but maybe you could confirm): After uploading a statement PDF with images, in the Überprüfen/Import bestätigen views, the images were not displayed and the links did not work, and in the split statement view, the statement didn't load at all and there was this error in the console: 
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'marks')
    at replaceMarkInRange
```
If you don't have these issues, could you maybe provide the pdf that you uploaded?

client/js/components/procedure/StatementSegmentsList/SegmentComment.vue
meissnerdemos: Removed in b66aa4dc3306f2c812a8aeec45826b4a0d0b66bc

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
meissnerdemos: Removed in 71d3b26a4cb1e2eff691f427a010f013d166b31e

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,11 +1,99 @@
...
+ *   2. Manually entered: bare `<img src="…">` (e.g. from DpEditor's
+ *      imageButton on caseworker / citizen forms).
+ *
+ * Stored HTML is never mutated — the transform runs on the display string
+ * only. The result is idempotent: re-running on already-wrapped HTML is a
+ * no-op.
  */
-export function inlineImageAnchors (html, className = DEFAULT_CLASS) {
-  if (typeof html !== 'string' || !html.includes(className)) {
+export function inlineImageAnchors (html, className = DEFAULT_CLASS, fallbackLabel = Translator.trans(DEFAULT_FALLBACK_LABEL_KEY)) {
```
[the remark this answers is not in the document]
meissnerdemos: Added a typeof Translator guard in 3c2458c7e8db495eeccdf6cdaaf02ba62c363055

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,11 +1,99 @@
...
+    const path = src.split(/[?#]/)[0]
+    const slashIndex = path.lastIndexOf('/')
+    if (slashIndex === -1) {
+      return ''
+    }
+    const segment = path.substring(slashIndex + 1)
+    if (segment === '' || !segment.includes('.')) {
+      return ''
+    }
+    return decodeURIComponent(segment)
```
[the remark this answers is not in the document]
meissnerdemos: Fixed in 3c2458c7e8db495eeccdf6cdaaf02ba62c363055

client/js/components/statement/splitStatement/SegmentationEditor.vue
```diff
@@ -156,6 +156,26 @@ export default {
...
+            img.className = 'block'
+            const link = document.createElement('a')
+            link.className = 'pdf-importer-image-link block mt-1'
+            link.setAttribute('href', node.attrs.src)
+            link.setAttribute('target', '_blank')
+            link.setAttribute('rel', 'noopener noreferrer')
+            link.textContent = (node.attrs.alt || '').trim() || Translator.trans('image.open')
+            wrapper.appendChild(img)
+            wrapper.appendChild(link)
+            return { dom: wrapper }
```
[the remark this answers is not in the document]
meissnerdemos: Unified via shared buildInlineImageFigure in 3c2458c7e8db495eeccdf6cdaaf02ba62c363055

client/js/lib/shared/inlineImageAnchors.js
meissnerdemos: Changed in 7988e1e58532c35980fabaf7fa5371539e915528

client/js/lib/shared/inlineImageAnchors.js
meissnerdemos: Renamed to basename in 5b20cda1634769732d5c5436429164a70b035b89

client/js/lib/shared/inlineImageAnchors.js
meissnerdemos: Extracted into helpers in 7988e1e58532c35980fabaf7fa5371539e915528

tests/frontend/unit/inlineImageAnchors.test.js
meissnerdemos: Changed to global.Translator in f382ba7c2123b8315c35a0a1975a2526ddcac038

tests/frontend/unit/inlineImageAnchors.test.js
meissnerdemos: Renamed in 9d23c2e179711715b67dd1fd446c76671d0b9ccb

tests/frontend/unit/inlineImageAnchors.test.js
meissnerdemos: Replaced with structural checks in b1644bcf2abfa74592e8e2f7cec0015d0c877216

client/js/lib/shared/inlineImageAnchors.js
meissnerdemos: Added in f89ee89105f36eac1ce20c4f9ba9db263daa66fc through eslint fix

hwiem (approved):
FE lgtm + works

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,10 +1,226 @@
...
+    anchor.replaceWith(wrapper)
+  })
+}
+
+const wrapBareImages = (root, doc, fallbackLabel) => {
+  /*
+   * Run after wrapImporterAnchors so that importer <img> nodes produced by
+   * that pass are already inside a wrapper and skipped by isInsideWrapper.
+   */
+  root.querySelectorAll('img').forEach((img) => {
```
mussbach, line 175: An `` that is already inside an `<a>` gets wrapped here too, which produces nested anchors. The parser then hoists the inner one out of the wrapper:

input `<a href="/page">Pic</a>`
after `innerHTML` → `<a href="/page"><span class="…wrapper"></span></a><a href="/file/h?disposition=inline">Pic</a>`

Two links, and the label is no longer inside the centered wrapper. `if (img.closest('a')) return` covers it.

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,10 +1,226 @@
...
 
 /**
- * Replace `<a class="<className>" href="…">label</a>` anchors with `<img>` tags
- * so image links render inline. Display-time only — stored HTML is untouched.
+ * The file routes serve attachment by default, so a plain link click would
+ * download the image. The parameter asks the backend to serve raster images
+ * inline so the new tab actually displays the image.
  */
-export function inlineImageAnchors (html, className = DEFAULT_CLASS) {
+const withInlineDisposition = (href) => {
```
mussbach, line 30: This appends the param to any href that isn't `data:`/`blob:`, so absolute third-party URLs get it too — e.g. `https://cdn.example.com/x.png?X-Amz-Signature=abc&disposition=inline`. Only `core_file`/`core_file_procedure` read the param, and on a pre-signed URL the extra query arg breaks the signature. Worth scoping to same-origin `/file/` hrefs.

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,10 +1,226 @@
...
+ * Stored HTML is never mutated — the transform runs on the display string
+ * only. The result is idempotent: re-running on already-wrapped HTML is a
+ * no-op.
+ */
+export function inlineImageAnchors (html, className = DEFAULT_CLASS, fallbackLabel = null) {
+  if (typeof html !== 'string') {
+    return html
+  }
+
+  if (!html.includes(className) && !html.includes('<img')) {
```
mussbach: `includes('` skips the transform entirely and renders without the link. Same guard at line 253 in `stripInlineImageAnchors`. `/<img/i.test(html)` covers both.

client/js/lib/shared/inlineImageAnchors.js
```diff
@@ -1,10 +1,226 @@
...
+    const slashIndex = path.lastIndexOf('/')
+    const basename = slashIndex === -1 ? path : path.substring(slashIndex + 1)
+
+    if (basename === '' || !basename.includes('.')) {
+      return ''
+    }
+
+    return decodeURIComponent(basename)
+  } catch (error) {
+    console.error('filenameFromSrc: failed to decode src', error)
```
mussbach: `decodeURIComponent` throwing here is expected for a filename with a stray `%` (e.g. `50%_plan.png`), and the fallback already handles it. Logging at error level puts it in the console on every render — a comment or `console.debug` would satisfy the Sonar rule without the noise.

client/js/lib/shared/inlineImageAnchors.js
hwiem: Fixed in fc4a4e9050be8a1f104361298622b0cd709db9df

client/js/lib/shared/inlineImageAnchors.js
hwiem: Fixed in 428dcbf5dc8854433dff20d6d6166e23abdec0da

client/js/lib/shared/inlineImageAnchors.js
hwiem: Done in 177c73dc365437b7ebe8914f8979efe6a791b8da

client/js/lib/shared/inlineImageAnchors.js
hwiem: Done in 160a1661ef92d9ee8805b63ff56a9da1cf902047

mussbach (approved):
lgtm, code review only, thx for this one

PR #6334, opened 2026-06-01
fix(refs: DPLAN-17502): add dissolving of visibility groups with only…

huellnerdemos:
### Ticket
DPLAN-17502

This PR adds the behavior, that dissolves visibility groups, when one of the last two members is deleted completely. 

### How to review/test

1. Login as FP-Admin and choose a procedure
2. Open "Planungsdokumente und Planzeichnungen"
3. Go to the dialog, where you can define map layers
4. Define at least two map layers and group them
5. Delete one of the map layers and check if the group was dissolved too

### PR Checklist

- [x] Run `yarn lint`

### DIFF
--- a/client/js/components/map/admin/AdminLayerListItem.vue
+++ b/client/js/components/map/admin/AdminLayerListItem.vue
@@ -792,6 +792,28 @@ export default {
           route: 'layer',
           relationshipType: 'gisLayers',
         }
+
+        const groupId = this.layer.attributes.visibilityGroupId
+        if (groupId) {
+          const groupMembers = this.$store.getters['Layers/elementsListByAttribute']({
+            type: 'visibilityGroupId',
+            value: groupId,
+          })
+          if (groupMembers.length <= 2) {
+            const dissolvePromises = groupMembers
+              .filter(member => member.id !== this.layer.id)
+              .map(member => this.$store.dispatch('Layers/save', {
+                ...member,
+                attributes: { ...member.attributes, visibilityGroupId: null },
+              }))
+
+            Promise.all(dissolvePromises).then(() => {
+              this.$store.dispatch('Layers/deleteElement', deleteData)
+            })
+
+            return
+          }
+        }
       }
       this.$store.dispatch('Layers/deleteElement', deleteData)
     },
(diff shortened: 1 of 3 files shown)


riechedemos (changes requested):
The dissolve logic is correct and well placed. Thank you 🎉 

But I think the error handling in the new block won't work as intended.

Layers/save catches its own error internally (.catch(err => console.error(...)) without re-throwing), so the returned promise always resolves — even when the PATCH fails. As a result:

- The .catch() here never fires, so error.gislayer.visibility.group.dissolve is effectively dead code.
- The .then() runs regardless, so the layer gets deleted even if dissolving the group failed — leaving the remaining member with a stale visibilityGroupId, i.e. exactly the inconsistent state this PR is meant to prevent.

To make the error path actually catch failures, Layers/save would need to re-throw instead of swallowing the error — but that's a store change affecting all other callers, so it should be coordinated rather than done silently here.

The happy path works as described in the test steps; only the failure path is affected.

huellnerdemos:
> The dissolve logic is correct and well placed. Thank you 🎉
> 
> But I think the error handling in the new block won't work as intended.
> 
> Layers/save catches its own error internally (.catch(err => console.error(...)) without re-throwing), so the returned promise always resolves — even when the PATCH fails. As a result:
> 
>     * The .catch() here never fires, so error.gislayer.visibility.group.dissolve is effectively dead code.
> 
>     * The .then() runs regardless, so the layer gets deleted even if dissolving the group failed — leaving the remaining member with a stale visibilityGroupId, i.e. exactly the inconsistent state this PR is meant to prevent.
> 
> 
> To make the error path actually catch failures, Layers/save would need to re-throw instead of swallowing the error — but that's a store change affecting all other callers, so it should be coordinated rather than done silently here.
> 
> The happy path works as described in the test steps; only the failure path is affected.

Thank you for your review! :shipit:  And thank you for the point you found, it's a good one. Without a fix the group member layer would be deleted, without dissolving the group, in the error case. I checked the other callers, there is just one, that has a similar problem with it's catch path. So i decided to re-throw the error. That fixes both places. Changed it in 01c9743cc4fe47cda4a62b9c5df54e18a722c6a4

riechedemos (approved):
Thank you <3

PR #6342, opened 2026-06-02
chore: bump webpack and resolve minimatch to close high security alerts

hwiem:
- bump webpack from 5.104.1 to 5.107.2 to remove the last vulnerable `serialize-javascript@6.0.2`
- pin minimatch via resolutions - this forces `editorconfig@1.0.4`'s exact `minimatch@9.0.1` pin up to `^9.0.7` (resolves to `9.0.9`). editorconfig, which is a dependency of js-beautify, hard-pins the vulnerable version and the latest js-beautify still depends on editorconfig 1.x, so bumping it wouldn't close the security alert.
- dedupe @typescript-eslint/* onto the already-present `8.57.2`, dropping the stale `8.40.0` duplicate using a vulnerable minimatch version

Closes
- https://github.com/demos-europe/demosplan-core/security/dependabot/157
- https://github.com/demos-europe/demosplan-core/security/dependabot/215
- https://github.com/demos-europe/demosplan-core/security/dependabot/156

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`


huellnerdemos (approved):
Looks good, as far as i can say 😄 Thank you! 🎉

PR #6347, opened 2026-06-02
feat(refs DPLAN-17527, #AB50748): Edit custom procedure phases

rafelddemos:
### Ticket
[DPLAN-17527](https://demoseurope.youtrack.cloud/issue/DPLAN-17527/NEU-ADO-50748-Bearbeiten-von-custom-Verfahrensschritten)

This PR adds edit functionality for the core table phase fields. 
Phase name has to be unique within one target group.
Configuration phase (oder `0`) --> only the phase name and code can be edited.

### How to review/test

1. **please review only the FE changes**
2. log in as mandanten admin -->  define procedure phases
3. edit a phase (name, permissionSet, participationState, cockpit code)
4. if a duplicate name is used (within the target group organisation/public), an error should throw

### Tasks
 For the invalid border styles and the flyout-width to be applied:
- [x] Merge https://github.com/demos-europe/demosplan-ui/pull/1513
- [x] Merge https://github.com/demos-europe/demosplan-ui/pull/1501
- [x] Update demosplan-ui in demosplan 

### Linked PRs 

BE-PR: https://github.com/demos-europe/demosplan-core/pull/6394
BE-PR: https://github.com/demos-europe/demosplan-core/pull/6396

### PR Checklist

- [x] Run `yarn lint`
- [x] Update changelog


sierrazamudiodemos:
Holiii I just tested quickly, and I am allowed to insert an empty phase name. maybe we should validate this

rafelddemos:
> Holiii I just tested quickly, and I am allowed to insert an empty phase name. maybe we should validate this

@sierrazamudiodemos good catch, thanks! added FE validation in bebf7317f07543bcd3aacaf472f52b578ff78320

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
```diff
@@ -109,50 +116,128 @@ All rights reserved
...
+              <template v-slot:permissionSet="phase">
+                <dp-multiselect
+                  v-if="editingRowId === phase.id && phase.orderInAudience !== 0"
+                  :id="`phasePermissionSet-${phase.id}`"
+                  :data-cy="`procedurePhases:editPermissionSet:${phase.id}`"
+                  :options="permissionSetOptions"
+                  :value="findPermissionSetOption(draftCoreRowValue.permissionSet)"
+                  label="label"
+                  track-by="value"
+                  @input="option => updateCoreRowValue('permissionSet', option?.value ?? '')"
```
riechedemos: dp-multiselect here allows deselection (default allow-empty="true"), so clearing the selection emits null and option?.value ?? '' sends an empty string — which the backend rejects (Assert\Choice). Other required single-selects in the project (e.g. SelectStatementCluster, DpFaqItem, fragment Edit) set :allow-empty="false" so the value can never be cleared. Suggest doing the same here; the ?? '' fallback then becomes unnecessary.

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
```diff
@@ -256,45 +325,134 @@ export default {
...
-      id: phase.id,
-      name: phase.name,
-      orderInAudience: phase.orderInAudience,
-      permissionSetLabel: permissionSetOptions.find(option => option.value === phase.permissionSet)?.label || phase.permissionSet,
-      participationStateLabel: phase.participationState === 'finished' ? Translator.trans('yes') : Translator.trans('no'),
-    })
+    const flyoutWidth = ref('80px')
+    const hasAttemptedSubmit = ref(false)
+
+    const showInvalidInputStyle = computed(() =>
```
riechedemos: Nitpick: This computed ORs three flags spanning both contexts (isNewPhaseNameDuplicate from the create form, isEditedPhaseNameDuplicate / isEditedPhaseNameEmpty from the edit row) and is bound to both the create input and the edit input. It only works because the modes reset each other's state (openCreateForm nulls editingRowId, startEdit calls resetForm). This coupling is fragile — consider separate computeds per context so create and edit validation can't bleed into each other.

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
```diff
@@ -256,45 +325,134 @@ export default {
...
+        phase.id !== excludeId &&
+        phase.audience === audience &&
+        phase.name.trim().toLowerCase() === trimmedName.toLowerCase(),
+      )
+    }
+
+    /*
+     * ==============================
+     *  Phase list logic
+     * ==============================
```
riechedemos: Nitpick: The ==== banner section comments aren't used anywhere else in the client codebase. Maybe consider a plain single-line comment (e.g. // --- Edit phase logic ---) to stay consistent with the rest of the project, or rely on the function-level doc comments that are already there.

riechedemos (dismissed):
I could not test it jet, but found a little bug in the multiselect and a nitpick.

riechedemos:
I tested it and it works :tada:

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
rafelddemos: good catch, thanks! adjusted in 937eb6a5f2e5cd2ecc35daa8ab0eabde4b60c060

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
rafelddemos: good idea, changed in 2a047494ad4b75652f5bb265d2a116c67a614931

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
rafelddemos: Yeah, there are not many files in core that have most of the reactive state in the setup() function, where the logic is grouped differently, so we still have to agree upon how we want to do it. I added ====  so that the sections are directly well visible, but it was rather clumsy ;) - changed it to single-line and used capital letters for the eye-catching effect in 2a047494ad4b75652f5bb265d2a116c67a614931

riechedemos (approved):
Thank you!

PR #6349, opened 2026-06-03
chore: prefix Keycloak/OAuth login logs with oauthAuthenticator

mussbach:
## What

Prefixes log messages across the Keycloak/OAuth authentication flow with `oauthAuthenticator:` so they can be filtered easily in Kibana.

Touched: authenticators, Keycloak/Ozg user & role data mappers, session manager, OAuth token lifecycle services (storage/refresh/expiration), and the Keycloak logout path. 99 messages across 15 files.

## Why

Login-via-Keycloak log lines were scattered with no common token, making them hard to isolate when debugging auth issues in Kibana. A consistent prefix makes them greppable.

## Notes

- The prefix is inserted right after the opening quote, so structured context arrays are unchanged.
- The Azure AD logout log line is deliberately left untouched (not part of the Keycloak/OAuth path).


graupnerdemos (approved):
lgtm, we probably should think about logger namespacing in other places too

PR #6352, opened 2026-06-03
fix(refs: DPLAN-17776): catch Error from uninitialized lazy proxy properties on procedure update

breierDemos:
### Ticket
DPLAN-17776

Saving a procedure pictogram (alt text + copyright) failed on procedure update with:

> `Error: Typed property Symfony\Component\VarExporter\Internal\LazyObjectState::$realInstance must not be accessed before initialization`

`PostProcedureUpdatedEvent::determineModifiedValues()` compares the procedure state before/after an update via reflection. When recursing into a nested object that is a Doctrine lazy proxy (now backed by Symfony's `LazyObjectState` since the ORM 3 upgrade), reflection enumerates the proxy's uninitialized typed property `$realInstance`. Reading it throws a PHP `Error` — but the existing guard only caught `Exception`, so the error escaped and aborted the entire procedure save.

This regressed when Doctrine ORM was upgraded from 2.x to 3.x: ORM 3 replaced the old proxy mechanism with Symfony native lazy objects, which throw `Error` (not `Exception`) on uninitialized typed-property access.

Fix:
- Skip uninitialized typed properties up front via `ReflectionProperty::isInitialized()` (recorded in `fieldsNotPresentInNewProcedure`, the existing "couldn't compare" mechanism).
- Broaden the `catch` from `Exception` to `Throwable` as a safety net.

### How to review/test
On a procedure with a pictogram, set an alternative text and copyright and save. Before the fix this throws the `LazyObjectState` error and the values are not persisted; after the fix the save succeeds.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
+++ b/demosplan/DemosPlanCoreBundle/Event/Procedure/PostProcedureUpdatedEvent.php
@@ -16,8 +16,8 @@ use DateTime;
 use DemosEurope\DemosplanAddon\Contracts\Events\PostProcedureUpdatedEventInterface;
 use demosplan\DemosPlanCoreBundle\Entity\Procedure\Procedure;
 use demosplan\DemosPlanCoreBundle\Event\DPlanEvent;
-use Exception;
 use ReflectionClass;
+use Throwable;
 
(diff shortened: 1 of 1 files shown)


MoritzMandler (approved):
lgtm

PR #6371, opened 2026-06-04
feat (DPLAN-12697): merge adjacent segment-marks into one segment, preserve HTML

MoritzMandler:
### Ticket

### Description
merge adjacent segment-marks into one segment, preserve HTML

SegmentMarkParser now treats <segment-mark> as the inline mark it is: a
single segment can be represented by several adjacent marks sharing one
data-segment-id (the mark closes and reopens at every block boundary and
around inline formatting). The parser groups all marks of an id and
reconstructs the segment's HTML from the original tree — keeping each
mark's content plus its ancestor chain, pruning everything else — instead
of emitting one entry per mark with strip_tags.

This fixes two issues in the new exchange format:
- multiple marks per id no longer create duplicate Segment primary keys
- segment text keeps its block/inline structure ```(<p>, <ul><li>, <strong>)``` ,  matching how segment text is stored elsewhere and rendered in the export

DraftsInfoToSegmentTransformer needs no change: it receives one merged
entry per id, so the per-id metadata/tag lookup stays 1:1.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/SegmentMarkParser.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/SegmentMarkParser.php
@@ -45,7 +45,7 @@ class SegmentMarkParser
      * Parse HTML with <segment-mark data-segment-id="..."> elements.
      *
      * @return array<int, array{segmentId: string, text: string}> One entry per
-     *                                                             segment, ordered by document position
+     *                                                            segment, ordered by document position
      */
     public function parse(string $html): array
     {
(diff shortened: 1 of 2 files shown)


sierrazamudiodemos (commented):
Thx for this.
I have only one concern: I think there might be performance issues because there are recursive methods.
If we have like statements with more than 50 segments, and we bulk import them. then we might need to look for another approach

sierrazamudiodemos (approved):
thx for researching and testing with a big amount of segments in a statement.
we will solve performance isues when they show up! thanks!! 👍

PR #6373, opened 2026-06-04
fix: Protect customFields under permission

sierrazamudiodemos:
Skip loading custom field definitions on the institution list when the `feature_organisations_custom_fields` permission is disabled. Previously the API call was always made, causing an error notification when the feature is intentionally turned off.


huellnerdemos (approved):
Looks good, thank you!
But i think we should adjust it a bit later. Like we talked about it, to put the check around the function calls

PR #6375, opened 2026-06-05
fix: avoid deadlock on blueprint-based procedure creation (DPLAN-11634)

mussbach:
### Ticket
DPLAN-11634

Creating a procedure from a blueprint copies its sub-entities (elements, paragraphs, topics, ...) within a single transaction, which takes FK locks on the shared blueprint rows. When two creations from the same blueprint run concurrently, InnoDB can deadlock (`SQLSTATE[40001]` / `1213`); the creation then fails and the request returns `200` instead of the expected redirect. This was observed intermittently for real users and constantly under parallel Cypress runs.

This wraps `ProcedureService::addProcedureEntity()` in a per-blueprint lock (`procedure-create-from-blueprint-<id>`), so concurrent creations from the same blueprint queue instead of contending for the same locks.

Note: `LOCK_DSN=flock` is node-local, so this serializes within a php-fpm node rather than cluster-wide. It removes the deadlock for the single-node case and substantially reduces it otherwise; moving the blueprint copy into a deferred/async step remains the longer-term structural fix.

The second commit re-enables 18 previously CI-skipped tests in `ProcedureServiceTest` that cover this path. All fixes were test-only: adding the missing `procedureType` to creation data (mandatory for non-master procedures in production), replacing removed accessors (`getDoctrine()`, `getPublicProcedureRepository()`), seeding the in-memory `default.storage` so file copies succeed, aligning the master-blueprint authorized-users assertion with current behavior (T15644), and comparing copied tags/topics by title instead of object identity.

### How to review/test
Run the re-enabled tests:

```
vendor/bin/phpunit --filter testElementsOnCreateProcedure tests/backend/core/Procedure/Functional/ProcedureServiceTest.php
```

18 tests / 711 assertions pass. A number of unrelated delete/purge and count-drift tests in the same file remain skipped (they need separate domain decisions, not mechanical fixes).

### PR Checklist
- [x] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
@@ -118,6 +118,7 @@ use Psr\Log\LoggerInterface;
 use ReflectionException;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\Lock\LockFactory;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
@@ -225,6 +226,7 @@ class ProcedureService implements ProcedureServiceInterface
         private readonly CustomFieldConfigurationRepository $customFieldConfigurationRepository,
         private readonly LoggerInterface $logger,
         private readonly ProfilerService $profilerService,
+        private readonly LockFactory $lockFactory,
     ) {
         $this->contentService = $contentService;
         $this->elementsService = $elementsService;
@@ -838,6 +840,19 @@ class ProcedureService implements ProcedureServiceInterface
      */
     public function addProcedureEntity(array $data, string $currentUserId): Procedure
     {
+        // DPLAN-11634: Creating a procedure from a blueprint copies its sub-entities
+        // (elements, paragraphs, topics, ...) within a single transaction, taking FK
+        // locks on the shared blueprint rows. Concurrent creations from the same
+        // blueprint can therefore deadlock (SQLSTATE 40001 / 1213). Serialize them with
+        // a per-blueprint lock so they queue instead of contending for the same locks.
+        $blueprintId = $data['copymaster'] ?? null;
+        $blueprintId = $blueprintId instanceof Procedure ? $blueprintId->getId() : $blueprintId;
+        $lock = null;
+        if (null !== $blueprintId) {
+            $lock = $this->lockFactory->createLock('procedure-create-from-blueprint-'.$blueprintId, ttl: 300);
+            $lock->acquire(blocking: true);
+        }
+
         try {
             // T15853 + T10976: default while allowing complete deletion of emailTitle by customer:
             $data['settings']['emailTitle'] ??= '';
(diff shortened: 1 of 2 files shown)


demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
```diff
@@ -838,6 +840,19 @@ public function getAuthorizedUsers(
...
+        // (elements, paragraphs, topics, ...) within a single transaction, taking FK
+        // locks on the shared blueprint rows. Concurrent creations from the same
+        // blueprint can therefore deadlock (SQLSTATE 40001 / 1213). Serialize them with
+        // a per-blueprint lock so they queue instead of contending for the same locks.
+        $blueprintId = $data['copymaster'] ?? null;
+        $blueprintId = $blueprintId instanceof Procedure ? $blueprintId->getId() : $blueprintId;
+        $lock = null;
+        if (null !== $blueprintId) {
+            $lock = $this->lockFactory->createLock('procedure-create-from-blueprint-'.$blueprintId, ttl: 300);
+            $lock->acquire(blocking: true);
```
sierrazamudiodemos: is it possible that acquire might throw an exception and maybe the lock is never released? then we could put it inside the try.. but i am not sure if it is a realistic case

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
```diff
@@ -838,6 +840,19 @@ public function getAuthorizedUsers(
...
+        // DPLAN-11634: Creating a procedure from a blueprint copies its sub-entities
+        // (elements, paragraphs, topics, ...) within a single transaction, taking FK
+        // locks on the shared blueprint rows. Concurrent creations from the same
+        // blueprint can therefore deadlock (SQLSTATE 40001 / 1213). Serialize them with
+        // a per-blueprint lock so they queue instead of contending for the same locks.
+        $blueprintId = $data['copymaster'] ?? null;
+        $blueprintId = $blueprintId instanceof Procedure ? $blueprintId->getId() : $blueprintId;
+        $lock = null;
+        if (null !== $blueprintId) {
+            $lock = $this->lockFactory->createLock('procedure-create-from-blueprint-'.$blueprintId, ttl: 300);
```
sierrazamudiodemos: ttl is 300, but here in .env it says `LOCK_DSN=flock`
I understand then that means that ttl is ignored
but i guess ttl is introduced here because on the client's server they will have redis or other mechanism, and then TTL is relevant?

sierrazamudiodemos (commented):
I wrote 2 concerns. thanks for this and reviving the tests!!

bakridemos (approved):
lgtm, thx

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
mussbach: Good catch to flag it. As written it was already safe: if `acquire(blocking: true)` throws, the lock was never held, so there is nothing to release. A leak could only happen if acquire *succeeded* and release then did not run — and release lives in `finally`, which covers that.

That said, I moved the acquisition inside the `try` in 3a6839432 to make the intent obvious and cover any acquire-time edge case. `release()` on an unheld lock is a safe no-op.

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureService.php
mussbach: Exactly right. `flock` (filesystem locks) is released automatically by the OS when the fd/process goes away, so it ignores the TTL. The 300s TTL is the safety net for store backends that do *not* auto-release — Redis, PDO, etc. — which is what clients running a different `LOCK_DSN` may use. There it prevents a crashed process from leaving the blueprint locked indefinitely.

PR #6377, opened 2026-06-05
chore: resolve per-request DebugClassLoader and api-platform deprecations

mussbach:
## What

Follow-up to the per-request deprecation cleanup. Resolves the remaining **core-owned** deprecations that fire on every request (observed via the Symfony profiler).

### ValueObjects — "Class X should implement method Y(): type"
Replaced `@method` magic-getter docblocks with explicit typed getters that delegate to the existing locked accessor, so the classes satisfy the interface contracts DebugClassLoader checks:
- `FileInfo` (7 getters)
- `Map/CoordinatesViewport` (4 getters)
- `CommonUserData` (11 getters) — covers both `KeycloakUserData` and `OzgKeycloakUserData`

### Implementations — "might add native return type"
Added explicit `@return` / native return types matching the addon interface methods:
- `FileService::getFileById`, `FileUploadService::prepareFilesUpload`, `StatementHandler::newStatement` / `updateStatementObject`, `UserService::findDistinctUserByEmailOrLogin`, `TusCors::handle`, `HasSegmentsClause::apply` (native `never` — always throws), `APIPagination::getSort`, `DraftStatement::getCategories`, `GlobalConfig::getMapPublicAvailableScales` / `getProcedureEntrypointRoute`

### api-platform
- `serializer.hydra_prefix: true` — preserves current JSON-LD output before the 4.0 default change.

## Verification

- Cache rebuilds cleanly; `hydra_prefix` applied.
- Force-loaded all touched classes under `DebugClassLoader`: none of them emit "should implement" or "native return type" deprecations anymore.
- All changed files pass PHPStorm inspections.

## Risk

Behavior-neutral: the new getters delegate to the same locked accessor the magic `__call` used; the return-type annotations match what the methods already return. No schema or runtime changes.

### DIFF
--- a/config/packages/api_platform.yaml
+++ b/config/packages/api_platform.yaml
@@ -21,3 +21,6 @@ api_platform:
             rfc_7807_compliant_errors: true
     keep_legacy_inflector: false
     use_symfony_listeners: false
+    serializer:
+        # Preserve the current JSON-LD output; the "hydra:" prefix default flips in api-platform 4.0.
+        hydra_prefix: true
--- a/demosplan/DemosPlanCoreBundle/Entity/Statement/DraftStatement.php
+++ b/demosplan/DemosPlanCoreBundle/Entity/Statement/DraftStatement.php
@@ -1032,6 +1032,9 @@ class DraftStatement extends CoreEntity implements UuidEntityInterface, DraftSta
         return $this->uStreet;
     }
 
+    /**
+     * @return array
+     */
     public function getCategories()
     {
         if ($this->categories instanceof Collection) {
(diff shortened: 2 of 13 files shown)


bakridemos (approved):
lgtm thx, tested as far i could:
- Force-loaded all touched classes under DebugClassLoader: no return-type / "should implement" deprecations from any of them (remaining hits are pre-existing `MasterToeb`, out of scope).
- `cache:clear` rebuilds cleanly; `debug:config api_platform` confirms 
`serializer.hydra_prefix: true` applied.
- ValueObject getters verified safe — they only return always-initialized properties.

PR #6278, opened 2026-05-18
feat(refs DPLAN-17476): statement export via planner-uploaded DOCX template

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17476/Exportfunktion-fur-Antwortschreiben-je-Einwendung-inkl.-Layoutupload

### Description
statement export via planner-uploaded DOCX template

Planners can render a single statement's segments into any .docx layout they upload via TUS. New endpoint `GET /verfahren/{procedureId}/{statementId}/abschnitte/export/vorlage` (?uploadedDocxTemplate=<hash>) opens the template, validates every ${...} placeholder gainst a fixed whitelist, picks a potentially given segment-rendering marker (cloneBlock for {AbschnitteAlsAbsätze}…${/AbschnitteAlsAbsätze}, fills the values, and streams the result. The local copy of the uploaded template is removed via FileService::deleteLocalFile on every exit path.
  
Permission ```feature_statement_via_template_export``` is registered in core and granted only in projects that opt in.
Ships an example DOCX template under ```demosplan/DemosPlanCoreBundle/Resources/public/files/statement_template_example_export.docx``` and 6 translation keys under docx.export.via_template.error.* in the core messages file so other projects can enable the feature without further work.

Backend only — the modal-side FE integration is a follow-up.

### How To Test:
if you want to test before the FE has implemented this feature you can create this smoke test controller ```demosplan/DemosPlanCoreBundle/Controller/Segment/SegmentsExportSmokeTestController.php``` [SegmentsExportSmokeTestController.php.txt](https://github.com/user-attachments/files/28272401/SegmentsExportSmokeTestController.php.txt)

1. Path matters — drop it at exactly
demosplan/DemosPlanCoreBundle/Controller/Segment/SegmentsExportSmokeTestController.php for
Symfony's controller autodiscovery to pick it up (your services.yml does the rest — no
manual wiring).
2. Cache clear — after creating the file they'll need docker compose exec development 
bin/<project> cache:clear so the router registers the new ```/verfahren/{procedureId}/{statementId}/abschnitte/export/vorlage-debug``` route.
3. Prereqs — logged in as an admin/worker whose org owns the target procedure (so
ownsProcedure() passes for the feature_statement_via_template_export grant), and the
statement must have segments. The example DOCX is already in core
(demosplan/.../Resources/public/files/statement_template_example_export.docx), no extra file
needed.
4. Cleanup reminder — the file's own header says DO NOT COMMIT.

[...]

PR #6395, opened 2026-06-09
feat(refs: DPLAN-17476): add DpUploadfiles to modal

meissnerdemos:
### Ticket
DPLAN-17476

Adds file upload to `StatementExportModal`: planners can upload a DOCX template that gets connected to the export request. The file hash persists in session storage (namespaced per procedure) so the selection survives page reloads; clearing the file also clears the hash. Shows a warning when the template's column headers would override the default export columns.

### Related PR
backend #6278 

### How to review/test
1. Navigate to "Erwiderung verfassen" view and click export
3. Upload a DOCX template and reload - file should still be there
4. Export and check that the file uses the template
5. Remove file, export and check that export is back to normal
6. Type something in column header input with template uploaded — warning appears

### PR Checklist
- [x] Create/Update tests
- [ ] Run `yarn lint`
- [x] Run `yarn test`


client/js/components/statement/StatementExportModal.vue
```diff
@@ -152,6 +155,40 @@
...
+          icon="download"
+          icon-size="medium"
+          variant="subtle"
+        />
+        <dp-upload-files
+          id="uploadTemplate"
+          allowed-file-types="import"
+          data-cy="exportModal:uploadTemplate"
+          :get-file-by-hash="hash => Routing.generate('core_file_procedure', { hash, procedureId })"
+          :max-file-size="5 * 1024 * 1024"
```
hwiem: Why this exact number?

translations/messages+intl-icu.de.yml
```diff
@@ -979,10 +979,15 @@ docx.export.header.custom.placeholder: |
       other {Export Synopse {procedureName}}
   }
 docx.export.landscape: "Docx Exp. (Querformat)"
+docx.export.via_template.column.headers.warning: "Ihre individuell angepassten Spaltennamen wurden durch die Bezeichnungen aus der hochgeladenen Briefvorlage ersetzt."
```
hwiem: I would use `werden` instead of `wurden` because so far nothing has happened; the column names will be replaced once the user starts the export
```suggestion
docx.export.via_template.column.headers.warning: "Ihre individuell angepassten Spaltennamen werden durch die Bezeichnungen aus der hochgeladenen Briefvorlage ersetzt."
```

tests/frontend/unit/StatementExportModal.spec.js
```diff
@@ -410,4 +413,117 @@ describe('StatementExportModal', () => {
...
+  })
+
+  it('resets uploadedHash when resetExportModalState is called', async () => {
+    await wrapper.setData({ uploadedHash: 'some-hash' })
+    wrapper.vm.resetExportModalState()
+
+    expect(wrapper.vm.uploadedHash).toBe('')
+  })
+
+  it('hasLayoutFileAndModifiedColumnHeaders is true when single statement export, hash uploaded, permission granted, and a column title is set', async () => {
```
hwiem: I'm not sure that this test case (and the ones below) is/are useful, because the computed is very simple and we're just making sure that `true` is returned if the four variables are `true`. I think I'd rather test if the component behaves the way it should and if it displays the info it should, so here we could check "displays column header warning only if single statement export, hash uploaded, permission granted, and column title is set" and then instead of checking the computed, we check if the notification is rendered. What do you think?

hwiem (approved):
Just some nitpicks and a question. I also tested it and it works as expected. :)

client/js/components/statement/StatementExportModal.vue
meissnerdemos: The max file size is supposed to be 5MB and I saw it written like this in other places so I copied it. I will add a small comment to explain.

tests/frontend/unit/StatementExportModal.spec.js
meissnerdemos: Yes! I agree. I wasn't sure but Claude talked me into it :D I will make it just one test.

client/js/components/statement/StatementExportModal.vue
meissnerdemos: 399a139fe85f9986ed00e1f48df163bd440ebcd2

translations/messages+intl-icu.de.yml
meissnerdemos: 76da2b58c7441a0e2b7d1bcf4c164325f4eab469

tests/frontend/unit/StatementExportModal.spec.js
meissnerdemos: changed it to one test here f8246524ac19967ab1f6ea152cdc66f5cf8a6b48

PR #6394, opened 2026-06-09
feat(refs DPLAN-17527): make procedure phase definitions editable

breierDemos:
### Ticket
[DPLAN-17963](https://demoseurope.youtrack.cloud/issue/DPLAN-17963)

Custom procedure phase definitions ("Verfahrensschritte") could only be created and read. This makes them editable via the `ProcedurePhaseDefinition` resource type:

- `name`, `permissionSet` and `participationState` are now updatable (gated by `area_customer_procedure_phase_definitions`).
- The configuration phase (`orderInAudience` 0) only allows changing its `name`; `permissionSet` and `participationState` stay fixed.
- `permissionSet` is validated against its allowed values ('write', 'read', 'hidden')
- `participationState` is validated against its allowed values (`null`, `finished`, `participateWithToken`); `participateWithToken` additionally requires the `area_customer_procedure_phase_participation_token` permission.

Backend Part 2: [6396](https://github.com/demos-europe/demosplan-core/pull/6396)
Frontend Part: [6347](https://github.com/demos-europe/demosplan-core/pull/6347)

### How to review/test
PATCH `/api/2.0/ProcedurePhaseDefinition/{id}` (JSON:API, id matching in URL and body) and verify:
- editing `name`/`permissionSet`/`participationState` on a regular phase persists;
- editing `permissionSet`/`participationState` on the configuration phase (order 0) is rejected (400), while `name` succeeds;
- an invalid `participationState` value returns 400;
- `participateWithToken` without the permission returns 403.

### PR Checklist
- [ ] Create/Update tests
- [x] Link all relevant tickets
- [ ] Move the tickets on the board accordingly

PR #6396, opened 2026-06-09
feat(refs ADO-50748): report procedure phase definition changes in Verfahrensprotokoll

breierDemos:
### Ticket
https://www.dev.diplanung.de/DefaultCollection/EfA%20DiPlanung/_workitems/edit/50748

When a `ProcedurePhaseDefinition` is updated (name, permissionSet, participationState), a report entry is now created and shown in the Verfahrensprotokoll — both in the UI and the PDF export.

Backend Part 1: [6394](https://github.com/demos-europe/demosplan-core/pull/6394)
Frontend Part: [6347](https://github.com/demos-europe/demosplan-core/pull/6347)

### How to review/test
1. PATCH `/api/2.0/ProcedurePhaseDefinition/{id}`
When Frontend is finished:
1. Go to `/procedure/phase-definitions` and edit a phase definition (name, permission set or participation state)
2. Open the Verfahrensprotokoll of any procedure under that customer — the change should appear in the "Verfahrensschritt" section
3. Export the Verfahrensprotokoll as PDF — the entry should be included there as well

### PR Checklist
- [x] Create/Update tests
- [ ] Update documentation
- [ ] Update changelog


sierrazamudiodemos (changes requested):
I found a little bug:
1. Make changes of name in the Verfahrenschritt
2. then as FP admin and go to the verfahren UI and select a verfahnre and click export
3. ZIP export crashes

`demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php:757`

$reportInfo = $this->exportReportService->getReportInfo($procedureId, $this->permissions);

getReportInfo now requires 3 arguments but this call only passes 2. PHP throws ArgumentCountError at this line the moment addReportToZip is invoked.

demosplan/DemosPlanCoreBundle/Logic/Report/ReportMessageConverter.php
```diff
@@ -507,6 +517,49 @@ protected function getProcedureFinalMailMessage(ProcedureFinalMailReportEntryDat
...
+            return '';
+        }
+
+        if (ProcedurePhaseDefinitionUpdatableField::PERMISSION_SET->value === $field) {
+            $oldValue = $this->translator->trans('permissionset.'.$oldValue);
+            $newValue = $this->translator->trans('permissionset.'.$newValue);
+        } elseif (ProcedurePhaseDefinitionUpdatableField::PARTICIPANT_STATE->value === $field) {
+            $oldValue = $this->translator->trans('finished' === $oldValue ? 'yes' : 'no');
+            $newValue = $this->translator->trans('finished' === $newValue ? 'yes' : 'no');
+        }
```
sierrazamudiodemos, line 544: do we want to take into account the tokenWithParticipation case here?

sierrazamudiodemos (commented):
I found a little corner case: The phase name is fetched live from the DB to displaye it in the VerfahrenProtokol, so if the phase is renamed after a change is logged, old protocol entries will show the new name instead of the name at the time of the change. 
What do you think about that scenario?

demosplan/DemosPlanCoreBundle/ResourceTypes/ProcedurePhaseDefinitionResourceType.php
```diff
@@ -229,4 +274,24 @@ private function resolvePermissionSet(mixed $value): string
...
+        if ($oldValue !== $newValue) {
+            $reportEntry = $this->procedurePhaseDefinitionReportEntryFactory->createProcedurePhaseDefinitionUpdateEntry(
+                $procedurePhaseDefinition,
+                $field,
+                $oldValue,
+                $newValue,
+            );
+            $this->reportService->persistAndFlushReportEntry($reportEntry);
+        }
+    }
```
sierrazamudiodemos: What about moving this to a dedicate service? and perhaps other parts related to vlidataing stuff?that is because given the case that move to 3.0 api platform, already this piece of code is already isolated and the migration is easier. it is an idea

demosplan/DemosPlanCoreBundle/Logic/Report/ReportMessageConverter.php
sierrazamudiodemos, line 60: this method is a good candidate for the StrategyPattern, that said, it is out of the scope, i wanted to highlight the example.
with the strategy pattern, we would remove the if-else stuff, and everytime that we need to add a new way to convert a message, we would add a new strategy, and avoid adjusting existing files

breierDemos:
> I found a little bug:
> 
> 1. Make changes of name in the Verfahrenschritt
> 2. then as FP admin and go to the verfahren UI and select a verfahnre and click export
> 3. ZIP export crashes
> 
> `demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php:757`
> 
> $reportInfo = $this->exportReportService->getReportInfo($procedureId, $this->permissions);
> 
> getReportInfo now requires 3 arguments but this call only passes 2. PHP throws ArgumentCountError at this line the moment addReportToZip is invoked.

@sierrazamudiodemos thanks for spotting this. I was not aware of this export and that it adds the Verfahrensprotokoll. I adjusted it in [9fead20](https://github.com/demos-europe/demosplan-core/pull/6396/commits/9fead2029a8948d278b96a15df6b1ba2b75ca5a5)

demosplan/DemosPlanCoreBundle/Logic/Report/ReportMessageConverter.php
breierDemos: No I think we should it take into acount when we make "tokenWithParticipation" as a editable value in the interface. For now the user can only switch between 'finished' and null. Also I think we should replace this entire field maybe with two boolean fields later on.

demosplan/DemosPlanCoreBundle/ResourceTypes/ProcedurePhaseDefinitionResourceType.php
```diff
@@ -108,9 +137,16 @@ protected function getProperties(): ProcedurePhaseDefinitionResourceConfigBuilde
             ->addUpdateBehavior(
                 CallbackAttributeSetBehavior::createFactory(
                     [],
-                    function (ProcedurePhaseDefinition $phaseDefinition, mixed $value): array {
-                        $this->guardConfigurationPhaseNotEditable($phaseDefinition);
-                        $phaseDefinition->setPermissionSet($value);
+                    function (ProcedurePhaseDefinition $procedurePhaseDefinition, string $newPermissionSet): array {
+                        $this->guardConfigurationPhaseNotEditable($procedurePhaseDefinition);
+                        $oldPermissionSet = $procedurePhaseDefinition->getPermissionSet();
+                        $procedurePhaseDefinition->setPermissionSet($this->resolvePermissionSet($newPermissionSet));
```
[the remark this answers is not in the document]
breierDemos: adjusted in [74ab1c7](https://github.com/demos-europe/demosplan-core/pull/6396/commits/74ab1c76184e6c64cf3b8ea6be8dc9baf937eec0)

breierDemos:
> I found a little corner case: The phase name is fetched live from the DB to displaye it in the VerfahrenProtokol, so if the phase is renamed after a change is logged, old protocol entries will show the new name instead of the name at the time of the change. What do you think about that scenario?

Yes therefore we need to store the old name. I adjusted the report entry to store also the old name.
[5313a06](https://github.com/demos-europe/demosplan-core/pull/6396/commits/5313a069842f64b71dc3cbd54e953a329c6f0965)

demosplan/DemosPlanCoreBundle/Logic/Report/ReportMessageConverter.php
breierDemos: Yes thats true, but for now I think its out of scope.

demosplan/DemosPlanCoreBundle/ResourceTypes/ProcedurePhaseDefinitionResourceType.php
breierDemos: I moved it to a new service: [bbaad6e](https://github.com/demos-europe/demosplan-core/pull/6396/commits/bbaad6e0261d4264e5ff6a219e78d0af5028bd76)

sierrazamudiodemos (commented):
it looks good! the fixes work and thanks for the separation of concerns.
I found one more corner case that I would like to hear your opinion. I am sorry i somehow skipped this case before:
When the Verfahrehschritt name is updated: when i export the Original STN pdf, in the verfahrenschritt it will displayed the new name (not the one to which the original stn was created.. )

sierrazamudiodemos (approved):
thank for your job!

breierDemos:
> it looks good! the fixes work and thanks for the separation of concerns. I found one more corner case that I would like to hear your opinion. I am sorry i somehow skipped this case before: When the Verfahrehschritt name is updated: when i export the Original STN pdf, in the verfahrenschritt it will displayed the new name (not the one to which the original stn was created.. )

That should be okay, because we decided for Variante 2 in the ticket. So the change of a procedure phase definition is taken over on all entities where it is used. But for this case I will ask the PO's.

PR #6400, opened 2026-06-10
refactor(DPLAN-4633): extract consideration table docx filename builder

breierDemos:
### Ticket
[DPLAN-4633](https://demoseurope.youtrack.cloud/issue/DPLAN-4633)

Eliminates duplicated filename construction logic in `addAssessmentTableToZip` and `addAssessmentTableAnonymousToZip` by extracting a private `buildConsiderationTableDocxFilename` method. The two callers differed only in the `_Anonym` suffix, which is now driven by a `bool $anonymous` parameter. Also replaces the `switch` with a `match` expression.

### How to review/test
Read the diff — no behaviour change.


MoritzMandler (approved):
lgtm - code review only

PR #6401, opened 2026-06-10
feat (DPP-1023): Introduce and dispatch new SegmentTagsChangedEvent

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPP-1023

### Description
Introduce and dispatch new SegmentTagsChangedEvent that gets dispatched during rpc bulk segment changes on tags
this event depends on the also new implemented interface and therefore depends on new contract layer version....

### Linked PRs (optional)
https://github.com/demos-europe/demosplan-addon/pull/177

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ## UNRELEASED
 
+- new SegmentTagsChangedEvent implements new interface added to contract layer in:
+- This core version will depend on the contract layer being > v...
```
mussbach, line 11: Entry needs finishing before merge: dangling "in:" on the first line and the version placeholder.

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
```diff
@@ -70,7 +74,10 @@ class RpcSegmentsBulkEditor implements RpcMethodSolverInterface
 
     final public const SEGMENTS_BULK_EDIT_METHOD = 'segment.bulk.edit';
 
-    public function __construct(protected CurrentProcedureService $currentProcedure, protected CurrentUserInterface $currentUser, protected LoggerInterface $logger, protected JsonSchemaValidator $jsonValidator, protected PlaceService $placeService, protected ProcedureService $procedureService, protected RpcErrorGenerator $errorGenerator, protected SegmentHandler $segmentHandler, protected SegmentValidator $segmentValidator, protected TagService $tagService, protected TagValidator $tagValidator, private readonly TransactionService $transactionService, protected UserHandler $userHandler, protected SegmentBulkEditorService $segmentBulkEditorService)
+    /** @var SegmentInterface[] */
+    private array $segmentsWithTagChanges = [];
```
mussbach: This turns a stateless RPC solver into a stateful service. A local variable captured by reference does the same job without service-level state:

```php
$segmentsWithTagChanges = [];
$resultResponse = $this->transactionService->executeAndFlushInTransaction(
    function (EntityManager $entityManager) use ($procedure, $rpcRequests, &$segmentsWithTagChanges): array { ... }
);
```

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
```diff
@@ -140,6 +149,10 @@ public function execute(?ProcedureInterface $procedure, $rpcRequests): array
                         $customFields
                     );
 
+                    if ([] !== $addTagIds || [] !== $removeTagIds) {
```
mussbach: This checks that valid tags were *requested*, not that any segment's tag set actually changed — e.g. adding a tag all segments already carry still fires the event. Probably fine, but if listeners do expensive work it may be worth tightening.

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
```diff
@@ -157,6 +170,33 @@ public function execute(?ProcedureInterface $procedure, $rpcRequests): array
...
+     */
+    private function dispatchSegmentTagsChangedEvent(array $segments): void
+    {
+        $statements = [];
+        foreach ($segments as $segment) {
+            $statement = $segment->getParentStatementOfSegment();
+            $statements[$statement->getId()] = $statement;
+        }
+
+        if ([] === $statements) {
```
mussbach: Unreachable: the method is only called with a non-empty segment list and every segment has a parent statement. This guard (or the caller-side one) can go.

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
```diff
@@ -157,6 +170,33 @@ public function execute(?ProcedureInterface $procedure, $rpcRequests): array
...
+            return;
+        }
+
+        $this->eventDispatcher->dispatch(
```
mussbach, line 233: Dispatch happens after commit but before the response is returned, so an exception in a listener surfaces as an RPC error even though the DB changes are committed. If listeners can fail, consider wrapping the dispatch in try/catch + log.

mussbach (commented):
A unit test for the new dispatch logic would be cheap and valuable here: event dispatched once with deduplicated statements when tags change, not dispatched when only assignee/place/recommendation change.

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6401/commits/82bc71246e290b60254120b4a80004f0eb6626d5

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
MoritzMandler: It would be very expensive for many statements to run a diff for each contained segment and its contained tags...

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
MoritzMandler: removed in https://github.com/demos-europe/demosplan-core/pull/6401/commits/82bc71246e290b60254120b4a80004f0eb6626d5

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6401/commits/82bc71246e290b60254120b4a80004f0eb6626d5

demosplan/DemosPlanCoreBundle/Logic/Segment/RpcBulkEditor/RpcSegmentsBulkEditor.php
bakridemos: I agree the current check only reflects what was requested, not what actually changed. I think we can adjust this, and I have an idea for doing it cleanly without the expensive per-statement diff:

Statement::addTag() already returns false when the tag is already present, and we can gate removals with a contains() check. So we can detect the effective change right at the mutation site in updateSegments() and only collect the segments whose tag set truly changed — no extra queries, since those tag collections are already loaded during the update anyway.

I'll push an update along those lines so the event only fires on real tag changes. Let me know if you see any issue with that approach!    https://github.com/demos-europe/demosplan-core/pull/6401/changes/e10e56f13ec5caea53ccaed2c23f2ecfd391a738

demosplan/DemosPlanCoreBundle/Logic/Segment/SegmentBulkEditorService.php
```diff
@@ -115,12 +116,29 @@ public function assertBatchEditable(array $segmentIds, string $procedureId): voi
...
+        array $customFields,
+        ?DateTime $deadline = null,
+    ): array {
+        $segmentsWithTagChanges = [];
+
         foreach ($segments as $segment) {
-            /* @var Segment $segment */
-            $segment->addTags($addTagIds);
-            $segment->removeTags($removeTagIds);
+            if ($this->applyTagChanges($segment, $addTagIds, $removeTagIds)) {
```
mussbach, line 139: If `updateOrAddCustomFieldValues()` throws further down for segment N, `RpcSegmentsBulkEditor::execute` catches it inside the transaction, so the tag changes on segments 1..N are still saved. But `updateSegments()` never returns, so those segments don't end up in `segmentsWithTagChanges` and no event is sent for them. Maybe validate the custom fields before touching any segment, or collect the changed segments somewhere that survives the exception?

MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6401/commits/3010fe4cf2ddd822716c4cc93ff090c1161746d2

PR #6402, opened 2026-06-11
Prevent doubled subdomain in Keycloak logout redirect URI

mussbach:
### Ticket
ado-55757

`OzgKeycloakSessionManager::getLogoutUrl()` injects the current customer's subdomain into the `post_logout_redirect_uri` of the configured Keycloak logout route via a blind `str_replace` on the prefix `post_logout_redirect_uri=https://`. When the configured logout route template (global parameter or per-customer config) already contains the subdomain-qualified host, the subdomain is prepended a second time (e.g. `https://xx.xx.example.com`), which Keycloak rejects as an invalid redirect URI.

The injection is now skipped when the route already contains the subdomain-qualified redirect host. The check is anchored on the full parameter prefix including the trailing dot, so hosts merely starting with the same letters as the subdomain still get the subdomain injected correctly.

### How to review/test
Unit tests cover injection, idempotency, and the letter-collision edge case:

```
./vendor/bin/phpunit tests/backend/core/Core/Unit/Logic/User/OzgKeycloakSessionManagerTest.php
```

### PR Checklist
- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/User/OzgKeycloakSessionManager.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/User/OzgKeycloakSessionManager.php
@@ -139,11 +139,16 @@ class OzgKeycloakSessionManager
     {
         $currentCustomer = $this->customerService->getCurrentCustomer();
 
-        $logoutRoute = str_replace(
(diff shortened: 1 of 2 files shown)


breierDemos (approved):
lgtm thanks

PR #6316, opened 2026-05-27
F dplan 17129 draft statement to api platform

sierrazamudiodemos:
Replace EDT for ApiPlatform

PR #6405, opened 2026-06-11
Improve access conditions of File resource type

mussbach:
### Ticket
n/a (code improvement, no ticket)

The JSON:API `File` resource type returned every undeleted file regardless of procedure context, exposing file metadata (including the hash used by the file download routes) across procedures.

The access conditions now mirror the procedure scoping already used for statements: accessible are files without a procedure (global assets such as logos, news images or sign language videos) and files belonging to the current procedure, including procedures configured for cross-procedure segment access. Without a procedure context only procedure-less files are returned.

The current procedure can be trusted here because `AccessProcedureListener` already denies the request if the user may not enter the procedure given in the request.

### How to review/test
- `tests/backend/core/JsonApi/Functional/FileResourceTypeTest.php` covers both cases (with and without procedure context); the assertions fail against the previous access conditions.
- As a user with assessment table access, `GET /api/2.0/File` should only return files of the current procedure and procedure-less files.

### PR Checklist
- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/ResourceTypes/FileResourceType.php
+++ b/demosplan/DemosPlanCoreBundle/ResourceTypes/FileResourceType.php
@@ -13,13 +13,17 @@ declare(strict_types=1);
 namespace demosplan\DemosPlanCoreBundle\ResourceTypes;
 
 use DemosEurope\DemosplanAddon\Contracts\Entities\FileInterface;
+use DemosEurope\DemosplanAddon\Contracts\Entities\ProcedureInterface;
 use DemosEurope\DemosplanAddon\Contracts\Events\IsFileAvailableEventInterface;
 use DemosEurope\DemosplanAddon\Contracts\Events\IsFileDirectlyAccessibleEventInterface;
 use DemosEurope\DemosplanAddon\Contracts\ResourceType\FileResourceTypeInterface;
 use demosplan\DemosPlanCoreBundle\Entity\File;
+use demosplan\DemosPlanCoreBundle\Entity\Procedure\Procedure;
+use demosplan\DemosPlanCoreBundle\Entity\User\User;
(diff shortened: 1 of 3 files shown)


demosplan/DemosPlanCoreBundle/ResourceTypes/FileResourceType.php
```diff
@@ -65,13 +75,38 @@ public function isAvailable(): bool
...
+            $procedureConditions[] = $this->conditionFactory->propertyHasAnyOfValues(
+                $allowedProcedureIds,
+                $this->procedure->id
+            );
+        }
+
+        return [
+            $this->conditionFactory->propertyHasValue(false, $this->deleted),
+            $this->conditionFactory->anyConditionApplies(...$procedureConditions),
+        ];
```
sierrazamudiodemos: hey, do you think we could move this to an isolated filed? for example like in [this](https://github.com/demos-europe/demosplan-core/pull/6316/changes#diff-a05b697596acde2af9ddb6532ba14c0efe07db2c031fbdab050f999d0694118b) PR with the file `demosplan/DemosPlanCoreBundle/ResourceAccess/DraftStatementAccessChecker.php` ?
that is because when we move this resource type to 3.0, then it will not be so hard? what do you think ?

breierDemos: Good idea, done. Extracted the access logic (isAvailable, isDirectlyAccessible and getAccessConditions) into a dedicated `demosplan/DemosPlanCoreBundle/ResourceAccess/FileAccessChecker.php`, mirroring the `DraftStatementAccessChecker` pattern so it is decoupled from the EDT resource type ahead of the 3.0 migration. `FileResourceType` now just delegates to it. Tests still green.

sierrazamudiodemos (approved):
thx Loki for taking care!

PR #6406, opened 2026-06-11
fix: enforce procedure access on statement copy/move endpoints

mussbach:
The `copyStatement` / `moveStatement` API handlers loaded the source statement
by the route id and the target procedure by the `targetProcedureId` query string
without checking either against the user's procedure access. The kernel-level
procedure gate only validates the procedure named in the route, which the caller
chooses — so a user could copy/move a foreign statement into a procedure they
control (reading foreign content) or write into a procedure they do not own.

Both handlers now assert, right after loading:
1. the source statement belongs to the current (route) procedure, and
2. the user owns the target procedure (unless they hold the respective
   `feature_statement_{copy,move}_to_foreign_procedure` permission).

The ownership result also replaces the previous `permissions->setProcedure()`
switching used to pick the linked confirmation message.

### How to review/test
`DemosPlanStatementAPIControllerTest` covers both endpoints (foreign source
statement and not-owned target procedure) plus the ownership predicate. The
HTTP tests were verified to fail against the unpatched controller — a
foreign-statement move returned 200 and actually moved the statement
cross-procedure — and now return a rejection.

### PR Checklist
- [x] Create/Update tests


breierDemos (approved):
lgtm

PR #6407, opened 2026-06-11
fix(DS-548): only clear segment assignee while workflow actions are visible

mussbach:
Follow-up to DS-548. The explicit clear-assignee PATCH fired on every segment save when `selectedAssignee` was not initialized, even if the workflow actions were never opened. It is now guarded with `showWorkflowActions`, mirroring the existing condition in `updateRelationships()`.

Ticket: [DS-548](https://demoseurope.youtrack.cloud/tickets/DS-548/VKE-705.3-Vorgang-lasst-sich-nicht-abschliessen)


bakridemos:
The `save` method can only be called if `isAssignedToMe && (isEditing || showWorkflowActions)`. If `isAssignedToMe` is true, `this.selectedAssignee?.id` is defined and `isUnassigning` is false, so the unassign PATCH request is not executed. At the same time, `showWorkflowActions = true` doesn't mean the assignee has actually been changed—only that the checkbox for displaying the assignee and step has been checked.

PR #6408, opened 2026-06-12
F ds 490 reject authored after submitted on statement create

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/tickets/DS-490/BOB-SH-ROBOB-Stage-Beim-nachtraglichen-Andern-einer-STN-kann-Verfassungsdatum-grosser-Einreichungsdatum-eingetragen-werden

Description:

The cross-field date rule "authoredDate (Verfassungsdatum) must not be later than submittedDate (Einreichungsdatum)" was only enforced on the statement **edit** path. The manual **creation** route (`/statement/new/manual/{procedureId}`) had no backend check, so a user could bypass the datepicker's min/max range by typing dates directly into the inputs and create a statement with authoredDate later than submittedDate.

Such records then appear with **empty date fields** in the statement detail view — not because the data is missing, but because the datepicker refuses to render values that fall outside its allowed`[minDate, maxDate]` range (those bounds being the other date).

This PR enforces the rule on creation as well, and extracts the comparison into a single `StatementDateOrderValidator` that both the create and edit routes call, so the rule can no longer drift between them. Each route keeps its own input parsing and skip-logic (the edit route still allows editing other fields on a pre-existing record with already-invalid dates); only the comparison is shared. The existing `error.date.authored.after.submitted` translation key is reused.

### How to review/test
At `/statement/new/manual/{procedureId}`:
- Type a Verfassungsdatum **later** than the Einreichungsdatum (type directly into the field to bypass the datepicker flyout), then save → red error *"Das Verfassungsdatum darf nicht nach dem Einreichungsdatum liegen."*, no statement created, typed values preserved on re-render.
- Valid order (authored ≤ submitted) → statement is created normally.
- Confirm the edit path in the assessment table still behaves as before (DS-490 original fix).

- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanAssessmentController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanAssessmentController.php
@@ -74,7 +74,7 @@ class DemosPlanAssessmentController extends BaseController
         StatementHandler $statementHandler,
         UserService $userService,
         string $entityId,
-        string $assignOrUnassign = 'assign'
+        string $assignOrUnassign = 'assign',
     ): Response {
         $statementToUpdate = $statementHandler->getStatement($entityId);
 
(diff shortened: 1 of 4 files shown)


hwiem (approved):
To me it looks good, and I tested it in the interface and it works

PR #6409, opened 2026-06-12
fix(refs: DPLAN-18002): guard GdprConsent user lookup against null or empty submitter id

breierDemos:
### Ticket
DPLAN-18002

After the Doctrine ORM upgrade, `MissingIdentifierField` no longer extends `ORMException` (it now extends `LogicException`), so passing `null` to `EntityManager::find()` was no longer silently caught. Anonymous public statement submissions always have a null submitter id — the null check was missing, causing the entire statement submission to fail.

Added a null and empty-string guard before calling `find()` so the lookup is skipped for anonymous submitters. The now-unreachable `catch (ORMException)` was removed; any unexpected DB errors are handled by the outer `catch (Exception)` in `submitDraftStatement`.

### How to review/test
Submit a public statement (Stellungnahme) as an anonymous citizen via the public participation form.

### PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ## UNRELEASED
 
+### Fixed
+- Submitting a public statement as an anonymous citizen no longer fails with a Doctrine `MissingIdentifierField` exception (DPLAN-18002)
+
 ## v4.43.0 (2026-06-05)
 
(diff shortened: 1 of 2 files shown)


sierrazamudiodemos (approved):
thx

PR #6412, opened 2026-06-12
fix(refs: DPLAN-18002): guard GdprConsent user lookup against null or empty submitter id

breierDemos:
### Ticket
DPLAN-18002

Cherry-pick of commit a95edf9f2 from main.

After the Doctrine ORM upgrade, `MissingIdentifierField` no longer extends `ORMException`, so passing `null` to `EntityManager::find()` was no longer silently caught. Anonymous public statement submissions always have a null submitter id — the null check was missing, causing the entire statement submission to fail.

### How to review/test
Submit a public statement as an anonymous citizen via the public participation form.

### PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ## UNRELEASED
 
+### Fixed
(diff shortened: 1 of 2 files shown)


sierrazamudiodemos (approved):
thx for this

PR #6413, opened 2026-06-12
feat(DPLAN-16354): Delete custom fields only when it does not have st…

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-16354/ADO-43310-Konfigurierbare-Felder-in-Stellungnahmen-Part-5-FP-kann-Multiselect-Custom-Fields-loschen

The same restriction that already applies to creating and updating statement custom fields now also applies to deletion: if the procedure has statements, the delete request is rejected with a validation error.

### How to review/test

1. Open a procedure that already has at least one statement.
2. Go to the custom fields configuration for that procedure.
3. Try to delete a multiselect custom field that targets statements.
4. Expected: deletion is blocked with an error message.
5. Open a procedure that has **no** statements.
6. Try to delete a multiselect custom field targeting statements.
7. Expected: deletion succeeds.

### Linked PRs (optional)
- Related: #6351 (adds statement custom field deletion support — this PR adds the guard on top of it)

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Repository/StatementRepository.php
+++ b/demosplan/DemosPlanCoreBundle/Repository/StatementRepository.php
@@ -2186,4 +2186,18 @@ class StatementRepository extends CoreRepository implements ArrayInterface, Obje
             'completed'  => $completed,
         ];
     }
+
+    public function findStatementsWithCustomField(string $customFieldId): array
+    {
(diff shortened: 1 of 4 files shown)


breierDemos (approved):
lgtm :)

PR #6416, opened 2026-06-12
feat (DPLAN-17744): create log entries on procedure deletion.

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17744/Protokollierung-von-Verfahrensloschung

### Description
Adds a deletion log for procedures. Every soft-delete (via UI) and hard-delete (via purge) writes a log entry to a new `procedure_deletion_log` table. The soft-delete entry holds a nullable FK to the procedure which is nulled when the procedure is physically purged. User details (id, name, email) are stored as snapshots so the log survives user-account deletion. Hard-delete entries are attributed to "System".

### How to review/test
- Delete a procedure via the UI → a `soft` row appears in `procedure_deletion_log` with full user details
- Run the purge job → a `hard` row appears and the `procedure_fk` on the soft row is set to null
- Backend tests: `ProcedureDeletionLogServiceTest`

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/Entity/Procedure/ProcedureDeletionLog.php
```diff
@@ -0,0 +1,228 @@
...
+    private $id;
+
+    /**
+     * Set while the procedure still exists; nulled before the procedure is hard-deleted
+     * so the FK constraint does not fire during purge.
+     *
+     * @var Procedure|null
+     */
+    #[ORM\JoinColumn(name: 'procedure_fk', referencedColumnName: '_p_id', nullable: true, onDelete: 'SET NULL')]
+    #[ORM\ManyToOne(targetEntity: Procedure::class)]
```
breierDemos: Do we really need this fk? And why not OneToOne Relation. Can a procedure be soft deleted more than once? In the ProcedureDeletionLogRepository you use also findOneBy.

demosplan/DemosPlanCoreBundle/Entity/Procedure/ProcedureDeletionLog.php
```diff
@@ -0,0 +1,228 @@
...
+     * @var string|null
+     */
+    #[ORM\Column(type: 'string', length: 255, nullable: true)]
+    private $deletedByUserEmail;
+
+    /**
+     * @var string
+     */
+    #[ORM\Column(type: 'string', length: 10)]
+    private $deleteType;
```
breierDemos: There are only two types of deletion, so we could save it as boolean instead of string. What do you think? Something like $isHardDeleted.

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureHandler.php
```diff
@@ -652,6 +653,7 @@ public function purgeDeletedProcedures($limit = 1, int $retentionDays = 0): int
         $proceduresPurged = 0;
         foreach ($this->procedureService->getDeletedProcedures($limit, $deletedBefore) as $deletedProcedure) {
             $procedureId = $deletedProcedure->getId();
+            $this->procedureDeletionLogService->logHardDelete($deletedProcedure);
```
breierDemos: When the try block below is failing, than we have a hard log, but the procedure is still there, or not?

demosplan/DemosPlanCoreBundle/Repository/ProcedureDeletionLogRepository.php
```diff
@@ -0,0 +1,74 @@
...
+use DateInterval;
+use DateTime;
+use demosplan\DemosPlanCoreBundle\Entity\Procedure\ProcedureDeletionLog;
+
+/**
+ * @template-extends CoreRepository<ProcedureDeletionLog>
+ */
+class ProcedureDeletionLogRepository extends CoreRepository
+{
+    public function findSoftDeleteEntryForProcedure(string $procedureId): ?ProcedureDeletionLog
```
breierDemos: We only use this method, or not? So we dont need the other repo method currently, right?

breierDemos (changes requested):
If a log cant be created because of an error, but the procedure is still deleted (soft or hard) is that a problem? Or should the deletion of the procedure than fail?

demosplan/DemosPlanCoreBundle/Entity/Procedure/ProcedureDeletionLog.php
MoritzMandler: I agree - removed in https://github.com/demos-europe/demosplan-core/pull/6416/commits/05fbc3cc3c45b681594485ea67bef2afa7d5bb8b

demosplan/DemosPlanCoreBundle/Entity/Procedure/ProcedureDeletionLog.php
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6416/commits/05fbc3cc3c45b681594485ea67bef2afa7d5bb8b

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedureHandler.php
MoritzMandler: moved inside the try block https://github.com/demos-europe/demosplan-core/pull/6416/commits/05fbc3cc3c45b681594485ea67bef2afa7d5bb8b

demosplan/DemosPlanCoreBundle/Repository/ProcedureDeletionLogRepository.php
MoritzMandler: I think the getOlderThen method could be useful for later usage and does no harm - removed all the other unused ones. https://github.com/demos-europe/demosplan-core/pull/6416/commits/05fbc3cc3c45b681594485ea67bef2afa7d5bb8b

breierDemos (approved):
lgtm code review only

PR #6421, opened 2026-06-15
fix(refs: DPLAN-17957): notification about faulty api request in insti list

huellnerdemos:
### Ticket
DPLAN-17957

This PR adds permission checks for the institution custom fields in the institution list. It prevents calling the custom fields for the list, if the permission is not set. It also adjusts the permission handling for loading the custom fields definitions a bit that was fixed in https://github.com/demos-europe/demosplan-core/pull/6373 before. This is done to separate concerns and be nearer to the rest of the code base.

### How to review/test

1. Login as M-A and open "Institutionen verschlagworten"
2. Check for notifications

1. Login as FP-A  open "Institutionen Verwalten"
2. Check for notifications

### PR Checklist

- [x] Run `yarn lint`

### DIFF
--- a/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
+++ b/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
@@ -462,6 +462,7 @@ export default {
     },
 
     getInstitutionsByPage (page) {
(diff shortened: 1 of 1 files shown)


rafelddemos (approved):
lgtm, thx! :)
There are no error notifications now when the custom fields permission is not enabled.

PR #6429, opened 2026-06-16
fix: include place in fetch of updated segments

meissnerdemos:
Assigning a segment to a locked place didn't update the lock badge until reload. The post-save refetch used a sparse fieldset missing `locked`, which wiped the attribute from the store. Added `locked` to the fieldset.

### How to review/test
Go to "Erwiderungen verfassen", set the place for a segment to 'Abgeschlossen' -> ui should show locked icon after save without reload.


sakutademos (approved):
thx, code review, lgtm!

PR #6430, opened 2026-06-16
feat(DPLAN-18005): Fix default values

sierrazamudiodemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-18005/ADO-55976-Default-Einstellung-zu-Beteiligungsschritt-Offentlichkeit-anpassen

Two checkboxes in the public participation settings had wrong default values when creating a new procedure:

- "Veröffentlichung von Stellungnahmen" was checked by default — it should be unchecked (planners must opt in)
- "Rückmeldung zu Stellungnahme" was unchecked by default — it should be checked (required by BauGB)

Both are fixed via a migration that corrects the column defaults and backfills existing data.

**Note for deployers: After running the migration, blueprint procedures in all projects will have the publication checkbox unchecked. Projects that want it checked by default (e.g. robobsh) need to update their blueprint once via the UI.**

**How to review/test**

1. Create a new procedure in diplanbau
2. Open the procedure settings → "Verfahrensschritt Öffentlichkeit"
3. "Veröffentlichung von Stellungnahmen" → should be unchecked
4. "Rückmeldung zu Stellungnahme" → should be checked

### DIFF
--- a/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/06/Version20260615120000.php
+++ b/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/06/Version20260615120000.php
@@ -1,0 +1,101 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * This file is part of the package demosplan.
+ *
+ * (c) 2010-present DEMOS plan GmbH, for more information see the license file.
+ *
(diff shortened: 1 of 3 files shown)


breierDemos (approved):
lgtm thank you

PR #6370, opened 2026-06-04
feat(refs: DPLAN-17747): Create statement groups

riechedemos:
### Ticket

DPLAN-17747

This PR lets a planning-agency admin select statements from the statement list and combine them into a group (cluster). Grouping runs through a multi-step page (choose action → group name + main statement → confirmation) and creates the group via POST /api/3.0/StatementGroup.

The list reflects the grouping: the group gets an icon, members are hidden, and the expanded row shows the associated group (name + count). Statements that cannot be grouped (already split into segments, synchronized, or a group themselves) are shown with a reason-specific tooltip.

### How to review/test

1. Open the statement list, select several non-split statements → Group → redirects to /verfahren/{id}/einwendungen/gruppe-erstellen.
2. Step 1: review/remove selected statements. Step 2: group name + main statement (required). Step 3: confirmation → POST /api/3.0/StatementGroup succeeds.
3. Back in the list: the group head shows a folder icon, members are gone, the expanded row shows the associated group with name + "(x statements)".
4. Split / synchronized / group rows: locked checkbox whose tooltip states the reason.

### Linked PRs

- demosplan-ui [#1506](https://github.com/demos-europe/demosplan-ui/pull/1506) (json:api headers fuare-out icons, lockMessageBy prop, button reorder) — must be released and the ui version bumped in package.json, otherwise the library changes are missing from the build. --> merged

### PR Checklist

- [ ] Create/Update tests
- [ ] Add/Update data-cy attributes
- [x] Run yarn lint
- [ ] Run yarn test
- [x] Link all relevant tickets
- [ ] Move the tickets on the board accordingly
- [x] Update changelog

PR #6432, opened 2026-06-16
F dplan 17747 create statement group

sierrazamudiodemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-17747/Ado-issue-54799-Erstellen-einer-Gruppe

**What is this about:**
- Now it is possible to create Statements Groups using 3.0 api (api platform)
- This PR is about that
- You can test this by actually checking out the FE PR: https://github.com/demos-europe/demosplan-core/pull/6370
- And adjust the vendor folder to apply the changes of the addon pr here: https://github.com/demos-europe/demosplan-addon/pull/178 (ask one of the asignees)
- If you do not want to checkout the FE branch, then you can test using PHP storm using this: 

Test BE:

```
POST {{ProjectUrl}}api/3.0/StatementGroup
Accept: application/vnd.api+json
Content-Type: application/vnd.api+json
Cookie: {{Cookie}}
X-Demosplan-Procedure-Id: {{X-Demosplan-Procedure-Id}}
X-CSRF-Token: {{X-CSRF-Token}}

{
  "data": {
    "type": "StatementGroup",
    "attributes": {
      "headStatementId": "ebf7e18a-adfe-4539-9954-74564c5b0128",
      "groupName": "My group name"
    },
    "relationships": {
      "statements": {
        "data": [
          { "type": "Statement", "id": "ebf7e18a-adfe-4539-9954-74564c5b0128" },
          { "type": "Statement", "id": "d081a70d-feb2-48b7-b1e8-824370558460" },
          { "type": "Statement", "id": "a8b5adc9-4236-4e8a-bfa5-e1d91c5e7eb3" }
        ]
      }
    }
  }
}
```
Addon PR: https://github.com/demos-europe/demosplan-addon/pull/178
FE PR: https://github.com/demos-europe/demosplan-core/pull/6370


breierDemos (commented):
Hey @sierrazamudiodemos, @winderdemos could you please create this branch also for the project repos so it becomes a common branch?

sierrazamudiodemos:
> Hey @sierrazamudiodemos, @winderdemos could you please create this branch also for the project repos so it becomes a common branch?

@breierDemos yep, it is there but it works only if you checkout the FE branch (tbecause the repos are related to the FE branch) sorry

demosplan/DemosPlanCoreBundle/ApiResources/StatementGroupResource.php
```diff
@@ -0,0 +1,80 @@
...
+#[ApiFilter(PropertyFilter::class)]
+class StatementGroupResource
+{
+    #[ApiProperty(readable: false, identifier: true)]
+    public string $id;
+
+    #[ApiProperty(readable: true, writable: true)]
+    public ?string $groupName = null;
+
+    #[ApiProperty(readable: true, writable: true)]
```
bakridemos: Nit: headStatementId is readable but never populated on read

headStatementId is declared readable: true, but fromStatement() doesn't set it, so GET responses always return headStatementId: null. Nothing reads it back today (it's only used as a create input on the POST side), so nothing's broken — it just means the API advertises a field that's always null, which could trip someone up later.

Since it only really makes sense as an input, maybe simplest to make it write-only? I mean (readable: false, writable: true) ?

demosplan/DemosPlanCoreBundle/StateProvider/StatementGroupStateProvider.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $statements = $this->statementRepository->getEntities(
+            $this->clusterConditions->forProcedureById($procedure->getId(), $id),
+            []
+        );
+
+        return [] === $statements ? null : StatementGroupResource::fromStatement($statements[0]);
+    }
+
+    public function isAvailable(): bool
```
bakridemos, line 65: Why not applying the feature_statement_cluster check on cluster creation too ?

I noticed while reading through StatementGroupProcessor::process(): the create path doesn't seem to check feature_statement_cluster. Since the Post operation uses read: false, the provider's isAvailable() check isn't invoked there, so as far as I can tell any authenticated user with a procedure context could create a cluster. In practice this is well covered on the UI side — users without feature_statement_cluster don't see the "group" button at all, so it's effectively not reachable through the interface. The concern is more of a technical/defense-in-depth one: the endpoint itself is still callable directly, so without a server-side check any authenticated user with a procedure context could technically create a cluster even though the UI wouldn't let them.

The ClusterStatementResourceType replaces gates on feature_statement_cluster in isAvailable(), and StatementGroupStateProvider already mirrors that on the read side — it'd be great to apply the same check on create for consistency.

may be extracting that check into a single shared method (maybe on StatementClusterConditions) and reusing it from both the provider and the processor (same way as 'demosplan/DemosPlanCoreBundle/ResourceAccess/DraftStatementAccessChecker.php' ) ? That way the rule lives in one place.

demosplan/DemosPlanCoreBundle/StateProcessor/StatementGroupProcessor.php
```diff
@@ -0,0 +1,71 @@
...
+        $statementIds = array_map(static fn (StatementResource $s): string => $s->id, $data->statements);
+
+        try {
+            $cluster = $this->statementHandler->createStatementCluster(
+                $procedure->getId(),
+                $statementIds,
+                $data->headStatementId,
+                $data->groupName
+            );
+        } catch (InvalidArgumentException $e) {
```
bakridemos: The processor currently catches only InvalidArgumentException:

} catch (InvalidArgumentException $e) {
    throw new BadRequestHttpException($e->getMessage(), $e);
}

But createStatementCluster() can also throw a couple of other expected validation cases that aren't caught here:

- NotAllStatementsGroupableException (it extends Exception, not InvalidArgumentException) — raised by assertAllStatementsGroupable() when a statement isn't in the procedure, isn't assigned to the current user, or otherwise isn't groupable.
- Exception('A Cluster cannot be used as HeadStatement') when the chosen head is itself a cluster.

Since these aren't caught, they'd surface as HTTP 500s rather than a 4xx. The most likely one in practice is the "not assigned" case via select-all: we rely on the backend to validate that path (which it does, nicely 👍), but right now a valid rejection comes back looking like a crash — it adds noise to error monitoring and the specific message gets lost, so the user just sees a generic error.

What do you think about also catching NotAllStatementsGroupableException (and ideally the cluster-as-head case) and mapping them to BadRequestHttpException / UnprocessableEntityHttpException? That way the genuine validation failures return a clean status and a meaningful message. it will make the endpoint a bit more robust.

bakridemos (changes requested):

bakridemos (approved):
All three previous concerns are resolved: headStatementId is correctly write-only, the feature_statement_cluster permission is enforced on the create path via StatementClusterAccessChecker, and the broad catch (Exception $e) now maps all validation failures to clean 4xx responses. LGTM from my side thx

PR #6447, opened 2026-06-17
feat (DPLAN-17697): introduce parameter to set as reply to address for account deletion mails

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17697

### Description
swap out reply to address for account deletion mails - also plug that address into the mail twig templates - adjust wording

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/config/parameters_default.yml
+++ b/config/parameters_default.yml
@@ -63,6 +63,9 @@ parameters:
     account_deletion.first_warning_days: ~
     account_deletion.warning_step_days: 30
     account_deletion.additional_protected_user_ids: []
+    # Support e-mail shown in warning mails and used as replyTo.
+    # Override per project in parameters_default_project.yml.
+    account_deletion.support_email: 'youtrack.helpdesk@demos-deutschland.de'
 
     # Route and parameters for the public index page
     #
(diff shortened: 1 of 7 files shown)


config/parameters_default.yml
```diff
@@ -63,6 +63,9 @@ parameters:
     account_deletion.first_warning_days: ~
     account_deletion.warning_step_days: 30
     account_deletion.additional_protected_user_ids: []
+    # Support e-mail shown in warning mails and used as replyTo.
+    # Override per project in parameters_default_project.yml.
+    account_deletion.support_email: 'youtrack.helpdesk@demos-deutschland.de'
```
kijewskidemos: I think this should be a .env variable

kijewskidemos (changes requested):
Only one change needed in my opinion, otherwise it looks good to me.

config/parameters_default.yml
MoritzMandler: done in https://github.com/demos-europe/demosplan-core/pull/6447/commits/bfab0554578b18701c07653ad14b5016df71fd6a
you need to overwrite .env value `ACCOUNT_DELETION_SUPPORT_EMAIL` with the desired value.
Without a value the system address fallback is used and the contact phrases are striped from email texts

kijewskidemos (approved):

PR #6459, opened 2026-06-18
feat(refs: DPLAN-17918): add tests for custom field components

huellnerdemos:
### Ticket
DPLAN-17918

This PR adds tests for all currently existing, custom fields related components.

### How to review/test

1. Check out this branch
2. Check if the tests work with the current implementation without errors

### PR Checklist

- [x] Run `yarn test`
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## UNRELEASED
 
 ### Added
+- Add tests for the custom field components
 - Make procedure phase definitions editable
 - Segment lock feature: workflow places can now be marked as locked, making segments on them read-only for users without the `feature_administrate_segment_lock` permission. Enforced on JSON:API PATCH, `segment.bulk.edit` RPC, and place updates; transitions are recorded in the segment Versionsverlauf.
 - Permission `feature_segment_lock_by_workflow_place` (exposed, login-required) toggles the segment lock feature per project. Grant it in `projects/<name>/.../Permissions.php` to every role that should see the feature. Requires `demos-europe/demosplan-addon` ^0.68.1.
--- a/client/js/VueConfigLocal.js
+++ b/client/js/VueConfigLocal.js
@@ -48,7 +48,11 @@ const dplan = {
   },
   debug: false,
   notify: {
+    confirm: jest.fn(),
+    error: jest.fn(),
+    info: jest.fn(),
     notify: jest.fn(),
+    warning: jest.fn(),
   },
 }
 
--- a/client/js/VueConfigLocal.js
+++ b/client/js/VueConfigLocal.js
@@ -47,6 +47,9 @@ const dplan = {
     publicCSSClassPrefix: 'dp-',
   },
   debug: false,
+  notify: {
+    notify: jest.fn(),
+  },
 }
 
 const dpApi = jest.fn(() => Promise.resolve())
--- a/tests/frontend/unit/CustomField.spec.js
+++ b/tests/frontend/unit/CustomField.spec.js
@@ -1,0 +1,361 @@
+/**
+ * (c) 2010-present DEMOS plan GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+import { afterEach, beforeEach, describe, expect, it } from '@jest/globals'
+import { enableAutoUnmount } from '@vue/test-utils'
+import CustomField from '@DpJs/components/customFields/CustomField'
+import shallowMountWithGlobalMocks from '@DpJs/VueConfigLocal'
+
+describe('CustomField', () => {
+  let wrapper
+
+  /*
+   * Pass definition directly to bypass the composable's API fetch.
+   * enrichFieldValue runs for real (pure function) — no mocking needed.
+   */
+  const singleSelectDefinition = {
+    id: 'field-1',
+    attributes: {
+      name: 'Status',
+      description: '',
+      isRequired: false,
+      fieldType: 'singleSelect',
+      options: [
+        { id: 'opt-1', label: 'Active' },
+        { id: 'opt-2', label: 'Inactive' },
+      ],
+    },
+  }
+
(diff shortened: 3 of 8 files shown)


client/js/VueConfigLocal.js
```diff
@@ -47,6 +47,9 @@ const dplan = {
     publicCSSClassPrefix: 'dp-',
   },
   debug: false,
+  notify: {
+    notify: jest.fn(),
+  },
 }
```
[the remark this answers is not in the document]
huellnerdemos: added the other methods to the mockup in 6f720458d305fb2a768031ff34836af53a5d3ea4

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ## UNRELEASED
 
 ### Added
+- Add tests for the custom field components
```
hwiem, line 10: I'd say we mostly want to add features to the changelog, while tests are internal and therefore not so relevant. I believe that's how we used the changelog so far?

tests/frontend/unit/CustomField.spec.js
```diff
@@ -0,0 +1,361 @@
...
+    },
+  }
+
+  const singleSelectFieldData = { id: 'field-1', value: 'opt-1' }
+  const multiSelectFieldData = { id: 'field-1', value: ['a'] }
+  const textFieldData = { id: 'field-1', value: 'Some text' }
+
+  enableAutoUnmount(afterEach)
+
+  describe('type dispatch – direct mode', () => {
```
hwiem, line 66: What is "type dispatch - direct mode"? I think if one is not super familiar with the component, it could be helpful to have a more explicit description (same for the following `describe`s)

tests/frontend/unit/CustomField.spec.js
```diff
@@ -0,0 +1,361 @@
...
+      wrapper = shallowMountWithGlobalMocks(CustomField, {
+        props: {
+          definition: singleSelectDefinition,
+          fieldData: singleSelectFieldData,
+          enableToggle: true,
+        },
+      })
+    })
+
+    it('should render the renderer in readonly mode', () => {
```
hwiem, line 146: I'd add this to make it clearer why it should render in readonly mode (it's the initial state for toggle mode)
```suggestion
    it('should render the renderer in readonly mode initially', () => {
```

tests/frontend/unit/CustomField.spec.js
```diff
@@ -0,0 +1,361 @@
...
+          definition: singleSelectDefinition,
+          fieldData: singleSelectFieldData,
+          enableToggle: true,
+        },
+      })
+      wrapper.vm.startEditing()
+    })
+
+    it('should store the incoming value in editingValue', () => {
+      wrapper.vm.handleEditingValueUpdate({ id: 'field-1', value: 'opt-2' })
```
hwiem, line 218: I think `{ id: 'field-1', value: 'opt-2' }` could be stored in a const, because it's used several times throughout the file

tests/frontend/unit/CustomField.spec.js
```diff
@@ -0,0 +1,361 @@
...
+          definition: singleSelectDefinition,
+          fieldData: singleSelectFieldData,
+          enableToggle: true,
+        },
+      })
+      wrapper.vm.startEditing()
+    })
+
+    it('should emit update:value with the editing value', () => {
+      wrapper.vm.saveEdit()
```
hwiem, line 274: I'd suggest to change the value before saving:
` wrapper.vm.handleEditingValueUpdate({ id: 'field-1', value: 'opt-2' })`
That way we ensure that the correct value is emitted if it has actually been changed (saving without changing the value is not a case we need to test)

tests/frontend/unit/CustomField.spec.js
```diff
@@ -0,0 +1,361 @@
...
+
+      expect(wrapper.emitted('update:value')).toBeTruthy()
+      expect(wrapper.emitted('update:value')[0][0]).toBe('opt-1')
+    })
+
+    it('should emit edit:save with the editing value', () => {
+      wrapper.vm.saveEdit()
```
hwiem, line 281: see above

tests/frontend/unit/MultiselectCustomField.spec.js
```diff
@@ -0,0 +1,154 @@
...
+
+  describe('readonly mode', () => {
+    beforeEach(() => {
+      wrapper = shallowMountWithGlobalMocks(MultiselectCustomField, {
+        props: { field: defaultField },
+      })
+    })
+
+    it('should display the field name in dt', () => {
+      expect(wrapper.find('dt').text()).toContain('Tags')
```
hwiem, line 51: Instead of duplicating the string (`'Tags'`) here, I think it's better to use `defaultField.attributes.name` - it reduces the error risk (e.g. because of a typo)

tests/frontend/unit/TextCustomField.spec.js
```diff
@@ -0,0 +1,164 @@
...
+      expect(wrapper.find('dl').exists()).toBe(true)
+    })
+
+    it('should display the field name in dt', () => {
+      expect(wrapper.find('dt').text()).toContain('Comment')
+    })
+
+    it('should display the text value in dd', () => {
+      expect(wrapper.find('dd').text()).toContain('Some text content')
+    })
```
hwiem, line 60: Just as a side note, I think it's not important here that dl, dt, and dd are present (those are implementation details). What's important is that field name and value are displayed. Maybe this is mostly a matter of the test descriptions here, so it's not super important

hwiem (changes requested):
Thank you so much for adding these tests ❤️ I have mostly nitpicks, but I also noticed that when running the tests, there are warnings for TextCustomField.spec.js, I think we should fix them before merging. 
I'd say I see a little danger of focusing too much on implementation details rather than on how the component should behave when the user interacts with it, but mostly in the sense that it's something to keep in mind. At the same time I think these tests can give us a lot of confidence that we don't break anything when working on these components :)

PR #6466, opened 2026-06-18
feat(refs DPLAN-18064): Add isDeleted and deletedDate columns to procedure_phase_definition table.

breierDemos:
### Ticket
[DPLAN-18064](https://demoseurope.youtrack.cloud/issue/DPLAN-18064/NEU-ADO-52178-Loschen-von-custom-Verfahrensschritten)

For now implements soft-delete for ProcedurePhaseDefinition entities.

- Migration: Adds is_deleted and deleted_date columns to procedure_phase_definition
- API: Exposes isDeleted as a readable, filterable, and writable attribute on the ProcedurePhaseDefinition JSON:API resource type. A PATCH with isDeleted: true is blocked if the definition is still referenced by a non-deleted procedure (active or designated phase slot), with an error message returned via messageBag. Deleting the configuration phase (orderInAudience = 0) is also blocked
- Soft-delete logic: setDeleted() on the entity now manages deletedDate internally. The editor handles the reference guard and calls the setter
- Filtering: Deleted definitions are excluded from all active-selection contexts (procedure settings dropdowns, statistics, statement phase dropdowns) via a new excludeDeleted parameter on findByCustomerOrderedByAudience(). The phase name resolver explicitly includes deleted definitions so existing statements continue to display their phase name correctly
- Dropdown integrity: If a statement's current phase has been deleted, it is still included in the available phase options (at its natural orderInAudience position) so the correct value remains selected — both in the server-rendered statement detail view and in the JSON:API availableProcedurePhases attribute
- Report log: Adds IS_DELETED to ProcedurePhaseDefinitionUpdatableField and wires up two distinct protocol messages ("wurde gelöscht" / "wurde wiederhergestellt") in ReportMessageConverter
- Guard on configuration phase: Deleting the configuration phase (orderInAudience = 0) is blocked — it reuses guardConfigurationPhaseNotEditable
- Error handling: The reference guard now throws BadRequestException (in addition to messageBag) so the frontend receives a 400 and can display the error notification
- Report log: Adds IS_DELETED to ProcedurePhaseDefinitionUpdatableField with two distinct protocol messages ("wurde gelöscht" / "wurde wiederhergestellt") in ReportMessageConverter
- Dropdown integrity: If a statement's current phase has been deleted, it is still included in the available phase options at its natural position — both in the server-rendered statement detail view and in the JSON:API availableProcedurePhases attribute
- Soft-delete event: Dispatches ProcedurePhaseDefinitionMarkedAsDeletedEvent (via demosplan-addon interface PR [#179](https://github.com/demos-europe/demosplan-addon/pull/179)) after a successful soft-delete so addons can react — the xbeteiligung addon uses this to delete its XBeteiligungPhaseDefinitionCode mapping atomically
- Unique name constraint: DB-level unique constraint on (name, customer_id, audience) is replaced with an application-level check that allows multiple soft-deleted definitions with the same name while still enforcing uniqueness among active ones

### Linked PRs
Frontend Part: [#6482](https://github.com/demos-europe/demosplan-core/pull/6482)

### PR Checklist

- [x] Create/Update tests


bakridemos (approved):

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedurePhaseDefinitionEditor.php
```diff
@@ -34,6 +42,38 @@ public function guardConfigurationPhaseNotEditable(ProcedurePhaseDefinition $pha
         }
     }
 
+    /**
+     * @throws JsonException
+     */
+    public function setDeleted(ProcedurePhaseDefinition $phaseDefinition, bool $newIsDeleted): void
+    {
+        if ($newIsDeleted && $this->procedurePhaseDefinitionRepository->isReferencedByActiveProcedure($phaseDefinition)) {
```
winderdemos, line 50: We removed the `uniq_name_customer_audience` constraint, so name uniqueness among active definitions is now enforced only in app code (create/rename). 
On restore (deleted → active), should we also check the name isn't already taken by another active definition?

winderdemos (changes requested):

demosplan/DemosPlanCoreBundle/Logic/Procedure/ProcedurePhaseDefinitionEditor.php
breierDemos: Good point! Currently its not possible to set a procedure phase definition active again via the interface, but yes the resourcetype alows it. I will check.

breierDemos: Thank you here is a edit: [d126102](https://github.com/demos-europe/demosplan-core/pull/6466/commits/d126102c5893068c3b7ce8191c7f31447b97333e)

winderdemos (approved):
Looks good 👍

PR #6482, opened 2026-06-23
feat(refs DPLAN-18064, #AB52178): delete procedure phase

rafelddemos:
### Ticket
[DPLAN-18064](https://demoseurope.youtrack.cloud/issue/DPLAN-18064/NEU-ADO-52178-Loschen-von-custom-Verfahrensschritten)

This PR implements deletion of a procedure phase.
Phases that are currently being used in a procedure cannot be deleted. Configuration phase (`orderInAudience = 0`) is also not deletable.

### How to review/test

1. log in as mandanten-admin --> define procedure phases
2. configuration phase shouldn't have a delete button (it is not deletable)
3. delete a phase (with and without a code):
- click on delete-button should trigger a confirm dialog
- successfull deletion --> success notification
- if a phase cannot be deleted ('cos actively used) --> error notification
- create a new phase with the name of the phase you deleted and for the same audience (names are unique per audience) --> it should be possible to use the name again

### Linked PRs

BE: https://github.com/demos-europe/demosplan-core/pull/6466

### PR Checklist

- [x] Run `yarn lint`
- [x] Update changelog


client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
```diff
@@ -824,6 +841,34 @@ export default {
         })
     }
 
+    // *** DELETE PHASE LOGIC ***
+    const confirmDeleteDialog = ref(null)
+
+    const deletePhase = async (id) => {
+      const isConfirmed = await confirmDeleteDialog.value.open()
```
riechedemos, line 848: I think a null check would make sense here:
```suggestion
      const isConfirmed = await confirmDeleteDialog.value?.open?.()
```

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
```diff
@@ -824,6 +841,34 @@ export default {
         })
     }
 
+    // *** DELETE PHASE LOGIC ***
+    const confirmDeleteDialog = ref(null)
+
+    const deletePhase = async (id) => {
```
riechedemos, line 847: The async keyword is unnecessary here. Either:
- Keep promise chain, remove async, OR
- Refactor to try/catch with await

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
```diff
@@ -824,6 +841,34 @@ export default {
         })
     }
 
+    // *** DELETE PHASE LOGIC ***
```
riechedemos, line 844: Nitpick: // Handle phase deletion workflow

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
```diff
@@ -824,6 +841,34 @@ export default {
...
+        Routing.generate('api_resource_update', { resourceType: 'ProcedurePhaseDefinition', resourceId: id }),
+        {},
+        { data: { type: 'ProcedurePhaseDefinition', id, attributes: { isDeleted: true } } },
+      )
+        .then(() => {
+          phaseDefinitions.value = phaseDefinitions.value.filter(phase => phase.id !== id)
+          dplan.notify.confirm(Translator.trans('procedure.phase.delete.success'))
+        })
+        .catch(err => {
+          if (!err.data?.meta) {
```
riechedemos: I think this is silent error swallowing. When backend returns an error with err.data.meta.messages, nothing is shown.

Other components in the codebase use handleResponseMessages() from demosplan-ui:

dpApi.patch(...)
  .then(() => {
    phaseDefinitions.value = phaseDefinitions.value.filter(phase => phase.id !== id)
    dplan.notify.confirm(Translator.trans('procedure.phase.delete.success'))
  })
  .catch(err => {
    handleResponseMessag  // Centralizes errorhandling
    dplan.notify.error(Teneric'))
  })

See: DpBoilerPlateModal.vue or StatementExportModal.vue

riechedemos (changes requested):
Thank you I tested it and it works well 🎉  Just some small issues.

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
rafelddemos: The comment uses capital letters and the asterisk to be straight away visible - it's specifically for grouping semantic sections within the composition API file that tends to be organised differently than options API. Since this is a new pattern in the codebase (meaning not many files yet using it), we still need to agree upon how exactly we want to write these comments (discussed it with Heidrun before and we agreed to try out things and then we agree within the team on a pattern). I guess it makes sense to wait what we agree upon and then adjust all the comp API files that need it.

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
rafelddemos: `async` is not unnecessary here, since the dialog confirmation is awaited.
I took the `try/catch` option and made the API call consistent with it. 63b1d2dd391815ddc207733ae045b7f1d4d135cd

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
rafelddemos: It is not needed here (the pattern in the codebase is to call `open()` directly). The reason is that `DpConfirmDialog` defines `open()` as a normal component method (that returns a Promise), and the confirmation dialog is rendered unconditionally (no `v-if`), so: 
- the ref is always populated once the component is mounted
- `deletePhase` can only be triggered by clicking a delete button inside the table, which only exists after mount, by which point the dialog is mounted too

client/js/components/procedurePhasesDefinition/ProcedurePhasesDefinition.vue
rafelddemos: I was confused with the usage of `handleResponseMessages()`at first too.. I can't actually see any usages of it or confirm dialog in  `DpBoilerPlateModal.vue` or `StatementExportModal.vue`&#129300;, but other files like `DpBulkEditFragment.vue` and `DpSelectedItemsStatements.vue` do use it -  that seems to be the old stale pattern, which doesn't apply to `dpApi`.

`dpApi`'s `checkResponse()` calls `handleResponseMessages`(response.data.meta) on every response (even 400), before it rejects, so `meta.messages` is shown automatically - re-importing/calling it would cause double-toast-display.

Good catch on the swallow risk, though, thanks! I added explicit check for `meta.messages` instead of just `meta`, so that an error with a `meta` without `messages` still gets the generic fallback. added in 0fa82111e2699c332d273e89de69495a3a38c7b6

riechedemos (approved):
Looks good now :)

PR #6492, opened 2026-06-26
feat: enable queue-based fos:elastica:populate via enqueue-elastica-bundle

mussbach:
Activates the multi-worker Elasticsearch populate that was already scaffolded (`dplan:elasticsearch:populate`, `elasticsearch_populate_workers`) but inert because the queue backend was never wired up.

## What changed
- Require an Elasticsearch 8–compatible `enqueue/elastica-bundle` and register `EnqueueBundle` + `EnqueueElasticaBundle`
- Restore the `enqueue` / `enqueue_elastica` config (filesystem transport under the cache dir; `null` transport in test)
- Fix `dplan:elasticsearch:populate`: the availability gate checked a class that never exists (so the command always failed), and the consumer workers were spawned twice — leaking half the processes

## Result
~4.8× faster populate of a ~295k-document index with 8 workers vs the sequential `in_place` persister.

## Dependency note (needs a decision)
The bundle is pulled from a fork pinned to a `dev-master` commit, because upstream does not yet support ES8. Upstream PRs are open (https://github.com/php-enqueue/enqueue-elastica-bundle/pull/41 for ES8, https://github.com/php-enqueue/enqueue-elastica-bundle/pull/42 for the idle-timeout fix). Once merged/tagged we should repoint to the upstream release. A team-owned fork may be preferable to a personal one for a production dependency.


demosplan/DemosPlanCoreBundle/Command/ElasticsearchPopulateCommand.php
```diff
@@ -48,11 +49,11 @@ public function configure(): void
...
         $io = new SymfonyStyle($input, $output);
-        // output a message that this commmand does not work without enqueue bundle, which does not yet support elasticsearch 8
-        // and advise to use the fos:elastica:populate command instead
-        if (!class_exists('Enqueue\Symfony\ConsumptionExtension')) {
+        // The multi-worker populate relies on the enqueue elastica bundle. If it is not installed,
+        // advise to use the plain fos:elastica:populate command instead.
+        if (!class_exists(QueuePagerPersister::class)) {
             $io->writeln(
-                '<error>This command does not work without the enqueue bundle. Please use the fos:elastica:populate and fos:elastica:populate command instead.</error>'
+                '<error>This command does not work without the enqueue elastica bundle. Please use the fos:elastica:populate command instead.</error>'
```
graupnerdemos: Very minor nitpick, you could just use `$io->error()` and drop the `<error>` tag. (https://symfony.com/doc/current/console/style.html#result-methods)

composer.json
```diff
@@ -5,6 +5,12 @@
   "version": "v4.47.0",
   "license": "EUPL-1.2",
   "minimum-stability": "stable",
+  "repositories": [
+    {
+      "type": "vcs",
+      "url": "https://github.com/mussbach/enqueue-elastica-bundle"
```
graupnerdemos: as posed in the pr description, I'd plus one the idea of moving the fork to the demos-europe org. That enables more of us to integrate upstream fixes if necessary.

graupnerdemos (approved):
Works and is - as advertised - faster. My remarks are non-blocking.

demosplan/DemosPlanCoreBundle/Command/ElasticsearchPopulateCommand.php
mussbach: indeed,, thx c0ee552d

composer.json
mussbach: indeed [613eaf3](https://github.com/demos-europe/demosplan-core/pull/6492/commits/613eaf350094b45bd5e8565fa09a105ce2390592)

PR #6493, opened 2026-06-26
fix add on-the-fly migration of old-format draftsListJson at read time

bakridemos:
Description: 

The segment editor frontend exclusively uses the new format where `textualReference` contains `<segment-mark data-segment-id="...">` wrappers around each segment's text. Statements segmented before this format was introduced have their data stored in the old format: segments carry `charStart`/`charEnd` Prosemirror position offsets and `textualReference` is plain HTML without any `<segment-mark>` tags. Opening such a statement in the editor crashes because the frontend can no longer work with the old format at all.

This PR adds `DraftsListJsonMigrator`, a service that converts old-format `draftsListJson` to the new format transparently at read time in `StatementResourceType`. The migrator finds each segment's `text` verbatim in `textualReference` using string
matching and wraps it with the corresponding `<segment-mark>` tag, then strips the position fields (`charStart`, `charEnd`, `charStartInit`, `charEndInit`, `hasProsemirrorIndex`) from the segment objects.

The migration happens only once per statement: on first open, the editor detects the new format and persists it back to the database automatically. All subsequent reads return the already-migrated data and the migrator is bypassed entirely.

### How to review/test

1. Find a statement that was segmented before the new format was introduced (its `drafts_list_json` contains segments with `charStart`/`charEnd` fields and no `<segment-mark>` tags in `textualReference`)
2. Open it in the segment editor — it should load without console errors and all segment cards should be visible
3. Save once — verify the database now stores the new format with `<segment-mark>` tags and no position fields
4. Reopen — confirm the editor still works and the migrator is no longer triggered

- [X] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * (c) 2010-present DEMOS plan GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+namespace demosplan\DemosPlanCoreBundle\Logic\Statement;
+
+/**
+ * Converts old-format draftsListJson (segments with charStart/charEnd positions, no <segment-mark>
+ * tags in textualReference) to the new format expected by the segment editor.
+ *
+ * Old format: segments carry charStart/charEnd Prosemirror positions; textualReference is plain HTML.
+ * New format: textualReference contains <segment-mark data-segment-id="..."> wrappers; no positions.
+ *
+ * Runs on-the-fly at read time — does not write back to the database.
+ */
+class DraftsListJsonMigrator
+{
+    public function needsMigration(array $data): bool
+    {
+        $segments = $data['data']['attributes']['segments'] ?? [];
+        $textualReference = $data['data']['attributes']['textualReference'] ?? '';
+
+        return !empty($segments)
+            && isset($segments[0]['charStart'])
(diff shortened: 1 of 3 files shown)


demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
```diff
@@ -0,0 +1,78 @@
...
+        // charStart may be a Prosemirror position (not an HTML offset), but relative order is preserved.
+        usort($segments, static fn (array $a, array $b): int => ($a['charStart'] ?? 0) - ($b['charStart'] ?? 0));
+
+        foreach ($segments as $segment) {
+            $text = $segment['text'] ?? '';
+            if ('' === $text) {
+                continue;
+            }
+
+            $pos = strpos($textualReference, $text);
```
mussbach, line 51: `strpos` always searches from offset 0, so duplicate or substring segment text gets mis-wrapped. Once the first segment is wrapped, its raw `text` still exists literally inside the new `<segment-mark>…</segment-mark>`. A later segment whose `text` is byte-identical (repeated boilerplate, two `<p>&nbsp;</p>` paragraphs) or a substring of an earlier segment's text will resolve back to that already-wrapped region and wrap it again — producing nested `<segment-mark>` tags with a wrong segment-id mapping, corrupting the `textualReference` this migration is meant to repair.

Segments are already sorted into document order, so track a running offset and search forward: `strpos($textualReference, $text, $offset)`, then set `$offset = $pos + strlen($marked)` after each replacement. Worth adding a test with identical/overlapping segment text — the current suite doesn't cover it.

bakridemos:
This changes are remade in https://github.com/demos-europe/demosplan-core/pull/6494 with the needed Frontend adjustments. this pr will is not needed anymore and will be closed

PR #6494, opened 2026-06-26
B dplan 18094 old exchnage format backend transformation

sakutademos:
### Ticket
[DPLAN-18094](https://demoseurope.youtrack.cloud/issue/DPLAN-18094) Refacturing Austauschformat

Description: This PR fixes an issue that occurred when a statement had previously been opened in the Split Statement view with an old exchange format, but the split process was not completed. In this case, the intermediate state was stored in drafts_info_json using the old exchange format.

1. After migrating to the new exchange format, reopening such a statement caused the frontend to break because it no longer supports the legacy format. To maintain backward compatibility, the backend now checks whether the stored draft still uses the old format and automatically transforms it into the new segment-mark format before returning it to the frontend --> **this compatibility layer is temporary (can be removed once all statements have been migrated)?**

2. Another refactoring that was done with this PR: it also removes the obsolete frontend conversion from HTML character positions to ProseMirror indices (`hasProsemirrorIndex`). With the new exchange format, ProseMirror processes segment-marks natively and determines the positions itself. Since the backend now converts legacy drafts to the new format, this frontend transformation is no longer needed.

### How to review/test
**First Bug:**
1. Find a statement that was previously opened in the Split Statement view using the old exchange format.
2. Click Aufteilen.
3. Verify that all segment boxes are displayed correctly and can be edited.

**Second Bug:**
1. Open an existing statement (or create a new one) and click Aufteilen.
2. Verify that the frontend does not automatically update the draft in the database before the user makes any changes or saves.
3. All boxes and segments should be displayed correctly.

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
```diff
@@ -0,0 +1,80 @@
...
+        // charStart may be a Prosemirror position (not an HTML offset), but relative order is preserved.
+        usort($segments, static fn (array $a, array $b): int => ($a['charStart'] ?? 0) - ($b['charStart'] ?? 0));
+
+        $offset = 0;
+        foreach ($segments as $segment) {
+            $text = $segment['text'] ?? '';
+            if ('' === $text) {
+                continue;
+            }
+
```
mussbach, line 51: `strpos` always returns the first occurrence, and an inserted `<segment-mark>` keeps its inner text searchable. Two segments with identical `text` (e.g. a repeated paragraph) → segment 2 re-finds and re-wraps segment 1's already-marked text, producing nested marks while the real second occurrence stays unmarked. Same failure when one segment's text is a substring of an earlier one. The `usort` by `charStart` doesn't help since the lookup ignores position entirely. Inserting at a tracked running offset in `charStart` order would be robust.

demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
```diff
@@ -0,0 +1,80 @@
...
+ */
+class DraftsListJsonMigrator
+{
+    public function needsMigration(array $data): bool
+    {
+        $segments = $data['data']['attributes']['segments'] ?? [];
+        $textualReference = $data['data']['attributes']['textualReference'] ?? '';
+
+        return !empty($segments)
+            && isset($segments[0]['charStart'])
```
mussbach: `isset()` is false for an explicit `charStart: null`, so a legacy draft that serialized a null position would skip migration and reach the frontend in the old shape. `array_key_exists('charStart', $segments[0])` would cover that case.

client/js/store/statement/SplitStatementStore.js
```diff
@@ -566,7 +533,7 @@ const SplitStatementStore = {
     isBusy: (state) => state.isBusy,
     procedureId: (state) => state.procedureId,
     segments: (state) => state.segments,
-    sortedSegments: (state) => state.segments.concat().sort((a, b) => a.charStart - b.charStart),
+    sortedSegments: (state) => state.segments,
```
mussbach: Getter no longer sorts but keeps the `sortedSegments` name, so it reads as if it still does. Consider renaming (or inlining at the `CardPane.vue` call sites) to avoid the impression that ordering is guaranteed here.

demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
bakridemos: @mussbach i did already fixed this in https://github.com/demos-europe/demosplan-core/pull/6494/changes/7f7cecee2be7c458883d6647986d09e3117c4cee

demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
bakridemos: thx done in https://github.com/demos-europe/demosplan-core/pull/6494/changes/7a6d4689e2b67b397ffa3ee0ed065fc80d17943a

client/js/store/statement/SplitStatementStore.js
sakutademos: totally right, done here https://github.com/demos-europe/demosplan-core/pull/6494/commits/320d133021291182e6b28895bc392bb0bf445c98

mussbach (approved):
lgtm, code review only, thx

PR #6504, opened 2026-07-01
fix(refs DPLAN-18125): correct manual sort event shape in element admin edit

breierDemos:
## Summary
- Fixes a crash when manually reordering documents in the element admin edit list: `saveManualSort` was reading `oldIndex`/`newIndex` from a nested `moved` object that the current drag component no longer emits.
- Updated to destructure `oldIndex`/`newIndex` directly from the event, matching how other consumers of the same `changed-order` event already handle it.

## Ticket
[DPLAN-18125](https://demoseurope.youtrack.cloud/issue/DPLAN-18125)

## Test plan
- [x] Manually reorder documents via drag-and-drop in the element admin edit list and confirm the sort saves without a console error

### DIFF
--- a/client/js/components/document/ElementAdminEdit.vue
+++ b/client/js/components/document/ElementAdminEdit.vue
@@ -154,10 +154,10 @@ export default {
       this.closeOnSelect = true
     },
 
-    saveManualSort (val) {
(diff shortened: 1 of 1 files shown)


rafelddemos (approved):
lgtm, thx for fixing it :)

PR #6506, opened 2026-07-01
fix(refs DPLAN-18125): Correct manual sort event shape in element admin edit

rafelddemos:
### Ticket
[DPLAN-18125](https://demoseurope.youtrack.cloud/issue/DPLAN-18125)

- Fixes a crash when manually reordering documents in the element admin edit list: `saveManualSort` was reading `oldIndex`/`newIndex` from a nested `moved` object that the current drag component no longer emits.
- Updated to destructure `oldIndex`/`newIndex` directly from the event, matching how other consumers of the same `changed-order` event already handle it.

### How to review/test

Manually reorder documents via drag-and-drop in the element admin edit list (FP-A) and confirm the sort saves without a console error

### Linked PRs

the fix was originally introduced in this PR: https://github.com/demos-europe/demosplan-core/pull/6504

### PR Checklist

- [x] Run `yarn test`
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ## UNRELEASED
 
+### Fixed 
+- Manual drag-and-drop reordering of documents in the plan document category admin now saves correctly
+
 ## v4.45.0 (2026-06-19)
 ### Added
 - Delete Institution Text CustomFields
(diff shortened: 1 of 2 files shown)


mussbach (approved):
lgtm, thx, code review only

PR #6507, opened 2026-07-01
fix(DPLAN-18094): fix position-matching bugs found in review of draft…

bakridemos:
### Ticket
DPLAN-18094

Some legacy `draftsListJson` statements never stored a `text` snapshot per segment — only `charStart`/`charEnd`/`charStartInit`/`charEndInit` positions. `DraftsListJsonMigrator` only knew how to wrap segments by searching for their verbatim `text` in `textualReference`, so for these records it silently did nothing: `needsMigration()` said "yes", but `migrate()` skipped every segment (no `text` to search for) and still stripped the position fields — leaving
`textualReference` with zero `<segment-mark>` tags and no way to recover the segment boundaries afterwards. The segment editor then had nothing to render.

**Fix:** added `DraftsListJsonPositionMigrator`, a dedicated migrator for this variant. It rebuilds `textualReference` directly from position offsets instead of a text search. Investigation showed `charStartInit`/`charEndInit` are the reliable raw string offsets — `charStart`/`charEnd` had been overwritten by since-removed frontend code (`transformHTMLPositionsToProsemirrorPositions`,
deleted in an earlier DPLAN-18094 commit) with Prosemirror document positions, which don't map onto `textualReference` as string offsets. So the new migrator prefers Init fields and falls back to the legacy pair only when Init is unavailable. Segments with unusable/overlapping positions are left unwrapped rather than guessed at — no character of the original text is ever dropped.

Both migrators are wired into `StatementResourceType`'s `draftsListJson` read closure and are mutually exclusive by design (one requires a text snapshot, the other requires its absence).

A follow-up review pass caught three real bugs in the first version, all fixed:
- Segments were sorted by `charStart` but placed by `charStartInit`/`charEndInit` — when the two  disagreed on order, a valid segment could be silently skipped as "overlapping". Now both sorting  and placement use the same resolved value.
- `charStartInit`/`charEndInit` were resolved independently per field, so a segment with only one  of the two could combine an Init value with a legacy value into a wrong boundary. Init is now all-or-nothing.
- Format detection (`needsMigration`) only inspected the first segment, so a record where segment 0 lacked `text` but a later segment had it would be misrouted to the position migrator, losing a reliable text snapshot. Now checks all segments.

The shared strip/detection logic between the two migrators was extracted into `DraftsListJsonSegmentFields`, injected into both rather than duplicated.

### How to review/test
- Run the two migrator test suites:
  vendor/bin/phpunit tests/backend/core/Logic/Statement/DraftsListJsonMigratorTest.php tests/backend/core/Logic/Statement/DraftsListJsonPositionMigratorTest.php
- Manually: open a legacy statement whose `drafts_info_json` has segments with only position fields (no `text`) in the segment editor and confirm all segments render with correct

- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/DraftsListJsonMigrator.php
@@ -16,13 +16,22 @@ namespace demosplan\DemosPlanCoreBundle\Logic\Statement;
  * Converts old-format draftsListJson (segments with charStart/charEnd positions, no <segment-mark>
  * tags in textualReference) to the new format expected by the segment editor.
  *
- * Old format: segments carry charStart/charEnd Prosemirror positions; textualReference is plain HTML.
- * New format: textualReference contains <segment-mark data-segment-id="..."> wrappers; no positions.
+ * Old format: segments carry charStart/charEnd Prosemirror positions plus a verbatim `text`
+ * snapshot; textualReference is plain HTML. New format: textualReference contains
+ * <segment-mark data-segment-id="..."> wrappers; no positions.
+ *
+ * Handles only records whose segments carry a `text` snapshot. Records without `text` (only
+ * positions) are handled by {@see DraftsListJsonPositionMigrator}.
  *
  * Runs on-the-fly at read time — does not write back to the database.
  */
 class DraftsListJsonMigrator
 {
+    public function __construct(
+        private readonly DraftsListJsonSegmentFields $segmentFields,
+    ) {
+    }
+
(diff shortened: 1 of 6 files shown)


mussbach (changes requested):
@bakridemos did you forget to include `DraftsListJsonSegmentFields` in your commit?

bakridemos:
> @bakridemos did you forget to include `DraftsListJsonSegmentFields` in your commit?

jup my bad, added in https://github.com/demos-europe/demosplan-core/pull/6507/changes/8e08e923025b2bcfa74e3167b1dcd0e46d3bb8b9

mussbach (approved):
lgtm, thx a lot. Code review only

PR #6509, opened 2026-07-02
fix(cors): expose tus upload headers so resumable uploads succeed

mussbach:
### Ticket
DPLAN-17129

The global nelmio CORS policy set `Access-Control-Expose-Headers` to `['Link']` on **every** path (`^/`), including the tus resumable-upload endpoint. This overwrote the CORS headers that tus-php (+ the `TusCors` middleware) sets, hiding `Upload-Offset` from the browser. As a result `tus-js-client` (Uppy) failed cross-origin uploads with `tus: invalid or missing offset value` — the server sent the offset, but JS could not read it.

This adds a dedicated `^/_tus` rule, matched **before** the catch-all `^/` (nelmio is first-match), that:
- exposes the tus headers (`Upload-Offset`, `Upload-Length`, `Location`, `Tus-Resumable`, …) plus the `x-demosplan-file-*` headers, and
- sets a correct preflight `allow_methods`/`allow_headers` (incl. `HEAD`/`PATCH` and the `Tus-*`/`Upload-*` request headers).

Explicit header lists are used rather than `*` because the uploader sends credentials, which forbids wildcard `allow-headers`/`allow-origin` under the CORS spec.

### How to review/test
1. Upload a file via the frontend uploader (e.g. Planunterlagen) on an environment where the tus endpoint is served cross-origin.
2. Without the fix the upload aborts with `invalid or missing offset value`; with it the upload completes.
3. `curl -I` a tus upload URL and confirm `access-control-expose-headers` now includes `upload-offset` (previously only `link`).

### PR Checklist

### DIFF
--- a/config/packages/nelmio_cors.yaml
+++ b/config/packages/nelmio_cors.yaml
@@ -7,4 +7,16 @@ nelmio_cors:
         expose_headers: ['Link']
         max_age: 3600
     paths:
+        # The tus resumable-upload endpoint sets its own CORS headers (see TusCors
+        # middleware + tus-php defaults). Without this dedicated rule the global '^/'
+        # policy below overwrites Access-Control-Expose-Headers with just ['Link'],
+        # hiding Upload-Offset from the browser so uploads fail with
+        # "tus: invalid or missing offset value". Wildcards are avoided because the
+        # uploader sends credentials, which forbids '*' for allow-headers/origin.
+        '^/_tus':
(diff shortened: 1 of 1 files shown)


graupnerdemos (approved):
Would it make sense to incorporate a "if nelmio then add rule" block to the tus bundle loading? This solves our problem here though, therefore lgtm.

PR #6510, opened 2026-07-02
fix HDDP-25: sanitize zip entry names for Windows extraction

bakridemos:
### Ticket
HDDP-25

Procedure exports produced ZIP archives that Windows refused to open ("Der zip-komprimierte Ordner ist ungültig" / PowerShell `Expand-Archive` reporting missing paths). Not every export was affected, and 7-zip opened them fine.

**Cause:** ZIP entry paths are built from user-entered planning-document / element titles. When a title carries a leading/trailing space or a trailing dot, that leaks verbatim into a path segment. Windows silently strips trailing spaces and dots from file/directory names on extraction, so the path it actually creates no longer matches the archive's entry name — and
its built-in extractor then declares the whole archive invalid. 7-zip does not normalize names, which is why it works. The existing ASCII-folding (`9a7d47aec`) fixed non-ASCII characters but did not trim whitespace/dots, so this survived it.

**Fix:** A new public `ZipExportService::sanitizeZipPath()` folds to ASCII and trims each path segment of surrounding whitespace and trailing dots.
Interior dots (e.g. `Kapitel 4.5.1`, the `.pdf` extension) are preserved. All ZIP-entry-name construction is routed through it at every site: `ZipExportService` (both choke points), `ZipResponseGenerator`, and the plandocument bulk download in `DemosPlanDocumentController`.

Only newly created archives are corrected; existing broken exports must be re-created 

### How to review/test
1. Create a procedure and give a planning-document category/element a title with a trailing space (e.g. `Amtliche  Beteiligungsverfahren Beispiele `).
2. Add a document/file under it and run the procedure export.
3. On Windows, open the resulting ZIP via Explorer or  `Expand-Archive` — it should extract without the "ungültig" error.
4. Automated coverage:  
ZipExportServiceTest.php t
ExportServiceTest.php`

- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/ZipExportService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/ZipExportService.php
@@ -248,7 +248,7 @@ class ZipExportService
         $zipPath = (new UnicodeString($zipPath))->ascii()->toString();
 
         $segments = array_map(
-            static fn (string $segment): string => rtrim(trim($segment), " ."),
+            static fn (string $segment): string => rtrim(trim($segment), ' .'),
             explode('/', $zipPath)
         );
 
--- a/demosplan/DemosPlanCoreBundle/Logic/ZipExportService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/ZipExportService.php
@@ -97,8 +97,12 @@ class ZipExportService
         ZipStream $zip,
         string $fileNamePrefix,
     ): void {
-        $path = (new UnicodeString($folderPath.$fileNamePrefix.$fileInfo->getFileName()))->ascii()->toString();
-        $pathHash = md5((string) $path);
+        // Hash the sanitized path so the dedup guard matches the entry name that
+        // is actually written. Otherwise two titles differing only by a trailing
+        // space (or dot) would hash differently, both pass this guard, then
+        // sanitize to the same entry name and collide as a duplicate ZIP entry.
+        $path = $this->sanitizeZipPath($folderPath.$fileNamePrefix.$fileInfo->getFileName());
+        $pathHash = md5($path);
         if (in_array($pathHash, $this->filesAdded, true)) {
             $this->logger->warning('File already present in Zip', ['path' => $path]);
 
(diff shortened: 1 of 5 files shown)


demosplan/DemosPlanCoreBundle/Logic/ZipExportService.php
```diff
@@ -226,4 +226,32 @@ private function checkIfSavable(object $writer): void
...
+    public function sanitizeZipPath(string $zipPath): string
+    {
+        $zipPath = (new UnicodeString($zipPath))->ascii()->toString();
+
+        $segments = array_map(
+            static fn (string $segment): string => rtrim(trim($segment), ' .'),
+            explode('/', $zipPath)
+        );
+
+        return implode('/', $segments);
```
mussbach, line 268: A segment consisting only of spaces/dots (e.g. a title `"   "` or `"..."`) trims to `""`, so `implode("/")` yields a `//` entry like `proc//file.pdf`. Since these paths come from arbitrary user-entered titles, an empty directory component can re-trigger the invalid-path issue on Windows that this fix targets. Consider dropping empty segments (`array_filter` after the map).

demosplan/DemosPlanCoreBundle/Logic/ZipExportService.php
```diff
@@ -73,7 +73,7 @@ public function buildZipStreamResponse(string $name, callable $fillZipFunction):
      */
     public function addFileToZipStream(string $filePath, string $zipPath, ZipStream $zip): void
     {
-        $zipPath = (new UnicodeString($zipPath))->ascii()->toString();
+        $zipPath = $this->sanitizeZipPath($zipPath);
```
mussbach, line 76: Related, in `addFileToZip()` above: its dedup key `md5($path)` is computed on the untrimmed ASCII path, while the entry written through here is trimmed. Two titles differing only by a trailing space hash differently, both pass the dedup guard, then sanitize to the same entry name → a duplicate ZIP entry. Computing the hash on the sanitized path would avoid it.

demosplan/DemosPlanCoreBundle/Logic/ZipExportService.php
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/6510/changes/984c34d1dea91929cb063946eb90c95ef9c3019d thx

demosplan/DemosPlanCoreBundle/Logic/ZipExportService.php
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/6510/changes/984c34d1dea91929cb063946eb90c95ef9c3019d thx

bakridemos: i had to adjust it once again in https://github.com/demos-europe/demosplan-core/pull/6510/changes/8c8ac5ab68b58b6d2ca73ba6cf62c7772541a79c to resolve Sonarcloud issues

mussbach (approved):
lgtm, thx, code review only

PR #6513, opened 2026-07-03
fix(refs DPLAN-18064): use PersistResourceException to suppress generic error when specific error already added

breierDemos:
### Ticket
DPLAN-18064

When a specific error message is added to the message bag before throwing (e.g. duplicate name on create, or referencing a phase still in use), the generic fallback error was also shown to the user.

`handleApiError` already has a special case for `PersistResourceExceptionInterface` that skips adding the generic message — switching `BadRequestException` to `PersistResourceException` in these two places ensures only the specific, user-facing error is shown.

### How to review/test
- Try creating a `ProcedurePhaseDefinition` with a duplicate name → only the specific duplicate-name error should appear
- Try soft-deleting a `ProcedurePhaseDefinition` that is still referenced by an active procedure → only the specific referenced error should appear


MoritzMandler (approved):
lgtm - nice solution :)

PR #6517, opened 2026-07-06
F ds 563 fix segment export oom tag filter

bakridemos:
Ticket

DS-563

When exporting the segment list ("Einwendungsmanagement" / Abschnitte) filtered by a tag (Schlagwort) to Excel, Word, or ZIP, users received the generic "Ein Fehler ist aufgetreten" page instead of a file. This PR fixes the crash and makes the export substantially faster.

1. Fix: OutOfMemoryError on large procedures

The export crashed with OutOfMemoryError (2 GB exhausted, thrown deep in the DBAL PDO layer). Root cause: StatementExportTagFilter::initializeRelationships() pre-warmed the relationships with a single query that fetch-joined two independent to-many collections (segmentsOfStatement and tags) plus topic. Joining two to-many relations in one query produces a cartesian product — the row count multiplies (segments × tags-per-segment) and every row drags along the full, wide _statement columns of both the parent statement and the segment, duplicated once per tag. On a large procedure the buffered result set exceeded the memory limit.

Fix: split the one fetch-join into two single-collection queries sharing the same Doctrine identity map (statements ⋈ segments, then segments ⋈ tags ⋈ topic). Row counts become additive instead of multiplicative and the wide parent row is no longer repeated per tag, while the in-PHP filter still runs without N+1 queries.

2. Improvement: only load the statements that match the filter

Even after the crash was fixed, the export was slow because the tag filter was applied in PHP after loading — so it always loaded every statement of the procedure (and all their segments) and then discarded the non-matching ones. Filtering to a single Schlagwort did not reduce the query at all.

We now push the tag filter into the query using the already-filterable segments.tags.* paths, so only statements that actually carry a matching tag are ever loaded. Permission scoping and Elasticsearch search are preserved: getObjectsByQueryParams() gained an optional conditions argument that is merged into the parsed request filters on both the Doctrine and Elasticsearch paths. The in-PHP filter is kept, but now only trims each (already-narrowed) statement to its matching segments and collects the tag titles for the export header.

Measured on the reporting procedure filtered by "Landschaftsbild": statements loaded dropped from 969 to 140 (~7×), with a corresponding drop in hydration and document-generation work.

---
How to review/test

1. Open a procedure with many segmented statements ("Abschnitte"), filter by a Schlagwort, and export as Excel, Word, and ZIP — all should complete instead of showing the error page, and noticeably faster than before.
2. Verify a full export (no tag filter) is unchanged and still contains everything.
3. Verify export with a search term typed in the list still works (Elasticsearch path).
4. Open an exported file and confirm no statement/segment appears twice (to-many filter de-duplication).

- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Statement/Exporter/StatementExportTagFilter.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Statement/Exporter/StatementExportTagFilter.php
@@ -18,8 +18,11 @@ use DemosEurope\DemosplanAddon\Contracts\Entities\TagInterface;
 use DemosEurope\DemosplanAddon\Contracts\Entities\TagTopicInterface;
 use demosplan\DemosPlanCoreBundle\Entity\Statement\Segment;
 use demosplan\DemosPlanCoreBundle\Entity\Statement\Statement;
+use demosplan\DemosPlanCoreBundle\ResourceTypes\StatementResourceType;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ORM\EntityManagerInterface;
+use EDT\DqlQuerying\ConditionFactories\DqlConditionFactory;
+use EDT\DqlQuerying\Contracts\ClauseFunctionInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
 use function in_array;
@@ -29,6 +32,7 @@ class StatementExportTagFilter
     public function __construct(
         private readonly TranslatorInterface $translator,
         private readonly EntityManagerInterface $entityManager,
+        private readonly DqlConditionFactory $conditionFactory,
     ) {
     }
     private const TAG_IDS_FILTER_KEY = 'tagIds';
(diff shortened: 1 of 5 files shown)


demosplan/DemosPlanCoreBundle/Logic/Statement/Exporter/StatementExportTagFilter.php
```diff
@@ -85,6 +90,49 @@ public function filterStatementsByTags(array $statements, array $tagsFilter): ar
...
+        }
+        if ([] !== $tagTopicTitles) {
+            $criteria[] = $this->conditionFactory->propertyHasAnyOfValues($tagTopicTitles, $statementResourceType->segments->tags->topic->title);
+        }
+
+        if ([] === $criteria) {
+            return [];
+        }
+
+        return [1 === count($criteria) ? $criteria[0] : $this->conditionFactory->anyConditionApplies(...$criteria)];
```
mussbach, line 133: The pushed-down `segments.tags.*` conditions filter on a **to-many** path, and EDT's `DoctrineOrmEntityProvider` builds the query without `DISTINCT`/`GROUP BY` (`SELECT s … LEFT JOIN seg … LEFT JOIN t … WHERE t.id IN (…)`, see `edt-dql/src/Utilities/QueryBuilderPreparer.php`). ORM hydration then returns the same `Statement` object once per matching (segment, tag) row, and neither `getObjectsByQueryParams` nor `applyTagFilter` (`new ArrayCollection($statements)->filter()`) deduplicates. A statement with two segments carrying the filtered tag ends up in the export twice (Word/Xlsx/ZIP) — the exact case called out in review point #4. Recommend deduplicating by statement id after loading.

tests/backend/core/Statement/Functional/SegmentsExportControllerTagFilterTest.php
```diff
@@ -55,7 +56,9 @@ protected function setUp(): void
         // Instantiate the filter service
         /** @var TranslatorInterface $translator */
         $translator = $this->getContainer()->get(TranslatorInterface::class);
-        $this->sut = new StatementExportTagFilter($translator, $this->getEntityManager());
+        /** @var DqlConditionFactory $conditionFactory */
+        $conditionFactory = $this->getContainer()->get(DqlConditionFactory::class);
+        $this->sut = new StatementExportTagFilter($translator, $this->getEntityManager(), $conditionFactory);
```
mussbach: The only test change here is the constructor wiring. There is no test for `buildStatementTagConditions()` (OR-combination, empty criteria → `[]`, single vs. multiple criteria), no equivalence test that the pushed-down query filter returns the same statements as the in-PHP filter, and no de-duplication test. A "no statement appears twice" assertion would catch the to-many join duplication noted on `StatementExportTagFilter.php`.

mussbach (commented):
Automated review — 2 inline findings (1 correctness, 1 test-coverage).

demosplan/DemosPlanCoreBundle/Logic/Statement/Exporter/StatementExportTagFilter.php
bakridemos: Thanks for the careful read — we dug into the EDT read path and also tested it on our side against the exact scenario you flagged (a statement with two segments both carrying the filtered tag), and the statement is not duplicated in the export. The premise about hydration doesn't hold for this query shape. Details:

The SELECT is root-only, not fetch-joined. QueryBuilderPreparer::fillQueryBuilder selects only the root alias when no select expressions are set (edt-dql/src/Utilities/QueryBuilderPreparer.php:155-156, $queryBuilder->select($entityAlias)), and setSelectExpressions() is never called on the read path (it appears only in EDT's own test). So the generated DQL is:

SELECT s FROM Statement s LEFT JOIN s.segmentsOfStatement seg LEFT JOIN seg.tags t WHERE t.id IN (:ids)

The seg/t joins exist purely for the WHERE; they're not in the SELECT.

Hydration is HYDRATE_OBJECT. DoctrineOrmEntityProvider::getEntities uses ->getQuery()->getResult(). With a root-only SELECT, Doctrine's ObjectHydrator keys on the root id via the identity map and emits each Statement exactly once — the cartesian (segment, tag) rows collapse. Duplicate roots are the array/scalar-hydration (or fetch-joined-SELECT) problem, not object hydration, which is why no DISTINCT/GROUP BY is needed here.

No LIMIT on the export path. Export enters listObjects with pagination = null, so no setMaxResults is applied — the one case where a missing DISTINCT actually bites (fetch-joined rows miscounting a LIMIT) isn't reached either.

So getObjectsByQueryParams already returns unique Statement objects, and applyTagFilter never sees duplicates — no extra dedup required. We confirmed this by testing: the two-matching-segments case exports the statement once.

tests/backend/core/Statement/Functional/SegmentsExportControllerTagFilterTest.php
bakridemos: done in https://github.com/demos-europe/demosplan-core/pull/6517/changes/fe9769f0a402f8d45f7ddd87dd639fa3f7556daa

mussbach (approved):
thx lgtm, worx

PR #6522, opened 2026-07-07
fix: skip unconfirmed and non-UUID segment marks during finalization

bakridemos:
Description:
When a split is finalized, `DraftsInfoToSegmentTransformer` turns every `<segment-mark>` found in `textualReference` into a `Segment` entity, using the mark's id verbatim as the primary key (via `AssignedGenerator`).

Two problems resulted:
- Marks whose id is absent from the schema-validated `segments` metadata were still persisted. These are leftovers for content that could not be classified (e.g. images or un-OCR-able tables) and are not real segments.
- Their ids are not UUIDs (e.g. `21929_24762`), so an invalid value reached the primary key and caused errors when the id was later used as a UUID.

This PR gates segment creation: a mark is only materialized when its id is present in the `segments` metadata **and** is a valid UUID; otherwise it is skipped and logged. It also fixes a latent bug in the post-persist tag loop, which indexed `$parsedMarks` by `$segments` position and would mis-assign tags as soon as any mark is skipped — it is now keyed by the segment's own id.


mussbach (changes requested):
pls add a test. Is a part of the statement that is visually marked as segment now silently dropped, or is the dropped part not event visible in the interface beforehand?

bakridemos:
> pls add a test. Is a part of the statement that is visually marked as segment now silently dropped, or is the dropped part not event visible in the interface beforehand?

1. Real segments that happen to carry a corrupted id — these are not dropped. The invalid id is replaced with a valid one (id substitution) upstream, so the segment is still materialized and the user can see and work with it normally. Nothing visible disappears.

2. Marks that aren't real segments — e.g. content wrapped in a class="pdf_importer_image" element. These are not shown because they are not real segments and have no metadata, so the frontend does not display them. Their presence is a parsing artifact from the imported data, not a real segment, so skipping them is exactly what we want.

So the skip logic only removes non-segments (parsing artifacts) and never a real, user-facing segment — corrupted ids on real segments are healed via replacement rather than dropped.

test added in https://github.com/demos-europe/demosplan-core/pull/6522/changes/61c10ea6be98e49fd6aee9701cbfb66041fdbe2f

demosplan/DemosPlanCoreBundle/Transformers/Segment/DraftsInfoToSegmentTransformer.php
```diff
@@ -118,10 +119,26 @@ private function getSegments(array $draftsInfoArray, Statement $statement): arra
...
         foreach ($parsedMarks as $segmentMark) {
-            $metadata = $metadataById[$segmentMark['segmentId']] ?? [];
+            $segmentId = $segmentMark['segmentId'];
+
+            // Only materialize segments the planner actually confirmed. Confirmed
+            // segments are the ones listed (and schema-validated) in the segments
+            // metadata; a mark whose id is missing there is a leftover the pipeline
+            // could not classify (e.g. images or un-OCR-able tables) and must not
+            // become a Segment. The id is also persisted verbatim as the entity
+            // primary key, so reject any non-UUID id as a safeguard.
```
mussbach: The gate doesn't check confirmation — the FE auto-confirms every remaining proposal in `confirmAllUnconfirmedSegments()` before the final save, and no status field is read here. The condition is "id is present in the segments metadata". Please reword the comment and the commit subject ("skip unconfirmed…"), otherwise the next reader looks for a confirmation check that doesn't exist.

demosplan/DemosPlanCoreBundle/Transformers/Segment/DraftsInfoToSegmentTransformer.php
```diff
@@ -118,10 +119,26 @@ private function getSegments(array $draftsInfoArray, Statement $statement): arra
...
+            if (!isset($metadataById[$segmentId]) || !Uuid::isValid($segmentId)) {
```
mussbach, line 136: `Uuid::isValid($segmentId)` can only ever reject a 36-char non-UUID, since `drafts-info-schema.json` pins `segments[].id` to `minLength`/`maxLength` 36 and the first condition already covers everything else. Fine as a safeguard — worth one line saying so, plus a test that actually hits it.

tests/backend/core/Statement/Functional/DraftsInfoToSegmentTransformerTest.php
```diff
@@ -0,0 +1,145 @@
...
+        self::assertSame('ST-1-2', $segments[1]->getExternId());
+    }
+
+    /**
+     * Answers the review question "is a visually marked part now silently
+     * dropped?": only marks that are absent from the metadata or carry a
+     * non-UUID id are dropped. A confirmed mark interleaved with them still
+     * survives, and the extern-id counter does not leave gaps for skipped marks.
+     */
+    public function testUnconfirmedAndNonUuidMarksAreDroppedButConfirmedSurvives(): void
```
mussbach: Two gaps here:

- The non-UUID gate isn't exercised: `NON_UUID_ID` is also absent from the metadata, so it's dropped by `!isset(...)` before `Uuid::isValid` is reached. Needs a 36-char non-UUID id *present* in metadata.
- The tag rekeying — the actual latent bug this PR fixes — isn't covered. Both tests use `'tags' => []`, so tags mis-assigned after a skipped mark would still pass. A case with distinct tags on interleaved confirmed/skipped marks is the one that guards the fix.

tests/backend/core/Statement/Functional/DraftsInfoToSegmentTransformerTest.php
```diff
@@ -0,0 +1,145 @@
...
+ * schema-validated `segments` metadata AND is a valid UUID. A planner-confirmed
+ * segment always satisfies both (the FE writes a matching metadata entry and the
+ * schema forces those ids to be 36 chars), so it is never dropped. Marks that
+ * fail either gate are leftover pipeline placeholders (content the pipeline could
+ * not classify, e.g. images or un-OCR-able tables) or corrupted ids that would
+ * otherwise be persisted verbatim as an invalid primary key.
+ */
+class DraftsInfoToSegmentTransformerTest extends FunctionalTestCase
+{
+    protected $sut;
```
mussbach: `protected ?DraftsInfoToSegmentTransformer $sut = null;` per convention.

mussbach (commented):
Thanks — the gating logic and the tag-loop fix look right, and the test answers my second question. Three follow-ups inline: the test doesn't actually reach the UUID check or cover the tag rekeying, and the "unconfirmed" wording doesn't match what the code checks.

On your point 1: I couldn't find any upstream id substitution — FE ids come from `uuid()`. What actually makes that case unreachable is the schema, which pins `segments[].id` to 36 chars, so a corrupted id in the metadata fails validation loudly in `DraftsInfoValidator` instead of being silently skipped. Same conclusion, different mechanism — worth correcting in the PR description so the reasoning survives.

On point 2: agreed that a mark without a metadata entry gets no card, tags or place. Small caveat — `mergeSegmentMarksAndSegments` already treats that as an anomaly (`console.warn('A segment was updated in Prosemirror but no corresponding segment found in store.')`), and the text stays wrapped in a rendered span in the editor, so it can still look marked. Not blocking; skipped content also stays in the statement via `draftsInfoHandler->save()`, and `confirmDrafts` still errors with `statement.has.no.segments` if everything is skipped.

bakridemos:
> Thanks — the gating logic and the tag-loop fix look right, and the test answers my second question. Three follow-ups inline: the test doesn't actually reach the UUID check or cover the tag rekeying, and the "unconfirmed" wording doesn't match what the code checks.
> 
> On your point 1: I couldn't find any upstream id substitution — FE ids come from `uuid()`. What actually makes that case unreachable is the schema, which pins `segments[].id` to 36 chars, so a corrupted id in the metadata fails validation loudly in `DraftsInfoValidator` instead of being silently skipped. Same conclusion, different mechanism — worth correcting in the PR description so the reasoning survives.
> 
> On point 2: agreed that a mark without a metadata entry gets no card, tags or place. Small caveat — `mergeSegmentMarksAndSegments` already treats that as an anomaly (`console.warn('A segment was updated in Prosemirror but no corresponding segment found in store.')`), and the text stays wrapped in a rendered span in the editor, so it can still look marked. Not blocking; skipped content also stays in the statement via `draftsInfoHandler->save()`, and `confirmDrafts` still errors with `statement.has.no.segments` if everything is skipped.

thx, all concerns are addressed in https://github.com/demos-europe/demosplan-core/pull/6522/changes/2f94fe33c9d53ccbc2ca0bb7c95550953997879c

mussbach (approved):
lgtm, thx, code review only

PR #6529, opened 2026-07-08
Fix(DPLAN-18100): ghost boxen when deleting segments in the split statement view

sakutademos:
### Ticket
[DPLAN-18100](https://demoseurope.youtrack.cloud/issue/DPLAN-18100) Fehlerhafte Abschnitts-Box beim Aufteilen

**Description:** This PR fixes an issue in the **Split Statement view** where ghost boxes could appear when users removed multiple segments in quick succession.

The root cause was that every removeSegment action triggered a full synchronization with the backend, even though the frontend already had the latest state. After each removal, the following requests were executed:

1. Remove the segment and its textual reference from the local store.
2. PATCH the updated segmentDraftList.
3. GET the updated statement.
4. GET the segmentDraftList again.

The last two requests were redundant because the frontend had already updated its local state and had just sent the same data to the backend. Besides the unnecessary network traffic, these overlapping requests could overwrite newer local changes with stale data, causing deleted segments to reappear as ghost boxes.

This PR removes the redundant synchronization. After a successful PATCH, saveSegmentsDrafts now updates initialData and initialSegments directly from the request payload instead of fetching them again. If the request fails, the editor restores the previous state from the saved initial data.

**As a result:**

- fixes the race condition that caused ghost boxes
- reduces the number of API requests

**Additional Performance Improvement**

This PR also [removes](https://github.com/demos-europe/demosplan-core/pull/6529/changes#diff-c3a27babad1569297407c6180509b943c468cf3c5a2408a3d8f84fcc4ce88cfaR17) the use of

`:key="'card' + segment.id + Math.random()"`

and replaces it with a stable key:

`:key="'card' + segment.id"`

Using `Math.random()` as part of a Vue key is an anti-pattern because it generates a new key on every render. As a result, Vue cannot reuse existing components and is forced to destroy and recreate every card on each update. This leads to unnecessary DOM updates, repeated lifecycle hooks, recalculation of card positions, and reduced performance, especially for statements with many segments.

Using a stable key based on the segment ID allows Vue to correctly track component identity, reuse existing components, and update only the segments that have actually changed, resulting in significantly better rendering performance.

**As a result:**
- significantly improves performance, especially for statements with many segments (in an STN with 200 segments, the re-render time after clicking the Remove button was reduced from about 2 seconds to 39 milliseconds)

### How to review/test

- Create a new STN with a large text.
- Click Aufteilen to generate a large number of AI segments.
- Quickly delete 10 or more segments one after another > No ghost boxes should appear.

**Verify the following operations:**
- Edit a segment > Save > succeeds
- Remove a segment > succeeds
- Create a new segment > succeeds
- Leave the page and open Aufteilen for the same STN again > the editor and segments should be restored correctly
- Split the statement > the statement is split successfully and all segments are created correctly

If any request fails:

- The previously saved initial state should be restored
- An error notification should be displayed

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Update changelog


client/js/store/statement/SplitStatementStore.js
```diff
@@ -92,7 +92,7 @@ const SplitStatementStore = {
     },
 
     resetSegments (state) {
-      state.segments = state.initialSegments
+      state.segments = structuredClone(state.initialSegments)
     },
```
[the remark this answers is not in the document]
sakutademos: segments is a reference directly from the API response data (in the fetchInitialData()). So the segments come fresh from the API there.

client/js/components/statement/splitStatement/CardPane.vue
```diff
@@ -14,7 +14,7 @@
   >
     <card-pane-card
       v-for="segment in segments"
-      :key="'card' + segment.id + Math.random()"
+      :key="'card' + segment.id"
```
hwiem: nitpick: we decided at some point to use template literals (not sure if it was with `:` or `_` though 🤔 ). in any case thanks for fixing this :)
```suggestion
      :key="`card:${segment.id}`"
```

client/js/store/statement/SplitStatementStore.js
```diff
@@ -390,41 +390,47 @@ const SplitStatementStore = {
...
+          commit('setProperty', { prop: 'needsEditorRefresh', val: true })
+        })
     },
 
     saveSegmentsFinal ({ dispatch, state, commit }) {
       const dataToSend = JSON.parse(JSON.stringify(state.initialData))
 
-      dataToSend.attributes.segments = state.segmentsWithText
+      dataToSend.attributes.segments = state.segments
       dataToSend.attributes.statementText = state.statementText
```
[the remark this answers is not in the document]
sakutademos: thx, refactored it

client/js/components/statement/splitStatement/SplitStatementView.vue
```diff
@@ -192,36 +192,39 @@ import { mapActions, mapGetters, mapMutations } from 'vuex'
...
- * This function synchronizes that state with the segment metadata (tags, place, assignee) stored in Vuex.
+ * SegmentMarks are ProseMirror marks that track segment positions in the editor.
+ * This function synchronizes segment metadata (tags, place, assignee, status) from the Vuex store
+ * with the ProseMirror marks.
  *
  * @param {Array} segmentMarks - Array of segmentMark objects from ProseMirror with {segmentId, isConfirmed}
  * @param {Array} segments - Array of segment metadata objects from Vuex store
- * @return {Array} Array of merged segment objects with updated status
+ * @return {Array} Array of merged segment objects preserving store status
  */
```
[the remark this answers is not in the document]
sakutademos: refactored

client/js/components/statement/splitStatement/SplitStatementView.vue
```diff
@@ -873,6 +881,26 @@ export default {
...
+
+      const newState = EditorState.create({
+        doc: newDoc,
+        plugins: this.prosemirror.view.state.plugins,
+      })
+
+      this.ignoreProsemirrorUpdates = true
+      this.prosemirror.view.updateState(newState)
+      this.ignoreProsemirrorUpdates = false
+    },
```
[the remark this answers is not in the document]
sakutademos: right! done here https://github.com/demos-europe/demosplan-core/pull/6529/commits/5ae431b5d1e5470ca0341f190c21f8f2a881c3ac

client/js/components/statement/splitStatement/SplitStatementView.vue
```diff
@@ -192,42 +192,14 @@ import { mapActions, mapGetters, mapMutations } from 'vuex'
 import AddonWrapper from '@DpJs/components/addon/AddonWrapper'
 import CardPane from './CardPane'
 import dayjs from 'dayjs'
-import { DOMSerializer } from 'prosemirror-model'
+import { DOMParser, DOMSerializer } from 'prosemirror-model'
```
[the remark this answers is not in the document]
sakutademos: dine here https://github.com/demos-europe/demosplan-core/pull/6529/commits/ea48bdf1e0e10e8a377ba2e42952e2a48d863f73

client/js/store/statement/SplitStatementStore.js
```diff
@@ -92,7 +92,7 @@ const SplitStatementStore = {
     },
 
     resetSegments (state) {
-      state.segments = state.initialSegments
+      state.segments = structuredClone(state.initialSegments)
```
hwiem, line 95: Is it safe to use `structuredClone` here and below? (it fails if `initialSegments`/`initialData`/`segments` contains functions)

client/js/components/statement/splitStatement/SplitStatementView.vue
```diff
@@ -826,15 +837,16 @@ export default {
...
+     * 4. Enable prosemirror change listeners by setting ignoreProsemirrorUpdates to false
      */
     runPostInitTasks (prosemirrorState) {
       this.prosemirror = prosemirrorState
       const { state } = this.prosemirror.view
       const segmentMarks = Object.values(this.prosemirror.keyAccess.rangeTrackerKey.getState(state))
-      const updatedSegments = mergeSegmentMarksAndSegments(segmentMarks, this.segments)
 
-      this.locallyUpdateSegments(updatedSegments)
+      this.ignoreProsemirrorUpdates = true
```
hwiem, line 866: Why can we remove 
```
const updatedSegments = mergeSegmentMarksAndSegments(segmentMarks, this.segments)

this.locallyUpdateSegments(updatedSegments)
```

Don't we have to remove this

>  2. Synchronize segment metadata with segmentMarks parsed from HTML

 from the doc block then?

hwiem (commented):
Just a nitpick and some questions (code review only for now)

client/js/components/statement/splitStatement/SplitStatementView.vue
sakutademos: it could be removed from here, but we still need to add this a bit different way.. the `status` property has to be added to segments that do not have it, such as AI-generated segments. 

Therefore, before updating the `segment-mark`s with the status from the store, we need to check whether the corresponding segment has a status property. If it does not, the status should be initialized with false. done here https://github.com/demos-europe/demosplan-core/pull/6529/commits/dd5c6ddb983e164ce7f31fd7c4df354ef61ef079

client/js/store/statement/SplitStatementStore.js
sakutademos: also, `structuredClone()` does not work with functions - it throws an error if the object contains one. thats true.

However, our segments from `segmentDraftList` are plain data objects and do not contain any functions. In this case, `structuredClone()` is safe to use and is the recommended modern alternative to JSON.parse(JSON.stringify(...))

client/js/components/statement/splitStatement/SplitStatementView.vue
hwiem: Why should it be initialized with `false`?

sakutademos: during initialization, if a segment does not yet have a `status` property (AI generated segments), it first checks whether the status exists. If it does not exist, it falls back to the marks confirmation state and initializes the status as follows: `status: mark.isConfirmed ? 'confirmed' : false` (same way as it was before). This means that if the mark is not confirmed, the segment status is initialized to `false`. So that AI segments receive `status: false` by the first initialization

hwiem (approved):
Lgtm + works!

PR #6530, opened 2026-07-08
f DS-568 exclude identity-provider users from inactivity deletion

bakridemos:
Ticket : DS-568

Description:
Users provisioned by an external identity provider (`provided_by_identity_provider = true`) were being picked up by the inactivity-based account deletion workflow and soft-deleted after the warning cascade. Their lifecycle is owned by the IdP, so they must never be deleted by this job.

This PR excludes IdP-provided users from the deletion workflow on two levels:

- **Primary guard** — `UserRepository::findInactivityDeletionCandidates()` now filters out users with `providedByIdentityProvider = true`. They never become candidates, so they receive no warning mails, get no tracking row, and are never soft-deleted.
- **Defensive guard** — `AccountDeletionRunMessageHandler::processCandidate()` short-circuits for IdP users in case such a user reaches the handler by another path (e.g. the `dplan:account-deletion:prepare-test` QA command, which invokes the same handler).

No dedicated CLI deletion command exists for this feature; the only deletion path is the handler's soft-delete, which both guards now cover.

- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/MessageHandler/AccountDeletionRunMessageHandler.php
+++ b/demosplan/DemosPlanCoreBundle/MessageHandler/AccountDeletionRunMessageHandler.php
@@ -108,6 +108,15 @@ final class AccountDeletionRunMessageHandler
 
     private function processCandidate(UserInterface $user): void
     {
+        // Accounts provisioned by an external identity provider are never soft-deleted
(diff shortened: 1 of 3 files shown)


MoritzMandler (approved):
lgtm - thx

PR #6534, opened 2026-07-09
feat(DPLAN-17966): log aborted statement imports and enrich import er…

bakridemos:
Ticket: DPLAN-17966

Description:
When a statement xlsx / participation import is rolled back due to per-row validation errors (e.g. an attachment reference that isn't present in the ZIP), the importer returned silently: nothing was logged and the whole batch was discarded with no server-side trace, which made such failures very hard to diagnose.

This PR improves observability only — it does **not** change import outcomes:

- `XlsxStatementImport::importFromFile()` now logs a `warning` (error count, per-row errors, skipped counts) before the rollback, instead of returning silently.
- `ImportError::toArray()` now also exposes `field` (the violation property path) and a scalar-safe `invalidValue`, so both the JSON error response and  the new log identify the offending column and value.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Import/Statement/ImportError.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Import/Statement/ImportError.php
@@ -49,16 +49,23 @@ class ImportError
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mussbach (approved):
lgtm, thx for this!

PR #6535, opened 2026-07-09
build(docker): use project-local var/tmp for PHP temp dir

mussbach:
## Description
Point PHP's temp directory at the project-local, writable `var/` tree instead of the shared `/tmp` in the production image.

- Set `sys_temp_dir` (and `upload_tmp_dir`) to `/srv/www/var/tmp` in the fpm and cli PHP config.
- Create `var/tmp` and set `www-data` ownership during the image build.
- Add `var/tmp/.gitkeep` so the directory ships with the repo.

## Why
`sys_get_temp_dir()` previously resolved to `/tmp`, which is shared and not guaranteed writable/owned by the runtime user. Keeping temp files in the app-owned `var/` tree matches the existing `var/cache` / `var/log` convention.

Ref: DPLAN-17966


bakridemos (approved):
lgtm thx

PR #6537, opened 2026-07-09
fix(refs: DPLAN-18169): add gap between checkboxes

meissnerdemos:
### Ticket
[DPLAN-18169](https://demoseurope.youtrack.cloud/issue/DPLAN-18169/Einzelexport-Abstand-der-Auswahlkastchen-anpassen)

Adds a gap between the checkboxes.

### How to review/test
navigate to single statement export and check gap between checkboxes. bug only showed on larger screens, so resize the window until there are no line breaks in the checkbox labels.


hwiem (approved):
Lgtm, thanks!

PR #6543, opened 2026-07-10
fix(refs DS-573): guard segmentation editor against segments without a document mark

mussbach:
### Ticket
DS-573

### Description
Segments that exist in the draft segments array but have no `segmentMark` in the textual reference have no range in the ProseMirror document. The range-editing code assumed every edited segment has a range, so opening, saving or scrolling such a segment threw uncaught `TypeError`s (`marks.selection is undefined`, `getBoundingClientRect` on `undefined`), blocking metadata edits (e.g. adding a tag).

This adds guards so a segment without a document mark can still have its metadata edited:

- `applySelectionChange`: when there is no active range selection, stop range editing cleanly and return success instead of applying a boundary change — so metadata-only edits still persist.
- `enableEditMode`: open the metadata editor without activating range editing when the segment has no range.
- `disableEditMode`: only reset range-editing state when the segment actually has a range.
- `determineDisplayScrollButton`: show no scroll button when there is no active segment span.

Normal (marked) segments are unaffected; the guards only trigger when a range is absent.

### How to review/test
Open a statement in the segmentation editor whose draft contains a segment that is present in the segments list but not marked in the text. Opening that segment, scrolling, and saving a metadata change (e.g. adding a tag) must no longer crash and must persist the change. Unit tests cover each guard for both the range-less path and the normal path.

### PR Checklist
- [x] Create/Update tests
- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/statement/splitStatement/SplitStatementView.vue
+++ b/client/js/components/statement/splitStatement/SplitStatementView.vue
@@ -422,6 +422,15 @@ export default {
 
     determineDisplayScrollButton () {
       const segmentSpans = document.querySelectorAll('span[data-range-active="true"]')
+
+      /**
+       * Without an active segment span in the document (e.g. a segment without a mark is being
+       * edited), there is nothing to scroll to, so no scroll button is shown.
+       */
+      if (!segmentSpans.length) {
+        return false
+      }
+
       const lastSegmentSpan = segmentSpans[segmentSpans.length - 1]
       const segmentBottom = lastSegmentSpan.getBoundingClientRect().bottom
       const headerHeight = document.getElementById('header').getBoundingClientRect().height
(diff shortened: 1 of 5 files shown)


client/js/components/statement/splitStatement/SplitStatementView.vue
hwiem: This 
```
return segmentSpans.length ?
        segmentIsAtTop || segmentIsAtBottom :
        false
```
can now be changed to
```
return segmentIsAtTop || segmentIsAtBottom
```

client/js/components/statement/splitStatement/SplitStatementView.vue
```diff
@@ -422,6 +422,15 @@ export default {
 
     determineDisplayScrollButton () {
       const segmentSpans = document.querySelectorAll('span[data-range-active="true"]')
+
+      /**
```
hwiem, line 426: I'd find it helpful if the comment explained that normally, there should be no segments without a mark, or that we're handling a case when the data is broken, because right now it leaves me wondering why there would be a segment without a mark (here and below)

client/js/components/statement/splitStatement/SplitStatementView.vue
```diff
@@ -507,6 +523,14 @@ export default {
       this.setProperty({ prop: 'editingSegment', val: this.segmentById(id) })
       this.setProperty({ prop: 'editModeActive', val: true })
 
+      /**
+       * Segments whose mark is missing from the document have no range to edit. Allow editing their
```
hwiem: Should it really be possible to edit the metadata? Maybe it's outside the scope of this PR, but I think from a user perspective, it would be better to show a warning message in the sidebar when the user is editing a segment that has no text

hwiem (commented):
Thanks a lot for this one, especially the tests ❤️ I have some small remarks, but they are not blockers.

mussbach:
Thank you @hwiem I tried to fix your concerns in [8dfe34e](https://github.com/demos-europe/demosplan-core/pull/6543/commits/8dfe34e140675b76fb7eb6047831e2c4a1e82b56)

client/js/components/statement/splitStatement/SplitStatementView.vue
mussbach: Yes, not sure how to handle that either

PR #6549, opened 2026-07-13
fix(DS-576): reset unconvertible legacy segment draft to null

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/tickets/DS-576

Description:
When a legacy drafts_info_json cannot be fully converted to the new <segment-mark> format, the stored draft is reset to null and null is returned to the client instead of a partially converted document.

Why

Some legacy drafts have a segment text snapshot that no longer matches textualReference (e.g. re-serialized  tags), or unusable positions. The read-time migrators then leave one or more segments unwrapped. The segment editor renders a card with no matching mark in the ProseMirror document, and deleting that card dispatches a null transaction → can't access property "before", tr is null crash.

How

- DraftsListJsonSegmentFields::allSegmentsWrapped() — checks every segment ended up wrapped (count(segments) === count of <segment-mark data-segment-id=).
- StatementResourceType (segmentDraftList readable): after attempting migration, if it was a legacy format and not fully wrapped → setDraftsListJson(null) + flush() and return null. Nulling the raw column makes the record look like a never-drafted statement, so the segmentation pipeline (demospipes) requests fresh proposals in the new format; the pipeline result overwrites the draft unconditionally.
- Statement::setDraftsListJson() widened string → ?string to match the nullable drafts_info_json column.

Testing

- DraftsListJsonSegmentFieldsTest (unit).

- [X] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Entity/Statement/Statement.php
+++ b/demosplan/DemosPlanCoreBundle/Entity/Statement/Statement.php
@@ -3999,7 +3999,7 @@ class Statement extends CoreEntity implements UuidEntityInterface, StatementInte
         return null;
     }
 
-    public function setDraftsListJson(string $json): void
+    public function setDraftsListJson(?string $json): void
     {
         $this->draftsListJson = $json;
     }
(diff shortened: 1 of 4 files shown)


mussbach:
We are currently discussing whether this is the best way

PR #6563, opened 2026-07-14
feat (DPLAN-18102): Remove conditional that fails when editing name o…

sierrazamudiodemos:
https://demoseurope.youtrack.cloud/issue/DPLAN-17975/Neu-Ado-issue-54804-Bearbeiten-einer-Gruppe

**This is a bug fix found in the Edition Stn group story**
1, Fixes PATCH requests to `/api/3.0/StatementGroup/{id}` failing when only updating the group name. 
2. The statements field was being validated as if the client sent it, but it was actually pre-filled by the API with the group's existing members before validation ran, so the update always got rejected. 
**The fix:** Removed the incorrect validation rule and made sure the name change is actually saved to the database.


mussbach (approved):
lgtm, thx code review only

PR #6565, opened 2026-07-14
fix(DPLAN-18071): show detach toast before dissolve toast on last member removal

riechedemos:
### Ticket
DPLAN-18071

Removing a group's last member redirected straight to the statement list with only the "group resolved" toast — the "statement detached" toast was silently skipped, because a plain `location.href` change unloads the page before a toast fired there could ever be seen.

Queues the detach toast via the same lscache mechanism already used for the "group resolved" toast, and shows it first on the statement list.

### How to review/test
- Open a group with 2+ members, remove members one by one down to the last one
- Expect: after the redirect to the statement list, first the "Stellungnahme X wurde aus der Gruppe Y gelöst" toast, then the "Gruppe Y wurde aufgelöst" toast
- Removing a non-last member should behave as before (single toast, no redirect)
- Using the "Gruppe auflösen" button should still show only the "aufgelöst" toast (no detach toast, since there's no single removed member in that flow)

### PR Checklist
- [ ] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Update changelog


client/js/components/statement/listStatements/ListStatements.vue
```diff
@@ -683,6 +683,20 @@ export default {
 
     // Show the "group resolved" toast once when arriving here after the last group member was detached
     notifyClusterResolved () {
+      const detachedRaw = lscache.get(`${this.procedureId}:clusterElementDetached`)
```
rafelddemos: maybe this name could be more informative as to what `'raw'` it refers to for readability - sth like `detachedToastData` or similar

rafelddemos (commented):
works well, thx! :-)

one nitpick (didn't notice it before): maybe it would make sense to name the JS file and helper in the **shared** folder `redirectToStatementListWithGroupResolvedToast`? it is then even longer, but straight away clear it is about dissolving a STN group

client/js/components/statement/listStatements/ListStatements.vue
riechedemos: fixed 9b0d2d8628aed783dfd89174dbe3883f21602624

rafelddemos (approved):
thx!:)

PR #6581, opened 2026-07-16
F hddp 27 async assessment table export

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/HDDP-27

Description:
Exporting the assessment table (PDF / ODT / ZIP) fails on large procedures (hundreds to ~1600 statements). The export builds one PDF per statement serially inside a single web request, so it hits the fixed gateway timeout (~180s) and is cut off — small procedures squeak under it, large ones never finish.

Raising timeouts or optimising the in-request render was ruled out by measurement: the dominant cost is the per-statement document build, which cannot be brought under the request timeout for large procedures. The only robust fix is to run the export without a web-request timeout.

This PR moves the export into a background job (Symfony Messenger, Doctrine transport). The existing synchronous exporter is reused unchanged; only the request-scoped context it needs (session filter hash, current user, current procedure, permissions) is rebuilt inside the worker. The browser starts the job, polls its status, and downloads the result once it is ready.

What is added:
- Symfony Messenger + Doctrine transport (async transport backed by the app's Doctrine connection; failed transport for retries).
- A job entity (status, procedure, user, result file) with its migration.
- A message + handler that runs the exporter off-request and stores the resulting file (streamed to disk in chunks so large zips stay off-heap).
- Start / status / download endpoints (procedure-scoped, ownership-checked).
- Frontend: the export modal starts the job, polls, and triggers the download, with a "processing" notice while it runs.

Session access in the exporter is made lazy so it constructs without an active
request. No behaviour change for existing synchronous callers.

### Note: async export now survives page refresh / navigation ( see commit https://github.com/demos-europe/demosplan-core/pull/6581/changes/6c4b63384bd886d1fd11c3483c7ab961d1fe1b6c )

The async export runs in a Symfony Messenger worker and always finishes regardless of
the browser. However, status polling previously lived in an in-memory `setTimeout` loop
inside the Vue component, so refreshing the page or navigating away killed the loop and
the finished file was never downloaded — it *looked* like the export was cancelled.

**Fix (frontend only — the backend already persists the job):** each job's status/download
URLs are stored in `localStorage`, and polling is resumed **globally on every page load**
(`InitVue.js` root mount). The export now downloads wherever the user is once the worker
finishes — even after a refresh, navigating to an unrelated page, opening a new tab, or a
browser restart. Stale entries self-heal (a purged job returns `not_found` and is cleared),
and polling survives transient network blips.

Applies to both async exports: the Abwägungstabelle export (`ExportModal.vue`) and the
procedure export (`AdministrationProceduresList.vue`).

Behavior note: the download is triggered on whatever page the user is on when the job
completes (a file-attachment download, so it does not navigate away).

### How to review/test
1. Run the new migration to create the job table.
2. Start a worker: `bin/<project> messenger:consume async --time-limit=3600 --memory-limit=1024M -v`
   (the queue table is auto-created by the Doctrine transport).
3. In a procedure with assessment-table access, open the export modal and export to PDF / ODT / ZIP. The request returns immediately; the browser polls and downloads the file when the worker finishes.
4. Verify the job row transitions pending → processing → completed, and that status/download are rejected for another user or a mismatched procedure.


mussbach:
PR itself is fine, it just cannot be merged into release_robobsh, as this would lead to an unmergeable release_robobsh into main, which we cannot do. This Branch should live as long as it needs to and then be dropped after the next basic update

mussbach:
merged into main via #6665

PR #6584, opened 2026-07-17
refactor(refs DPLAN-15761, #AB-27466): replace template tag [vue 3 migration]

hwiem:
- in vue 3, non-root `<template>` tags without special directives (like `v-if`, `v-else`, `v-for`, `v-slot`) are treated as plain elements and will result in native <template> elements instead of rendering their inner content - so the `<template>` tag is replaced with a `<div>` element
see https://v3-migration.vuejs.org/breaking-changes/
- also, `v-html` is replaced with `v-cleanhtml` to make eslint happy, and `c-styled-html` is used to to display `<ul>` and `<ol>` in non-edit mode

### Linked PRs
- [x] ui cleanup: https://github.com/demos-europe/demosplan-core/pull/6589
- [x] BE fix for editing/creating support contacts: https://github.com/demos-europe/demosplan-core/pull/6590

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`


riechedemos (approved):
Thank you, lgtm - code review only.

PR #6598, opened 2026-07-21
feat (DPLAN-18102): Custom field filter with ES

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-18102/Neu-ADO-55605-Konfigurierbare-Felder-in-Stellungnahmen-Part-7-Suchen-und-Filtern
**Title: Custom field filter with ES**

**What this does**

Adds filtering (and live option counts) by custom fields in the assessment table. Custom field values are indexed into Elasticsearch as a nested list, and both the filter dropdown counts and the actual "apply filter" query reuse the existing statement-search pipeline instead of a separate lookup.

**Files changed**

**Backend**

- `config/packages/fos_elastica.yaml` — tells Elasticsearch about a new field on each indexed statement: a list of {id, value} pairs, one per custom field that statement has a value for. This is the raw data ES needs before it can filter on anything.
- `Entity/Statement/Statement.php `— adds one getter method that produces that {id, value} list from the statement's existing custom field data. ES's indexing tool calls this getter automatically by name.
- `Logic/CustomField/CustomFieldElasticaQueryBuilder.php `(new) — the only file that knows Elasticsearch's query syntax for this feature. Given a field ID and option ID(s), it builds either "find statements matching this option" (for filtering) or "count statements per option" (for the dropdown numbers).
- `Logic/CustomField/CustomFieldFilterResolver.php` (new) — reads the customField_* entries out of a filter request and decides what they mean (which fields, which options, ignoring "just opened this dropdown" placeholders). Also checks the permission that gates this whole feature.
- `Logic/CustomField/CustomFieldFilterResponseBuilder.php` (new) — builds what the filter modal actually displays: for every custom field the user has touched, works out its available options plus a live count for each, and packages that into the same object type the UI already uses for other filters.
- `Logic/Statement/ElasticsearchResultCreator.php` — the shared class that already builds the search query for the assessment table. Now it also asks the classes above whether to (a) filter the real result list by custom fields, or (b) attach count-only aggregations, depending on what the caller needs.
- `Logic/Statement/StatementFilterHandler.php` — when a saved/shared filter link is reopened, custom field selections need to be read back out and shown as "currently selected" — this file adds that one extra step.
- `Controller/Procedure/DemosPlanProcedureAPIController.php` — the API endpoint that returns "what filters are available right now." Adds the custom field filter items to that response, alongside the existing ones.
-` Transformers/Procedure/AssessmentTableFilterTransformer.php` — small tweak so the API response includes which field ID a custom-field filter belongs to (the frontend needs this to know which dropdown an option belongs to).

**Frontend**

- `client/js/components/shared/DpCustomFieldsFilter.vue` (new) — the actual dropdown component for one custom field: shows its options with counts, lets you sort alphabetically or by count.
- `client/js/components/statement/assessmentTable/DpFilterModal.vue` — the filter modal that already lists institution/phase/etc. filters. Now also renders one DpCustomFieldsFilter per custom field, and handles selecting/opening/submitting them alongside the existing filters.
- `client/js/components/statement/assessmentTable/DpTable.vue` — small wiring change so the table knows about custom field filters too.
- `client/js/components/statement/assessmentTable/FilterModalSelectItem.vue` — minor tweak (accessibility label) unrelated to the custom-field logic itself.
- `client/js/store/statement/Filter.js` — the Vuex store that holds filter state. Adds storage for custom field options/counts and the currently active custom field selections.
- `client/js/types/filters.ts` — TypeScript types describing the shape of a custom field filter definition and its options, so the components above get type-checking.

**How to review/test**

- Enable feature_statements_custom_fields, open the assessment table filter modal on a procedure with singleSelect/multiSelect custom fields, and confirm: options show counts, selecting one narrows the list, counts update as other filters change, and selections survive a page reload.


breierDemos (approved):
BE lgtm, I tested the filtering with statements and original statements and it works :)

client/js/components/statement/assessmentTable/DpFilterModal.vue
```diff
@@ -433,13 +538,20 @@ export default {
...
+            if (standardOptions.length > 0) {
+              this.selectedOptions = standardOptions
+              this.loadSelectedFilterOptions(standardOptions)
+              this.loadAppliedFilterOptions(standardOptions)
+            }
+
+            this.restoreCustomFieldFilterValue()
           }
         })
+        .then(() => this.getFilterOptionsAction({ filterHash: this.filterHash }))
```
hwiem: Why is this (`getFilterOptionsAction`) needed here?

client/js/components/statement/assessmentTable/DpFilterModal.vue
```diff
@@ -395,15 +437,56 @@ export default {
...
+      Object.entries(this.customFieldFilterValue).forEach(([fieldId, optionIds]) => {
+        optionIds.forEach(optionId => {
+          entries.push({ name: `filter_customField_${fieldId}[]`, value: optionId })
+        })
+      })
+
+      return entries
+    },
+
+    applyCustomFieldFilter () {
```
hwiem: When triggering this on @close, I think we should check if updating the filters is needed, i.e. if selected options changed

huellnerdemos: Thank you for this point! Yes it makes sense to check this. Changed it in 3504cdec90165865e412e1ae6560d15b159de86f

client/js/components/statement/assessmentTable/DpFilterModal.vue
huellnerdemos: For already added filters, the count has to be called, to render the custom field filters in the correct initial state. I moved getFilterOptionsAction as a return, to the method and added a comment in 3504cdec90165865e412e1ae6560d15b159de86f

hwiem (approved):
FE lgtm

PR #6608, opened 2026-07-22
feat(refs DPLAN-18219): update assignee on tag change

hwiem:
### Ticket
DPLAN-18219

When splitting a statement, changing a segment's tag did not update the Bearbeiter*in field: once an assignee was set from a tag, it stayed on the old value even after the tag was removed or replaced.

Now, on every tag change the assignee field is recomputed from the currently selected tags: the first selected tag that has a default assignee wins, and the field resets to "Nicht zugewiesen" when no selected tag provides one. The value is applied even if an assignee was already set, and is persisted through the regular save flow.

How to review/test

1. Start splitting a statement and edit a segment proposal.
2. Assign a tag that has a default assignee -> the Bearbeiter*in field is filled with that user
3. Remove that tag and assign a different tag that has another default assignee -> the field updates to the new user
4. Assign a tag without a default assignee (or remove all tags) -> the assignee field resets to "Nicht zugewiesen"
5. With two tags that both have a default assignee, the first selected one wins

### PR Checklist

- [x] Create/Update tests
- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/statement/splitStatement/SideBar.vue
+++ b/client/js/components/statement/splitStatement/SideBar.vue
@@ -335,10 +335,6 @@ export default {
       return this.deadline !== this.initialDeadline
     },
 
-    hasSelectedAssignee () {
-      return Boolean(this.selectedAssignee?.id) && this.selectedAssignee.id !== 'noAssigneeId'
-    },
-
     initialAssignee () {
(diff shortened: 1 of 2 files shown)


riechedemos (approved):
Thank you! I tested it and it works as expected.

PR #6610, opened 2026-07-23
feat(DPLAN-18225): place resource to api platform

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-18225/Place-Resource-edt-to-API-Platfrom-Ressource

**Description**

- Adds a GET ApiPlatform resource for `Place` (`Get`/`GetCollection` at `/3.0/Place`) as a first step to remove the legacy EDT `PlaceResourceType`. 
- Permission/access-condition logic is extracted into `AccessChecker`, used by both the old and new resource. 
- FE's `Place` list action now points at `/3.0/Place` to validate it against the "Abschnitte" admin screen
- Writes (create/update/reorder) still go through the EDT resource.

### How to review/test
- Check the "Abschnitte" list's "Schritt" column/filter still works.
- Select multiple abschnitte and check at the moment of selecting the update of places, that the places itself are coming correctly
- Go to Verfahren > Bearbeitungsschritte and also make sure the steps are displayed correctly
### PR Checklist
- [ ] Run `yarn lint`
- [ ] Run `yarn test`


demosplan/DemosPlanCoreBundle/Api/Place/AccessChecker.php
```diff
@@ -0,0 +1,55 @@
...
+    public function getAccessConditions(): array
+    {
+        $procedure = $this->currentProcedureService->getProcedure();
+        if (!$procedure instanceof Procedure) {
+            return [$this->conditionFactory->false()];
+        }
+
+        // for now all places can be read by anyone if they are available
+        return [
+            $this->conditionFactory->propertyHasValue($procedure->getId(), ['procedure', 'id']),
```
breierDemos: Maybe we can use Paths::place()->procedure->id here instead.
ProcedurePath/PlacePath/Paths are auto-generated (dplan:generator:entity:companion) and live in demosplan-addon — for a new entity/field not yet covered, that generator has to be re-run and copied over.

demosplan/DemosPlanCoreBundle/Api/Place/Provider.php
```diff
@@ -0,0 +1,104 @@
...
+use demosplan\DemosPlanCoreBundle\Api\Place\Resource as PlaceResource;
+use demosplan\DemosPlanCoreBundle\Entity\Workflow\Place;
+use demosplan\DemosPlanCoreBundle\Repository\Workflow\PlaceRepository;
+use EDT\DqlQuerying\Contracts\OrderBySortMethodInterface;
+use EDT\DqlQuerying\SortMethodFactories\SortMethodFactory;
+use InvalidArgumentException;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+use Webmozart\Assert\Assert;
+
+class Provider implements ProviderInterface
```
breierDemos: I put together a draft (#6628) showing what extracting this could look like: a shared AbstractDoctrineResourceProvider (so provide() lives once, and each concrete provider just supplies the resource class, sortable properties, and an entity→DTO mapper) plus a small AccessCheckerInterface, since Place's and Tag's AccessChecker are separate classes with identical shape but no common contract. Place/Provider.php is refactored onto it there as a worked example. Open to discussion on that PR

breierDemos (commented):
Thank you for pushing that further - here are some thoughts of mine.

Small naming nit: I'd consider naming these for what they are — PlaceProvider, PlaceAccessChecker, PlaceResource — rather than the generic Provider/AccessChecker/Resource relying on the Api\Place\ namespace to disambiguate. It cuts against how the rest of the codebase already names things (PlaceResourceType, TagRepository, StatementGroupResource, DraftStatementResource all bake the resource name into the class itself), and already forces an alias on import (use ...Place\Resource as PlaceResource). A bare Provider/Resource in a stack trace or global search is also ambiguous without the namespace in view. Not a blocker, just a consistency preference.

hwiem (approved):
FE lgtm, I adjusted a few things: urls are now created dynamically (I wanted to prevent hardcoding 3.0 routes like 1.0 routes) and urls are generated in one place rather than in every call site.
I also tested in the app and places load as expected 🎉

demosplan/DemosPlanCoreBundle/Api/Place/PlaceResource.php
```diff
@@ -0,0 +1,71 @@
...
+use ApiPlatform\Metadata\Get;
+use ApiPlatform\Metadata\GetCollection;
+use ApiPlatform\Serializer\Filter\PropertyFilter;
+use demosplan\DemosPlanCoreBundle\ApiResources\ApiPlatformConstants;
+use demosplan\DemosPlanCoreBundle\Entity\Workflow\Place as PlaceEntity;
+
+#[ApiResource(
+    shortName: 'Place',
+    operations: [
+        new GetCollection(uriTemplate: '/Place'),
```
winderdemos: Worth explicitly setting `paginationEnabled: false` here as a visual cue that pagination is deliberately off, not just missing.

tests/backend/core/JsonApi/PlaceResourceApiTest.php
winderdemos, line 1: I see we have the positive case `testGetReturnsPlaceScopedToCurrentProcedure` but should we catch a negative-case scenario?

Something along the lines of

```
$ownerProcedure = ProcedureFactory::new()->withDefaultSettings()->create();
$otherProcedure = ProcedureFactory::new()->withDefaultSettings()->create();
$place = PlaceFactory::createOne(['procedure' => $ownerProcedure]);
...
$response = $this->sendRequest('/api/3.0/Place/'.$place->getId(), 'GET', $user, $otherProcedure);
self::assertSame(Response::HTTP_NOT_FOUND, $response->getStatusCode());

demosplan/DemosPlanCoreBundle/Api/Place/AccessChecker.php
sierrazamudiodemos: holi, thanks for the feedback, applied here be921d15a980cda22564fb5103e1f1433fc44fd0

sierrazamudiodemos:
> Thank you for pushing that further - here are some thoughts of mine.
> 
> Small naming nit: I'd consider naming these for what they are — PlaceProvider, PlaceAccessChecker, PlaceResource — rather than the generic Provider/AccessChecker/Resource relying on the Api\Place\ namespace to disambiguate. It cuts against how the rest of the codebase already names things (PlaceResourceType, TagRepository, StatementGroupResource, DraftStatementResource all bake the resource name into the class itself), and already forces an alias on import (use ...Place\Resource as PlaceResource). A bare Provider/Resource in a stack trace or global search is also ambiguous without the namespace in view. Not a blocker, just a consistency preference.

feedback applied 8a07dfe786129f3d77e1046a48be3283d994b782

demosplan/DemosPlanCoreBundle/Api/Place/PlaceResource.php
sierrazamudiodemos: thx for the feedback - done 2389ead279fc4fd6afcd65afa88b95d330f04607

tests/backend/core/JsonApi/PlaceResourceApiTest.php
sierrazamudiodemos: thx fir the feedback, I added another test case: 48a500659b8596d77164c9b485a031bb43c30824

and the thing is that real 404 isn't returned here — access errors just redirect to an HTML page, not a proper error response. So instead I check that the place's name never appears in the response

demosplan/DemosPlanCoreBundle/Api/Place/Provider.php
sierrazamudiodemos: I think it is a good idea as more resources are coming.
thank you for opening the PR. i have few comments:
1. what about renaming AbstractDoctrineResourceProvider to AbstractResourceProvider?
2. what about returning by default sorting with emtpy array? not all the resources offer sorting
3. do u feel this Draft PR is a blocker for merging this current PR?

Thx!

breierDemos: No I think no blocker, but I think we should definitely look into that to avoid duplication. Maybe you can put your comments to the PR.

breierDemos (approved):
lgtm, thank you!

PR #6611, opened 2026-07-24
fix(refs DPLAN-18200): confirm unconfirmed segments on 'save and finish'

hwiem:
### Ticket
DPLAN-18200

- when the user clicks the 'finish segmentation' button, all segment proposals that have not been confirmed by the user are now auto-confirmed on save, meaning that they are sent with `status: 'confirmed'` to the BE
- previously, segments that were unconfirmed when saving didn't get a `pmId`, which lead to the error `TypeError: Cannot read properties of undefined (reading 'marks')`

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`

### DIFF
--- a/client/js/components/statement/splitStatement/SplitStatementView.vue
+++ b/client/js/components/statement/splitStatement/SplitStatementView.vue
@@ -776,6 +776,46 @@ export default {
       this.ignoreProsemirrorUpdates = false
     },
 
+    /**
+     * Confirms every segment that is still an unconfirmed proposal.
+     *
+     * Clicking the 'save and finish' button implicitly accepts all remaining proposals. For each one the
+     * segmentMark is set to confirmed, which assigns the pmId that the drafts-list ProseMirror logic
+     * later relies on, and its store status is updated to 'confirmed'. Confirming a mark does not
+     * change document positions, so the ranges read once up front stay valid across the loop
+     */
+    confirmAllUnconfirmedSegments () {
+      const ranges = this.prosemirror.keyAccess.rangeTrackerKey.getState(this.prosemirror.view.state)
+      const confirmedSegments = []
+
+      this.ignoreProsemirrorUpdates = true
+
+      this.segments.forEach(segment => {
+        if (segment.status === 'confirmed') {
+          return
+        }
+
+        const range = ranges[segment.id]
+
+        /**
+         * A proposal without a mark in the document (e.g. a broken draft) cannot be confirmed in
+         * ProseMirror, so it is skipped rather than throwing on the missing range
+         */
+        if (!range) {
+          return
+        }
+
+        setRange(this.prosemirror.view)(range.from, range.to, { segmentId: segment.id, isConfirmed: true })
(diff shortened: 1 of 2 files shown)


client/js/components/statement/splitStatement/SplitStatementView.vue
```diff
@@ -776,6 +776,46 @@ export default {
...
+        if (!range) {
+          return
+        }
+
+        setRange(this.prosemirror.view)(range.from, range.to, { segmentId: segment.id, isConfirmed: true })
+        confirmedSegments.push({ ...segment, status: 'confirmed' })
+      })
+
+      this.ignoreProsemirrorUpdates = false
+
```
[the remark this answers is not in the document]
hwiem: d46821315fe5f442ac5aeb38a447ede5b863e649
Restoring the previous value seems overengineered; when calling `confirmAllUnconfirmedSegments`, `ignoreProsemirrorUpdates` should always be `false`

client/js/components/statement/splitStatement/SplitStatementView.vue
```diff
@@ -776,6 +776,48 @@ export default {
...
+          /**
+           * A proposal without a mark in the document (e.g. a broken draft) cannot be confirmed in
+           * ProseMirror, so it is skipped rather than throwing on the missing range
+           */
+          if (!range) {
+            return
+          }
+
+          setRange(this.prosemirror.view)(range.from, range.to, { segmentId: segment.id, isConfirmed: true })
```
huellnerdemos, line 809: I'm not completely sure about this, but i think maybe it is possible that segments gets confirmed for the document inside the for loop, when the range is correct. But lets say there is a segment following, that has a range, but it is overlapping with another one or something. In this case  replaceRange, that is called by setRange, throws an error. setRange catches this error and thorws it up to the try block in confirmAllUnconfirmedSegments from there it goes to saveAndFinish. The problem at this point could be, that we get some confirmed segments in the document, but the `if`-block that follows the `try` is never reached and the state is possibly not saved into the store, because the error interrupted the saving to the store. Put the
`      if (confirmedSegments.length) {
        this.locallyUpdateSegments(confirmedSegments)
      }`
into the `finally` could solve such situation and keep the store consistent with the so far processed segments in the document.

huellnerdemos (commented):
Found something, i am not sure about, but it could be important. We can discuss this in the next days if you want. :shipit:

client/js/components/statement/splitStatement/SplitStatementView.vue
hwiem: I looked into this and I think that it's a rather unlikely edge case - we would have to receive nested segments from the backend, because in the UI they can't be created. Since this is a hotfix and we haven't run into this issue (afaik), I'd prefer to leave it as is for now.

huellnerdemos (approved):
Code looks good. Thank you!
Tested the segmentation of a statement and it worked. 🎉

PR #6612, opened 2026-07-24
fix DS-589: strip read-only relationships from segment update payload

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/tickets/DS-589

Description
Changing a segment's assignee (for example when forwarding a segment for technical review) could fail with a JSON:API validation error and the change was rejected:

[data][relationships][comments]: Unexpected field(s) in the context of the list of relationships, allowed fields are: parentStatement,assignee, place, tags

On the StatementSegment resource, `comments` and `recommendationVersions` are read-only relationships — they are managed through their own resources and can never be written through a segment update. The backend rejects any
relationship in an update payload that is not updatable, so as soon as one of them reached the PATCH request the whole update failed and the assignee could not be changed.

The component already tried to drop these fields (`excludeComments()` / `excludeRecommendationVersion()`), but it deleted
them from a detached copy of the relationships object. The `save` action of vuex-json-api builds the request body by diffing the *stored* item, so the stripping had no effect. Whether the fields ended up in the payload depended entirely on the store state (the diff baseline vs. the current item's comments), which is why the problem was intermittent and only showed up in some environments.

This PR removes the read-only relationships in `updateRelationships()`, before the item is written to the store, so they can never enter the diff or the request body. Comments are still restored to the store after saving, so the UI keeps displaying them.

### How to review/test
1. Open a segment that has comments in the segment view.
2. Change its assignee (e.g. forward it for technical review) and save.
3. The assignee change is persisted without a validation error, and the comments remain visible on the segment.

The reliable trigger for the old behaviour is a diverging store state:
the segment's `comments` in the store differing from the diff baseline (e.g. after the comments are (re)loaded into the item but not the baseline). The fix makes the update robust regardless of that state.

### DIFF
--- a/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
+++ b/client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
@@ -989,27 +989,6 @@ export default {
       }
     },
 
-    /**
-     * Remove non-updatable comments from segments relationships for update request
-     * @param relations {Object}
-     */
-    excludeComments (relations) {
-      if (relations.comments) {
-        this.setProperty({ prop: 'isLoading', val: true })
-        delete relations.comments
-      }
-    },
-
-    /**
-     * Remove non-updatable recommendationVersions from segments relationships for update request
-     * @param relations {Object}
-     */
-    excludeRecommendationVersion (relations) {
-      if (relations.recommendationVersions) {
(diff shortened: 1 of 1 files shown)


hwiem (approved):
I made a few adjustments so the comments don't disappear from the interface when saving - now it looks good to me, thank you!

PR #6615, opened 2026-07-24
chore: bump brace-expansion and js-yaml to patched versions

hwiem:
fixes these security alerts:
- brace-expansion: DoS via exponential-time expansion of consecutive non-expanding {} groups (high)
- js-yaml: YAML merge-key chains can force quadratic CPU consumption (high)


huellnerdemos (approved):
Looks good, as far as i can say. Thank you! 🎉

PR #6621, opened 2026-07-27
chore: dedupe transitive postcss to patched 8.5.23

hwiem:
fixes a high security alert:
PostCSS: Path Traversal in Previous Source Map Auto-Loading (sourceMappingURL) leads to Arbitrary .map File Disclosure

Replaces https://github.com/demos-europe/demosplan-core/pull/6619


huellnerdemos (approved):
Looks good, thank you!

PR #6623, opened 2026-07-27
feat(DPLAN-17488): segment deadline reminder emails

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17488/E-Mail-Benachrichtigung-fur-Wiedervorlage

### Description
Add a daily cron that reminds each assignee about their segments due 
 - in one week 
 - and on the deadline day
, bundled into one mail per assignee and deadline;
past deadlines are never matched. 
Show the deadline in the existing assignment digest mail as well.

Add a per-user opt-out (default on, SEGMENT_DEADLINE_REMINDER_ENABLED)
surfaced as a settings checkbox. Gate the feature with two new permissions
(companions to the pre-existing procedure-scoped field_statement_deadline):
- feature_statement_deadline_mail — global backend/cron gate (evaluable by
  the cron's anonymous user)
- feature_statement_deadline_mail_setting — role-bound, gates the settings
  opt-out checkbox

### How To Test
Requires a project with the two permissions enabled.

1. Give an assignee a segment whose deadline is today or today + 7 days,
     assignee must have the reminder enabled — default on
2. Generate the reminder mails:
     `dplan:scheduler:trigger send-segment-deadline-reminder-emails`
3. Flush the mail queue to Mailpit:
     `dplan:scheduler:trigger send-emails`
   → one bundled mail per (assignee, deadline).

The opt-out checkbox appears under user settings (/portal/user) for
planner and hearing-authority roles; unchecking it stops the mails.

Teh links within the mail should work if these parameters are set: 
image

image


client/js/components/user/portalUser/PersonalData.vue
```diff
@@ -85,6 +85,23 @@
       </p>
     </template>
 
+    <template v-if="hasPermission('feature_statement_deadline_mail_setting')">
+      <dp-checkbox
+        id="segmentDeadlineReminderEnabled"
+        v-model="isSegmentDeadlineReminderChecked"
+        class="u-mb-0_25"
```
hwiem: nitpick: We should use tailwind classes when possible
```suggestion
        class="mb-1"
```

client/js/components/user/portalUser/PersonalData.vue
```diff
@@ -85,6 +85,23 @@
...
+        :label="{
+          bold: true,
+          text: Translator.trans('email.segment.deadline.reminder.subscribe')
+        }"
+        name="segmentDeadlineReminderEnabled"
+        value-to-send="on"
+      />
+      <p>
```
hwiem: To make this more accessible (and in my opinion, it would also look better), instead of displaying this text as a paragraph below the checkbox, we could pass it as a hint to `DpLabel` (via the `DpCheckbox` `label` prop) - but then we would have to do the same for the `assignedTaskNotification` checkbox, so I consider it an optional improvement:
```
:label="{
  bold: true,
  hint: Translator.trans('email.segment.deadline.reminder.explanation')
  text: Translator.trans('email.segment.deadline.reminder.subscribe')
}"
```

hwiem (approved):
FE code lgtm, just 2 small nitpicks

client/js/components/user/portalUser/PersonalData.vue
MoritzMandler: thx, done in https://github.com/demos-europe/demosplan-core/pull/6623/commits/ebfdbfae1ac235543a8e239cca9fd9026d1eb1b1

sierrazamudiodemos:
holii, thx for doing this, i a curious if this can happen (if it is a bug)
Editing any unrelated profile field (e.g. last name) and saving triggers `DemosPlanUserController.php:538-539`, which writes false for `segmentDeadlineReminderEnabled` whenever that checkbox of the email is absent from the form — even though you never touched the checkbox. This silently overwrites the intended `"on by default"` state with a persisted false in the database.

kijewskidemos (approved):
looks good to me and email notification worked. the bug cata mentioned i could not reproduce.

PR #6637, opened 2026-07-29
feat(DPLAN-18226): Add StatementSegment api platform only for GET and…

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-18226/Segment-Resource-Eingang-Nr-edt-to-API-Platfrom-Ressource

**Description**
- Enable sorting by eingangsnummer in Abschnitte table
- That said, the main endpoint that load that abschnitte table had to be migrated to 3.0

**What this PR does**

- Adds a new backend API endpoint for fetching segments, replacing the old one.
- Adds permission checks so this new endpoint only returns segments from procedures the current user is allowed to see (same rules as before, just moved over: AccessChecker decides who can see what, SegmentDoctrineAccessExtension applies that as a filter on the database query, SegmentRepository got a small helper to support it).
- Adds more fields to the Statement data (submitter name, address, dates, status, etc.) and makes sure the API actually fills them in — this is what the segment list needs to show its columns.
- Updates the segment list page in the frontend to use the new endpoint. Paging (10/25/50/100 per page) and sorting work the same as before.
- Small unrelated fix: a tooltip no longer breaks if it can't find a segment's assignee or place in the loaded data.

**How to test**

- Open a procedure's segment list.
- Check you only see segments from procedures you have access to.
- Try changing pages and sorting.
- Select a few segments and copy them to clipboard.


demosplan/DemosPlanCoreBundle/StateProvider/StatementStateProvider.php
```diff
@@ -56,10 +56,7 @@ private function provideSingle(string $id): ?StatementResource
             return null;
         }
 
-        $statementResource = new StatementResource();
-        $statementResource->id = $statement->getId();
-
-        return $statementResource;
+        return StatementResource::fromEntity($statement, $this->statementService->getProcessingStatus($statement));
```
mussbach, line 59: `StatementService::getStatement()` is a bare `statementRepository->get($id)` with no procedure/customer scoping. Until now the response body was just `id`, so that didn't matter. With `fromEntity()` it returns submitter name, street, house number, postal code, city, memo, internId and dates — so anyone with `area_admin_assessmenttable`, `feature_json_api_statement`, `area_admin_consultations` or `area_search_submitter_in_procedures` in any procedure can read any statement in the instance by id.

`StatementResourceType::getAccessConditions()` covers this on the EDT side. Could this provider do what the segment side does — build access conditions and pass them into a `getEntityByIdentifier($id, $conditions, ['id'])` lookup?

demosplan/DemosPlanCoreBundle/ApiResources/StatementResource.php
```diff
@@ -40,4 +44,61 @@ class StatementResource
...
+    public ?string $status = null;
+
+    public static function fromEntity(StatementEntity $statement, ?string $status = null): self
+    {
+        $resource = new self();
+        $resource->id = $statement->getId();
+        $resource->externId = $statement->getExternId();
+        $resource->internId = $statement->getInternId();
+        $resource->isSubmittedByCitizen = $statement->isSubmittedByCitizen();
+        $resource->authorName = $statement->getAuthorName();
```
mussbach, line 88: These are filled unconditionally, but the EDT type gates them:

- `authorName` / `submitName` — readable only under `hasAssessmentPermission()` (`StatementResourceType.php:317-318`)
- `memo` and `initialOrganisation{Street,HouseNumber,PostalCode,City,DepartmentName}` — no `readable()` in `StatementResourceType` at all, only `updatable()` under `area_admin_statement_list` / `area_admin_consultations`

Since the segment endpoint only needs `feature_json_api_statement_segment | area_admin_statement_list | feature_statements_import_excel`, this widens the audience for all of it. Worth double-checking the memo/address row — the old frontend did request those fields from 2.0, so there may be a path I missed; if there isn't, this is new exposure.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Resource.php
```diff
@@ -0,0 +1,131 @@
...
+    #[ApiProperty(readable: true, writable: false)]
+    public int $orderInProcedure = 0;
+
+    #[ApiProperty(readable: true, writable: false)]
+    public string $recommendation = '';
+
+    #[ApiFilter(SearchFilter::class, properties: [
+        'parentStatementOfSegment.id'            => 'exact',
+        'parentStatementOfSegment.procedure.id'  => 'exact',
+    ])]
```
mussbach, line 97: `SegmentsList.vue` sends five sort keys, but only two are declared here. `order[parentStatementOfSegment.externId]` and `order[orderInProcedure]` are dropped — `OrderFilter::filterProperty()` returns early for undeclared properties without erroring.

So the baseline order degrades from `submitDate, externId, orderInProcedure` to `submit` alone, which isn't unique. Combined with pagination that means rows can repeat or disappear between pages. Adding `orderInProcedure` and `parentStatementOfSegment.externId` should restore it.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Provider.php
```diff
@@ -0,0 +1,160 @@
...
+            ? iterator_to_array($result)
+            : $result;
+
+        $currentVersionNumbers = $this->currentUser->hasPermission('feature_enable_recommendation_versions')
+            ? $this->recommendationVersionService->getCurrentVersionNumbersForSegments($segments)
+            : [];
+
+        $map = fn (Segment $segment): StatementSegmentResource => StatementSegmentResource::fromEntity(
+            $segment,
+            $this->statementService->getProcessingStatus($segment->getParentStatementOfSegment()),
```
mussbach, line 128: `getProcessingStatus()` per segment lazy-loads the parent statement, then its whole `segmentsOfStatement` collection, then each sibling segment's `place`. On top of that `Resource::fromEntity()` walks place→procedure, tags→topic/boilerplate/defaultAssignee→orga, assignee→orga and statement→meta for every row, unconditionally — the old endpoint used `fields[…]`/`include` so only requested columns were hydrated.

Same batching trick as `getCurrentVersionNumbersForSegments()` would work here: group the segments by parent statement id and resolve the status in one query.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Resource.php
```diff
@@ -0,0 +1,131 @@
...
+use demosplan\DemosPlanCoreBundle\ApiResources\StatementResource;
+use demosplan\DemosPlanCoreBundle\Entity\Statement\Segment as SegmentEntity;
+use demosplan\DemosPlanCoreBundle\Entity\Statement\Tag as TagEntity;
+
+#[ApiResource(
+    shortName: 'StatementSegment',
+    operations: [
+        new GetCollection(
+            uriTemplate: '/StatementSegment',
+            paginationEnabled: false,
```
mussbach, line 39: With pagination off by default, a request without `pagination=true` returns every segment of the procedure with the full nested hydration from `fromEntity()`. The frontend's "fetch all segments for bulk edit" path goes down exactly that road. Enabling pagination by default and letting clients opt out would bound it.

mussbach (commented):
Backend review. The migration follows the Place/Tag precedent well and the batched `getVersionCountsForStatementIds` lookup is a nice touch. Inline comments cover the rest; two things that have no line to hang off:

**Tests** — a new access-control extension, access checker and provider land with no coverage. Minimum I'd want: a functional test that a user scoped to procedure A gets no segments of procedure B from `GET /3.0/StatementSegment`, plus one covering `allowedSegmentAccessProcedures`.

**Search consistency** — the page query now searches via a SQL `ipartial` LIKE on `text` only, while `fetchSegmentIds()` still uses the 2.0 search with `fieldsToSearch`. With a search term and non-default search fields, `allItemsCount` and the bulk-select id set can disagree with the rows actually displayed.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/AccessChecker.php
```diff
@@ -0,0 +1,70 @@
...
+    public function __construct(
+        private readonly CurrentUserInterface $currentUser,
+        private readonly CurrentProcedureService $currentProcedureService,
+        private readonly ProcedureAccessEvaluator $procedureAccessEvaluator,
+        private readonly DqlConditionFactory $conditionFactory,
+    ) {
+    }
+
+    /**
+     * Mirrors StatementSegmentResourceType::isAvailable().
```
mussbach, line 33: This is a verbatim copy of `StatementSegmentResourceType.php:83-121`. The Place and Tag migrations made the ResourceType delegate to the AccessChecker instead of leaving both copies around — same here would avoid the drift.

It also carries over the unreachable `[] === $procedureIds` branch in `getAccessConditions()`: `$procedureIds[] = $procedureId` on the line above guarantees it's non-empty.

config/services.yml
```diff
@@ -14,6 +14,10 @@ services:
         tags:
             - name: dplan.json_api_transformer
 
+    demosplan\DemosPlanCoreBundle\Api\StatementSegment\Extension\SegmentDoctrineAccessExtension:
```
mussbach, line 17: `_defaults: autoconfigure: true` is set and API Platform autoconfigures `QueryCollectionExtensionInterface`. `TagDoctrineAccessExtension` isn't registered anywhere and works, so this looks redundant — unless segments differ for a reason worth a comment.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Resource.php
```diff
@@ -0,0 +1,131 @@
...
+        $resource->recommendation = $segment->getRecommendation();
+        $resource->parentStatement = StatementResource::fromEntity($segment->getParentStatementOfSegment(), $parentStatementStatus);
+        $assignee = $segment->getAssignee();
+        $resource->assignee = null !== $assignee ? AssignableUserResource::fromEntity($assignee) : null;
+        $resource->place = PlaceResource::fromEntity($segment->getPlace());
+        $resource->tags = array_values($segment->getTags()->map(
+            static fn (TagEntity $tag): TagResource => TagResource::fromEntity($tag)
+        )->toArray());
+        $resource->deadline = $segment->getDeadline()?->format('Y-m-d');
+        $resource->customFields = $segment->getCustomFields()?->toJson();
```
mussbach, line 150: `customFields` is gated behind `field_segments_custom_fields` in the EDT type (`StatementSegmentResourceType.php:197`) but is always set here.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Resource.php
```diff
@@ -0,0 +1,131 @@
...
+        $resource->id = $segment->getId();
+        $resource->text = $segment->getText();
+        $resource->externId = $segment->getExternId();
+        $resource->internId = $segment->getInternId();
+        $resource->orderInProcedure = $segment->getOrderInProcedure();
+        $resource->recommendation = $segment->getRecommendation();
+        $resource->parentStatement = StatementResource::fromEntity($segment->getParentStatementOfSegment(), $parentStatementStatus);
+        $assignee = $segment->getAssignee();
+        $resource->assignee = null !== $assignee ? AssignableUserResource::fromEntity($assignee) : null;
+        $resource->place = PlaceResource::fromEntity($segment->getPlace());
```
mussbach, line 145: `Segment::getPlace()` returns `PlaceInterface`, `PlaceResource::fromEntity()` takes the concrete `Place`. PHPStan will likely flag this.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Resource.php
```diff
@@ -0,0 +1,131 @@
...
+
+    #[ApiProperty(readable: true, writable: false)]
+    public ?string $internId = null;
+
+    #[ApiProperty(readable: true, writable: false)]
+    public int $orderInProcedure = 0;
+
+    #[ApiProperty(readable: true, writable: false)]
+    public string $recommendation = '';
+
```
mussbach, line 84: The property is `parentStatement` but every filter and sort key is `parentStatementOfSegment.…`, so clients read one name and filter by another — and the generated OpenAPI/JSON:API docs will show the mismatch. If it's unavoidable because of `DoctrineOptions(entityClass: Segment::class)`, a short comment would help the next reader.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Provider.php
```diff
@@ -0,0 +1,160 @@
...
+
+        return array_map($map, $segments);
+    }
+
+    /**
+     * Because this resource supports sorting, API Platform stops forwarding plain
+     * `page`/`itemsPerPage`/`pagination` query params on its own, so we read them
+     * from the URL ourselves and add them to `$context['filters']`, where API
+     * Platform expects to find them.
+     */
```
mussbach, line 144: `order`, `text` and `exists` clearly do reach the provider via `$context['filters']`, so "because this resource supports sorting, API Platform stops forwarding page/itemsPerPage" probably isn't the real cause. Worth pinning down, because the method silently no-ops when `$context['request']` is absent — pagination would then be ignored rather than fail.

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -801,9 +735,11 @@ export default {
       selectedSort: '',
       selectionCopiedToClipboard: false,
       sortOptions: [
-        { value: '-deadline', label: Translator.trans('sort.deadline.descending') },
-        { value: 'deadline', label: Translator.trans('sort.deadline.ascending') },
+        { value: 'internId-desc', label: Translator.trans('sort.internId.descending') },
+        { value: 'internId-asc', label: Translator.trans('sort.internId.ascending') },
       ],
+      v3Segments: [],
+      v3Included: {},
```
hwiem: Why should we include `v3` in the name here?

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1127,43 +1015,81 @@ export default {
...
+        pagination: 'true',
+        page,
+        itemsPerPage: this.pagination.perPage,
+      })
+
+      if (this.searchTerm !== '') {
+        params.set('text', this.searchTerm)
       }
 
+      if ('internId-desc' === this.selectedSort) {
```
hwiem: nitpick: no yoda
```suggestion
      if (this.selectedSort === 'internId-desc') {
```

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1127,43 +1015,81 @@ export default {
...
+        params.append('order[parentStatementOfSegment.original.internId]', 'desc')
+      } else if ('internId-asc' === this.selectedSort) {
+        params.append('order[parentStatementOfSegment.original.internId]', 'asc')
+      } else {
+        // Baseline order so the list stays stable when no sort is selected.
```
hwiem: Is this needed? It's not possible to select 'no sort'. Or should that option be added to the select?

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1127,43 +1015,81 @@ export default {
...
+        params.append('order[parentStatementOfSegment.submit]', 'asc')
+        params.append('order[parentStatementOfSegment.externId]', 'asc')
+        params.append('order[orderInProcedure]', 'asc')
+      }
+
+      /*
+       * Translate the filter-flyout selections (place/assignee/tags) into v3 SearchFilter/ExistsFilter params.
+       * Multiple values on the same path[] naturally OR together; different paths naturally AND together.
+       */
+      Object.values(this.getFilterQuery).forEach((entry) => {
```
hwiem: nitpick
```suggestion
      Object.values(this.getFilterQuery).forEach(entry => {
```

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1127,43 +1015,81 @@ export default {
...
+      } else {
+        // Baseline order so the list stays stable when no sort is selected.
+        params.append('order[parentStatementOfSegment.submit]', 'asc')
+        params.append('order[parentStatementOfSegment.externId]', 'asc')
+        params.append('order[orderInProcedure]', 'asc')
+      }
+
+      /*
+       * Translate the filter-flyout selections (place/assignee/tags) into v3 SearchFilter/ExistsFilter params.
+       * Multiple values on the same path[] naturally OR together; different paths naturally AND together.
```
hwiem: Is this true for tags?

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1127,43 +1015,81 @@ export default {
...
+      const params = new URLSearchParams({
+        'parentStatementOfSegment.procedure.id': this.procedureId,
+        include: 'parentStatement,assignee,place,tags',
+        pagination: 'true',
+        page,
+        itemsPerPage: this.pagination.perPage,
+      })
+
+      if (this.searchTerm !== '') {
+        params.set('text', this.searchTerm)
```
hwiem: If I understand correctly (and I confirmed in the interface), search now only works for the segment text, while in the search modal, you can select a lot of different fields to search

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1093,10 +983,6 @@ export default {
       fetchPlaces: 'list',
     }),
 
-    ...mapActions('StatementSegment', {
```
hwiem, line 1056: Can't we keep using the vuex json api library?

hwiem (changes requested):
I have 2 critical concerns: 
- search is not working as before
- we should keep using vuex json api library instead of writing all the code by hand

demosplan/DemosPlanCoreBundle/ApiResources/StatementResource.php
sierrazamudiodemos: good point

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Resource.php
winderdemos: Fixed - added orderInProcedure and parentStatementOfSegment.externId as OrderFilter properties, plus deadline as a 5th sortable field. Baseline sort is parentStatementOfSegment.submit, parentStatementOfSegment.externId, orderInProcedure (compounds into one multi-column ORDER BY when all three are sent).

We also introduced a tiebreaker - every collection query appends ORDER BY id ASC after whatever the client requests, so pagination stays stable even if a future sort selection turns out not to be unique.

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Resource.php
winderdemos: Good point; I've switched this to true.

demosplan/DemosPlanCoreBundle/StateProvider/StatementStateProvider.php
winderdemos: Currently looking into this high priority concern

demosplan/DemosPlanCoreBundle/Api/StatementSegment/Provider.php
winderdemos: Good catch - batched it. Added SegmentRepository::getSegmentSolvedCountsForStatementIds() (one grouped query) and StatementService::getProcessingStatuses(), mirroring getCurrentVersionNumbersForSegments(). provideCollection() now dedupes segments by parent statement id and resolves status once per page instead of once per row.

config/services.yml
winderdemos: Good catch, and it's worse than it looked - traced it to double-tagging: the bundle-wide resource glob already autoconfigures QueryCollectionExtensionInterface at default priority, so the explicit registration here was adding a second tag, causing applyToCollection() to run twice per request. Removed the explicit entry for this one and SegmentTextSearchExtension; SegmentDeterministicOrderExtension needs autoconfigure: false since it requires a non-default priority. Verified via debug:container - each extension now shows exactly one tag.

PR #6650, opened 2026-08-03
feat: register addon API Platform resources for discovery

breierDemos:
### Ticket
No ticket

Addons had no way to expose ApiPlatform `#[ApiResource]` classes: `api_platform.mapping.paths` only scanned two fixed core directories (`DemosPlanCoreBundle/ApiResources` and `DemosPlanCoreBundle/Api`), so any addon-defined resource attribute was silently never discovered and its routes never registered.

Adds `AddonApiResourceMappingPass`, a compiler pass that appends each installed addon's `src/Api` directory (if present) to `api_platform.resource_class_directories`, mirroring the existing `AddonDoctrineMigrationsPass` pattern for migrations directories.

### How to review/test
- `bin/<project> cache:clear` and `bin/<project> debug:router` should now list routes for any addon resource under `src/Api` (e.g. `_api_/3.0/DcatApPluStandardCode_get_collection`)

### PR Checklist
- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`


demosplan/DemosPlanCoreBundle/Application/DemosPlanKernel.php
```diff
@@ -280,6 +281,7 @@ protected function build(ContainerBuilder $container): void
         $container->addCompilerPass(new OptionsLoaderPass(), PassConfig::TYPE_AFTER_REMOVING, 0);
         $container->addCompilerPass(new VirusCheckPass(), PassConfig::TYPE_BEFORE_OPTIMIZATION, 0);
         $container->addCompilerPass(new AddonResolveTargetEntity(), PassConfig::TYPE_BEFORE_OPTIMIZATION, 1000);
+        $container->addCompilerPass(new AddonApiResourceMappingPass(), PassConfig::TYPE_BEFORE_OPTIMIZATION, 0);
```
graupnerdemos, line 284: This should be moved into the block below, after the doctrine pass.

graupnerdemos (changes requested):
One minor change request. Mostly to make sure all subsystem already know about addons when this runs and to not fail core tests if this fails, because available addons are unrelated to core testing. Otherwise, I like this solution.

demosplan/DemosPlanCoreBundle/Application/DemosPlanKernel.php
breierDemos: Moved in 43218f0 — now runs alongside the other addon passes in the test-excluded block, after AddonDoctrineMigrationsPass.

graupnerdemos (approved):

PR #6654, opened 2026-08-04
chore: bump transitive brace-expansion to 2.1.4 and 5.0.9

hwiem:
closes these security alerts:
- DoS via unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation
- DoS via unbounded expansion length causing an out-of-memory process crash


huellnerdemos (approved):
Looks good, thank you! 🎉

PR #6655, opened 2026-08-04
chore(refs DPLAN-16770): allow typescript in test files

hwiem:
### Ticket
DPLAN-16770

Test files could be written in TypeScript but nothing was type-safe: no typecheck ran anywhere, and two things broke as soon as a spec was actually typed.

- Add type to `shallowMountWithGlobalMocks` so `wrapper.vm.<prop>` is resolvable in test files
- Require the `.vue` extension on imports in `.ts` files only; TypeScript cannot resolve an extensionless import to a `.vue` file, and `.js` files keep the existing rule
- Add `typecheck` script running `vue-tsc`, so `<script lang="ts">` blocks are checked; plain `tsc` skips `.vue` files entirely
- Declare globals on `ComponentCustomProperties` so they are resolved in vue templates (template expressions are checked against the component instance, not script scope)
- Add missing properties to the `dplan` object in the test setup
- Migrate StatementExportModal spec to typescript as POC

### How to review/test
- `yarn typecheck` succeeds
- `yarn test` is unchanged from the base branch
- A `*.spec.ts` mounting via `shallowMountWithGlobalMocks` should typecheck only with the `.vue` extension on the component import


elmasdemos (approved):
Nice, lgtm !!

PR #6660, opened 2026-08-05
feat(refs DPLAN-18003, #AB55977): Add drag-and-drop sorting for procedure phases

sierrazamudiodemos:
PR: Add drag-and-drop sorting for procedure phases
https://demoseurope.youtrack.cloud/issue/DPLAN-18003/ADO-55977-Verfahrensschritte-sortieren

What this does: Lets a tenant admin drag-and-drop reorder the "Verfahrensschritte" (procedure phases) list, and saves the new order.

**Files changed:**

- ProcedurePhasesDefinition.vue — Turns on drag-and-drop in the phases table and sends the new order to the backend when a row is dropped.
- ProcedurePhaseDefinition.php (entity) — Teaches the phase entity how to report and update its own position, so it can plug into our existing generic reordering tool.
- RpcProcedurePhaseDefinitionListReorder.php — New backend endpoint the frontend calls when a row is dropped; checks permissions and loads the right phases, then asks the reorderer to do the work.
- ProcedurePhaseDefinitionAudienceReorderer.php — New helper class that does the actual position math (moving one phase and shifting the others); kept separate so it's easy to test on its own.
- ProcedurePhaseDefinitionResourceType.php — Small addition so our tools (PHPStan) understand two fields the new endpoint reads from this class.
- rpc-procedurePhaseDefinition-list-reorder-schema.json — Defines what a valid "reorder" request looks like, so bad requests get rejected early.
- messages+intl-icu.de.yml — Adds the German error text shown if someone tries to move the phase that must always stay first.
- ProcedurePhaseDefinitionAudienceReordererTest.php — Tests that prove the reordering math works correctly (moving up, down, front, back) and never breaks the fixed first phase.

**Linked PRs**
demosplan-UI: https://github.com/demos-europe/demosplan-ui/pull/1552

**Tasks**
- [x] merge the UI PR first
- [x] release new demosplan-ui version
- [ ] merge this PR

**How to test this**

1. Log in as Mandanten admin
2. Go to Verfahrensschritte definieren in the admin menu.
3. Check the list shows your procedure phases, with "Konfiguration" always first.
4. Grab a row (not the first one) by its drag handle and drop it in a new position.
5. Confirm the row moves to the new spot and you see a "saved" success message.
6. Try dragging a row across tables --> shouldn't be possible, the item should fall back into its original position.
7. Reload the page and confirm the new order is still there (it was really saved, not just visual).
8. Konfigurationsschritt shoudn't have the drag handle, but instead a tooltip info that is cannot be moved.
9. Try dropping another row into the "Konfiguration" row — it lands on the 2nd position (the last allowed index when dragging --> consistent with the segments list column-drag, where the 1st column is locked)
10. Repeat steps 4–6 in the other audience group (Internal vs. External) to check both lists work independently.
11. Open a procedure's Grundeinstellungen (basic settings) and check the phase dropdown now shows phases in your new order.


rafelddemos (commented):
I read Cata's FE changes, but since I made adjustments, it would be good if sby else had a look at those (and the demosplan-ui PR) &#x1F64F;

hwiem (approved):
FE lgtm!

breierDemos (approved):
lgtm

PR #6666, opened 2026-08-07
refactor(DPLAN-17722): rename UserFilterSet to Bookmark

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-17722/Speicherung-von-Favoritenansicht-in-der-Abschnittsliste

### Description
rename UserFilterSet to Bookmark
The entity is a named, per-user, per-procedure pointer at a stored query.
That query now also carries column selection, order and sort, so "filter set"
described less than the thing does.

Entity, repository, service and resource type renamed, plus the table
user_filter_set -> bookmark. Index names are hash-derived from the table name
and are renamed with it; foreign keys keep their names, since Doctrine compares
their definitions rather than their names and dropping them would rewrite
definitions that differ between
user_filter_set -> bookmark. Inom the table name
and are renamed with it; foreign keys keep their names, since Doctrine compares
their definitions rather than their names and dropping them would rewrite
definitions that differ between

Unchanged on purpose, all fronte identifier
returned by getName(), the perm_filter_sets, the
route dplan_api_procedure_delete_statement_filter, the savedFilterSet
translation keys, and the $filtilter_set_id column.

Also included:
- BookmarkRepository::updateObjss
- MapHandler: assign getRepositupdateObject()
  resolves to MapRepository insg class
- AssessmentHandler::saveBookmarface instead of
  the concrete User entity, matce::getUser() returns

Requires demosplan-addon with Bookmark().

### related prs:
will rely on this version when released: https://github.com/demos-europe/demosplan-addon/pull/183

### HowToTest:
run the migration
Try out existing implementations of stored bookmarks aka userFilter - should work as before without any visible change.

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


breierDemos (approved):
lgtm

PR #6674, opened 2026-08-10
feat(DPLAN-18275): add tooltip explaining automatic assignee assignment

riechedemos:
### Ticket
DPLAN-18275

The assignee field in the split-statement sidebar gives no indication that an assignee
can be filled in automatically. When a tag with a linked default assignee is added to a
segment, that user is silently set as the segment's assignee, which is surprising if you
do not know the feature exists.

This adds an explanatory tooltip to the "Bearbeiter*in" label. The tooltip is bound to
`feature_tag_default_assignee`, so it only shows where automatic assignment is actually
available and stays hidden everywhere else.

In the process, the two plain `<label>` elements for "Schritt" and "Bearbeiter*in" were
replaced by `dp-label`, which is what the rest of the sidebar already uses and what
provides the tooltip slot. Both keep `mb-0` to reset the global `label` margin from
`objects/_form.scss` and preserve the previous spacing.

### How to review/test
1. Open a procedure and start segmenting a statement (split statement view).
2. Expand the "Schritt und Zuweisung" section in the sidebar.
3. With `feature_tag_default_assignee` enabled, a help icon appears next to
   "Bearbeiter*in" explaining the automatic assignment; without the permission the
   label renders unchanged.
4. Check that the spacing between the labels and their selects is unchanged.

### PR Checklist

- [x] Run `yarn lint`


rafelddemos (approved):
lgtm (code review only)

PR #6675, opened 2026-08-10
fix(DPLAN-18254): align profile page heading with content width

riechedemos:
### Ticket
DPLAN-18254

The heading "Meine Daten" and its subheading in the profile page header were indented by 25% while the content below rendered full width.

`portal_user.html.twig` only passed `full_width: true`, which affects the content section. The pageheader has its own flag, `full_width_pageheader`, which defaulted to `false` — so the header rendered an empty `u-1-of-4` spacer column and pushed the heading into `u-3-of-4`. Passing `full_width_pageheader: true` drops the spacer and makes the heading column full width.

### How to review/test
Open `/portal/user` (Meine Daten). The heading and subheading should now be left-aligned with the form fields below.

### PR Checklist


rafelddemos (approved):
lgtm, 'Meine daten' is now aligned correctly, thx :)

PR #6658, opened 2026-08-05
feat(refs: DPLAN-18152): add copy segment selection to clipboard

huellnerdemos:
### Ticket
DPLAN-18152

This PR adds the possibility to copy items from the segments list to the clipboard, to paste them into Excel.

### How to review/test

1. Login as admin and go to a procedure with segmented statements
2. Go to the segments list
3. Choose which columns should be shown in the list
4. Select at least one segment
5. The bulk edit header appears, including a button to copy to clipboard
6. Push the button and past the clipboard to an Excel table
7. Check if all selected segments are pasted
8. Check if only the columns were copied, that were chosen before
9. No headlines should be copied
10. Check if it works with selections through different pages
11. Maybe check if it works with a selection with over 1000 items

### Linked PRs
Demosplan-ui: https://github.com/demos-europe/demosplan-ui/pull/1549

### PR Checklist

- [x] Run `yarn lint`
- [x] Link all relevant tickets
- [x] Move the tickets on the board accordingly
- [x] Update changelog

PR #6582, opened 2026-07-17
Feat(refs: DPLAn-17487): Add sortable deadline column in segments list

riechedemos:
### Ticket
[DPLAN-17487](https://demoseurope.youtrack.cloud/issue/DPLAN-17487/Bearbeitungsfrist-als-Spalte-in-der-Abschnittsliste-Sortierfunktion)

Adds a sortable deadline column to the segments list. Sorting is handled entirely on the frontend (ascending/descending) and the chosen sort order is persisted in localStorage, so it survives a page reload. Pagination for this list were removed as part of this change — all segments (up to the server's 1000-item page-size cap) are now loaded in a single request instead of paged.

### How to review/test
- Add deadline to a segment in "Erwiderung verfassen -> weitere Felder bearbeiten"
- go to segments list and try sorting 
- Reload to page, change should be persisted 

### PR Checklist

- [ ] Create/Update tests
- [x] Update documentation
- [ ] Add/Update data-cy attributes ([conventions](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/twig_html.html#guideline-for-naming-cypress-hooks))
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Link all relevant tickets
- [ ] Move the tickets on the board accordingly
- [ ] Update changelog

PR #6678, opened 2026-08-10
DPLAN-18277: permission wrapping

graupnerdemos:
### Ticket
DPLAN-18277

The segment list sorting has been wrapped into `feature_segments_manualsort`, not enabled anywhere. The segment list pagination has been restored.

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [ ] Run `yarn test`
- [x] Update changelog


riechedemos (approved):
Looks good! Please rebase onto main before merging — 3bcd4ba02 touches the same
file (dp-bulk-edit-header area).

riechedemos:
Updated: everything PR #6582 changed is now behind `feature_segments_manualsort`, so with the permission off the list
behaves exactly as it did before that PR.

- Sort dropdown, client-side deadline sort and the localStorage sort restore are gated (was already the case).
- The pager is hidden when the permission is on, and the request then asks for `{ number: 1, size: 1000 }` instead of the paged `perPage`. Without this the client-side sort would only sort the currently visible page — #6582 removed pagination precisely because the sort needs the full list.
- The deadline column is now behind a `hasDeadlineColumn` computed combining `field_statement_deadline` and the new permission, replacing three separate inline checks.
- The Excel import button stays in the position #6582 gave it; it works in both states and needs no conditional.

Also merged main to pick up #6658 (copy segment selection to clipboard), which touches the same file. Two conflicts, both resolved in favour of main: the `dp-bulk-edit-header` attributes and the `fields` extraction into `buildSegmentFetchOptions()`.

riechedemos:
> Looks good! Please rebase onto main before merging — [3bcd4ba](https://github.com/demos-europe/demosplan-core/commit/3bcd4ba02788978d067a9a72aca0fc10ecaebc89) touches the same file (dp-bulk-edit-header area).

done :tada:

PR #6679, opened 2026-08-10
build(ado-56925): base production image on official php image

mussbach:
### Ticket
ado-56925

The `php-fpm` binary from `packages.sury.org` has DNS telemetry compiled in (`-DTELEMETRY_HOST=telemetry.sury.org -DTELEMETRY_PORT=53`, see `/usr/lib/php/php-common.mk` in those packages). On every master start it resolves that host and sends an encrypted TXT query **directly to it on UDP/53**, bypassing the configured resolver — so a DNS-level block upstream does not stop it. There is no opt-out switch, and `/etc/hosts` and `/etc/resolv.conf` cannot be baked into the image because the container runtime replaces both.

The production image is therefore based on the official `php:8.4.24-fpm-bookworm` image, which compiles PHP from source and carries no telemetry. The sury keyring and apt source are gone.

Notable points for review:

- Build and runtime stages share a new `php-base` stage, so the extension set is installed and cached once. `composer` now comes from the `composer:2.8.12` image.
- The official image scans a **single conf.d for every SAPI**, while CLI and FPM need different settings — most importantly `disable_functions`, which the CLI must keep unset for Symfony Process. FPM-only values moved into the pool config as `php_admin_value` (which also makes them final against `ini_set()`), and `zzz-dplan-cli.ini` merged into `zzz-dplan.ini`. `opcache.preload` stays in conf.d because it is only honoured at master startup.
- The official image ships no `php.ini`, only templates, so `php.ini-production` is copied into place — otherwise `display_errors` defaults to On.
- Two compat symlinks: `/usr/sbin/php-fpm8.4` because the k8s deployment overrides the container args with that absolute path, and `/usr/bin/php` because `client/fe/config/projectConfig.js` spawns a hardcoded interpreter path. **No chart change is required and no coordinated deploy is needed.**
- `ftp` is the only extension dropped.

### How to review/test

Build the image and compare against the previous release image. Verified on a full build:

- **No telemetry**: zero DNS packets on `php-fpm` startup, measured against a sinkhole resolver. No `deb.sury.org` apt source, no keyring package, no telemetry string in any PHP binary or extension.
- **Extension parity**: 56 vs 57 modules, the only delta being `ftp`.
- **Assets unchanged**: `public/` 42 MB, same js/css/manifest counts, `client/js/generated` contains `routes.json` and `translations.json`. nginx image 55.5 MB vs 55.4 MB.
- **Runtime parity**, probed side by side over FastCGI on both images, identical on every value: `preload[scripts=1] | mem=2G disable_fn=31 opcache_mem_used=38MB`. Matching opcache usage confirms `opcache.preload` still compiles the same amount of application code.
- **CLI stays permissive** in the same container: 6G, `exec`/`shell_exec`/`proc_open`/`system` callable.
- Starts cleanly via the exact command the chart uses (`envsubst ... && /usr/sbin/php-fpm8.4 -F`).

Note: the image grows ~250 MB because the official base ships the `$PHPIZE_DEPS` C toolchain (gcc, g++, cpp, binutils, libc6-dev ~220 MB) plus `/usr/src/php.tar.xz`. Purging it in a later layer reclaims nothing, since layers are additive. A follow-up can copy `/usr/local` into a slim runtime with only the 21 required shared libraries — measured at 202 MB vs 508 MB for the PHP base, with byte-identical module output — which would bring the final image below its current size. Kept separate so both changes stay independently revertible.

### PR Checklist


graupnerdemos (approved):
Interesting catch. I guess we should re-evaluate our usage of sury-packages elsewhere too.

PR #6632, opened 2026-07-28
Feat( ref: DPLAN-17637): Add Submitter search and possibility to delete statements

riechedemos:
### Ticket
[DPLAN-17637](https://demoseurope.youtrack.cloud/issue/DPLAN-17637/Einreichende-Suche-aus-ROBOB-in-EWM-freischalten-anpassen)

This Pr introduces a new component DpSubmitterSearch that allows to search for submitter in all procedures and display it's statements and metadata. This also add the possibility to delete those statements.

### How to review/test
- check out corresponding ewm and demosplan-ui branches
- login as Masteruser
- Click on "Einreichende suchen" in side menu ans search a username
- this should show you all statements made by this user in all procedures
- delete a statement
- it should work if the statement isn't claimed by an other user

### Linked PRs (optional)

- Other PR #{PR-1543}](https://github.com/demos-europe/demosplan-ui/pull/1543)

### PR Checklist

- [x] Create/Update tests
- [x] Update documentation
- [x] Add/Update data-cy attributes ([conventions](https://dplan-documentation.demos-europe.eu/development/guidelines-conventions/coding-styleguides/twig_html.html#guideline-for-naming-cypress-hooks))
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [x] Link all relevant tickets
- [ ] Move the tickets on the board accordingly
- [x] Update changelog

PR #6684, opened 2026-08-11
fix(refs: DPLAN-17637): fix AdminProcedure relationship resolution in cross-procedure search

breierDemos:
## Summary
- `OwnsProcedureConditionFactory::isEitherTemplateOrProcedure()` compared `Procedure::master` with `!==`, which fails once Doctrine hydrates the property as `int` instead of the `bool` the condition was built with — dropping every real procedure when the relationship is resolved in PHP.
- `isAuthorizedViaPlanningAgency()` and `userIsExplicitlyAuthorized()` checked to-many collection membership via `propertyHasStringAsMember()`, which reads the property with DIRECT access; when evaluated in PHP against an already-loaded entity this returns the raw `PersistentCollection` instead of an array, crashing `in_array()`.
- The `procedure` relationship on `AdminStatementCrossProcedureSearchResourceType` pointed at `ProcedureResourceType`, contradicting its own docblock, which requires `AdminProcedureResourceType` for consistent access conditions.
- The frontend never requested a sparse fieldset for the sideloaded `AdminProcedure` resource, so grouping headings fell back to a generic label instead of the real procedure name.

## Related PR
https://github.com/demos-europe/demosplan-core/pull/6632

## Test plan
- [x] `OwnsProcedureConditionFactoryTest` (21 tests) — new regression tests hydrate real `Procedure` entities via Doctrine and evaluate the conditions in PHP, reproducing both crashes before the fix and passing after
- [x] `AdminStatementCrossProcedureSearchResourceTypeTest` (8 tests) — green
- [x] PHPStan level 5 on changed files — no new findings

### DIFF
--- a/client/js/components/procedure/DpSearchSubmitters.vue
+++ b/client/js/components/procedure/DpSearchSubmitters.vue
@@ -216,6 +216,7 @@ const search = term => {
     },
     include: 'procedure',
     fields: {
+      AdminProcedure: 'name',
       AdminStatementCrossProcedureSearch: [
         'authoredDate',
         'authorName',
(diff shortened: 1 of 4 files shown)


MoritzMandler (approved):
lgtm

PR #6686, opened 2026-08-12
fix(refs DPLAN-18292): grant master blueprint ownership to owning org…

bakridemos:
Ticket
  DPLAN-18292

  The platform master blueprint ("Plattform-Blaupause") shows up in the procedure template list but cannot be opened: clicking it redirects back with `?status=accessdenied`, logging `AccessDeniedException` on  `area_admin_preferences` for users holding an FP role.

  `area_admin_preferences` is only ever enabled inside the `ownsProcedure()` branch, so "may open the settings page" is equivalent to "owns the procedure". With `procedure_user_restricted_access` enabled, ownership additionally requires an explicit `procedure_user` row, and `userIsExplicitlyAuthorized()` resolves an empty list to a hard `false()` — an empty list therefore means *nobody* owns the procedure rather than *everybody*.

  The master blueprint is seeded by migration rather than created through the UI, so it never receives the creator row that
  `ProcedureService::setAuthorizedUsersToProcedure()` writes for every other procedure. Its list stays permanently empty and the blueprint stays permanently unowned. Consistent with that, inheriting authorized users from a master blueprint is deliberately refused when copying it (T15644 / T23583).

  Commit 3a9895205 (ADO #44507) added the master blueprint to the list, described as *"visible and accessible to all users in that organization"*, but changed only `ProcedureService` and not the ownership layer. Visibility became org-scoped while ownership stayed user-scoped — the two diverge wherever `procedure_user_restricted_access` is on. Visibility shipped, accessibility did not; this PR is the missing half.

  **Change:** exempt the master blueprint from the explicit-authorization requirement in `OwnsProcedureConditionFactory::isAuthorizedViaOrgaOrManually()`, which now reads *owning-organisation match AND (master blueprint OR explicitly authorized)*. The organisation match and the role check are untouched, so the grant is exactly "users with an FP role in the owning organisation" — the group ADO #44507 made it visible to.

Two review notes:

- The exemption is expressed as a **condition**, not as an early `return` inside one of the factory's two shapes. `isAuthorizedViaOrgaOrManually()` is shared by `ProcedureAccessEvaluator::isOwningProcedure()` (permission checks) and
  `getOwnsProcedureConditions()` (list queries), which carry a documented "keep in sync" contract. Expressing it as a condition satisfies that contract by construction instead of relying on both branches being edited together.

- `getMasterTemplatesForUser()` and its merge block in `ProcedureService` are  **intentionally kept** and only gain a comment. The master blueprint has `customer = NULL` while the list query is restricted to the current customer
  whenever a `customer` filter key is present — which the blueprint list handler always sets. It can therefore never come from the regular query, independently of ownership, so removing the merge would drop it from the list again. The comment exists so this isn't deleted later as apparent dead weight.

**Scope of the behaviour change:** only projects with `procedure_user_restricted_access: true` are affected — the core default is `false`, where ownership is already organisation-only and nothing changes.

### How to review/test

Requires a project with `procedure_user_restricted_access: true`.

1. Pick a user with an FP role whose organisation owns the master blueprint (`master_template = 1`) and who has **no** `procedure_user` row for it:

       SELECT p._p_id, p._p_name, p.master_template, p.customer, p._o_id,
              (SELECT COUNT(*) FROM procedure_user pu
                WHERE pu.procedure_id = p._p_id) AS authorized_total
       FROM   `_procedure` p
       WHERE  p.master_template = 1;

2. Before: open *Verfahrensvorlagen* → click the master blueprint → redirect with `?status=accessdenied`.
3. After: the settings page opens. No `procedure_user` row was added.
4. Regression check: an ordinary blueprint with zero authorized users must still **not** be openable, and the master blueprint must not be openable for a user from a different organisation.

Unit tests:

    docker compose exec development su -l $USER -s /usr/bin/zsh -c 'cd /srv/www && SYMFONY_DEPRECATIONS_HELPER=disabled php -d xdebug.mode=off ./vendor/bin/phpunit tests/backend/core/Logic/OwnsProcedureConditionFactoryTest.php'

Four cases added: master blueprint without authorized users (owned), ordinary blueprint otherwise identical (not owned), master blueprint with a different organisation (not owned), plus the same check in the factory's other shape (conditions built from a `User`, evaluated against procedures).

Heads-up for reviewers: the 20 pre-existing tests in that file only assert `assertNotNull($condition)` on methods that cannot return null, so they pass with the reported behaviour fully present. The new cases evaluate the produced
conditions via `EntityFetcher`. Rewriting the existing ones is left out of scope here.

PHPStan level 5 on the touched files: 37 errors before and after — none introduced. The one hit inside `OwnsProcedureConditionFactory` (`identical.alwaysFalse`, in `hasProcedureAccessingRole()`) is pre-existing.

### PR Checklist

- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/OwnsProcedureConditionFactory.php
@@ -87,7 +87,8 @@ class OwnsProcedureConditionFactory
      *
      * If {@link GlobalConfigInterface::hasProcedureUserRestrictedAccess} is set to `true`,
      * then the user must be authorized manually for the procedure AND must be in the
-     * procedure's owning organization.
+     * procedure's owning organization. The platform master blueprint is exempt from the
+     * manual authorization part, see {@link self::isMasterTemplate()}.
      *
      * The returned condition will not apply role checks by itself. Use in conjunction with
      * {@link self::hasProcedureAccessingRole}.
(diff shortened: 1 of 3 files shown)


mussbach (commented):
Needs ProcedureService::isUserAuthorized() also to be adjusted? getAuthorizedUsers now cannot be set at the Master blueprint

bakridemos:
> Needs ProcedureService::isUserAuthorized() also to be adjusted? getAuthorizedUsers now cannot be set at the Master blueprint

i checked this many times and if i didnt miss somthing i dont think we need to adjust it, may be this small explanation can help:

1. It's not on this path. All 7 callers are statement-, segment-, fragment- or claim-scoped (bulk copy, segment/fragment assignee, claim validation, assignable-user lists). A blueprint has none of those entities, so the method is never called for one.
2. It can't regress. The fix touches neither isUserAuthorized() nor getAuthorizedUsers(). Their behaviour for the master blueprint is identical before and after — so whatever they do, this change didn't cause it.
3. It answers a different question. "May this user be assigned statements here?" ≠ "Does this user own this procedure?" Widening it would alter claim and assignee behaviour on every ordinary procedure — far bigger blast radius than the defect warrants.

And authorizedUsers can be set on the master blueprint:

- Reading the candidate list works because the settings page passes $excludeProcedureAuthorizedUsers = false (DemosPlanProcedureController.php:1434), which skips the restrictive filter entirely. All orga FP/HA members are offered even while the stored list is empty.
- Writing was blocked by the same root cause, not a separate one: feature_procedure_user_restrict_access_edit sits inside the identical ownsProcedure() branch as area_admin_preferences (core Permissions.php:551→:576; EWM :229→:361). Fix ownership, both unlock together. Verified end-to-end while testing locally.

mussbach (approved):
lgtm, thx. code review only

PR #6698, opened 2026-08-14
feat: upgrade api-platform to 4.3

bakridemos:
Description: 

Upgrades api-platform from 3.4.17 to 4.3.17.

**Why now:** api-platform 3.4.x is blocked by four security advisories (PKSA-8kfs-m8zw-ggzs, PKSA-3ncz-km6v-5vjr, PKSA-gs8r-6kz6-pp56, PKSA-gnn4-pxdg-q76m), so `composer update` fails on the old constraint.
The upgrade is not optional.

**What changed** — 3 files, no source changes:

- `composer.json` / `composer.lock` — api-platform 3.4.17 → 4.3.17, via demosplan-addon (which owns the `api-platform/core` constraint)
- `config/packages/api_platform.yaml` — two config keys that v4 removed or renamed:
  - `keep_legacy_inflector` no longer exists → removed
  - `validator.legacy_query_parameter_validation` → `validator.query_parameter_validation` (same behaviour; deprecated in 4.2, removal in 5.0)

No changes to the 8 `#[ApiResource]` classes, 5 state providers, 4 state processors, or the two EDT bridge classes. All API Platform classes we import exist unchanged in v4, and `ProviderInterface::provide()` is byte-identical.

#### Three things a reviewer will likely question

**1. Why is `symfony/type-info` 7.4 in a Symfony 6.4 lock?**

api-platform ≥4.1 requires it, and the component has no 6.x release — it was introduced in Symfony 7.1. It resolves cleanly because Flex's `PackageFilter` *exempts* symfony splits that have no release in the constrained range (`extra.symfony.require: "6.*.*"`): such a package is dropped from the splits manifest and passes through unfiltered. It is a leaf component (requires only
php, psr/container, deprecation-contracts), so there is no Symfony 7 cascade.

Please do **not** "fix" this by adding an explicit `symfony/type-info` root require or by loosening `extra.symfony.require`. Neither is needed. No Symfony upgrade is involved.

**2. Why does `api-platform/core` 4.3 not pull in split packages?**

`api-platform/core` v4.3.17 still `replaces` all 22 sub-packages, exactly like v3.4 — one package containing everything. Only the `v4.0.0` tag briefly shipped without those `replace` entries. The lock contains exactly one `api-platform/*`
entry and no splits.

Note: `composer require api-platform/core` **without a constraint** resolves to the newest stable and is the wrong command. The constraint lives in demosplan-addon.

**3. Why is `nette/utils` in the lock diff?**

Incidental transitive patch bump (4.1.4 → 4.1.5) swept in by `composer update ... -w`. Not required by api-platform 4.3, whose runtime `require` list contains no `nette/*`. The update was deliberately run with `-w` rather than `-W` to keep the diff to 4 packages instead of 8 — `-W` additionally moved `doctrine/doctrine-bundle` (2.18.2 → 2.19.0), `symfony/doctrine-bridge`, phpstan/phpdoc-parser` and `symfony/clock`, none of which are needed here.

- [X] Update changelog


composer.json
```diff
@@ -43,7 +43,7 @@
     "cebe/php-openapi": "^1.5",
     "cocur/slugify": "^4.0",
     "composer/composer": "^2.8",
-    "demos-europe/demosplan-addon": "^0.78",
+    "demos-europe/demosplan-addon": "dev-f_api-platform-4.3",
```
sierrazamudiodemos: holii, i think having approved this one https://github.com/demos-europe/demosplan-addon/pull/185/changes we can also adjust here oder?

bakridemos: Absolutly, this is only used for testing. Once this pr is approved and the pr ( in demosplan-addon ) is merged in i ll create a new version which will be used in core.

sierrazamudiodemos (approved):
thx for this

PR #6699, opened 2026-08-14
fix(refs DPLAN-18299): import all statement copies from assessment table exports

mussbach:
### Ticket
DPLAN-18299

An assessment table export writes one row per statement, and copies share their extern id. The import used the extern id as identity, so every copy after the first was skipped — silently, because the skipped-rows list is only logged when the import aborts for other reasons. The copy-and-shorten workflow used to split large statements therefore lost data without any message.

The export now carries the statement id in a "Referenz" column. Where the import finds that column, it stores the value on the imported statement and deduplicates on it: all copies arrive, each with its own reference, and re-importing a newer export adds only what is not there yet. Older exports without the column, and rows where the value is empty, keep the previous extern-id behaviour unchanged.

Note that the column is not gated by a permission, so it appears in the XLSX assessment table exports of every project.

Two details worth knowing while reviewing:

- Exported copies report the intern id of the statement they were copied from, and that has to stay unique per procedure — so only the first statement imported for a given intern id keeps it, mirroring what `StatementCopier` does within a procedure.
- The reminder mail for ending segment deadlines keyed its entries by segment extern id and collapsed segments sharing one, losing a link. Fixed in the second commit, since duplicate extern ids become legitimate with this change.

### How to review/test

- Export an assessment table from a procedure that contains copies of a statement, then import it into another procedure: every copy arrives, each with its own reference.
- Import the same file again: nothing is duplicated.
- Import an export created before this change: behaviour unchanged, copies still skipped.
- `SYMFONY_DEPRECATIONS_HELPER=disabled ./vendor/bin/phpunit tests/backend/core/Statement/Functional/StatementSourceReferenceImportTest.php tests/backend/core/Statement/Unit/AssessmentTableXlsExporterSourceReferenceTest.php tests/backend/core/Statement/Functional/SegmentDeadlineReminderServiceTest.php`

### PR Checklist

- [x] Create/Update tests
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## UNRELEASED
 
+### Fixed
+- Copies of a statement are no longer silently dropped when an assessment table export is imported. The export carries a reference to the statement each row originates from, so every copy arrives, and re-importing the same export adds only what is not there yet.
+- The reminder mail about ending segment deadlines lists every segment, also when several of them share an ID.
+
 ## v4.55.0 (2026-08-12)
 
 ### Changed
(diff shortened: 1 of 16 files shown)


bakridemos (approved):
thx code review and tested too

PR #6704, opened 2026-08-17
fix: Unconditionally register the import tabs hook

graupnerdemos:
### Ticket
(no ticket)

The import center was unable to register imports unless a feature unrelated to most import variants was activated.

This lead to some import addons not being able to be loaded without other import unrelated import addons also being existent.

Along the way, a longer in use older variant of the same hook was removed.

Additionally, references to the non-core feature have been removed elsewhere in the code base to reduce confusion. Please note that not all references are removed to keep this change minimal and locally contained.

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Run `yarn lint`
- [ ] Run `yarn test`
- [ ] Update changelog


client/js/components/procedure/AdministrationImport/AdministrationImport.vue
```diff
@@ -157,9 +166,14 @@ export default {
...
         },
-      ].filter((component) => {
-        return hasAnyPermissions(component.permissions)
-      }).concat(this.asyncComponents)
+      ]
+        .filter((component) => {
+          return hasAnyPermissions(component.permissions)
+        })
+        .concat(this.asyncComponents)
+        .sort((a, b) => {
```
hwiem, line 173: Should they be sorted alphabetically? I don't know if the previous order was random, and to me it would make sense to order them by how much they are used (or what POs consider a reasonable order)

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
```diff
@@ -170,53 +184,28 @@ export default {
...
-      promises.push(this.loadComponents('import.tabs'))
-    }
-
-    Promise.allSettled(promises)
-      .then(() => {
-        this.allComponentsLoaded = true
-        this.setActiveTabId()
-      })
+    const promises = [
+      loadAddonComponents('import.tabs').then((addons) => {
```
hwiem, line 195: Is it safe to just remove the permission (`feature_import_statement_pdf`) check?

graupnerdemos: Yes this permission is only available via an addon

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
graupnerdemos: I discussed this with UX and POs and yes, this is okay for now.

I like your idea, however, where would we track that? Depending on whether that is in use by the user or in the procedure or in the organisation that's very different questions to be answered.

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
hwiem: But does that really answer the question? If I understand correctly, this change means that the pdf import is now displayed to all users (if the addon is installed)

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
hwiem: If we want to track it, I'd start with a simple solution, just tracking tab clicks in general. I wouldn't overcomplicate it, I think the benefit is not big enough to justify that.

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
graupnerdemos: Oh. Yeah. That's a valid point, let me check if that needs a permission flag in the corresponding addon. In any case, this check should not be in the core.

hwiem (approved):
Code lgtm, thanks! 🙏

PR #6705, opened 2026-08-17
Feat(DPLAN-18296): redesign statement export modal

sakutademos:
### Ticket
[DPLAN-18296](https://demoseurope.youtrack.cloud/issue/DPLAN-18296) Redesign vom Export Modal

**Description:**

- adjust the modal according to the new design: [Figma design](https://www.figma.com/design/yNvo4JltT7c0k2bcEygPog/Geplante-Synopsenexporte?node-id=1-116&p=f&m=dev)
- capitalize the first letter of radio button and checkbox labels
- make the Export Statement modal narrower

**Refactored:**

- instead of using `position: absolute` for the checkboxes and radio buttons, use a more modern approach with flex: [refactored here](https://github.com/demos-europe/demosplan-core/pull/6705/changes#diff-0163ad915697df125cb2f2de32c2c0c84f6c91793b009ee31b9a7885689de820R493-R495). Without refactoring, it would be tricky to add this kind of styling to the radio buttons and not to break it in different places.
image

**How to Test:**

1. go to the STN List and click the Export button > check that the modal matches the design template
3. check that the other modals still look the same as before (except for the updated header (h2, instead of the h1 styling))

**How it looks now:**
image
image

There is a corresponding PR is in the UI library.

### PR Checklist

- [ ] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_form.scss
```diff
@@ -490,41 +490,45 @@ input.disabled {
     &__element {
```
hwiem, line 490: I don't know if it's these changes or changes in demosplan-ui, but I checked some checkboxes and radio buttons in the assesssment table export modal (there the 'x' button is now also right above a border, not sure we want to keep it like that), the statement modal and in the news admin area, and buttons and labels are not aligned properly; in the statement modal, buttons are quite big and far away from the labels - I don't think that's intended.

client/js/components/statement/StatementExportModal.vue
```diff
@@ -18,109 +18,139 @@
...
           v-text="Translator.trans('export.type')"
         />
-        <div class="grid grid-cols-4 mt-2 mb-5 gap-x-2">
+        <div class="grid grid-cols-4 gap-x-2">
           <dp-radio
             v-for="(exportType, key) in exportTypes"
             :id="key"
             :key="key"
+            class="p-2"
+            :class="{ 'bg-neutral-light-4 border-l-4 border-interactive rounded-sm' : active === key }"
```
hwiem: I guess it's because of the border - when you select a radio button, it jumps a little bit. Also, when you switch from docx to xlsx or csv, the modal shrinks a lot, so there's quite the layout jump. I guess that's how it was designed, but I wonder if we can do something about it? Maybe a smoother transition?
I'm also wondering if we could give the "Einzelne Docx-Dateien" button a little more space so the text fits on one line. The other buttons do have a lot of space they don't really need

hwiem (changes requested):
The export modal looks so much better now, thank you! ✨ 
The changes in `_form.scss` seem to have some unintended side effects though

demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_form.scss
sakutademos: what do you mean by "in the statement modal, buttons are quite big and far away from the labels"? I made some [changes](https://github.com/demos-europe/demosplan-core/pull/6705/commits/94843d7b4f0be9a6782750358fd33db94023559b) there that should fix the layout issues with the assessment table and other places where checkboxes/radio buttons are used.

hwiem: > what do you mean by "in the statement modal, buttons are quite big and far away from the labels"?

This is how it looks:
statement_modal_radio_btns
statement_modal_checkbox

client/js/components/statement/StatementExportModal.vue
sakutademos: yes, good ideas! I fixed the radio button jumps when selecting them, added more space around the “Einzelne Docx-Dateien” button, and added a height transition to the DpModal (in the corresponding project) to make the experience smoother for the user

image

demosplan/DemosPlanCoreBundle/Resources/client/scss/objects/_form.scss
sakutademos: the size of the inputs has not been changed in my PR. This is how they look on dev. However, some projects have overrides that add padding-left because the inputs had an absolute position before.

So I created corresponding PRs in those projects and assigned you as a reviewer. They remove the extra `36px` of left padding from the checkboxes and radio buttons

client/js/components/statement/StatementExportModal.vue
hwiem: Thanks, it's perfect now. :) I love the transition!

hwiem (approved):
Lgtm + thanks for putting in the extra effort to fix every last detail ❤️

PR #6706, opened 2026-08-17
fix HDDP-27: stream export download in chunks to avoid memory exhaustion

mussbach:
### Ticket
HDDP-27

Downloading the result of an asynchronous export could exhaust the memory limit and abort the
transfer mid-stream:

```
app.ERROR: OutOfMemoryError: Allowed memory size of 2147483648 bytes exhausted
(tried to allocate 2147483648 bytes) in ExportJobDownloadResponseFactory
```

`fpassthru()` handed the archive to the output layer without ever flushing it, so an active output
buffer accumulated the complete file in memory. The reported allocation is that buffer doubling from
1 GB to 2 GB — for a multi-gigabyte archive the download therefore died where the export had been
moved into the background to begin with.

The response now reads the archive in 1 MiB chunks and flushes each one to the client, and it sends a
`Content-Length`. Measured with a 200 MB file at `memory_limit=64M`: `fpassthru()` and a chunked copy
without flushing both die, the flushing loop peaks at 6 MB. Both `readStream()` and `fileSize()` go
through the `FilesystemOperator`, so local, S3 and the lazy default storage take the same path.

### How to review/test
Trigger an assessment table or procedure export large enough to matter, then download the finished
job and watch the memory of the serving process — it stays flat instead of growing with the file.
The unit test covers the chunked output and the new header; it captures through an output handler
because the response now flushes its own buffer.

### PR Checklist

- [x] Create/Update tests
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## UNRELEASED
 
 ### Fixed
+- Downloading the result of a background export works for large archives. The file is sent to the browser piece by piece instead of being held in memory as a whole, which could abort the download.
 - Copies of a statement are no longer silently dropped when an assessment table export is imported. The export carries a reference to the statement each row originates from, so every copy arrives, and re-importing the same export adds only what is not there yet.
 - The reminder mail about ending segment deadlines lists every segment, also when several of them share an ID.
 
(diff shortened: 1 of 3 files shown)


mussbach:
All changes landed meanwhile in main

PR #6279, opened 2026-05-19
refactor(refs DPLAN-16770): migrate from jest to vitest

hwiem:
### Ticket
DPLAN-16770

- replace `@vue/vue3-jest` with `vitest` - `vue3-jest` had its last release 3 years ago, and `vitest` is the officially recommended test runner for vue 3
- `@vitest/ui` is added, so when running `yarn tdd:js`, you can optionally view the vitest ui in the browser with the results for each individual test case
- also, tests are now run in @vue/compat runtime instead of in vue 3 (as before) to match the actual runtime of demosplan; compat warnings are suppressed in the test results though because we don't really need them there, and they create a lot of noise
- replace `eslint-plugin-jest` with `@vitest/eslint-plugin`

#### This will close the following dependabot PRs:
| PR | Dependency | Why |
| --- | --- | --- |
| [#6361](https://github.com/demos-europe/demosplan-core/pull/6361) | `jest` | removed |
| [#6355](https://github.com/demos-europe/demosplan-core/pull/6355) | `jest-environment-jsdom` | removed; could not be updated, it capped `jsdom` at `^26.1.0` even in 30.4.1 |
| [#6555](https://github.com/demos-europe/demosplan-core/pull/6555) | `@jest/globals` | removed |
| [#5036](https://github.com/demos-europe/demosplan-core/pull/5036) | `babel-jest` | removed |
| [#6364](https://github.com/demos-europe/demosplan-core/pull/6364) | `jsdom` | superseded; this PR updates `jsdom` from `^26.1.0` to `^29.1.0` |

### How to review/test

```bash
y
y test        # 35 files, 191 passing / 16 skipped
y lint .      # 0 errors under tests/frontend
y tdd:js      # vitest UI at http://172.22.255.2:51204/__vitest__/
```

### Tasks
- [x] adjust FE CI check script (from `yarn test --maxWorkers=1 --ci` to `CI=true yarn test --maxWorkers=1`) - otherwise, the FE check will fail

### PR Checklist

- [x] Create/Update tests
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [x] Link all relevant tickets

PR #6709, opened 2026-08-18
DPLAN-16770: use vitest mock in inline image anchors test

riechedemos:
### Ticket
DPLAN-16770

The frontend test runner was migrated from jest to vitest (#6279), but
`tests/frontend/unit/inlineImageAnchors.test.js` still calls `jest.fn()`. Vitest does not
provide a `jest` global, so the call throws a `ReferenceError` while the module is being
evaluated. That kills the whole file as a failed suite (`0 test`) and turns the FE test job
red on `release_kanban` and on every branch based on it.

The fix uses `vi.fn()` with an explicit `import { vi } from 'vitest'` and moves the
`global.Translator` assignment below the imports. The file is now byte-identical to the
version on `main`, where this was already corrected in #6665 — so no conflict is introduced
when `release_kanban` and `main` are reconciled later.

### How to review/test
Check that the FE test job passes on this PR. `tests/frontend/unit/inlineImageAnchors.test.js`
should be reported as a passing suite instead of a failed one.

No further `jest` references remain in `tests/frontend/`.

### PR Checklist

- [x] Run `yarn test`

### DIFF
--- a/tests/frontend/unit/inlineImageAnchors.test.js
+++ b/tests/frontend/unit/inlineImageAnchors.test.js
@@ -7,9 +7,10 @@
  * All rights reserved
  */
 
-global.Translator = { trans: jest.fn(key => key === 'image.open' ? 'Bild öffnen' : key) }
-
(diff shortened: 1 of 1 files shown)


hwiem (approved):
Thanks!

PR #6710, opened 2026-08-19
f_DPLAN-18151: first commit implementing a new ServerBannerLoader.php…

kijewskidemos:
### Ticket
[DPLAN-18151](https://demoseurope.youtrack.cloud/issue/DPLAN-18151) Zentrale User Information

### Description: 

This PR introduces a new service to customers: now a banner pops up for every user when a server downtime / maintenance tasks are expected. This enhances the user experience and guarantees that user are now able to plan accordingly without being taken by surprise. The banner persists over page changes but the user can close the tab.


mussbach (changes requested):
lgtm but test is missing

kijewskidemos:
> lgtm but test is missing

Thank you for the review! I took your comment to think about my way of handling the banner file. As of now its path is hardcoded into the ServerBannerLoader. Should I keep this or should it be injected instead?

graupnerdemos:
@kijewskidemos I think having this file name hard coded is fine for the time being.

breierDemos (approved):
lgtm

tests/backend/core/Core/Unit/Services/ServerBannerLoaderTest.php
```diff
@@ -0,0 +1,92 @@
...
+        parent::setUp();
+
+        $this->sut = new ServerBannerLoader();
+        $this->bannerPath = DemosPlanPath::getRootPath('SERVER_BANNER.md');
+
+        // Preserve any banner file that may already exist on disk so the test
+        // never clobbers real state, then start from a clean slate.
+        if (is_readable($this->bannerPath)) {
+            $this->preexistingBannerContent = file_get_contents($this->bannerPath);
+            unlink($this->bannerPath);
```
mussbach: it is better to change the contents than to delete the file and create it again, as this will change the permissions of the file to the user that performs the test, which may be a different one and leaves the file read only to the current user

mussbach (changes requested):
thank you for the test. One tiny remark, otherwise lgtm

tests/backend/core/Core/Unit/Services/ServerBannerLoaderTest.php
kijewskidemos: Thank you for the concern! I agree and adjusted this issues in this commit: fbc1dae92f79a326bdd354611032cc0de1bb35f9

mussbach (approved):
lgtm, thx

PR #6711, opened 2026-08-19
chore: dedupe dompurify to patched version (3.4.13)

hwiem:
fixes a moderate security alert:
IN_PLACE hook removal leaves a detached subtree executable, causing XSS

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7656,19 +7656,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"dompurify@npm:^3.0.2, dompurify@npm:^3.4.5":
-  version: 3.4.12
-  resolution: "dompurify@npm:3.4.12"
-  dependencies:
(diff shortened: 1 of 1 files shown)


mussbach (commented):
how is this dedupe done? Will it come back when lockfile is generated again?

hwiem:
> how is this dedupe done? Will it come back when lockfile is generated again?

It was done with `yarn dedupe` and since 3.4.13 satisfies all ranges required by demosplan-core and dependencies, it would not go back to a vulnerable version when `yarn.lock` is regenerated. The dependencies that need dompurify are demosplan-ui and cesium - demosplan-ui already uses 3.4.13 on main (and therefore in the next release), and the only way a vulnerable version could make its way back here is if a dependency used a different range, excluding 3.4.13. I find that unlikely enough to prefer this over a resolution pin in `package.json`.

mussbach (approved):
lgtm, thx

PR #6712, opened 2026-08-19
feat: generate CycloneDX SBOMs during the container build

mussbach:
Container images now ship CycloneDX SBOMs for their composer and yarn dependencies in
`/srv/www/sbom`, generated by two build stages on top of the shared dependency stage.
This replaces the two scheduled workflows that opened PRs with generated SBOM files, so
SBOMs are no longer committed to the repository.

Both cover production dependencies only, transitives included, so they describe what the
image actually ships:

- `sbom-php` installs the generator globally in its own stage and builds the SBOM from the
  prod vendor. Its component count matches the `packages` block of `composer.lock` exactly.
- `sbom-js` runs cdxgen, then reduces the result to the production closure with
  `docker/demosplan-production/sbom-prod-filter.php`. cdxgen's own `--required-only` is not
  usable here: it drops declared dependencies, among them cesium, core-js and the @uppy
  packages. The script refuses to write rather than emit a thinned SBOM if cdxgen's
  dependency graph is missing or yields no transitives.

`cyclonedx/cyclonedx-php-composer` is no longer a dev dependency; it is only needed inside
the build stage. That drops 6 packages from `packages-dev` and leaves the prod set untouched.

### How to review/test

    docker/build.sh <image> <version> <project>

Then inspect the result:

    docker run --rm --entrypoint php <image>:<version> -r 'foreach (["php","js"] as $e) {
      $d = json_decode(file_get_contents("/srv/www/sbom/sbom-$e.cdx.json"), true);
      printf("%s: %d components\n", $e, count($d["components"])); }'

Expect 275 for PHP (= `composer.lock` prod packages) and 535 for JS, with no dev packages
such as phpunit, rector, eslint or jest.

### Known inconsistency

The two SBOMs disagree on `metadata.component.version`: the PHP one carries the image
version, the JS one takes `v4.57.0` from `package.json`, because cdxgen has no equivalent of
`--mc-version`. Cosmetic, and fixable in the filter script if wanted.

### PR Checklist

- [x] Update documentation


bakridemos:
I tried to review the Dockerfile, the new filter script, and the docs. The approach looks solid, and I didn't find any dangling references after the cleanup. Two things stood out in the README:

1. The README says the cyclonedx-php-composer generator is installed "pinned", but the Dockerfile installs it with no version constraint at all — and even has a comment saying that's intentional. These two contradict each other, worth fixing one or the other.
2. The README mentions a file sbom-prod-filter.mjs, but the actual file shipped is sbom-prod-filter.php. Looks like a leftover from an earlier version.

One question (not my area, so tell me if it's a non-issue): both SBOM tools are installed unpinned, pulling whatever's latest at build time. Since this exists for compliance reasons, could an upstream update silently break or change the build without us noticing, given there's no CI job that builds this image on PRs?

mussbach:
@bakridemos thank you for the pointer to the obsolete documentation and the version pinning. Initially I did this by purpose to avoid the maintenance burden to update Versions where dependabot cannot find them. But that was a rabbithole that did not pay off against the version pinning. I reintroduced it.

bakridemos (approved):
I reviewed it as best I could.

graupnerdemos (approved):
Looks good. Thanks!

PR #6714, opened 2026-08-20
refactor: log scalars via Monolog context instead of varExport

mussbach:
### Ticket
Maintenance, no ticket.

First step of removing `DemosPlanTools::varExport()`. This PR covers only the mechanical
part: the 31 log calls that passed a **scalar**, where `var_export` did nothing but add
quotes around the value. Message becomes a constant string with `{placeholder}`s, the value
moves into the Monolog context array.

Monolog already covers this: `NormalizerFormatter` (base of `monolog.formatter.line`, which
`logging_formatter` resolves to in `config/parameters_default.yml`) normalizes and
JSON-encodes context with a depth cap, and `process_psr_3_messages` is enabled for all
handlers except `console`, so the placeholders are interpolated. Benefits over the old form:
no dump built when the level is filtered out, single-line records instead of `var_export`'s
multi-line output, and a stable Sentry fingerprint because the payload is no longer part of
the message.

Two changes beyond the plain swap:

- Numerically-keyed exception context (`[$e]`) and message-concatenated exceptions (`.$e`)
  become `['exception' => $e]`, which is why exceptions were previously unreadable in these
  records. Affects `StatementFragmentService`, `StatementService` and the three sites in
  `StatementToLegacyConverter`.
- `StatementDeleter` is not a log call but a `DemosException` message, so it uses `sprintf`
  rather than context.

The `DemosPlanTools` import is dropped from the 10 files that no longer reference it; it
stays in the 4 files that still have call sites left for the follow-ups below.

Deliberately **not** in this PR, because each needs a judgement call rather than a review
scan — the remaining ~80 call sites are grouped into follow-ups by theme: Elasticsearch query
dumps, PDF/RabbitMQ payload dumps, `StatementGeoService`, mail logging, the Dataport user/role
mapping, entity dumps, WMS responses, and exception messages. The helper itself is deleted at
the end of that sequence.

### How to review/test
Every hunk is the same transformation, so it should read quickly. Worth a second look:

- `UserService.php:154` — kept as `['login' => $login]`. Dropping information isn't a
  mechanical change, but this is a login/email and is a candidate for removal in the
  mail/user follow-up.
- `DraftStatementService.php:705` — `submitResult` is always `false` inside
  `if (false === $submitResult)`, so the context is dead weight. Converted faithfully rather
  than fixed here.

Verification: `php -l` clean on all 14 files. PHPStan level 5 on the changed files reports 204
errors, all pre-existing and none on a changed line. No test run: the changes are log calls
only and no test covers `DemosPlanTools`.

Unrelated finding while touching the file: `DemosPlanDocumentController.php:824` has
`md5(random_int(0, 9999))` without a `(string)` cast (PHPStan level 5 flags it); the duplicated
block in `ZipImportService.php:152` already has the cast.

### PR Checklist

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/AssessmentTable/DemosPlanAssessmentTableController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/AssessmentTable/DemosPlanAssessmentTableController.php
@@ -46,7 +46,6 @@ use demosplan\DemosPlanCoreBundle\Services\Breadcrumb\Breadcrumb;
 use demosplan\DemosPlanCoreBundle\Services\HTMLFragmentSlicer;
 use demosplan\DemosPlanCoreBundle\StoredQuery\AssessmentTableQuery;
 use demosplan\DemosPlanCoreBundle\Traits\DI\RefreshElasticsearchIndexTrait;
-use demosplan\DemosPlanCoreBundle\Utilities\DemosPlanTools;
 use demosplan\DemosPlanCoreBundle\ValueObject\AssessmentTable\StatementBulkEditVO;
 use Doctrine\Common\Collections\Collection;
 use Exception;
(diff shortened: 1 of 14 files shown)


sierrazamudiodemos (approved):
thx, it looks good!!

PR #6722, opened 2026-08-21
refactor(refs DPLAN-15761, #AB27466): replace inline-template with slot [vue 3 migration]

hwiem:
### Ticket
DPLAN-15761

Removes the inline-template in `fragment_statement.html.twig` since it's no longer supported in vue 3.

### How to review/test
- Open a project in which statements can be split into fragments
- Login as FPA, navigate to the assessment table, claim a statement and select "In Datensätze aufteilen"
- Make sure that tags, municipalities, etc. from the statement are applied to the multiselects, test that the multiselects still work properly
- Create a fragment and check that all the data is correctly saved

### PR Checklist

- [x] Run `yarn lint`

### DIFF
--- a/client/js/bundles/statement/fragmentStatement.js
+++ b/client/js/bundles/statement/fragmentStatement.js
@@ -11,17 +11,23 @@
  * This is the entrypoint for fragment_statement.html.twig
  */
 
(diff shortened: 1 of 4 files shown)


sakutademos (approved):
lgtm + works! Thx a lot!!!

PR #6725, opened 2026-08-21
feat(refs DPLAN-18326): reset two-factor authentication from the user administration

mussbach:
### Ticket
DPLAN-18326

Organisation administrators and support had no way to reset the two-factor
authentication of a user who lost access to their second factor (device gone,
technical problem). The only remedy so far was a manual database change.

The expanded entry in the user administration now offers a
"Zwei-Faktor-Authentifizierung zurücksetzen" button. It clears both second
factors (TOTP app and email code), so the user can log in with their password
alone and set 2FA up again.

Implemented as a `twoFactorEnabled` attribute on `AdministratableUser` rather
than a new endpoint, so the existing authorization applies unchanged:

- the attribute only exists when `feature_2fa` is granted, and writing needs
  `feature_user_edit` — the same permission the user administration already
  requires
- `getAccessConditions()` is merged into the entity lookup on the update path
  (`AbstractResourceType::updateEntity`), so an organisation administrator can
  only reset users of their own organisation and support only within the
  customer — a foreign id simply does not resolve
- the write behaviour accepts `false` only; sending `true` throws, so 2FA can
  never be activated for somebody else through this route

Every reset is written as a report entry (group `user`, category
`resetTwoFactor`) recording who reset it for whom.

`feature_2fa` was missing `expose: true`, so it was not available to
JavaScript-based permission checks at all. That is fixed here as well.

### How to review/test
1. Give a test user a second factor (TOTP or email) in their profile.
2. Log in as support or as an organisation administrator of that user's
   organisation and open the user administration.
3. Expand the user — the reset button appears above the save/cancel row. It
   stays hidden for users without an active second factor.
4. Confirm the dialog; the user can afterwards log in without a second factor.
5. Check the report entries for a `resetTwoFactor` record.

Also worth checking: as an organisation administrator, a `PATCH` on a user of
another organisation must fail rather than reset anything.

### Linked PRs
The `UserInterface` extension this builds on:

- demos-europe/demosplan-addon#186

### PR Checklist

- [x] Create/Update tests
- [x] Run `yarn lint`
- [x] Update changelog

### DIFF
--- a/client/js/components/user/DpUserList/DpUserListItem.vue
+++ b/client/js/components/user/DpUserList/DpUserListItem.vue
@@ -207,7 +207,7 @@ export default {
     },
 
     canResetTwoFactor () {
-      return hasPermission('feature_2fa') && true === this.user.attributes.twoFactorEnabled
+      return hasPermission('feature_2fa') && this.user.attributes.twoFactorEnabled
     },
 
     hasDepartment () {
--- a/client/js/components/user/DpUserList/DpUserListItem.vue
+++ b/client/js/components/user/DpUserList/DpUserListItem.vue
@@ -118,6 +118,22 @@
         @user:update="updateUser"
       />
 
+      <div
+        v-if="canResetTwoFactor"
+        class="mb-2"
+      >
+        <p class="lbl__hint mb-1">
+          {{ Translator.trans('2fa.reset.hint') }}
+        </p>
+        <dp-button
+          :busy="isResettingTwoFactor"
+          data-cy="resetTwoFactor"
+          :text="Translator.trans('2fa.reset')"
+          variant="outline"
+          @click="resetTwoFactor"
+        />
+      </div>
+
       <dp-button-row
         form-name="userForm"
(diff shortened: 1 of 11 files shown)


client/js/components/user/DpUserList/DpUserListItem.vue
```diff
@@ -188,6 +206,10 @@ export default {
       return Translator.trans(this.isOpen ? 'aria.collapse' : 'aria.expand')
     },
 
+    canResetTwoFactor () {
+      return hasPermission('feature_2fa') && true === this.user.attributes.twoFactorEnabled
```
sakutademos: it can be simplified

```suggestion
      return hasPermission('feature_2fa') && this.user.attributes.twoFactorEnabled
```

sakutademos (approved):
frontend code looks good (just a tiny nitpick), lgtm, thx!

bakridemos (approved):
lgtm thx, tested too

PR #6727, opened 2026-08-21
Convert addon frontend asset loading to ES modules

graupnerdemos:
### Ticket
No ticket

Related to [demosplan-addon-client-builder#264](https://github.com/demos-europe/demosplan-addon-client-builder/pull/264), this enables ES module imports for addon js assets in the core. In addition, some refactoring for readability  (in the backend) and towards TypeScript (in the frontend) was applied.

The general changes are:

- The RPC communication for addon hooks now returns their asset url(s) if it exists. For non-updated addons, the current UMD module build will still be sourced and passed directly in the rpc response.
- Warnings will be written to the log (and subsequently Sentry) if an addon hook does not have defined permissions or is not providable via ESM.

**Additional Notes**

- While this enables us to get rid of the  evil `eval`, we cannot directly do so since updates in all addons are required. 
- Embedding addon frontend code with dynamic imports has some impacts on how that code behaves. Most notably, the imported code getting it's own namespace in the runtime, i.e. being somewhat sandboxed. I could not find any negative side effects of this during testing, but we should be aware that there might be some. For more information on `import()` see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#description.

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`


sakutademos (commented):
ooo, its so a good improvement! thank you so much!!! I tested it both with and without the client-builder updates, and it works, it can execute both the esm.js and umd.js formats. The code looks good

composer.json
```diff
@@ -232,7 +231,8 @@
       "bamarni/composer-bin-plugin": true,
       "php-http/discovery": false,
       "endroid/installer": true,
-      "tbachert/spi": false
+      "tbachert/spi": false,
+      "cyclonedx/cyclonedx-php-composer": true
```
mussbach: this most likely does not belong here

mussbach (commented):
A few points not yet covered by the existing review.

Also: this branch forked before `5cb6c1f1a` ("Unconditionally register the import tabs hook"). As it stands, merging would re-gate `import.tabs` behind `feature_import_statement_pdf`, re-add the `email.import` hook, and drop the `.sort()` by translated tab title that main has in `AdministrationImport.vue`. Worth rebasing on main first.

demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
```diff
@@ -115,4 +240,34 @@ private function getAssetPathsFromManifest(string $manifestPath, string $entryNa
...
+
+        $manifestPath = DemosPlanPath::getRootPath($addonInfo->getInstallPath()).'/'.$uiData['manifest'];
+
+        // If the requested filename is a source map, validate against the corresponding JS file instead
+        $validationPath = $filename;
+        if (str_ends_with($filename, '.js.map')) {
+            $validationPath = substr($filename, 0, -4);
+        }
+
+        if (!file_exists($manifestPath) || !$this->manifestDeclaresJsAsset($manifestPath, $validationPath)) {
```
mussbach, line 274: `manifestDeclaresJsAsset()` scans **all** entrypoints of the manifest, so the filename check is not tied to the `$hookName` whose permission was just verified.

For an addon registering two hooks — `hookA` ungated, `hookB` behind an admin permission — any user can request `/addon-assets/vendor/name/hookA/hookB-entry.esm.js` and receive the restricted bundle. Validating against `$uiData['hooks'][$hookName]['entry']`'s assets only would close that.

demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
```diff
@@ -13,82 +13,207 @@
...
+        foreach ($javascriptEntries as $entry) {
+            if (!is_string($entry)) {
+                continue;
             }
 
-            $hookData = $uiData['hooks'][$hookName];
-            $manifestPath = DemosPlanPath::getRootPath($addonInfo->getInstallPath()).'/'.$uiData['manifest'];
+            if (str_ends_with($entry, '.esm.js')) {
+                // new-format addon build: serve by URL, frontend `import()`s it directly
+                list($addonVendor, $addonName) = explode('/', $addonInfo->getName(), 2);
```
mussbach, line 123: An addon name without a `/` gives an undefined-offset warning and `$addonName = null`, which makes `generate()` throw `InvalidParameterException` and turns the whole `addons.assets.load` RPC into a 500 instead of skipping that one addon.

demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
```diff
@@ -13,82 +13,207 @@
...
+                    ],
+                    UrlGeneratorInterface::ABSOLUTE_URL
+                );
+
+                continue;
             }
 
-            try {
-                $entries = $this->getAssetPathsFromManifest($manifestPath, $hookData['entry']);
+            $this->logger->warning(
```
mussbach, line 149: The loop walks every `js` entry, so an addon shipping both `x.umd.js` and `x.esm.js` — likely during the migration this PR enables — yields both `content` and `urls`. The frontend's `'urls' in addon` branch wins, so the UMD source is `file_get_contents`'d and shipped over RPC for nothing, and the "legacy UMD bundle" warning fires for an addon that has already migrated.

demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
```diff
@@ -115,4 +240,34 @@ private function getAssetPathsFromManifest(string $manifestPath, string $entryNa
...
         return $manifestContent['entrypoints'][$entryName]['assets'];
     }
+
+    private function checkFileAccess(?AddonInfo $addonInfo, string $hookName, string $filename)
+    {
+        if (!$addonInfo instanceof AddonInfo || !$addonInfo->isEnabled() || !$addonInfo->hasUIHooks()) {
+            return false;
+        }
+
+        $uiData = $addonInfo->getUIHooks();
```
mussbach, line 264: `$uiData['hooks'][$hookName]` goes straight into `hasHookPermission(array $hookData, …)` without the `is_array()` guard that `lookupAssetInfo` uses, so a malformed hook definition (string/null) gives a `TypeError` → 500 rather than a clean 404. The method is also missing a `: bool` return type.

client/js/components/procedure/AdministrationImport/AdministrationImport.vue
```diff
@@ -177,32 +178,13 @@ export default {
...
-
-            this.asyncComponents.push({
-              name: addon.entry,
-              title: addon.options.title,
-            })
-          }
+      return loadAddonComponents(hookName)
+        .then((addons) => {
+          this.asyncComponents.push(...addons.map((addon) => ({
+            component: addon.component,
```
mussbach: `asyncComponents` lives in `data()`, so `addon.component` gets wrapped in a deep reactive proxy — Vue warns "received a Component that was made a reactive object" and deep-tracks the whole component definition. `AddonWrapper.vue` and `RecommendationModal.vue` in this PR both wrap with `shallowRef`; same is needed here.

composer.json
graupnerdemos: yes, but should solve itself with the next main merge. this was missing on the sbom pr.

client/js/lib/addon/loadAddonComponents.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    return module.default
+  }
+
+  const contentKey = `${addon.entry}.umd.js`
+
+  eval(addon.content[contentKey])
+
+  return (window as unknown as Record<string, { default: unknown }>)[addon.entry].default
+}
```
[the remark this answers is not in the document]
graupnerdemos: Using fixed filenames is safe here since the naming conventions are set in demosplan-addon-builder which is developed in conjunction with the core addon handling.

demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
graupnerdemos: Since the esm-section above will `continue` out of that addon hook, the umd source will in fact not be `file_get_contents`-ed.

composer.json
```diff
@@ -1,244 +1,245 @@
...
-      "Tests\\Core\\": "tests/backend/core/",
-      "Tests\\CoreApplication\\": "tests/backend/core-application/src/",
-      "Tests\\Plugin\\": "tests/backend/plugins/",
-      "Tests\\Project\\": "tests/backend/projects/"
-    }
-  },
-  "extra": {
-    "aws/aws-sdk-php": [
-      "S3"
+    "name": "demos-europe/demosplan-core",
```
graupnerdemos, line 2: I have no clue why the composer.json was completely rewritten :'/

composer.json
graupnerdemos: composer.json changes got worse for some reason… @mussbach let me know if I can leave them in or if it's fine. I can't see any obvious changes apart from some re-ordering and I didn't manually touch that file so that has me a little worried.

demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
graupnerdemos: Nevermind, you're right. But, this cannot happen since any given addon may only have an older addon-builder version which yielded umd bundles or a newer version which yields esm bundles. Since every addon exists as a singleton in the codebase and there can't be differing versions of the same addon floating around this is fine.

client/js/lib/addon/loadAddonComponents.ts
```diff
@@ -0,0 +1,88 @@
...
+      /*
+       * If for some reason we don't receive a valid response object from the backend
+       * we'll just skip it.
+       */
+      console.debug('Skipping addon hook response evaluation for ' + key)
+      continue
+    }
+
+    addons.push({
+      component: await resolveComponent(addon),
```
hwiem: I think we should wrap this in `shallowRef()`, that way we don't need to remember doing it in the call sites (so if we add `shallowRef` here, we should remove it in `AddonWrapper.vue` and `RecommendationModal.vue`)
```suggestion
      component: shallowRef(await resolveComponent(addon)),
```

hwiem (changes requested):
Just a small change request, thanks for this! ❤️

client/js/lib/addon/loadAddonComponents.ts
graupnerdemos: Good catch, I lost track on where the correct `shallowRef`-place should be. :D

graupnerdemos: [2f0be91](https://github.com/demos-europe/demosplan-core/pull/6727/commits/2f0be911b773904a955678a00ea2f40278055a89)

client/js/components/procedure/Shared/RecommendationModal.vue
```diff
@@ -121,10 +121,10 @@ export default {
         this.$emit('addons:loaded')
 
         this.modalAddons = addons.map(addon => {
-          const { name, options } = addon
+          const { component, name, options } = addon
```
hwiem: If we remove the `shallowRef` here, doesn't that make the entire `.map()` redundant?

client/js/lib/addon/loadAddonComponents.ts
```diff
@@ -0,0 +1,89 @@
...
+  entry: string
+  options?: AddonHookOptions
+  urls: Record<string, string>
+}
+
+interface LegacyAddonEntry {
+  entry: string
+  options?: AddonHookOptions
+  content: Record<string, string>
+}
```
hwiem, line 16: Why do we differentiate these two types if they are identical?

client/js/lib/addon/loadAddonComponents.ts
```diff
@@ -0,0 +1,89 @@
...
+       * we'll just skip it.
+       */
+      console.debug('Skipping addon hook response evaluation for ' + key)
+      continue
+    }
+
+    addons.push({
+      component: await shallowRef(resolveComponent(addon)),
+      name: addon.entry,
+      options: addon.options ?? '',
```
hwiem: I know it was like this before, but why do the options default to an empty string? - Since you didn't change this, you can ignore my comment, I'm just wondering.

client/js/lib/addon/loadAddonComponents.ts
```diff
@@ -0,0 +1,89 @@
...
+      /*
+       * If for some reason we don't receive a valid response object from the backend
+       * we'll just skip it.
+       */
+      console.debug('Skipping addon hook response evaluation for ' + key)
+      continue
+    }
+
+    addons.push({
+      component: await shallowRef(resolveComponent(addon)),
```
hwiem: not `shallowRef` should be awaited, but `resolveComponent`
```suggestion
      component: shallowRef(await resolveComponent(addon)),
```

client/js/components/addon/AddonWrapper.vue
```diff
@@ -15,12 +15,17 @@
...
-<script>
+<script lang="ts">
 import * as demosplanUi from '@demos-europe/demosplan-ui'
+import { defineComponent, shallowRef, type ShallowRef } from 'vue'
 import loadAddonComponents from '@DpJs/lib/addon/loadAddonComponents'
-import { shallowRef } from 'vue'
 
-export default {
+interface WrappedAddon {
+  component: ShallowRef
```
hwiem: `ShallowRef` alone should not be the type here (`Component` will also need to be imported)

```suggestion
  component: ShallowRef<Component>
```

client/js/lib/addon/loadAddonComponents.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type AddonRpcEntry = EsmAddonEntry | LegacyAddonEntry
+
+interface AddonRpcResponse {
+  data: Array<{ result: Record<string, AddonRpcEntry | undefined> }>
+}
+
+export interface LoadedAddon {
+  component: ShallowRef
```
hwiem: `Component` will also need to be imported here
```suggestion
  component: ShallowRef<Component>
```

hwiem (changes requested):
Just some tiny change requests

client/js/lib/addon/loadAddonComponents.ts
graupnerdemos: The friendly little helper did that, I left it in because it made sense to me to keep the new code path completely separate from the old one.

hwiem: I agree to keep the code paths separate, but since this part didn't change, I think we should just create a common `AddonEntry` interface. Keeping the two seems like unnecessary code duplication and I actually find it confusing. I was staring at the two interfaces for a bit to find the difference. :smile:

client/js/lib/addon/loadAddonComponents.ts
graupnerdemos: You're right. It should be an empty object. I changed that and don't see any apparent errors.

client/js/lib/addon/loadAddonComponents.ts
graupnerdemos: They are not actually identical. But I must admit I didn't fully realise that until I tried deleting it. Esm Addons have `urls`, legacy umd ones have `content`. So it makes sense to keep those separated.

hwiem (approved):
FE lgtm, thanks!

mussbach:
@graupnerdemos I wasnt aware of this PR any more and I did changes in #6813 that conflict with this PR, I do not think it is a simple merge but we need to take care, that the fix dont accidentally gets overridden

PR #6728, opened 2026-08-21
fix: scope institution tags to the current customer

mussbach:
## Summary
- `InstitutionTagResourceType` returned all institution tags regardless of customer, so a tag created under one customer was readable via the API from any other customer, and could also be attached to organizations under a different customer through the same read-side gap.
- Scope access to the current customer via the tag's category, mirroring the existing pattern in `InstitutionTagCategoryResourceType`.
- Add missing Foundry factories for `InstitutionTag`/`InstitutionTagCategory` and a regression test covering both the same-customer and cross-customer cases.

## Test plan
- [x] `InstitutionTagResourceTypeTest` (new) passes
- [x] `phpstan -l5` on changed files shows no new errors (pre-existing template-covariance findings unchanged)


hwiem (approved):
FE code lgtm, thank you!

breierDemos (approved):
BE lgtm code review only

mussbach:
already merged via #6739

PR #6729, opened 2026-08-21
fix(orga): make showlist a per-customer setting

mussbach:
`showlist` gates whether an organisation appears in the invitation lists, but it lived on
`Orga` as a single value shared across every customer — unlike `status`, `orgaType` and
customer membership, which are all per customer on `OrgaStatusInCustomer`. Disabling it in
one customer therefore hid the organisation everywhere, and an organisation could not be
visible in one customer while hidden in another.

The flag moves to `OrgaStatusInCustomer` and is resolved for the current customer in the
backend, so the flat `showlist` API attribute keeps its contract and the frontend is
untouched.

**Grain:** `showlist` is one decision per customer, spanning every orgaType row the
organisation holds there — reads prefer the public-agency row (the grain
`InvitablePublicAgencyResourceType` filters on) and fall back to any other row of the
customer; writes cover all rows; a newly added orgaType row inherits the customer's current
value. This keeps it aligned with `getParticipants()`, which matches any orgaType row, and
keeps the flag settable for organisations that are not public agencies (roughly half of them
in a real customer).

`Orga._o_showlist` is retained and dual-written during the transition; a follow-up drops the
column together with the `canUpdateShowList` plumbing.

### How to review/test
- The migration adds the column and backfills it from `_o_showlist`, so existing settings
  carry over. It must run exactly once — re-running the `UPDATE … JOIN` would overwrite
  deliberately diverged per-customer values.
- `UserServiceTest` covers the query path end to end: per-customer split via
  `getParticipants()`, multi-orgaType writes, organisations without a public-agency row, and
  orgaType inheritance.

### PR Checklist
- [x] Create/Update tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
     "cebe/php-openapi": "^1.5",
     "cocur/slugify": "^4.0",
     "composer/composer": "^2.8",
-    "demos-europe/demosplan-addon": "^0.79",
+    "demos-europe/demosplan-addon": "^0.80",
     "doctrine/annotations": "^2.0",
     "doctrine/common": "^3.2",
     "doctrine/data-fixtures": "^1.8",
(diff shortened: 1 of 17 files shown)


breierDemos (approved):
lgtm code review only

PR #6739, opened 2026-08-24
fix: scope institution tags to the current customer

mussbach:
Backport of #6728.
Ado-57428

Institution tags are scoped to a customer only through their category, but
`InstitutionTagResourceType` returned all tags regardless of customer. An organisation that is a
member of several customers therefore exposed its tags from every one of them:

- Reading: the tag came back while its category did not (the category type does filter by
  customer), so the institution list received a tag with `category: null` and the row failed to
  render as soon as a category column was selected.
- Writing: the `assignedTags` payload only ever covers the current customer, but the update diffed
  it against the organisation's complete tag collection, so saving an institution silently deleted
  the tags it holds in other customers.

Fixed on both sides — access conditions scope reads via `category->customer`, and removals are
limited to tags of the current customer. The institution list additionally skips tags whose
category it cannot resolve, since those have no column to be displayed or edited in.

### How to review/test
- An organisation in two customers, tagged in both: the foreign tag must be invisible in the other
  customer, and editing the organisation there must leave that assignment intact.
- `InstitutionTagResourceTypeTest` covers same-customer and cross-customer reads,
  `InvitableInstitutionResourceTypeTest` covers the update keeping foreign tags while still
  removing own ones.

### PR Checklist
- [x] Create/Update tests

### DIFF
--- a/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
+++ b/client/js/components/procedure/admin/InstitutionTagManagement/InstitutionList.vue
@@ -95,7 +95,7 @@
             <div
               v-else
               :key="`tags:${idx}`"
-              v-text="separateByCommas(institution.tags.filter(tag => tag.category.id === category.id))"
+              v-text="separateByCommas(institution.tags.filter(tag => tag.category?.id === category.id))"
             />
           </template>
           <template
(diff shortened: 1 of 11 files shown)


hwiem (approved):
FE code lgtm, thank you!

bakridemos (approved):
lgtm, tested too thx

PR #6746, opened 2026-08-25
feat: add `feature_map_scales` permission

hwiem:
The permission controls whether it is possible to define scales for the public detail map in the admin area. Disabled by default, enabled in separate project PRs.

### PR Checklist

- [x] Create/Update tests


tests/backend/core/Core/Unit/Logic/PermissionsTest.php
```diff
@@ -367,6 +367,7 @@ public function permissionsTests(): array
                     'feature_map_deactivate',
                     'feature_map_layer_contextual_help',
                     'feature_map_new_statement',
+                    'feature_map_scales',
```
bakridemos: I think the permissionstest should not be adjuted here, this permission is not enabled in core permissions.

bakridemos (approved):

bakridemos (changes requested):
I think the permissionstest should not be adjuted here, this permission is not enabled in core permissions.

tests/backend/core/Core/Unit/Logic/PermissionsTest.php
hwiem: Done in 8a3f5c9524b9670ffb346c5f36bb92ee72268576.
To be sure I understand correctly (since I'm not 100% sure how permissions are handled in the BE): permissions are disabled in the core by default, then enabled in the projects that should have the feature, and tests only test if the permission is enabled for the desired roles, but we do not test if it is disabled for roles that shouldn't have the feature?

bakridemos (approved):

PR #6758, opened 2026-08-27
F dplan 18189 procedure export statement metadata

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-18189

Description: 

Requirement: in the procedure export ("Verfahrensexport"), the statement list DOCX documents should additionally show each statement's assigned Potenzialflächen and Schlagworte, directly below the statement text. This must not affect the standalone Abwägungstabelle export — the client explicitly wants the two kept separate.

What changed

The Verfahrensexport's Abwägungstabelle DOCX documents (Abwaegungstabelle_Liste.docx, _mit_Datensaetzen.docx, _Anonym.docx, _mit_Datensaetzen_Anonym.docx) and the standalone Abwägungstabelle export share the same rendering pipeline (AssessmentHandler/AssessmentTableDocxExporter → AssessmentTableServiceOutput::generateDocx → DocxExporter). Since they converge on one code path, a new opt-in flag ($includeStatementMetadataRow, default false) was threaded through the chain, set to true only at the two ExportService call sites (addAssessmentTableToZip, addAssessmentTableAnonymousToZip) that build the Verfahrensexport documents.

When the flag is set, DocxExporter::renderTableItem() appends a bold "Potenzialflächen: …" line and a bold "Schlagworte: …" line under the statement text in the "Stellungnahme" cell (each line only rendered if the statement actually has values). The underlying data (priorityAreaKeys, tagNames) was already available in the statement array used by this pipeline — it just wasn't copied into DocxExporter::formatStatementArray()'s output before.

The standalone Abwägungstabelle export (AssessmentTableDocxExporter) calls the same methods with its original argument list, so the flag defaults to false there and its output is unchanged.

**Additional fix: SQLite functional test schema rebuild**

Unrelated to the export feature, but found while running functional tests on this branch: any test using ORMSqliteDatabaseTool (Liip TestFixturesBundle) failed with SQLSTATE[HY000]: General error: 1 table messenger_messages already exists.

Cause: config/packages/doctrine.yaml excludes messenger_messages from schema_filter on the dplan connection to prevent spurious migration diffs against the real database (that table is owned by Symfony Messenger's Doctrine transport, not an ORM entity). That filter only affects introspection of the live database, not schema generation. The SQLite test-schema rebuild does both:
 - SchemaTool::dropDatabase() introspects the real DB (filter applies) → messenger_messages is invisible → never dropped.
 - SchemaTool::createSchema() builds from ORM metadata + Doctrine event listeners, and MessengerTransportDoctrineSchemaListener re-adds messenger_messages unconditionally (filter doesn't apply here) → CREATE TABLE fails because the table is still there.

Fix: null schema_filter for the test-env dplan connection only (config/packages/doctrine.yaml, when@test block). Dev/prod keep the filter, so real migration-diff behavior is unchanged.

 **Additional cleanup: SonarCloud findings on renderTableItem()**

Two SonarCloud issues were flagged on DocxExporter::renderTableItem() after the metadata-row change above:
 - php:S3776 — Cognitive Complexity 17 > 15. The method had a nested if (moved statement) {...} else { if (fragments) {...} else { if/if/if... } } structure. Flattened it into three sequential guard clauses inside the same method, each ending in an early return; (moved-statement check → fragment-rows check → regular statement row). No new methods were extracted and no new type hints were introduced — every branch now sits at nesting level 0 instead of being stacked inside the others, which is what drives the cognitive-complexity score down (roughly 17 → 6–7).

 - php:S107 — 8 parameters > 7. The three rendering-option flags ($numberStatements, $statementNumber, $includeStatementMetadataRow) are now bundled into a single array $renderOptions = [], destructured back into the same local variable names as the first lines of the method body. The rest of the method is unchanged. One of the two call sites needed no change (it already relied on defaults for these args); the other now passes the options array instead of 3 positional args.

Both changes are structural only — no behavior change, verified by re-running the procedure export manually after each step.

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php
@@ -408,7 +408,8 @@ class ExportService
                 $rParams,
                 $type,
                 AssessmentTableViewMode::DEFAULT_VIEW,
-                false
+                false,
+                true
             );
             $filename = $procedureName.'/'.$this->literals['statements'].'/'.$this->literals['considerationtable'].'/%s.docx';
             switch ($exportType) {
@@ -458,7 +459,8 @@ class ExportService
                 $rParams,
                 $type,
                 AssessmentTableViewMode::DEFAULT_VIEW,
-                false
+                false,
+                true
             );
             $filename = $procedureName.'/'.$this->literals['statements'].'/'.$this->literals['considerationtable'].'/%s.docx';
             switch ($exportType) {
--- a/demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php
@@ -408,8 +408,7 @@ class ExportService
                 $rParams,
                 $type,
                 AssessmentTableViewMode::DEFAULT_VIEW,
-                false,
-                true
+                includeStatementMetadataRow:  true
             );
             $filename = $procedureName.'/'.$this->literals['statements'].'/'.$this->literals['considerationtable'].'/%s.docx';
             switch ($exportType) {
--- a/demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php
@@ -408,7 +408,7 @@ class ExportService
                 $rParams,
                 $type,
                 AssessmentTableViewMode::DEFAULT_VIEW,
-                includeStatementMetadataRow:  true
+                includeStatementMetadataRow: true
             );
             $filename = $procedureName.'/'.$this->literals['statements'].'/'.$this->literals['considerationtable'].'/%s.docx';
             switch ($exportType) {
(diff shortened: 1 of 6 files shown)


demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php
```diff
@@ -408,7 +408,8 @@ public function addAssessmentTableToZip(
                 $rParams,
                 $type,
                 AssessmentTableViewMode::DEFAULT_VIEW,
-                false
+                false,
+                true
```
mussbach:
```suggestion
                includeStatementMetadataRow:  true
```

demosplan/DemosPlanCoreBundle/Logic/Procedure/ExportService.php
```diff
@@ -458,7 +459,8 @@ public function addAssessmentTableAnonymousToZip(
...
```
mussbach:
```suggestion
                includeStatementMetadataRow:  true
```

mussbach (changes requested):
Metadata is not shown on the fragments export. Oderwise lgtm, two minor suggestions

bakridemos:
> Metadata is not shown on the fragments export. Oderwise lgtm, two minor suggestions

done in https://github.com/demos-europe/demosplan-core/pull/6758/changes/ac04312930bcf5c63f88fb3c618482548a87c78d , thx

mussbach (approved):
lgtm, thx

PR #6767, opened 2026-08-31
feat(DPLAN-18213): First draft to moving CustomField get to 3.0 API

sierrazamudiodemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-18213/ADO-56769-Zusatzlicher-Filter-Konfigurierbare-Felder-Abschnittsliste
## What this PR does

Adds a new, read-only API Platform endpoint for custom fields: `GET /api/3.0/CustomField` (list) and `GET /api/3.0/CustomField/{id}` (single). Create/update/delete stay on the old endpoint for now.

## BE changes
**1. Client picks which fields it wants back → Sparse Fieldset**

If the frontend only needs `name` and `fieldType`, it can ask for just that:
GET /api/3.0/CustomField?fields[CustomField]=name,fieldType
This uses API Platform's built-in `SparseFieldset` filter. We just declare it, no custom code needed.

**2. Some filters must always be sent → Search Filter (ExactFilter), marked required**

A list of custom fields only makes sense scoped to something (e.g. "STATEMENT fields of procedure X"). Without that scope, we'd return every custom field in the whole app. So `sourceEntity` and `targetEntity` are declared as **required** filters:
GET /api/3.0/CustomField?sourceEntity=PROCEDURE&targetEntity=SEGMENT&sourceEntityId=...
Missing one of these → the request is rejected automatically (422), before our own code even runs.

Both of these are declared in one place: the `parameters` list on the `GetCollection` operation in `CustomFieldResource.php`. Nothing extra to configure per-property.

**3. Access control stays as it was**

Who's allowed to see which rows (e.g. customer-scoped fields) is unchanged — same rules as before, just plugged into the new endpoint.

**4. Note:**
One filter (`SparseFieldset`) needs a small service registration in `config/services.yml` — it's a vendor class not covered by our own autodiscovery, and Symfony (unlike Laravel) doesn't wire it up automatically. Explained in the file comment.

## FE changes

**1. Ask for only the fields we need → matches the Sparse Fieldset filter**

We already know which fields the UI needs, so we send them in the request:
```js
fields: {
  CustomField: ['name', 'description', 'options', 'fieldType', 'isRequired'].join(),
}
This is the same fields[CustomField]=... format the backend's SparseFieldset filter understands.

2. Send the required filters as simple params → matches the required Search Filter

The old code wrapped filters in a nested filter: { sourceEntity: { condition: { path, value } } } object (the old JSON:API style). The new endpoint just wants flat key/value pairs:
{
  sourceEntity: 'PROCEDURE',
  targetEntity: 'SEGMENT',
  sourceEntityId: '...',
}
Simpler, and it matches what the backend now requires.

3. New URL

Old: built via Routing.generate('api_resource_list', { resourceType: 'CustomField' }).
New: hardcoded /api/3.0/CustomField — this new API Platform route isn't a named Symfony route, so Routing.generate can't build it.

4. Small bug fix found along the way

One place (StatementModal.vue, the "new statement" form) was missing source-entity/target-entity on a component, so it never sent a scope at all. Added the missing props.

Files changed

- client/js/composables/useCustomFieldDefinitions.js
- client/js/components/statement/publicStatementModal/StatementModal.vue

### CODE REVIEW

**Review by rafelddemos — COMMENTED** [review]
Thanks for the FE changes! :)
I tested it, but unfortunately get the error '`Ein oder mehrere konfigurierbare Felder konnten nicht geladen werden`' on page load. And the selected CF are not displayed in the last step of the statement modal. Does it work for you? I guess it's because:
```
 source-entity="PROCEDURE"
 target-entity="STATEMENT"
```
is missing for `<custom-field>` in `StatementModalRecheck.vue`.

File: client/js/components/statement/publicStatementModal/StatementModal.vue [hunk 1]
```diff
@@ -1331,40 +1333,16 @@ export default {
       }
 
       try {
-        const url = Routing.generate('api_resource_list', {
-          resourceType: 'CustomField',
-        })
+        // Not a Symfony-named route: API Platform resources aren't exposed to Routing.generate().
+        const url = '/api/3.0/CustomField'
```
rafelddemos, line 1337: shouldn't it be 

```suggestion
        const url = `${Routing.getBaseUrl()}/api/3.0/CustomField`
```
here?

client/js/composables/useCustomFieldDefinitions.js
```diff
@@ -94,44 +94,20 @@ export function useCustomFieldDefinitions () {
       return pendingFetches.get(cacheKey)
     }
 
-    const url = Routing.generate('api_resource_list', {
-      resourceType: 'CustomField',
-    })
+    // Not a Symfony-named route: API Platform resources aren't exposed to Routing.generate().
+    const url = '/api/3.0/CustomField'
```
rafelddemos, line 98: shouldn't it be 

```suggestion
    const url = `${Routing.getBaseUrl()}/api/3.0/CustomField`
```
here?

PR #6768, opened 2026-08-31
Feat dplan 18271 link used boilerplates to corresponding segments

riechedemos:
### Ticket
DPLAN-18271

Links a boilerplate ("Textbaustein") to the recommendation text it was inserted into, instead of letting it become an untracked copy. A linked boilerplate renders as its own protected block in the editor, keeps working when future edits are made to the boilerplate elsewhere, and can be explicitly unlinked (with an undo option).

Key pieces:
- New Tiptap node represents a linked boilerplate as a distinct, hover-highlighted block with a pencil icon to unlink it. Text can be typed before/after the block, but its content is protected from accidental edits.
- The database only ever stores an empty reference tag (`<dp-boilerplate boilerplate-id="…">`). The editor fills it with the boilerplate's current text on load and strips that content back out before saving - so a later edit to the boilerplate automatically shows up wherever it's linked, without copying text into every recommendation that uses it.
- Unlinking dissolves the block back into plain paragraphs and offers a 15s "undo" toast.

### How to review/test
- do a migration
- Open a segment's recommendation editor with permission to link boilerplates, insert one via the toolbar button, and confirm it renders as its own block.
- Type text before/after the block, save, reload the page, and confirm the link survives.
- Click the pencil, confirm the dialog, and check that the block dissolves into plain text; use "undo" within 15s to restore it.
- Leave and re-enter edit mode without reloading the page, and confirm the content doesn't revert to a stale state.
- 
### Linked PR
https://github.com/demos-europe/demosplan-ui/pull/1566

### PR Checklist
- [x] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Update changelog


client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -159,20 +159,29 @@
         />
       </div>
       <div v-else>
+        <dp-loading
+          v-if="recommendationEmbeddedLoading"
+          data-cy="segmentEditor:loading"
```
hwiem, line 164: I think this could use some margin-bottom
```suggestion
        <dp-loading
          v-if="recommendationEmbeddedLoading"
          class="mb-2"
          data-cy="segmentEditor:loading"
```

client/js/components/procedure/StatementSegmentsList/utils/boilerplateTagContent.js
```diff
@@ -0,0 +1,56 @@
...
+ * (c) 2010-present DEMOS plan GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
+/**
+ * Fills empty `<dp-boilerplate boilerplate-id="…">` tags with the boilerplate's current text,
```
hwiem: Shouldn't this (and its corresponding test) file rather be in demosplan-ui?

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -1160,23 +1289,49 @@ export default {
...
-
-      if (boilerplateId && hasPermission('feature_boilerplate_usage_list')) {
-        return dpApi.post(
-          Routing.generate('dplan_boilerplate_usage_create', { procedureId: this.procedureId, boilerplateId }),
-          {},
-          { segmentId: this.segment.id },
-        ).catch(() => {
-          // Recording the usage is non-critical: the text was inserted regardless.
-        })
+    insertBoilerplateText (text, boilerplateId, insertBoilerplate, handleInsertText) {
```
hwiem, line 1313: Some of this code here and below should probably be extracted to a shared file, but maybe it's best to do that when we actually start using this feature in other views like the bulk edit.

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -1358,9 +1513,57 @@ export default {
...
+
+      const promises = [this.initBoilerplates()]
+
+      if (this.recommendationEmbedded === null) {
+        promises.push(this.loadRecommendationEmbedded())
+      }
+
+      Promise.all(promises).then(() => {
+        this.recommendationEmbeddedLoading = false
+      })
```
hwiem, line 1539: Shouldn't we add a catch block here?

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
```diff
@@ -1079,6 +1171,37 @@ export default {
...
+
+      this.$refs.editor.unlinkBoilerplate(pos)
+
+      const title = this.pendingUnlink.title || Translator.trans('boilerplate')
+
+      dplan.notify.confirm({
+        message: Translator.trans('boilerplate.link.dissolved', { title }),
+        actionText: Translator.trans('undo'),
+        hideTimer: 15000,
+        onAction: () => this.$refs.editor?.undo(),
```
hwiem, line 1201: Is this correct? Ultimately, `undo()` just undoes the last action, and if I unlink a boilerplate, then do something else, then click 'undo' in the notification, it will not undo the unlink, but the last action I did.

hwiem (commented):
Code looks good in general, just 2 small concerns

client/js/components/procedure/StatementSegmentsList/StatementSegment.vue
riechedemos: God idea I added it in the bulk edit ticket https://demoseurope.youtrack.cloud/issue/DPLAN-18261/Hinweis-und-Auflosen-von-Textbaustein-Verknupfung-in-der-Mehrfachbearbeitung-4

PR #6769, opened 2026-08-31
fix: Allow scalar or array hook options

graupnerdemos:
### Ticket
(No Ticket)

The manifest configuration parser failed on non-scalar options, however, other code like the permissions checking in `FrontendAssetProvider` already expects demosPlan core to be able to handle scalar and array options. This adds a variable node that is evaluated based on it's contents and allows both value types.

### How to review/test

Install an addon with defined hook permissions locally (--develop). It should not fail.


mussbach (approved):
lgtm, code review only

PR #6770, opened 2026-08-31
feat(DPLAN-18271): link boilerplates into recommendations via a reference in custom html tag

MoritzMandler:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-18271/Textbausteine-Erwiderungen-ver-und-entknupfen-Erwiderung-verfassen-View-1

### Description
Backend implementation for linking Textbausteine (boilerplates) into segment/statement recommendations ("Erwiderungen"). A caseworker inserts a boilerplate into a recommendation; the link stays intact while they edit around it, and can be explicitly dissolved via the editor's unlink flow.

Design: the DB column never stores boilerplate content — only a reference tag (<dp-boilerplate boilerplate-id="…">) at the insertion point. Every read path (exports, Elasticsearch, JSON:API, version history) substitutes the tag with the boilerplate's current text at read time; the setter parses tags on save and reconciles the link table accordingly. This avoids any content-drift comparison — link state is purely tag presence.

This PR is backend-only; the FE editor rework to match this tag format is tracked separately and isn't blocked by this PR, but the feature isn't end-to-end functional until that lands.

How to review/test

Large PR — recommend reviewing in this order rather than as one flat diff:

1. Substitution + accessors — BoilerplateTagSubstitutionService, Statement::getRecommendation()/getRecommendationEmbedded(), the postLoad listener wiring in services.yml.
2. Write-side reconciliation — BoilerplateUsageReconciliationService, BoilerplateUsage widened to cover both Statement and Segment.
3. Version history integrity — RecommendationVersionService (tag-form change detection, substituted snapshots) and, separately, EntityContentChangeService/EntityContentChangeDisplayService (the generic Versionsverlauf audit trail, an older/unrelated system also affected by this feature — fixed a phantom-diff bug and added a dedicated materialization notice).
4. Boilerplate deletion — BoilerplateDeletionService materializes content into every usage before the row is removed; runs asynchronously off the scheduler (PurgePendingBoilerplateDeletionsMessage), not inline in the delete request. Atomicity is covered by a test that forces a partial failure.
5. Copy/move across procedures — StatementCopier/StatementMover: three real bugs found and fixed here (raw clone() was bypassing the setter, silently dropping the link's reconciliation).
6. Legacy data migration — Version20260828094447 converts existing links into tag form where the text still matches, severs the relation otherwise.

Test suite: docker compose exec development su -l $USER -s /usr/bin/zsh -c 'cd /srv/www && SYMFONY_DEPRECATIONS_HELPER=disabled php -d xdebug.mode=off ./vendor/bin/phpunit tests/backend/core/Statement/Functional/Boilerplate* tests/backend/core/Procedure/Functional/Boilerplate* tests/backend/core/Core/Functional/EntityContentChange*'

PR Checklist

- [x] Create/Update tests
- [ ] Update documentation
- [ ] Update changelog


demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/08/Version20260828094447.php
```diff
@@ -0,0 +1,217 @@
...
+ *   text appearing more than once (not a realistic production case — a boilerplate's text
+ *   occurring twice in one recommendation has no legitimate reason to happen, so any
+ *   occurrence is treated as test data rather than something to guess at); or the
+ *   boilerplate's text being empty.
+ *
+ * Idempotent: a row already containing this boilerplate's tag is skipped entirely, so
+ * running this migration twice (or on a database where some rows were already migrated by
+ * hand) is safe.
+ */
+class Version20260828094447 extends AbstractMigration
```
mussbach, line 57: tbh: now that I see it i am a bit anxious. What if something goes wrong? Then the recommendation would be altered. I think we should not try to upgrade the existing text and only use the new mechanism at new inserts

PR #6771, opened 2026-08-31
f_DPLAN-18151: add systemwide information banner

riechedemos:
### Ticket
DPLAN-18151

Frontend part on top of #6710. Renders the `serverBanner` Twig variable as a dismissible warning banner on every page, including login.

- New `ServerBanner` Vue component (registered globally like `NotifyContainer`/`UnsavedChangesDialog`), mounted inside `#app` in `base.html.twig` and `base_oeb.html.twig`.
- Dismissal is session-scoped (`sessionStorage`), so the banner reappears on the next login/session, not permanently hidden like `localStorage`-based dismissals.

### How to review/test
1. Create a `SERVER_BANNER.md` file at the repository root (same level as `composer.json`), e.g.:
   ```
   **Downtime des Systems heute zwischen 12:00 und 12:15 Uhr**
   ```
2. Reload any page and the login page — banner should appear at the top, inside the yellow warning style.
3. Click the close (X) button — banner disappears; reload in the same tab — stays hidden.
4. Open a new session (e.g. new incognito window) — banner reappears.
5. Remove `SERVER_BANNER.md` — banner no longer renders.

### PR Checklist
- [x] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Update changelog

### DIFF
--- a/client/js/components/shared/ServerBanner.vue
+++ b/client/js/components/shared/ServerBanner.vue
@@ -1,0 +1,48 @@
+<license>
+  (c) 2010-present DEMOS plan GmbH.
+
+  This file is part of the package demosplan,
+  for more information see the license file.
+
+  All rights reserved
+</license>
+
+<template>
+  <div
+    v-if="isVisible"
+    :class="prefixClass('bg-message-warning text-message-warning border border-message-warning relative px-2 pt-2')"
+  >
+    <!-- message is server-controlled markdown (SERVER_BANNER.md via ServerBannerLoader), never user input -->
+    <div
+      v-html="message"
+    />
+    <dp-button
+      :class="prefixClass('absolute top-2 right-2')"
+      :text="Translator.trans('close')"
+      hide-text
+      icon="x"
+      variant="subtle"
+      @click="dismiss"
+    />
+  </div>
+</template>
+
+<script setup>
+import { DpButton, prefixClass } from '@demos-europe/demosplan-ui'
+import { ref } from 'vue'
+
+defineProps({
+  message: {
+    type: String,
+    required: true,
+  },
+})
+
+const storageKey = 'serverBannerDismissed'
+const isVisible = ref(sessionStorage.getItem(storageKey) === null)
+
+const dismiss = () => {
+  isVisible.value = false
+  sessionStorage.setItem(storageKey, '1')
+}
+</script>
--- a/client/js/components/shared/ServerBanner.vue
+++ b/client/js/components/shared/ServerBanner.vue
@@ -12,9 +12,9 @@
     v-if="isVisible"
     :class="prefixClass('bg-message-warning text-message-warning border border-message-warning relative px-2 pt-2')"
   >
-    <!-- message is server-controlled markdown (SERVER_BANNER.md via ServerBannerLoader), never user input -->
+    <!-- message originates from SERVER_BANNER.md; sanitize before rendering HTML -->
     <div
-      v-html="message"
+      v-html="sanitizedMessage"
     />
     <dp-button
       :class="prefixClass('absolute top-2 right-2')"
@@ -28,10 +28,11 @@
 </template>
 
 <script setup>
+import { computed, ref } from 'vue'
 import { DpButton, prefixClass } from '@demos-europe/demosplan-ui'
-import { ref } from 'vue'
+import DomPurify from 'dompurify'
 
-defineProps({
+const props = defineProps({
   message: {
     type: String,
     required: true,
@@ -40,6 +41,7 @@ defineProps({
 
 const storageKey = 'serverBannerDismissed'
 const isVisible = ref(sessionStorage.getItem(storageKey) === null)
+const sanitizedMessage = computed(() => DomPurify.sanitize(props.message))
 
 const dismiss = () => {
   isVisible.value = false
--- a/client/js/components/shared/ServerBanner.vue
+++ b/client/js/components/shared/ServerBanner.vue
@@ -12,7 +12,6 @@
     v-if="isVisible"
     :class="prefixClass('bg-message-warning text-message-warning border border-message-warning relative px-2 pt-2')"
   >
-    <!-- message originates from SERVER_BANNER.md; sanitize before rendering HTML -->
     <div
       v-html="sanitizedMessage"
     />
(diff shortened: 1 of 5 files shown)


client/js/components/shared/ServerBanner.vue
```diff
@@ -0,0 +1,48 @@
...
+</license>
+
+<template>
+  <div
+    v-if="isVisible"
+    :class="prefixClass('bg-message-warning text-message-warning border border-message-warning relative px-2 pt-2')"
+  >
+    <!-- message is server-controlled markdown (SERVER_BANNER.md via ServerBannerLoader), never user input -->
+    <div
+      v-html="message"
```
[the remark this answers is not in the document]
riechedemos: as the comments sais there is no input that could be abused.

client/js/components/shared/ServerBanner.vue
```diff
@@ -0,0 +1,48 @@
...
+    />
```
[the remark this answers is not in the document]
riechedemos: 36a86ecf5d6a2708c526a2586f6bd13ae7207b4c

client/js/components/shared/ServerBanner.vue
```diff
@@ -0,0 +1,50 @@
...
+</license>
+
+<template>
+  <div
+    v-if="isVisible"
+    :class="prefixClass('bg-message-warning text-message-warning border border-message-warning relative px-2 pt-2')"
+  >
+    <!-- message originates from SERVER_BANNER.md; sanitize before rendering HTML -->
+    <div
+      v-html="sanitizedMessage"
```
sakutademos: yes, exactly. You can use the v-cleanhtml directive from our demospla-ui library instead

client/js/components/shared/ServerBanner.vue
```diff
@@ -0,0 +1,50 @@
...
+const props = defineProps({
+  message: {
+    type: String,
+    required: true,
+  },
+})
+
+const storageKey = 'serverBannerDismissed'
+const isVisible = ref(sessionStorage.getItem(storageKey) === null)
+const sanitizedMessage = computed(() => DomPurify.sanitize(props.message))
```
sakutademos: the v-cleanhtml directive from demosplan-ui already sanitizes html, so you do not need to call DOMPurify here

sakutademos (changes requested):
i am a bit concerned about one thing in the current implementation: what happens if there is more than one server message? You use `serverBannerDismissed` which only stores whether the server banner has been dismissed.

Once it has been dismissed, **any** server banner will be hidden for the entire session, even if the message content changes.

client/js/components/shared/ServerBanner.vue
```diff
@@ -0,0 +1,50 @@
...
+    <!-- message originates from SERVER_BANNER.md; sanitize before rendering HTML -->
+    <div
+      v-html="sanitizedMessage"
+    />
+    <dp-button
+      :class="prefixClass('absolute top-2 right-2')"
+      :text="Translator.trans('close')"
+      hide-text
+      icon="x"
+      variant="subtle"
```
hwiem: I think it looks a bit weird to have a white button on the yellow background, I'd make it transparent
```suggestion
      variant="transparent"
```

client/js/components/shared/ServerBanner.vue
```diff
@@ -0,0 +1,50 @@
...
+  for more information see the license file.
+
+  All rights reserved
+</license>
+
+<template>
+  <div
+    v-if="isVisible"
+    :class="prefixClass('bg-message-warning text-message-warning border border-message-warning relative px-2 pt-2')"
+  >
```
hwiem, line 15: screen readers should announce this, too
```suggestion
  <div
    v-if="isVisible"
    :class="prefixClass('bg-message-warning text-message-warning border border-message-warning relative px-2 pt-2')"
    role="alert"
  >
```

client/js/components/shared/ServerBanner.vue
```diff
@@ -0,0 +1,50 @@
...
+  },
+})
+
+const storageKey = 'serverBannerDismissed'
+const isVisible = ref(sessionStorage.getItem(storageKey) === null)
+const sanitizedMessage = computed(() => DomPurify.sanitize(props.message))
+
+const dismiss = () => {
+  isVisible.value = false
+  sessionStorage.setItem(storageKey, '1')
```
hwiem: With this implementation, once the message is dismissed, it's dismissed (for the period of that session). I think we should make sure that if the message content changes, the banner is displayed again, e.g. by storing the message as the value of the storageKey, or as a separate entry in the storage?

hwiem (dismissed):

client/js/components/shared/ServerBanner.vue
riechedemos: Done in 9064bfe7e — switched to `variant="transparent"`.

client/js/components/shared/ServerBanner.vue
riechedemos: Done in 9064bfe7e — added `role="alert"`.

client/js/components/shared/ServerBanner.vue
riechedemos: Done in 9064bfe7e — the stored value is now the message itself, so a changed banner text reappears even if an older version was dismissed this session.

riechedemos:
@sakutademos Same underlying issue as hwiem's comment below — fixed in 9064bfe7e by storing the message itself as the storage value instead of a boolean flag, so a changed message reappears even if an older one was dismissed this session.

hwiem (commented):
My concerns have been addressed and I tested it and it all looks good and works 🎉  but there's still a concern from @sakutademos that needs to be addressed 🙏

client/js/components/shared/ServerBanner.vue
riechedemos: Done in 06b73d0c0 — switched to `v-cleanhtml`.

client/js/components/shared/ServerBanner.vue
riechedemos: Done in 06b73d0c0 — removed the DOMPurify call, now handled by `v-cleanhtml`.

sakutademos (approved):
my concerns have been addressed as well, thx!

PR #6785, opened 2026-09-03
feat(refs DPLAN-18348): allow setting individual procedure creation permission

bakridemos:
### Ticket
https://demoseurope.youtrack.cloud/issue/DPLAN-18348

### Description
- Procedure creation (`feature_admin_new_procedure`) can be granted to individual Fachplaner-Admin (RMOPSA) and Anhörungsbehörde-Admin (RMOPHA) users via a checkbox in the user form, as an alternative to the org-wide grant. Gated by the new `feature_manage_user_procedure_creation_permission`, so masterusers can manage it without access to organisation settings.
- While the org-wide grant is active, the per-user checkbox is shown checked and disabled with a contextual hint. Enabling the org-wide grant removes existing individual grants.
- Toggling the org-wide checkbox requires confirmation via a dialog in both directions.
- New organisations and newly accepted Kommune/Anhörungsbehörde types get the org-wide grant by default, unless the admin unticks it in the same save. A type accepted later inherits the orga's current setting.
- Fixes in the existing org-wide toggle: disabling it no longer gets silently reverted on the next unrelated save, and RMOPHA is now covered alongside RMOPSA.
- Removing RMOPSA/RMOPHA from a user also removes their individual grant.

### Linked PRs
- demos-europe/demosplan-project-ewm#284 grants the permissions in EWM; without it neither checkbox appears there.

### How to review/test
- Organisation list: untick "Darf Verfahren anlegen" on an orga with accepted Kommune or Anhörungsbehörde type, confirm the dialog, save. Save again with an unrelated change, reload: still unticked.
- Approve a pending Kommune orga: checkbox appears ticked. Save, reload: ticked. Repeat with the checkbox unticked before saving: stays unticked.
- Orga with checkbox off, accept a second type: checkbox stays off.
- User form, as support and as masteruser: with the org-wide grant off, assign RMOPSA or RMOPHA, toggle "Darf Verfahren anlegen", confirm the individual grant is created and removed. With the org-wide grant on, the checkbox is checked and disabled.
- Enable the org-wide grant on an orga with individual grants: the "Nutzende mit der Berechtigung" list empties after save.
- Tests: `tests/backend/core/User/Functional/UserHandlerEnsureAccessControlTest.php`, `tests/backend/core/Logic/UserAccessControlServiceTest.php`, `tests/backend/core/Repository/UserAccessControlRepositoryTest.php`.

### PR Checklist
- [x] Create/Update tests
- [x] Run `yarn lint`
- [x] Update changelog

### DIFF
--- a/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
+++ b/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
@@ -1038,7 +1038,7 @@ export default {
     },
 
     isPlanningOfficeOrMunicipalityAcceptedOrPending () {
-      return this.registrationStatuses.some(registration => (registration.status === 'accepted' || registration.status === 'pending') && (registration.type === 'OPAUTH' || registration.type === 'OLAUTH'))
+      return this.registrationStatuses.some(registration => (registration.status === 'accepted' || registration.status === 'pending') && (registration.type === 'OPAUTH' || registration.type === 'OLAUTH' || registration.type === 'OHAUTH'))
     },
 
     /**
--- a/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
+++ b/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
@@ -952,7 +952,8 @@ export default {
         return {
           attributes: {
             addressExtension: '',
-            canCreateProcedures: false,
+            // New organizations get the orga-wide procedure-creation grant unless the admin opts out
+            canCreateProcedures: true,
             ccEmail2: '',
             city: '',
             dataProtection: '',
--- a/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
+++ b/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
@@ -392,7 +392,7 @@
           :label="{
             text: Translator.trans('procedure.canCreate')
           }"
-          @change="emitOrganisationUpdate"
+          @change="onCanCreateProceduresChange"
         />
 
         <template v-if="hasPermission('feature_manage_procedure_creation_permission') && isMunicipalityOrHearingAuthorityAccepted && !localOrganisation.attributes.canCreateProcedures">
@@ -1024,6 +1024,7 @@ export default {
     'addon:update',
     'addonOptions:loaded',
     'organisation:update',
+    'procedureCreationToggled',
     'reset:complete',
   ],
 
@@ -1164,11 +1165,34 @@ export default {
       })
     },
 
+    /**
+     * Emitted in addition to the regular organization:update, so the parent can track the toggle
+     * independently of the store - a background refresh of the organizations list (triggered by
+     * an unrelated save elsewhere) can overwrite the `organization` prop with server state while
+     * this one is still being edited, which would otherwise make an in-progress enable/disable
+     * detection based on that prop flicker back to "unchanged".
+     */
+    onCanCreateProceduresChange (checked) {
+      this.emitOrganisationUpdate()
+      // Under @vue/compat this listener runs before the legacy v-model write, so localOrganisation is still stale here
+      this.$emit('procedureCreationToggled', checked)
+    },
+
     async fetchUsersWithIndividualProcedureCreationPermission () {
       const url = Routing.generate('dplan_api_organisation_procedure_creation_individual_grants', { id: this.organisation.id })
       const response = await dpApi.get(url)
 
       this.usersWithIndividualProcedureCreationPermission = response.data.data || []
+    },
+
+    /**
+     * Refetches the individual-grants list whenever the organization's data changes (e.g. after a
+     * save) so the list shown below the org-wide checkbox is updated
+     */
+    fetchUsersWithIndividualProcedureCreationPermissionIfNeeded () {
+      if (hasPermission('feature_manage_procedure_creation_permission') && this.isMunicipalityOrHearingAuthorityAccepted) {
+        this.fetchUsersWithIndividualProcedureCreationPermission()
+      }
     },
 
     hasChanged (field) {
@@ -1233,10 +1257,7 @@ export default {
 
   created () {
     this.setInitialOrganisation()
-
-    if (hasPermission('feature_manage_procedure_creation_permission') && this.isMunicipalityOrHearingAuthorityAccepted) {
-      this.fetchUsersWithIndividualProcedureCreationPermission()
-    }
+    this.fetchUsersWithIndividualProcedureCreationPermissionIfNeeded()
   },
 
   mounted () {
--- a/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
+++ b/client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
@@ -325,7 +325,7 @@
 
           <div
             v-if="showAddStatusForm"
-            class="layout"
+            class="layout flex items-end"
           >
             <!-- Select row  -->
             <div class="layout__item u-1-of-4">
@@ -362,7 +362,7 @@
             </div>
 
             <!-- Button row  -->
-            <div class="layout__item u-1-of-2 u-mt-0_5 space-inline-m">
+            <div class="layout__item u-1-of-2 space-inline-m">
               <button
                 class="btn btn--primary"
                 data-cy="orgaFormField:saveNewRegistrationStatus"
@@ -384,17 +384,61 @@
         </template>
 
         <dp-checkbox
-          v-if="hasPermission('feature_manage_procedure_creation_permission') && isPlanningOfficeOrMunicipalityAcceptedOrPending"
+          v-if="hasPermission('feature_manage_procedure_creation_permission') && isMunicipalityOrHearingAuthorityAccepted"
           :id="`${organisation.id}:procedureCreatePermission`"
           v-model="localOrganisation.attributes.canCreateProcedures"
           class="mt-2"
           data-cy="orgaFormField:procedureCreatePermission"
           :label="{
-            text: Translator.trans('procedure.canCreate'),
-            bold: true
+            text: Translator.trans('procedure.canCreate')
           }"
           @change="emitOrganisationUpdate"
         />
+
+        <template v-if="hasPermission('feature_manage_procedure_creation_permission') && isMunicipalityOrHearingAuthorityAccepted && !localOrganisation.attributes.canCreateProcedures">
+          <p
+            v-if="usersWithIndividualProcedureCreationPermission.length === 0"
+            class="mt-1 lbl__hint"
+            data-cy="orgaFormField:procedureCreatePermission:hintEmpty"
+          >
+            {{ Translator.trans('procedure.canManage.orga.none') }}
+          </p>
+          <template v-else>
+            <p
+              class="mt-1 u-mb-0_25"
+              data-cy="orgaFormField:procedureCreatePermission:usersLabel"
+            >
+              {{ Translator.trans('procedure.canManage.orga.users') }}
+            </p>
+            <ul
+              class="list-disc u-ml-0_75 u-mb-0"
+              data-cy="orgaFormField:procedureCreatePermission:usersList"
+            >
+              <li
+                v-for="user in visibleProcedureCreationUsers"
+                :key="user.id"
+                class="u-mb-0_25"
+              >
+                {{ user.firstname }} {{ user.lastname }}
+              </li>
+            </ul>
+            <dp-details
+              v-if="collapsedProcedureCreationUsers.length > 0"
+              data-cy="orgaFormField:procedureCreatePermission:usersMore"
+              :summary="Translator.trans('procedure.canManage.orga.users.more', { count: collapsedProcedureCreationUsers.length })"
+            >
+              <ul class="list-disc u-ml-0_75 u-mb-0">
+                <li
+                  v-for="user in collapsedProcedureCreationUsers"
+                  :key="user.id"
+                  class="u-mb-0_25"
+                >
+                  {{ user.firstname }} {{ user.lastname }}
+                </li>
+              </ul>
+            </dp-details>
+          </template>
+        </template>
       </div>
 
       <div
@@ -851,8 +895,9 @@
 </template>
 
 <script>
-import { CleanHtml, DpCheckbox, DpDetails, DpEditor, DpSelect, DpTextArea, hasOwnProp } from '@demos-europe/demosplan-ui'
+import { CleanHtml, dpApi, DpCheckbox, DpDetails, DpEditor, DpSelect, DpTextArea, hasOwnProp } from '@demos-europe/demosplan-ui'
 import AddonWrapper from '@DpJs/components/addon/AddonWrapper'
+import { isOrgaAcceptedAsType } from '@DpJs/lib/shared/isOrgaAcceptedAsType'
 
 export default {
   name: 'DpOrganisationFormFields',
@@ -999,11 +1044,13 @@ export default {
           label: Translator.trans('rejected'),
         },
       ],
+      procedureCreationUsersVisibleLimit: 5,
       showAddStatusForm: false,
       statusForm: {
         status: 'pending',
         type: 'TöB',
       },
+      usersWithIndividualProcedureCreationPermission: [],
     }
   },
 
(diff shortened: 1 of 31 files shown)


client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
```diff
@@ -384,17 +384,61 @@
...
-            bold: true
+            text: Translator.trans('procedure.canCreate')
           }"
-          @change="emitOrganisationUpdate"
+          @change="onCanCreateProceduresChange"
         />
+
+        <template v-if="hasPermission('feature_manage_procedure_creation_permission') && isMunicipalityOrHearingAuthorityAccepted && !localOrganisation.attributes.canCreateProcedures">
+          <p
+            v-if="usersWithIndividualProcedureCreationPermission.length === 0"
```
sakutademos, line 400: could be a bit simplified

```suggestion
            v-if="!usersWithIndividualProcedureCreationPermission.length"
```

client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
```diff
@@ -384,17 +384,61 @@
...
+              <li
+                v-for="user in visibleProcedureCreationUsers"
+                :key="user.id"
+                class="u-mb-0_25"
+              >
+                {{ user.firstname }} {{ user.lastname }}
+              </li>
+            </ul>
+            <dp-details
+              v-if="collapsedProcedureCreationUsers.length > 0"
```
sakutademos, line 426:
```suggestion
              v-if="collapsedProcedureCreationUsers.length"
```

client/js/components/user/DpOrganisationList/DpOrganisationListItem.vue
```diff
@@ -185,9 +202,116 @@ export default {
     icon () {
       return this.isOpen ? 'chevron-up' : 'chevron-down'
     },
+
+    isEnablingProcedureCreation () {
+      const current = this.canCreateProceduresCurrent ?? this.organisation.attributes.canCreateProcedures
+
+      return current === true && this.canCreateProceduresOnOpen !== true
```
sakutademos, line 209: could it be simplified? 

```suggestion
      return current && !this.canCreateProceduresOnOpen
```

client/js/components/user/DpOrganisationList/DpOrganisationListItem.vue
```diff
@@ -185,9 +202,116 @@ export default {
...
+    },
+
+    isDisablingProcedureCreation () {
+      const current = this.canCreateProceduresCurrent ?? this.organisation.attributes.canCreateProcedures
+
+      return current !== true && this.canCreateProceduresOnOpen === true
```
sakutademos, line 215: same here

translations/messages+intl-icu.de.yml
```diff
@@ -2875,7 +2874,22 @@ procedure.by.phase.public: |
...
         other {# Verfahren nach Phase (Öffentlichkeit)}
     }
-procedure.canCreate: 'Kommune oder Planungsbüro darf Verfahren anlegen'
+procedure.canCreate: 'Alle Nutzenden der Organisation mit der Rolle Fachplanung-Admin bzw. Anhörungsbehörde-Admin können Verfahren anlegen'
+procedure.canCreate.confirmDisable.header: 'Berechtigung deaktivieren?'
+procedure.canCreate.confirmDisable.message: "Die Berechtigung, Verfahren anzulegen, ist in dieser Organisation bisher allen Nutzenden mit der Rolle Fachplanung-Admin bzw. Anhörungsbehörde-Admin zugewiesen.\nWenn Sie diese Berechtigung deaktivieren, können Sie sie einzelnen Nutzenden in der Nutzendenverwaltung erteilen. Möchten Sie fortfahren?"
+procedure.canCreate.confirmEnable.header: 'Berechtigung aktivieren?'
+procedure.canCreate.confirmEnable.message: "Die Berechtigung, Verfahren anzulegen, ist in dieser Organisation möglicherweise nur einzelnen Personen mit der Rolle Fachplanung-Admin bzw. Anhörungsbehörde-Admin zugewiesen.\nWenn Sie die Berechtigung für alle Nutzenden der Organisation mit der Rolle Fachplanung-Admin aktivieren, können Sie sie einzelnen Nutzenden nicht mehr entziehen. Möchten Sie fortfahren?"
```
sakutademos, line 2881: in the first part of the text, two roles are mentioned, while the second part only refers to one role. That feels a bit inconsistent/confusing

sakutademos (approved):
it works and FE code looks nice! thank you! just small nitpicks

bakridemos (commented):
@hwiem While reworking the visibility check for the "Darf Verfahren anlegen" checkbox (commit 63ba163e2 on this branch), the Planungsbüro org type (OPAUTH) was accidentally dropped from the isMunicipalityOrHearingAuthorityAccepted() condition in DpOrganisationFormFields.vue — only Kommune (OLAUTH) and Anhörungsbehörde (OHAUTH) remain. The backend still fully supports Planungsbüro (there's even a dedicated test proving it).

Why this needs to be fixed: This Vue component lives in core (client/js/) and is shared by all projects — no project overrides it. So any change here affects every project, not just ewm. Concretely impacted: blp, bobsh, and robobsh — all three have the Support role actively using this permission, and all three have real Planungsbüro orgs. For Planungsbüro-only orgs there, the checkbox simply no longer renders — Support staff can no longer manage this permission through the UI (only via console command as a workaround). ewm itself is unaffected since it has no Planungsbüro orgs anymore — which is also why this wasn't caught during testing there, even though the regression reached ewm too through the shared core component.

Fix: Add 'OPAUTH' back to the type list in the same file:
isMunicipalityOrHearingAuthorityAccepted () {
  return isOrgaAcceptedAsType(this.registrationStatuses, ['OLAUTH', 'OHAUTH', 'OPAUTH'])
}
May the method 'isMunicipalityOrHearingAuthorityAccepted' name should be adjusted too.


This is the only change needed in this PR. No core translation change is required — the current wording of procedure.canCreate is a fine default for projects without Planungsbüro. **But** blp, bobsh, and robobsh should instead add their own project-level override of that translation key (in their own repos) to mention Planungsbüro again, the same way ewm already overrides this key for its own context — that's a separate, small follow-up outside this PR.

client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
hwiem: Done in e128e6d97c5fee2f7e1a963b6530805c1edcd231

client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
hwiem: Done in e128e6d97c5fee2f7e1a963b6530805c1edcd231

client/js/components/user/DpOrganisationList/DpOrganisationListItem.vue
hwiem: Done in e128e6d97c5fee2f7e1a963b6530805c1edcd231

client/js/components/user/DpOrganisationList/DpOrganisationListItem.vue
hwiem: Done in e128e6d97c5fee2f7e1a963b6530805c1edcd231

translations/messages+intl-icu.de.yml
hwiem: Thanks for spotting this! Fixed in 1bdcf97058660323557ab1470e1dd5142817c441

hwiem:
> @hwiem While reworking the visibility check for the "Darf Verfahren anlegen" checkbox (commit [63ba163](https://github.com/demos-europe/demosplan-core/commit/63ba163e23d116c380f01e2b0f3466f581c4e308) on this branch), the Planungsbüro org type (OPAUTH) was accidentally dropped from the isMunicipalityOrHearingAuthorityAccepted() condition in DpOrganisationFormFields.vue — only Kommune (OLAUTH) and Anhörungsbehörde (OHAUTH) remain. The backend still fully supports Planungsbüro (there's even a dedicated test proving it).
> 
> Why this needs to be fixed: This Vue component lives in core (client/js/) and is shared by all projects — no project overrides it. So any change here affects every project, not just ewm. Concretely impacted: blp, bobsh, and robobsh — all three have the Support role actively using this permission, and all three have real Planungsbüro orgs. For Planungsbüro-only orgs there, the checkbox simply no longer renders — Support staff can no longer manage this permission through the UI (only via console command as a workaround). ewm itself is unaffected since it has no Planungsbüro orgs anymore — which is also why this wasn't caught during testing there, even though the regression reached ewm too through the shared core component.
> 
> Fix: Add 'OPAUTH' back to the type list in the same file: isMunicipalityOrHearingAuthorityAccepted () { return isOrgaAcceptedAsType(this.registrationStatuses, ['OLAUTH', 'OHAUTH', 'OPAUTH']) } May the method 'isMunicipalityOrHearingAuthorityAccepted' name should be adjusted too.
> 
> This is the only change needed in this PR. No core translation change is required — the current wording of procedure.canCreate is a fine default for projects without Planungsbüro. **But** blp, bobsh, and robobsh should instead add their own project-level override of that translation key (in their own repos) to mention Planungsbüro again, the same way ewm already overrides this key for its own context — that's a separate, small follow-up outside this PR.

Thanks! I didn't realize Planungsbüro had to be supported, too. I restored the checkbox for that role in 379737bc6b63951b6acfeacc82739bcaa6864b80; the per-user list below it stays limited to Kommune/Anhörungsbehörde since individual grants only exist for those admin roles.

hwiem:
> This is the only change needed in this PR. No core translation change is required — the current wording of procedure.canCreate is a fine default for projects without Planungsbüro. **But** blp, bobsh, and robobsh should instead add their own project-level override of that translation key (in their own repos) to mention Planungsbüro again, the same way ewm already overrides this key for its own context — that's a separate, small follow-up outside this PR.

Thanks for pointing this out, I actually decided to restore the original translation key in the core and override it in ewm (9d3bab7ee6e50832c62343d3e3b17be2fa3b3a65, 3826ef48a438e1b41b90c7f44225be338eb15dde), and I added permission checks for the list of users with individual procedure creation permission and the toggle warning (1e684e44c23d5387c6da6af6bd0bd6f8642688a4), because these should only be visible if procedure creation permission can be granted to individual users.

bakridemos (commented):
lgtm, thx. 
Verified the fix for the Planungsbüro (OPAUTH) visibility regression on the org-level "Darf Verfahren anlegen" checkbox — the split between isProcedureCreationOrgaTypeAccepted and the individual-grant gating is correct, the added feature_manage_user_procedure_creation_permission check on the hint/confirm-dialog is a good catch, and the ewm translation override matches the reverted core wording. Diff scope is unchanged, no debug leftovers.

Follow-up needed (not blocking this PR): blp, bobsh, and robobsh each grant feature_manage_procedure_creation_permission to a role on real Planungsbüro orgas, so with core's procedure.canCreate wording reverted to mention Planungsbüro again, each of these three projects should add their own translation override for procedure.canCreate in their respective repos — same approach ewm just used — otherwise their users will see the generic core wording that no longer matches their org types shown. Please track this as a separate task in each of those repos.

client/js/components/user/DpOrganisationList/DpOrganisationListItem.vue
```diff
@@ -63,22 +64,34 @@
...
+      :header="procedureCreationToggleHeader"
+      icon="warning"
+      :message="procedureCreationToggleMessage"
+    />
   </dp-table-card>
 </template>
 
 <script>
-import { dpApi, DpButtonRow, DpIcon, dpValidateMixin } from '@demos-europe/demosplan-ui'
+import { DpButtonRow, DpConfirmDialog, DpIcon, dpValidateMixin } from '@demos-europe/demosplan-ui'
```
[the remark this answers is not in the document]
hwiem: Done in f23ed9eb81302bcc00331a6023a8be8eb5a65638

client/js/components/user/DpOrganisationList/DpOrganisationListItem.vue
```diff
@@ -185,9 +202,118 @@ export default {
...
+     * the form can still be saved unaffected. The dialog warns about individual grants, which only
+     * exist in projects exposing the per-user permission.
+     *
+     * @return {Promise<boolean>} whether the toggle (if any) was confirmed
+     */
+    async confirmProcedureCreationToggle () {
+      if (!hasPermission('feature_manage_user_procedure_creation_permission') ||
+        (!this.isEnablingProcedureCreation && !this.isDisablingProcedureCreation)) {
+        return true
+      }
```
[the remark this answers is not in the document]
hwiem: Intentional: both dialog texts explain the effect on individual grants in the user management, which only exist where the per-user permission is enabled, so showing the dialog without it would be misleading. Clarified in the PR description.

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
```diff
@@ -330,12 +332,59 @@ private function getAvailableOrgaRoles(Orga $preUpdateOrga): array
                 if (OrgaType::MUNICIPALITY === $orgaStatusInCustomer->getOrgaType()->getName()) {
                     $availableOrgaRoles[] = $this->roleHandler->getRoleByCode(RoleInterface::PLANNING_AGENCY_ADMIN);
                 }
+
+                if (OrgaType::HEARING_AUTHORITY_AGENCY === $orgaStatusInCustomer->getOrgaType()->getName()) {
+                    $availableOrgaRoles[] = $this->roleHandler->getRoleByCode(RoleInterface::HEARING_AUTHORITY_ADMIN);
```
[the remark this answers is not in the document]
hwiem: Done in 70f5243cdfb4b036a3606c957a09b27521f1b67b

demosplan/DemosPlanCoreBundle/Controller/User/DemosPlanOrganisationAPIController.php
```diff
@@ -330,12 +332,59 @@ private function getAvailableOrgaRoles(Orga $preUpdateOrga): array
...
+     */
+    #[DplanPermissions('feature_manage_procedure_creation_permission')]
+    #[Route(path: '/api/1.0/organisation/{id}/procedure-creation-permission/individual-grants', name: 'dplan_api_organisation_procedure_creation_individual_grants', options: ['expose' => true], methods: ['GET'])]
+    public function getIndividualProcedureCreationGrants(
+        string $id,
+        OrgaHandler $orgaHandler,
+        CustomerHandler $customerHandler,
+        UserAccessControlService $userAccessControlService,
+    ): APIResponse {
+        $orga = $orgaHandler->getOrga($id);
```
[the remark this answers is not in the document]
hwiem: Not an issue right now:
- The response is already scoped to the caller's customer. The query only returns grants that were written in the current customer, for roles the orga has accepted in that customer. For an orga that doesn't belong to the customer, the list is simply empty, so no names can leak across customers.
- The route permission (`feature_manage_procedure_creation_permission`) is only granted to platform support, which manages all organisations anyway. There is no role that may manage only its own orga and still holds this permission.

client/js/components/user/DpOrganisationList/DpOrganisationFormFields.vue
```diff
@@ -1171,6 +1257,7 @@ export default {
 
   created () {
     this.setInitialOrganisation()
+    this.fetchUsersWithIndividualProcedureCreationPermissionIfNeeded()
```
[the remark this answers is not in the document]
hwiem: Fixed in b893dc549247f8f76513c448175601143b12f4ed

demosplan/DemosPlanCoreBundle/ResourceTypes/AdministratableUserResourceType.php
```diff
@@ -356,14 +457,118 @@ public function updateEntity(string $entityId, EntityDataInterface $entityData):
...
+
+            if ($this->accessControlService->permissionExist(AccessControlService::CREATE_PROCEDURES_PERMISSION, $orga, $customer, [$roleCode])) {
+                continue;
+            }
+
+            if ($canManageProcedures) {
+                if (!$this->isOrgaAcceptedForRole($orga, $customer, $roleCode)) {
+                    continue;
+                }
+                $this->userAccessControlService->createUserPermission($user, AccessControlService::CREATE_PROCEDURES_PERMISSION, $role);
```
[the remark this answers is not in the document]
hwiem: Fixed in 0d5face6029f02c723d311d5b14388e4da322aaf

PR #6786, opened 2026-09-03
Refactor(DPLAN-18226): migrate segments list to API platform - FE part

sakutademos:
### Ticket
[DPLAN-18226](https://demoseurope.youtrack.cloud/issue/DPLAN-18226) Segment Resource Eingang-Nr edt to API Platfrom Ressource

**Description:**  This PR implements the FE migration of the segments list functionality to API Platform 3.0.

  **API Version Handling:**
  - Add per-page configuration to enable specific resources (`StatementSegment`) to use API Platform 3.0
  - Maintain backward compatibility with global default of API version 2.0
  - Allow bundles to override API version on a resource-by-resource basis

**IMPORTANT:** Vuex Store: It implements a per-page API version override system to enable gradual migration from EDT 2.0 to API Platform 3.0 on a page-by-page basis. The system allows different pages to use API Platform 3.0 for specific resource types while maintaining the global default of EDT 2.0 elsewhere (for example [administrationSegmentsList.js](https://github.com/demos-europe/demosplan-core/pull/6786/changes#diff-461a3b067e9a40ba62b041052f5430d678fe37f6a8e0c58fdbefaf9f602afc32R42-R50) entry point)

  **Pagination:**
  - Extend pagination mixin to support both API Platform 3.0 and legacy EDT pagination patterns
  - Handle different response structures between API versions

  **Component Updates:**
  - Refactor SegmentsList bundle for API Platform 3.0 compatibility
  - Remove unnecessary permission checks: `feature_segments_manualsort`
  - New requirement: remove `recommendationVersion` from included and request it as a StatementSegment's field `currentRecommendationVersionNumber`

--------------------------------------------------------------------------------------------------------------------------------------

**Key naming convention rules by 3.0 Api:**

1. **Main resource type** (the one being requested): Use PascalCase resource type name
   Example: 'StatementSegment' - matches ResourceType::getName() in backend
   Backend: StatementSegmentResourceType::getName() returns 'StatementSegment'

2. **Related resources** (via relationships): Use camelCase relationship property name
   Examples: 'parentStatement', 'place', 'tags' - match relationship definitions
   Backend: createToOneRelationship($this->parentStatement) creates 'parentStatement' relationship
            createToOneRelationship($this->place) creates 'place' relationship
            createToManyRelationship($this->tags) creates 'tags' relationship

This generates query parameters like:
 ```
 fields[StatementSegment]=text,externId,internId
 fields[parentStatement]=authorName,status,submitDate
 fields[place]=name,locked
 fields[tags]=title
```
---------------------------------------------------------------------------------------------------------------------------

**How to sort or filter in the new API 3.0:**

**Sorting:**
_Here you can see the different approaches to filtering with EDT and API Platform._

- **EDT 2.0**
```
sort=-internId
sort=internId
```

- **API Platform 3.0**
```
order[parentStatementOfSegment.original.internId]=desc
order[parentStatementOfSegment.original.internId]=asc
order[parentStatementOfSegment.submit]=asc
order[parentStatementOfSegment.externId]=asc
order[orderInProcedure]=asc
```

**Filtering**
_Here you can see the different approaches to filtering with EDT and API Platform, along with some example payloads. The `useApiPlatformFilters` composable transforms EDT-style filters into the API Platform 3.0 format._

- **EDT 2.0**

```
filter[tag-id-1][condition][path]=tags
filter[tag-id-1][condition][value]=tag-id-1
filter[tag-id-1][condition][operator]=ARRAY_CONTAINS_VALUE

filter[tag-id-2][condition][path]=tags
filter[tag-id-2][condition][value]=tag-id-2
filter[tag-id-2][condition][operator]=ARRAY_CONTAINS_VALUE
```
- **API Platform 3.0**
```
tags.id[]=<tag-id-1>
tags.id[]=<tag-id-2>
place.id[]=<place-id-1>
place.id[]=<place-id-2>
```
**Filtering with the `IS NULL` operator:**
`exists[<source>]=false`

**Group and OR Conjunction:**
If there is a group with an OR conjunction and one of the group elements contains `IS NULL`, the group ID is sent. Example for the assignee --> `<assignee>OrUnassigned[]=` and `<assignee>OrUnassigned[]=uuid` instaed of the `exists[assignee]=false` and `assignee.id[]=uiid`

-------------------------------------------------------------------------------------------------------
**Questions:**
Do we still need the permission `feature_segments_manualsort` --> NO

### How to review/test
  - Verify segments list loads correctly with API Platform 3.0
  - Confirm pagination works across different page sizes
  - Ensure backward compatibility with components still using API 2.0
  - Test sorting and filtering functionality

### PR Checklist
- [ ] Create/Update tests
- [ ] Update documentation
- [x] Run `yarn lint`
- [x] Run `yarn test`
- [ ] Update changelog

### CODE REVIEW

**Review by rafelddemos — COMMENTED** [review]
thanks for the work! &#x1F389;
and also adding the tests &#x2764;&#xFE0F;

I tested it and it works nicely. Checked a few examples for backward compatibility too.

Noticed two issues though in the filter panel:
1. if the user checks both `assigned` and `unassigned`, it returns an empty segments list.
2. topical tags behaviour is a bit different now, 
- `before`: when two filters are set, only the result matching both are returned 
- `now`: it's an OR behaviour

File: client/js/composables/useApiPlatformFilters.js [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+export function useApiPlatformFilters () {
+  /**
+   * Transforms EDT-style filters to API Platform 3.0 format.
+   * EDT: filter[uuid][condition][path/value/operator]
+   * AP3: path.id[]=uuid or exists[path]=false for IS NULL
+   *
+   * @param {Object} edtFilters - EDT 2.0 filter object
+   * @returns {Object} - API Platform 3.0 filter object
+   */
+  const transformFiltersToApiPlatform = (edtFilters) => {
```
rafelddemos, line 22: I guess this method is related to the problem with an empty segments-list if both `assigned` and `unassigned` checkboxes are checked, but I remember vaguely Cata discussing how OR / AND works differently across the APIs, so maybe the adjustment would need to be on the BE &#x1F914;

client/js/store/core/VuexApiRoutes.js
```diff
@@ -145,15 +148,31 @@ const buildRoute = (version, module, action) => {
...
+const generateApi3_0Routes = (additionalModules = {}) => {
+  bundleApi3Modules = additionalModules
+
+  const baseRoutes = Object.entries(api3_0Modules)
+    .flatMap(([typeName, actions]) => actions.map(action => buildRoute('3.0', typeName, action)))
+
+  const additionalRoutes = Object.entries(additionalModules)
+    .flatMap(([typeName, actions]) => actions.map(action => buildRoute('3.0', typeName, action)))
+
+  return [...baseRoutes, ...additionalRoutes]
```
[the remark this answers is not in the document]
sakutademos: each page is a separate bundle with its own entry point and navigation causes a full page reload, so the global state gets reset on each navigation. The JS context is completely fresh for each page, so `bundleApi3Modules` would not carry over between pages.

And `api3AdditionalModules` is temporary migration scaffolding that allows pages to gradually migrate to API Platform 3.0 on a per-page basis, and eventually all migrated resources will move directly into the base `api3_0Modules` object.

client/js/composables/useApiPlatformFilters.js
sakutademos: thx for spotting this! this feature has not been implemented in the backend yet, but we discussed it with the backend developer, and he will adjust the logic. So I can update the frontend accordingly once thats done

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -123,6 +111,19 @@
           @page-change="applyQuery"
           @size-change="handleSizeChange"
         />
+        <div class="ml-auto flex items-center space-inline-xs">
+          <dp-label
```
hwiem, line 115: Can't we use the label `DpSelect` provides?

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1079,32 +1082,33 @@ export default {
...
-        },
+        ...this.transformFiltersToApiPlatform(this.getFilterQuery),
+        ...defaultFilter,
       }
-      const { include, fields } = this.buildSegmentFetchOptions()
+
+      const defaultOrder = {
+        'parentStatementOfSegment.submit': 'asc',
+        'parentStatementOfSegment.externId': 'asc',
+        orderInProcedure: 'asc',
```
hwiem, line 1073: Shouldn't these appear as options in the select? 🤔

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -905,7 +907,7 @@ export default {
             ?.attributes?.locked,
       }))
 
-      if (this.selectedSort === '' || !hasPermission('feature_segments_manualsort')) {
+      if (this.selectedSort === '') {
```
hwiem: Maybe not in the scope of this PR, but shouldn't the client-side deadline sorting below have been removed already? 🤔 I believe it's now always applied when the user didn't select any sorting option in the select?

rafelddemos (approved):

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1203,37 +1182,34 @@ export default {
...
+       */
       const fields = {
-        File: [
-          'hash',
-        ].join(),
-        GenericStatementAttachment: [
-          'file',
-        ].join(),
-        Place: [
+        place: [
```
hwiem, line 1179: Apparently it hasn't worked for some time, but in the segment flyout, there's the option "Original-PDF", and that's what we need the source attachment for, so instead of removing it here, the new backend resource should provide it 🫠

client/js/components/procedure/SegmentsList/SegmentsList.vue
```diff
@@ -1373,6 +1336,41 @@ export default {
       })
     },
 
+    /**
+     * Get all segments (without pagination) to save them in localStorage for bulk editing.
```
hwiem:
```suggestion
     * Get all segment ids (without pagination) to save them in localStorage for bulk editing.
```

client/js/store/core/initStore.js
```diff
@@ -48,13 +48,23 @@ const handleResponse = async (response, messages = {}) => {
...
 }
 
-function initStore (storeModules, apiStoreModules, presetStoreModules) {
+/**
+ * Initialize Vuex store with API route configuration
+ *
+ * @param {Object} storeModules - Static Vuex modules
+ * @param {Array} apiStoreModules - API modules (defaults to 2.0 routes)
+ * @param {Object} presetStoreModules - Preset modules configuration
+ * @param {Object} api3AdditionalModules - Page-specific modules to use API 3.0 instead of 2.0
```
hwiem: nitpick: I wonder if we should rename `api3AdditionalModules` to something more descriptive, like `pageSpecificApi3Modules`. That way the name already hints at the reason why these modules are added

tests/frontend/unit/useApiPlatformFilters.spec.js
```diff
@@ -0,0 +1,117 @@
...
+        filter1: createFilter('tag', 'tag-uuid-123'),
+        filter2: invalidFilter,
+      }
+
+      expect(transformFiltersToApiPlatform(filters)).toEqual({
+        'tag.id': ['tag-uuid-123'],
+      })
+    })
+
+    it('handles multiple IS NULL filters', () => {
```
hwiem: I think this test case is a bit redundant, because we already have a test case above that asserts that a null filter is handled properly

hwiem (changes requested):
I think the page-specific overrides are a good solution 👍 There are 2 issues that I think need to be fixed: the obsolete deadline sort and the missing source attachment (needs to be fixed in the backend first though). the rest are just minor issues or nitpicks

client/js/components/procedure/SegmentsList/SegmentsList.vue
sakutademos: its a bit tricky because it sits above the select and breaks the layout a little. I thought it would be better to use the same layout as on the statement list page - inline

client/js/components/procedure/SegmentsList/SegmentsList.vue
sakutademos: hmm, do you mean the sorting select? As I understood it, these parameters are applied by default. And looking at the current implementation the default sorting seems to be: `parentStatement.submitDate%2CparentStatement.externId%2CorderInProcedure`

client/js/store/core/initStore.js
sakutademos: totally agree, i also wanted to find a better name for this, your suggestion is very nice!

tests/frontend/unit/useApiPlatformFilters.spec.js
sakutademos: done here https://github.com/demos-europe/demosplan-core/pull/6786/commits/5a14cd1c896cdfd626638531ff1cebc1968b8d14

client/js/store/core/initStore.js
sakutademos: done here: https://github.com/demos-europe/demosplan-core/pull/6786/commits/5f17635b796e3a012fab4d65f564a6c8244d2f8f

client/js/components/procedure/SegmentsList/SegmentsList.vue
sakutademos: right, this has to be adjusted in the BE because we no longer get `genericAttachments` and `sourceAttachment`, but we still need them

client/js/components/procedure/SegmentsList/SegmentsList.vue
sakutademos: correct, done here: https://github.com/demos-europe/demosplan-core/pull/6786/commits/048dcd218503bb0bcb8c908aa273f804a1efd7a7, but i have to check whether its implemented in the backend

hwiem (approved):
Since the 'Original-PDF' will be fixed separately: Lgtm, thanks for this effort 🙏 ✨

hwiem (approved):
Lgtm, thank you! 🙏

PR #6792, opened 2026-09-04
fix: scope global gis layers to their customer

mussbach:
### Ticket
DPLAN-18357
ado-55709

Creating a global gis layer copied it into **every** procedure in the database. The fan-out in `MapRepository::add()` used `Procedure::findAll()` without any customer filter, so a layer created by one customer's admin appeared in every other customer's procedures, in their organisation blueprints and in the platform master template — and, via the master template, in every procedure created afterwards. The edit path propagated changes into those copies as well.

This binds a global layer to the customer it is created in and limits both the fan-out and the global admin list to that customer:

- new `GisLayer::$customer` (`_gis.customer_id`, nullable), set in `MapService::addGis()` for layers without a procedure
- `MapRepository::add()` copies into procedures of that customer only; the copies themselves carry no customer, they are scoped by their procedure
- `MapRepository::findGlobalLayersForCustomer()` backs `MapService::getGisGlobalList()`

Layers with `customer_id IS NULL` predate the scoping and stay visible to every customer, so existing ones can still be administered and removed.

Not covered here: global gis layer create/update/delete still produce no report entry, so the acting user cannot be traced. Noted in the ticket as a follow-up.

### How to review/test
1. Run `vendor/bin/phpunit tests/backend/core/Map/` — includes the new regression test `MapServiceTest::testAddGlobalGisSkipsProceduresOfOtherCustomers`.
2. Create a global layer under "Globale GIS-Layer" and check that it appears in the procedures of the current customer only, and no longer in the master template.
3. `doctrine:migrations:diff` produces no further `_gis` statements after applying `Version20260904103000`.

### PR Checklist

- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/09/Version20260904103000.php
+++ b/demosplan/DemosPlanCoreBundle/DoctrineMigrations/2026/09/Version20260904103000.php
@@ -28,7 +28,7 @@ final class Version20260904103000 extends AbstractMigration
 {
     public function getDescription(): string
     {
-        return 'Add customer_id to _gis to scope global GIS layers to a single customer';
+        return 'chore: (refs: DPLAN-18357) Add customer_id to _gis to scope global GIS layers to a single customer';
     }
 
     /**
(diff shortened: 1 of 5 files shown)


breierDemos (approved):
lgtm code review only

PR #6801, opened 2026-09-07
feat(DPLAN-18363): expose pictogram alt text and copyright on public procedure list

mussbach:
### Ticket
DPLAN-18363

### Description
The pictogram alt text and copyright fields were previously collected in the procedure settings but only ever transmitted to the meinBerlin interface — they were never rendered publicly. Projects without that interface therefore save values nobody sees.

This exposes both fields on the public procedure list:

- Index `pictogramAltText` and `pictogramCopyright` in the procedure Elasticsearch documents (non-indexed keyword fields, like the existing `pictogram` field)
- Use the stored alt text as the `alt` attribute of the list image, falling back to the previous `Abbildung für {procedure name}` when empty
- Render the copyright as a small grey caption below the image when set

Existing procedures show no visual change unless values are set. After deployment the procedure index needs to be repopulated (or procedures resaved) for the new fields to appear in the documents.

### How to review/test
- Reindex procedures (`bin/<project> fos:elastica:populate --index=procedure` or resave a procedure)
- Open the public procedure list and check a procedure with a pictogram: alt attribute + copyright caption
- Check a procedure without copyright/alt values: no caption, fallback alt text

### PR Checklist

### DIFF
--- a/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/public_index_list.html.twig
+++ b/templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/public_index_list.html.twig
@@ -44,7 +44,7 @@
                             class="{{ 'w-full h-auto'|prefixClass }}"
                             src="{{ publicProcedure.pictogram|default ? path('core_logo', {'hash': publicProcedure.pictogram|getFile('hash')}) : asset('img/procedure-placeholder.gif') }}"
                             loading="lazy"
-                            alt="{{ publicProcedure.pictogramAltText|default ? publicProcedure.pictogramAltText : 'Abbildung für ' ~ procedureDisplayName }}">
+                            alt="{{ publicProcedure.pictogramAltText|default ? publicProcedure.pictogramAltText : 'Piktogramm für das Verfahren ' ~ procedureDisplayName }}">
                         {% if publicProcedure.pictogramCopyright|default %}
                             <div class="{{ 'text-sm color--grey u-mt-0_125 break-words'|prefixClass }}">
                                 © {{ publicProcedure.pictogramCopyright }}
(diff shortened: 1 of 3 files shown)


bakridemos (approved):
lgtm, code review only

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/public_index_list.html.twig
```diff
@@ -44,7 +44,12 @@
                             class="{{ 'w-full h-auto'|prefixClass }}"
                             src="{{ publicProcedure.pictogram|default ? path('core_logo', {'hash': publicProcedure.pictogram|getFile('hash')}) : asset('img/procedure-placeholder.gif') }}"
                             loading="lazy"
-                            alt="Abbildung für {{ procedureDisplayName }}">
+                            alt="{{ publicProcedure.pictogramAltText|default ? publicProcedure.pictogramAltText : 'Abbildung für ' ~ procedureDisplayName }}">
```
hwiem: nitpick: I think `'Piktogramm für das Verfahren ' ~ procedureDisplayName` could be a more precise alt text

hwiem (approved):
FE looks good + works, thanks! Just a minor suggestion

templates/bundles/DemosPlanCoreBundle/DemosPlanProcedure/public_index_list.html.twig
mussbach: agreed. I just took the existing text to not change anything unrelated, but as we are here we could improve it along the way [1ad9ec2](https://github.com/demos-europe/demosplan-core/pull/6801/commits/1ad9ec219a85831d0c3b333e7a36f98a152d6931)

PR #6803, opened 2026-09-09
F dplan 18189 fix fragment metadata export

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-18189

Description: 

The procedure export's assessment table DOCX documents (`Abwaegungstabelle_Liste_mit_Datensaetzen.docx` and its anonymized variant) were missing the priority areas (Potenzialflächen) and tags (Schlagworte) for statements that were split into fragments (Datensätze). This required two fixes:

**1. Export rendering (`DocxExporter`)**

- `addFragmentRows()` now renders the priority areas/tags line for every fragment row, not just the first.
- Each fragment row now shows its **own** priority areas/tags (fetched from the fragment entity), instead of incorrectly inheriting the parent statement's values (`formatFragmentArray()` previously never overrode these fields).

**2. Stale Elasticsearch data (`StatementFragmentService`)**

The assessment table export reads statement data from Elasticsearch, where each Statement document embeds its fragments as a nested field. Creating, updating or deleting a `StatementFragment` never triggered a refresh of the parent Statement's ES document (they are separate entities/indexes), so newly added, edited or removed fragments would not appear in any fragment-dependent export until a manual reindex was run.

`StatementFragmentService` now explicitly refreshes the parent Statement's Elasticsearch document (via the `statements` object persister — the same pattern already used in `ProcedureService`) after every fragment create, update and delete, so exports stay correct without requiring a manual reindex.

### How to review/test

- Add a fragment to a statement, assign it its own priority areas/tags (different from the parent statement's), then run the procedure export — the `_mit_Datensaetzen*` DOCX files should show each fragment's own values, with no manual reindex needed.
- Edit and delete a fragment, then re-export — changes should be reflected immediately.
- Confirm the standalone Abwägungstabelle export (outside the procedure export) is unaffected.
- Confirm regular (non-fragment) statement workflows are unaffected, since `StatementFragmentService` is used by `StatementHandler`, `StatementCopier`, `AssessmentHandler`, etc.

### DIFF
--- a/config/services.yml
+++ b/config/services.yml
@@ -947,6 +947,8 @@ services:
         lazy: true
         calls:
             - [ 'setEsStatementFragmentType', [ '@fos_elastica.index.statementFragments' ]]
+        arguments:
+            $esStatementPersister: '@fos_elastica.object_persister.statements'
 
 
     demosplan\DemosPlanCoreBundle\Logic\Statement\TagService:
--- a/demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php
+++ b/demosplan/DemosPlanCoreBundle/Logic/Export/DocxExporter.php
@@ -729,11 +729,12 @@ class DocxExporter
     }
 
     /**
-     * Appends the assigned Potenzialflächen and Schlagworte below the statement text.
+     * Appends the assigned Potenzialflächen and Schlagworte below the given item's text.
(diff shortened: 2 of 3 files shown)


mussbach (approved):
lgtm, code review only

PR #6804, opened 2026-09-09
fix(refs HDDP-27): delete assessment table export jobs with their procedure

mussbach:
### Ticket
HDDP-27

ProcedureDeleter did not touch `assessment_table_export_job`, so deleting a procedure left its export job rows behind until the maintenance sweep pruned them by age. The rows also pin no download file: the result file's `_files` row already carries the procedure id and is removed by the deleter's existing file sweep.

Adds the matching `deleteFromTableByIdentifierArray` call (covered on every deletion path, which all funnel through `ProcedureDeleter::deleteProcedures()`) plus tests for deletion and dry run.

### How to review/test
`vendor/bin/phpunit tests/backend/core/Procedure/Functional/ProcedureDeleterTest.php`

### PR Checklist
- [x] Create/Update tests


bakridemos (approved):
lgtm, thx

PR #6805, opened 2026-09-09
fix: set email address on ProcedurePerson from weitere Einreichende import

mussbach:
## Summary
- The 'weitere Einreichende' worksheet import validated the E-Mail column as required but never called `setEmailAddress()` on the created `ProcedurePerson`, so the email was silently discarded.
- Added a functional test covering `processWeitereEinreichendeEntry` to prevent regressions on this contact-data mapping.

## Test plan
- [x] `StatementExcelImporterTest::testProcessWeitereEinreichendeEntryCreatesProcedurePersonWithContactData` added and passing
- [x] Verified the new test fails without the fix and passes with it
- [x] Full `StatementExcelImporterTest` suite passes


bakridemos (approved):
lgtm, code review only

PR #6811, opened 2026-09-11
fix(refs DS-619): discard changes on negative report select

bakridemos:
Ticket: [DS-619](https://demoseurope.youtrack.cloud/tickets/DS-619/WG-28.09.-Einladung-zur-Beteiligung-232.-Anderung-des-Flachennutzungsplanes-Nordlich-Stroer-Allee-in-Koln-Surth)

- Previously, the 'Fehlanzeige' radio button was disabled as soon as any change was made to the fields in the form, even if "kein Ortsbezug" was selected
- the 'Fehlanzeige' radio button is now always enabled. If the user previously made any changes to the form fields, a warning explains that the entries are not saved
- submitting the form asks for confirmation; cancel switches back to "Stellung nehmen", confirm sends a payload without text, references, location and files. The store keeps the content so it reappears when switching back to "Stellung nehmen"
- "kein Ortsbezug" radio buttons are disabled while the 'Fehlanzeige' radio button is selected
- when 'Stellung nehmen' is selected, the user clicks into the editor, then selects 'Fehlanzeige', and then switches back to 'Stellung nehmen', the editor now no longer shows a red error border (this was previously the case because the editor is validated on blur; this validation state is now reset when selecting 'Fehlanzeige')

### How to review/test
1. Log in as an institution user on a procedure with the Fehlanzeige category, open the statement form
2. Enter text and pick a map point or county, then select Fehlanzeige: the fields are disabled and the warning appears
3. Save: a confirm dialog appears. Cancel switches back to "Stellung nehmen" with the content intact. Confirm saves a Fehlanzeige without content
4. Open a new statement modal; click into the empty editor, select Fehlanzeige, then "Stellung nehmen" again: there should be no red border on the editor
5. Reopen the draft from the drafts list: no warning, empty fields, Fehlanzeige selected
6. Select Fehlanzeige on an untouched form: no warning, no dialog, saving goes straight through

### DIFF
--- a/client/js/components/statement/publicStatementModal/StatementModal.vue
+++ b/client/js/components/statement/publicStatementModal/StatementModal.vue
@@ -115,7 +115,7 @@
               id="negative_report_true"
               data-cy="statementModal:indicationerror"
               :checked="formData.r_isNegativeReport === '1'"
-              @change="() => { setStatementData({ r_isNegativeReport: '1'}) }"
+              @change="() => { setStatementData({ r_isNegativeReport: '1'}); clearLocationValidationState() }"
               :label="{
                 hint: Translator.trans('link.title.indicationerror'),
                 text: Translator.trans('indicationerror')
@@ -1211,6 +1211,20 @@ export default {
         r_paragraph_title: ''
       }
       this.setStatementData(elementFields)
+    },
+
+    /*
+     * The vanilla dpValidate lib marks the whole location fieldset invalid on blur once it's
+     * empty, without rechecking whether it's still required - clear that leftover state once
+     * Fehlanzeige makes the location optional again.
+     */
+    clearLocationValidationState () {
+      const fieldset = document.getElementById('locationFieldset')
+      if (!fieldset) {
+        return
+      }
+      const errorClass = this.prefixClass('is-invalid')
+      fieldset.querySelectorAll(`.${errorClass}`).forEach(el => el.classList.remove(errorClass))
     },
 
     removeNotificationsFromStore () {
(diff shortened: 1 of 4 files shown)


client/js/components/statement/publicStatementModal/StatementModal.vue
```diff
@@ -1464,6 +1559,24 @@ export default {
...
     },
 
+    // After a Fehlanzeige save the server holds no content, so the form must not either
+    discardStatementContent () {
+      if (this.$refs.statementEditor) {
+        this.$refs.statementEditor.resetEditor()
+      }
+      if (this.$refs.uploadFiles) {
+        this.$refs.uploadFiles.clearFilesList()
+      }
```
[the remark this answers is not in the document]
hwiem: `handleRemove` is client-side and wouldn't remove any orphaned uploads from the server, so your suggested fix wouldn't work. Orphaned uploads are cleaned up server side.

client/js/components/statement/publicStatementModal/StatementModal.vue
```diff
@@ -184,6 +191,7 @@
             :class="prefixClass('c-statement__formblock u-ml u-mb-0_5 u-mt-0_5 inline-block')">
             <template v-if="formData.r_element_id !== ''">
               <button
+                :disabled="formData.r_isNegativeReport !== '0'"
```
sakutademos: are there only two options: 0 and 1? If so, would it be more readable to check for 1, since it expresses the actual condition explicitly?


```suggestion
                :disabled="formData.r_isNegativeReport === '1'"
```

and maybe it makes sense to create a variable for this value, since its used in so many places, because the repeated `!== '0'` makes the code a bit harder to read

sakutademos (approved):
just a tiny nitpick. Unfortunately, I cant test it right now, but the code review lgtm! But there are also some SonarCloud issues, as well as some TeamCity issues which should be fixed before merge

client/js/components/statement/publicStatementModal/StatementModal.vue
hwiem: Good point, thanks! Done in 0addfa081482ffb6b48487d97f35f77ea22a90f8

PR #6812, opened 2026-09-11
feat: add read-only state for procedures

mussbach:
### Ticket
DPLAN-18375

A procedure can be flagged read-only: it stays reachable for reference (statement
list, segment list incl. filters, submitter list, protocol) while no write
permission is granted inside it. A new `read_only` column on `Procedure` carries
the state; the flag is toggled from the admin procedure list via
`feature_procedure_read_only_toggle` and recorded in the procedure protocol.

**Permission handling is enable-only.** `setProcedurePermissions()` is now a `final`
dispatcher: for a read-only procedure it grants a fixed read set instead of the
normal procedure permissions, so writes are never enabled rather than enabled and
then revoked. Members entering through the phase permissionset get
`setProcedurePermissionsetRead()` and never the write variant.

- projects extend the new `protected setFullProcedurePermissions()`
- the read set is `READ_ONLY_PROCEDURE_PERMISSIONS` (only what core's own owner
  branch grants) plus an additive `projectReadOnlyProcedurePermissions()` hook — a
  project can add, never remove
- keeping the existing method name and making it `final` means all four callers are
  covered, including `ImportJobProcessor` and `ExportJobContextRestorer`, and no
  project override can bypass the gate
- addon permissions are resolved on demand and core must not know their names, so
  none are granted in a read-only procedure

**Fine-grained permissions split out** of area gates that also guard those reference
views, so a view can stay readable without its writes: `feature_segment_edit`,
`feature_segment_delete`, `feature_segmentation_start`,
`feature_segments_of_statement_list_export`, `feature_procedure_submitter_export`,
`feature_statement_claim`. Two of these previously had no gate at all
(`StatementSegmentResourceType` writes, `StatementResourceType::assignee`).
**Projects must grant them to keep the corresponding actions available.**

A dashboard deep link into a read-only procedure redirects to the statement list.

### How to review/test
Flag a procedure read-only from the admin list, then confirm the four reference
views render while export/delete/bulk-edit/assignment/import are gone. Deep-link
`/verfahren/{id}/uebersicht` → statement list; `/verfahren/{id}/einstellungen` →
denied. PATCH a Statement via JSON:API with `X-Demosplan-Procedure-Id` → 403.
Unflag and confirm full access returns with both flips in the protocol.

Worth exercising an import or export job in a read-only procedure: those call
`setProcedurePermissions()` directly and were previously ungated.

### PR Checklist
- [x] Create/Update tests


bakridemos (changes requested):
read-only toggle PATCH fails with 403 for the role meant to use it

ProcedureResourceType::isUpdateAllowed() correctly gates the readOnly field on feature_procedure_read_only_toggle, but that check never runs: the generic PATCH route (GenericApiController::updateResource, api_resource_update) is gated first by a controller-level #[DplanPermissions('feature_json_api_update')], checked in CheckPermissionEventSubscriber before any EDT ResourceType logic executes.

The toggle is invoked from the admin procedure list — i.e. outside any procedure context. feature_json_api_update is only granted to PLANNING_AGENCY_ADMIN inside a procedure (via projectProcedurePermissions()/setFullProcedurePermissions()), never at the global scope where feature_procedure_read_only_toggle itself lives. Result: every user with the toggle permission gets AccessDeniedException: Der Zugriff auf feature_json_api_update ist nicht gestattet on the PATCH, reproduced locally with multiple RMOPSA accounts.

This also isn't caught by the existing test suite: PermissionsTest's isInProcedure: false case for PLANNING_AGENCY_ADMIN (the exact scenario that matches the admin list) already asserts feature_procedure_read_only_toggle is allowed, but never asserted feature_json_api_update — so the missing pairing slipped through.

Suggested fix: grant feature_json_api_update at the same (global) scope everywhere feature_procedure_read_only_toggle is granted, and add a test asserting both together for the outside-procedure case — otherwise any other project adopting this feature will hit the same wall. Might also be worth a note in config/permissions.yml on feature_procedure_read_only_toggle documenting this dependency, since it's not obvious from the permission name alone.

bakridemos (changes requested):
ProcedureResourceType::isUpdateAllowed() authorizes the write purely on the permission feature_procedure_read_only_toggle, relying on the shared read-visibility access conditions (owning/invited/data-input/planning-office org) to scope which procedures are reachable — there's no explicit "must be the owning organisation" restriction on the write itself. In EWM this is a non-issue because the project doesn't enable institution participation or planning offices. But any other project that does (e.g. ones supporting TÖB participation or contracted planning offices) and later adopts feature_procedure_read_only_toggle for the same role should double-check this before relying on "only the owner can archive a procedure" as an assumption — it isn't enforced in code, it's incidental to which roles/relations exist in that project's config.

bakridemos (changes requested):
Read-only procedures → empty ZIP

Confirmed the mechanism. In ExportService::createProcedureExportJob() (ExportService.php:216-217), for every procedure being exported:

$this->permissions->setProcedure($procedureToExport);
$this->permissions->checkProcedurePermission();

checkProcedurePermission() → setProcedurePermissions(), which is exactly the dispatcher this PR introduced. For a read-only procedure it skips setFullProcedurePermissions() entirely and grants only the tiny read-only allowlist. But every content section in the export is gated by a feature_procedure_export_include_* permission (assessment table, cover page, news, member list, statements final/released/public, protocol, and segments):

// ExportService.php:289-292
if ($this->permissions->hasPermission('feature_procedure_export_include_segments')) {
    $zip = $this->addStatementSegmentsToZip($procedureId, $procedureName, $zip);
}

None of the feature_procedure_export_include_* permissions are in Permissions::READ_ONLY_PROCEDURE_PERMISSIONS (core) or in EWM's projectReadOnlyProcedurePermissions(). So for a procedure flagged read-only, every single one of those if checks fails, and the "Gesamtabzug" comes back as an almost-empty ZIP — not just missing segments, missing everything content-related. That directly contradicts the feature's own stated goal ("stays reachable for reference").

mussbach:
All three confirmed, fixed in ca32e309.

**1. `feature_json_api_update` on the toggle PATCH** — confirmed: `GenericApiController::update` is gated by it and `checkPermissions()` is AND-semantics, so `isUpdateAllowed()` never ran. Instead of pairing the two permissions, the toggle now has its own route — `POST /verfahren/{procedureId}/schreibschutz`, gated by `feature_procedure_read_only_toggle` alone. That removes the dependency rather than documenting it, and avoids opening the generic PATCH endpoint globally for the role. Side effect: `ProcedureResourceType` is back to its `main` state — the `readOnly` attribute, `isUpdateAllowed()` and the report-service dependencies existed only for the toggle, and the procedure list reads `AdminProcedure`.

**2. No owning-organisation restriction on the write** — confirmed, the access conditions also admit invited, planning-office and data-input organisations. The new action checks `ProcedureAccessEvaluator::isOwningProcedure()` before flipping the flag, so it no longer depends on which roles and relations a project happens to enable.

**3. Empty ZIP** — confirmed. Exports staying off inside a read-only procedure is intended; the silently near-empty archive was not. A selection containing a read-only procedure is now rejected with a message — `exportProcedures()` and `startAsyncExport()` both check, and the list blocks the submit client-side since it already has `readOnly` per row (a download response would not render a flash message).

Unrelated to the reviews: `setProcedurePermissions()` is `final` now, so project overrides of it have to move to `setFullProcedurePermissions()`. Already done on the matching project branches.

bakridemos (approved):
Re-tested after ca32e30 — the three confirmed bugs from the first review round (403 on the toggle PATCH, missing owning-org check, empty export ZIP for read-only procedures) are all fixed and verified working. the frontend part is working however it needs to be reviewed again.
open point: a small frontend component-misuse issue on the read-only badge.
 Read-only badge uses the wrong component
client/js/components/procedure/admin/AdministrationProceduresList.vue, lines 142-147

The "Schreibgeschützt" badge is built with class="flash flash-info u-ml-0_25" on a <span>. .flash is our page-level alert/notification-banner style (see the docblock in _flash.scss — "not related to a single user interaction," and every other usage in the codebase is on a <div>/<p>). It's not meant as an inline status chip, and its banner-sized padding + border is why the badge overlaps the second line and bleeds into the count column (see screenshot). We already have DpBadge in @demos-europe/demosplan-ui for exactly this — a small rounded-md pill with a matching color="info" variant and a size="small" preset. Could you swap this to <dp-badge color="info" size="small" :text="Translator.trans('procedure.read.only')" /> instead?

Suggested fix:

```
<dp-badge
  v-if="readOnly"
  color="info"
  size="small"
  data-cy="procedureReadOnlyBadge"
  :text="Translator.trans('procedure.read.only')"
  class="u-ml-0_25"
/>

```

PR #6813, opened 2026-09-11
fix(refs DPLAN-18366): resolve addon-declared permissions when gating UI hooks

mussbach:
### Ticket
DPLAN-18366

Addon UI hooks can be guarded by `options.permissions` in the addon manifest.
`FrontendAssetProvider` resolved those names with a plain core lookup, but
permissions declared by an addon live in a separate collection keyed by addon
and are invisible to the core permission array. Any hook guarded by an
addon-declared permission was therefore dropped for every user, regardless of
their actual access — the PDF import tab and its dashboard chart never
rendered. Hooks guarded by core permissions, and hooks with no `permissions`
option at all, were unaffected, which is why only some addon hooks went
missing.

The name is now resolved against the owning addon when that addon declares it,
falling back to core otherwise. Frontend exposure was already addon-aware, so
`hasPermission()` in JS returned `true` while the backend filter said no.

Two fixes found along the way:

- `getAssetPathsFromManifest()` built its exceptions but never threw them, so a
  malformed manifest fell through to a `TypeError` instead of being caught and
  degrading to "hook not delivered".
- `AddonRegistry::boot()` was annotated `Definition[]`, but the container
  resolves those before the call; the runtime type is `AddonInfo[]`.

### How to review/test
Open a procedure's import page as a user who owns the procedure, with the
PDF import addon installed. The PDF-Import tab should appear and accept
uploads. All three gating branches are covered by
`tests/backend/core/Core/Unit/Addon/FrontendAssetProviderTest.php`
(10 tests): addon-declared permission, core permission fallback, and hooks
with no permission option.

### PR Checklist
- [x] Create/Update tests

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Addon/AddonRegistry.php
+++ b/demosplan/DemosPlanCoreBundle/Addon/AddonRegistry.php
@@ -32,7 +32,10 @@ class AddonRegistry implements ArrayAccess
     }
 
     /**
-     * @param Definition[] $addonInfos
+     * The compiler pass passes {@link Definition}s here, which the container resolves into the
+     * actual instances before this method ever runs.
+     *
+     * @param AddonInfo[] $addonInfos
      */
     public function boot(array $addonInfos = [])
     {
--- a/demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
+++ b/demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
@@ -20,7 +20,7 @@ final readonly class FrontendAssetProvider
 {
     public function __construct(
         private PermissionEvaluatorInterface $permissionEvaluator,
-        private AddonRegistry $registry
+        private AddonRegistry $registry,
     ) {
     }
 
--- a/demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
+++ b/demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
@@ -10,70 +10,123 @@
 
 namespace demosplan\DemosPlanCoreBundle\Addon;
 
-use DemosEurope\DemosplanAddon\Contracts\PermissionsInterface;
+use DemosEurope\DemosplanAddon\Permission\PermissionEvaluatorInterface;
 use demosplan\DemosPlanCoreBundle\Exception\AddonException;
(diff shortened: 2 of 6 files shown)


bakridemos (commented):
RuntimePermissionIdentifier was a byte-for-byte duplicate of DemosEurope\DemosplanAddon\Permission\PermissionIdentifier (same two fields, same forCore()/forAddon() factories, same PermissionIdentifierInterface). Replaced the usage in FrontendAssetProvider with the addon's existing class and deleted the new one.

Why it was introduced, and why that reasoning doesn't hold up:

PermissionIdentifier carries @deprecated use enums implementing PermissionIdentifierInterface instead. That's presumably why a parallel class was written rather than reusing it — but the deprecation only makes sense for permission names that are fixed at compile time (i.e. a hardcoded set you could write out as enum cases). FrontendAssetProvider::isHookEnabled() reads $permissionName out of an addon's manifest at request time — the set of valid names isn't known until an addon is installed and its manifest is parsed, so there's no fixed case list to enumerate. An enum can't represent that.

Checked this isn't a one-off — same pattern holds everywhere:

- All 8 addons declare their compile-time-known permissions the same way: class Features extends AbstractPermissionMeta. None of them use PermissionIdentifier — it's never needed for permissions an addon author already knows about when writing code.
- AbstractPermissionMeta (the base class Features extends) does carry a legitimate // TODO: Replace with enum based approach — but for a different reason: its constructor uses PHP reflection to parse each permission method's docblock for label/description at runtime, which is genuinely slow, and a backed enum with a match() would remove that cost entirely.
- PermissionIdentifier has no such cost and no fixed case list — it exists specifically for permission names the codebase can't enumerate in advance (addon manifests, addon-declared permission collections). It's already used this way at DefaultTwigVariablesService.php:81.

Conclusion: PermissionIdentifier's @deprecated tag looks like it inherited AbstractPermissionMeta's TODO wording without inheriting the justification behind it. Suggest dropping the @deprecated tag from PermissionIdentifier in the addon repo (separate PR/release) and leaving AbstractPermissionMeta's TODO untouched, since only the latter has an actual problem (reflection cost) to solve.

mussbach:
@bakridemos good catch, thank you. Fixed in [f9a1126](https://github.com/demos-europe/demosplan-core/pull/6813/commits/f9a11268132992f8bfde0d7f1e85c880839c9466)

bakridemos (approved):
lgtm thx

PR #6814, opened 2026-09-11
fix: correct removed Doctrine ORM/DBAL exception FQCNs

mussbach:
## Summary
- `Doctrine\ORM\ORMException` moved to `Doctrine\ORM\Exception\ORMException`, and `Doctrine\DBAL\DBALException` was renamed to `Doctrine\DBAL\Exception` in the installed Doctrine version. The old FQCNs no longer exist.
- Updated the `use` imports in all affected files to the correct FQCNs. `@throws` docblocks keep using the short class name, which now resolves correctly.
- This was masking real issues: PHPStan treated any `@throws` union containing the unresolvable old FQCN as "not a Throwable" entirely, hiding genuine exception-doc problems on those lines.

## Test plan
- [ ] `bin/ewm d:php -l2` no longer reports `throws.notThrowable` for these files


bakridemos (approved):
lgtm thx

PR #6815, opened 2026-09-11
fix: correct malformed PHPDoc tags and Collection::add() void misuse

mussbach:
## Summary
- Fixed 13 malformed PHPDoc `@var`/`@return`/`@SuppressWarnings` tags (invalid syntax like `array()`, `array<>`, trailing `;`, `string{...}`) that PHPStan couldn't parse at all.
- Fixed 11 `add*()` methods (Statement/County/Municipality/PriorityArea/Tag/Orga entities) that assigned the result of `Doctrine\Common\Collections\Collection::add()` — which is `void` — to a "successful" flag. Since `add()` always succeeds once `contains()` has been checked, success is now tracked explicitly instead of relying on the (nonexistent) return value. Behavior is unchanged.
- Fixed two `FlashBagInterface::set()` (also `void`) calls in `UserHandler::handleSaveAllDepartments()` that were being `return`ed directly; the doc already allowed `array|null` at the call site, so these now return `null` after calling `set()`, matching prior runtime behavior.

## Test plan
- [ ] `bin/ewm d:php -l2` no longer reports `phpDoc.parseError` or `method.void` for these files

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
+++ b/demosplan/DemosPlanCoreBundle/Addon/FrontendAssetProvider.php
@@ -22,7 +22,7 @@ final class FrontendAssetProvider
     }
 
     /**
-     * @return array<string, array<string, mixed>>>
+     * @return array<string, array<string, mixed>>
      */
     public function getFrontendClassesForHook(string $hookName): array
     {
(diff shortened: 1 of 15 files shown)


bakridemos (approved):
lgtm, code review only

PR #6816, opened 2026-09-11
fix: correct stale PHPDoc types that conflicted with native type declarations

mussbach:
## Summary
- 35 files where a `@return`/`@var`/`@param` PHPDoc tag no longer matched the native type hint already declared on the same method/variable. Native type declarations were already correct and more specific in every case; the PHPDoc was stale (predates a return-type narrowing, e.g. `Response` -> `RedirectResponse`/`JsonResponse`/`APIResponse`) or simply wrong (copy-paste artifacts like `@return bool` on a `getCustomer(): ?CustomerInterface` getter).
- A few `@var` tags were misplaced (annotating a variable before its final assignment, or before a `foreach` loop instead of inside it) rather than wrong in content; moved to the correct statement.
- One generic template mismatch in `EntityGrouper::fillEntityIntoGroupStructure()` (`T` widened to `T&EntityInterface` for that parameter only, no change to the class-wide template bound).
- No behavior changes — doc-only and one generic type annotation.

## Test plan
- [ ] `bin/ewm d:php -l2` no longer reports `return.phpDocType`, `varTag.nativeType`, or `parameter.phpDocType` for these files

### DIFF
--- a/demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
+++ b/demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
@@ -1637,7 +1637,7 @@ class DemosPlanStatementController extends BaseController
         } else {
             $draftListFilterVO = $draftFilterList[$procedureId][$templateName];
         }
-        /** @var DraftStatementListFilters $draftListFilterVO */
+        /* @var DraftStatementListFilters $draftListFilterVO */
 
         if ($request->query->has('reset')) {
             $draftFilterList[$procedureId][$templateName] = new DraftStatementListFilters();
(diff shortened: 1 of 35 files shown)


demosplan/DemosPlanCoreBundle/Controller/Statement/DemosPlanStatementController.php
```diff
@@ -1639,13 +1631,13 @@ protected function saveDraftListFiltersInSession(Request $request, string $proce
...
         }
 
-        /** @var DraftStatementListFilters $draftListFilterVO */
         $draftListFilterVO = null;
         if (!$draftFilterList[$procedureId][$templateName] instanceof DraftStatementListFilters) {
             $draftListFilterVO = new DraftStatementListFilters();
         } else {
             $draftListFilterVO = $draftFilterList[$procedureId][$templateName];
         }
+        /* @var DraftStatementListFilters $draftListFilterVO */
```
bakridemos, line 1640: The relocated @var annotation lost its second asterisk  /* @var DraftStatementListFilters $draftListFilterVO */ is a plain comment, not a PHPDoc block, so PHPStan won't parse it and $draftListFilterVO stays mixed. Should be /** @var DraftStatementListFilters $draftListFilterVO */.

bakridemos (changes requested):

PR #6817, opened 2026-09-11
build: resolve @tiptap/* to 3.31.3 and align prosemirror pins

hwiem:
Replaces https://github.com/demos-europe/demosplan-core/pull/6778 which only updates @tiptap/core - we also need to add prosemirror-schema-basic as a direct dependency (it was only present transitively via @tiptap/pm, which dropped it in 3.31) and update the prosemirror resolutions to align them with demosplan-ui.

Closes 2 high and moderate security alerts:
- https://github.com/demos-europe/demosplan-core/security/dependabot/298
- https://github.com/demos-europe/demosplan-core/security/dependabot/288

### How to review/test
- test together with the PR in demosplan-ui, so `yarn link` the demosplan-ui branch
- run `yarn intall` in demosplan-core
- check the split statement view (creating/editing segments should still work) and some editors, play around with their buttons - all the features should still work

### Linked PRs
- [ ] https://github.com/demos-europe/demosplan-ui/pull/1576


rafelddemos (approved):
tested it symlinked with the UI-PR, it works fine as far as I can see, thanks :)

PR #6823, opened 2026-09-14
fix(refs DPLAN-18378): enable browser SDK tracing via configurable sample rate

mussbach:
### Ticket
DPLAN-18378

`Sentry.init()` in `client/js/loadSentry.js` ran without `tracesSampleRate`. Since SDK v8 that disables tracing entirely, so `browserTracingIntegration` was registered but collected nothing — no browser performance data has ever reached Sentry, while the backend traces normally. The options passed to the integration (`attachProps`, `tracing`, `tracingOptions.trackComponents`) belong to the Vue integration and were silently ignored.

This matters beyond pageload timings: the interaction metrics (INP, long animation frames) ride on the same integration, and they are what makes client-side input latency measurable at all. Without them there is no way to tell a slow backend response apart from a blocked main thread.

Adds `sentry_traces_sample_rate_frontend`, exposed to the client along the same path the DSN already takes (parameter → Twig global → `window.dplan`). Deliberately separate from `sentry_traces_sample_rate`: the backend can run at a high rate while the browser stays low, since every page view emits its own transaction. The default is `0`, so this changes nothing until a deployment opts in.

### How to review/test
With `sentry_dsn` and `sentry_traces_sample_rate_frontend` set in the local container config:

```js
window.dplan.sentryTracesSampleRate            // the configured rate
window.__SENTRY__                              // present
// client options: tracesSampleRate set, integrations include "BrowserTracing"
```

With the defaults (empty DSN, rate `0`) no Sentry client is initialised at all and the tracing integration is not registered. Both cases verified in the browser against a local instance.

### PR Checklist
- [x] Run `yarn lint`


hwiem (approved):
Lgtm, thank you!

PR #6838, opened 2026-09-16
perf: cache customer lookups by subdomain

mussbach:
### Ticket
—

`CustomerService::findCustomerBySubdomain()` goes through the entity persister, which bypasses
Doctrine's identity map. Every `getCurrentCustomer()` therefore issued another `SELECT` — and
`DoctrineUserListener::postLoad` calls it on every `User` that gets loaded.

The service now caches the customer id per subdomain and resolves it via `findCustomerById()`,
so repeat calls are served from the identity map. Caching the id rather than the entity keeps the
returned customer managed even after an `$em->clear()` (import job processing, mail queue), where a
cached entity reference would have gone detached and could fail on a later flush. `updateCustomer()`
drops the cache, which also covers a changed subdomain, and `reset()` (via `ResetInterface`) clears
it between messages in long running workers.

`lazy: true` is removed from the service definition. It deferred nothing in practice — the postLoad
listener instantiates the service on virtually every request — while a lazy ghost cannot hydrate a
private property that has an inline default, which made the cache property crash with
`Error: Cannot access property starting with "\0"`.

### How to review/test
`vendor/bin/phpunit tests/backend/core/User/Unit/CustomerServiceTest.php` — six cases covering the
single-query behaviour, per-subdomain keys, identity-map resolution, and invalidation on
`updateCustomer()` / `reset()`. Also worth running
`tests/backend/core/Core/Functional/OzgKeycloakUserDataMapperTest.php` and
`tests/backend/core/Procedure/Functional/SubdomainHandlerTest.php`, which exercise the changed paths.

### PR Checklist
- [x] Create/Update tests


bakridemos (approved):
lgtm thx

PR #6841, opened 2026-09-17
fix DS-589: re-apply stripping of read-only relationships from segment update payload

mussbach:
### Ticket
DS-589

Re-applies #6612, whose changes were discarded by a conflict resolution that took the other side of `StatementSegment.vue` wholesale. The commit is still reachable in the history, so `git branch --contains` looks fine, but its content is gone and the broken pre-fix helpers came back. The error it fixed is still being reported.

`comments` and `recommendationVersions` are read-only on the `StatementSegment` resource (managed through their own resources, never via a segment update). The BE rejects any non-updatable relationship in an update payload, so whenever one of them reached the PATCH the whole request failed with:

```
[data][relationships][comments]: Unexpected field(s) in the context of the list of
relationships, allowed fields are: parentStatement, assignee, place, tags
```

The restored-by-accident `excludeComments()` / `excludeRecommendationVersion()` deleted the fields from a detached copy of the relationships object, while the vuex-json-api save action builds the payload by diffing the *stored* item -- so the stripping had no effect and whether the fields ended up in the request depended purely on store state. Hence the intermittency.

This strips them in `updateRelationships()`, before the item is written to the store, so they never enter the diff or the request body, and restores them to the store around the save so comments stay visible while the request is in flight.

Not a plain cherry-pick: the file has since gained `completeSave()` / `rollbackFailedSave()`, the `showWorkflowActions` -> `showWorkflowFields` rename and the response-status branch in `save()`, all of which are preserved.

Also drops the `SegmentSlidebar` `isLoading` toggle, which only existed to mask the comments disappearing and was an unbalanced set-true/set-false pair. That flag is still owned and toggled as a proper pair by `CreateCommentForm.vue` and consumed by `SegmentCommentsList.vue`, so the comments-slidebar spinner is unaffected.

### How to review/test
Assign a segment to another user (e.g. forward it for technical review) on a segment that has comments, and save. Before: the PATCH fails with the validation error above and the assignee does not change. After: the assignee is saved and the comments stay visible throughout.

`tests/frontend/unit/StatementSegmentReadOnlyRelationships.spec.js` is a regression guard -- it fails 6 of 7 against the unfixed component and passes fully with the fix.

### Linked PRs
- Reapplies #6612

### PR Checklist
- [x] Create/Update tests
- [x] Run `yarn lint`
- [x] Run `yarn test`


hwiem (approved):
Lgtm, thanks!

PR #6844, opened 2026-09-17
build: disable webpack realContentHash pass

hwiem:
- with `realContentHash` on, prod builds could emit a core bundle with a syntax error; when a hash consists only of digits and that digit sequence appears inside a numeric literal of another chunk, the literal is replaced with the new hex hash and the bundle no longer parses


mussbach (approved):
lgtm, tested, works, thx

PR #6845, opened 2026-09-17
fix(refs DPLAN-18333): display button row for tags only, too

hwiem:
### Ticket
Related to [DPLAN-18333](https://demoseurope.youtrack.cloud/issue/DPLAN-18333/Schlagworte-in-Erwiderung-verfassen-anzeigen-und-bearbeiten) - cancel/save buttons were not displayed if only tags were changed, but the "Schritt und/oder Bearbeiter*in ändern" and "Weitere Felder bearbeiten" checkboxes were not checked.

Now, cancel/save buttons are displayed as disabled when the segment is assigned to the current user and both checkboxes are unticked; then enabled once a tag is added or removed, or any of the other fields is changed.

When any of the fields behind the checkboxes has been changed and the checkbox is then unchecked, these changes are now reverted, so the cancel/save buttons are disabled again and these changes are not saved accidentally.

### PR Checklist

- [x] Run `yarn lint`
- [x] Run `yarn test`


sakutademos (approved):
code review, looks good, thx!

PR #6846, opened 2026-09-17
perf: avoid self-join to resolve segment procedure

mussbach:
### Ticket
None — found via Sentry DB-span triage, not tied to a ticket.

`StatementSegmentResourceType::getAccessConditions()` and `RpcSegmentIdLoader::getConditions()` both resolved a segment's procedure via `parentStatement->procedure->id`, which self-joins the statement table through the parent Statement row to reach the procedure FK. Since Segment extends Statement, it already carries its own procedure association directly (same column) — no join required.

The old query's cost scaled with the total number of statements (not just segments) attached to a procedure. On procedures with several thousand statements, this pushed the query from tens of milliseconds into the hundreds; reproducing the same query shape against a representative dataset showed a 35-56% reduction in execution time after switching to the direct path, with identical result sets.

### How to review/test
- Compare `getAccessConditions()`/`getConditions()` before/after: same filtered result, no self-join needed since `Segment` inherits `procedure` from `Statement`.
- Load a procedure's Segments list and run a bulk-edit RPC call to confirm segment filtering/access still works as expected.

### PR Checklist
- [ ] Create/Update tests
- [ ] Update documentation
- [ ] Add/Update data-cy attributes
- [ ] Run `yarn lint`
- [ ] Run `yarn test`


bakridemos (approved):
lgtm, code review only

PR #6850, opened 2026-09-18
F hddp 16 customer files deleted as orphans

bakridemos:
Ticket: https://demoseurope.youtrack.cloud/issue/HDDP-16/RoBOB-Bilder-fur-Leichte-Sprache-werden-nicht-angezeigt-Prod-Stage

Description: 

Images embedded only in a Customer's rich-text fields (overview description in simple/sign language, imprint, data protection, terms of use, xplanning) were never checked by FileInUseChecker. The nightly maintenance job treated them as orphaned and soft-deleted them within a day or two of upload, permanently breaking the images (the img tag and alt text stayed in the saved content, but the file itself was gone).

FileInUseChecker only searched for a file's ident, but Customer's editor embeds images via a direct "/file/{hash}" URL rather than the ident-based filestring tag used by Paragraph/Elements text fields, so even a partial check would have missed them.

Adds a check for Customer's rich-text fields to FileInUseChecker, matching against both the file's ident and its hash.

### How to review/test
- Upload an image into one of the Customer rich-text fields (e.g. easy-language overview) via customer settings.
- Trigger the maintenance file check (`checkDeletedFiles`) and confirm the file is no longer flagged as unused.
- Run the new test suite: `FileInUseCheckerTest`.

- [X] Create/Update tests


mussbach (approved):
sonarcloud issues is valid, otherwise lgtm, thx

